PR #17, opened 2021-10-24
Create build.yml

faddat:
works towards #1


faddat (commented):
Just so you're aware, this ins't really here as a "super great github action".  Instead, this is my starting point when working on things like this.

PR #21, opened 2021-11-17
(draft) Staking hook integration

jtremback:
These are my changes to support integration with the staking module unbonding hooks. This mechanism allows CCV to stop certain delegations from unbonding if the unbonding period on any relevant consumer chains has not yet elapsed.

It's still in draft because it is not unit tested, but it doesn't break anything


x/ccv/parent/keeper/keeper.go
```diff
@@ -56,6 +62,7 @@ func NewKeeper(
 		connectionKeeper: connectionKeeper,
 		clientKeeper:     clientKeeper,
 		registryKeeper:   registryKeeper,
```
AdityaSripal, line 64: Pull #23 into this PR, and remove registryKeeper

x/ccv/parent/keeper/relay.go
```diff
@@ -60,6 +70,30 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
 		return err
 	}
 	k.registryKeeper.UnbondValidators(ctx, chainID, data.ValidatorUpdates)
```
AdityaSripal, line 72: Remove this line

x/ccv/parent/module.go
```diff
@@ -152,8 +152,14 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
...
 func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.ValidatorUpdate {
-	am.keeper.IterateBabyChains(ctx, am.keeper.EndBlockCallback)
+	// For each child chain, call the endblock method which gets relevant validator set changes and
+	// sends them to the child chain
+	valUpdateID := am.keeper.GetValidatorSetUpdateId(ctx)
+	am.keeper.IterateBabyChains(ctx, func(ctx sdk.Context, chainID string) (stop bool) {
+		return am.keeper.EndBlockCallback(ctx, chainID, valUpdateID)
+	})
 	am.keeper.IteratePendingClientInfo(ctx)
+	am.keeper.IncrementValidatorSetUpdateId(ctx)
```
AdityaSripal, line 162: Again this shouldn't be necessary since there is already an index that increments on every packet sent

AdityaSripal (changes requested):
initial pass review.

Main points:
- Let's use the sequence in IBC instead of storing our own index
- We definitely should not be storing a redundant sequence in the packet data
- Remove registryKeeper. See #23

jtremback:
Discussed @AdityaSripal's comments on a call:

- `valset_update_id` is currently the same as the packet sequence, and so could be eliminated, but if any other packet types were added into the protocol, it would break.
- UnbondingDelegationEntry could be renamed, but it is a record of the same logical entity as the one in staking, and they will have a 1-1 relationship, so it makes sense for the to have the same name

These changes are now in #24

PR #30, opened 2022-01-31
Fix Child Genesis

AdityaSripal:
- Include initial validator set in child genesis and verify it against provider client
- Return it in InitGenesis to provide Tendermint with initial validator set over ABCI


x/ccv/child/keeper/genesis.go
```diff
@@ -46,6 +76,8 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *types.GenesisState) {
 			k.SetUnbondingPacket(ctx, us.Sequence, us.UnbondingPacket)
 		}
 	}
+
+	return state.InitialValSet
```
jtremback, line 80: Am I correct in assuming that this is exactly the same as was included in the genesis.json, and is already in the format that chain startup needs?

x/ccv/child/keeper/genesis.go
```diff
@@ -38,6 +43,31 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *types.GenesisState) {
 		// set parent client id.
 		k.SetParentClient(ctx, clientID)
 	} else {
+		// verify that latest consensus state on parent client matches the initial validator set of restarted chain
+		// thus, IBC genesis MUST run before CCV child genesis
```
jtremback, line 47: Just wondering, could this be verified by looking at the genesis.json alone? That would eliminate that condition, not that it matters.

x/ccv/child/keeper/genesis.go
```diff
@@ -57,7 +89,12 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 	}
 
 	if channelID, ok := k.GetParentChannel(ctx); ok {
-		gs := types.NewRestartGenesisState(channelID, nil)
+		clientID, ok := k.GetParentClient(ctx)
+		if !ok {
+			panic("parent client does not exist")
+		}
+		// ValUpdates must be filled in off-line
```
jtremback, line 96: I'm not following this comment

jtremback: Oh, they need to paste the latest provider valset updates in themselves, right? Maybe some time we can go over the exact procedure involved in doing a consumer chain genesis export and restart

jtremback (approved):
LGTM

proto/interchain_security/ccv/child/v1/genesis.proto
```diff
@@ -8,23 +8,28 @@ import "ibc/lightclients/tendermint/v1/tendermint.proto";
...
 import "gogoproto/gogo.proto";
 import "interchain_security/ccv/child/v1/child.proto";
+import "tendermint/abci/types.proto";
 
 // GenesisState defines the CCV child chain genesis state
 message GenesisState {
     Params   params          = 1 [(gogoproto.nullable) = false];
-    string parent_channel_id = 2
+    string parent_client_id = 2
+    [(gogoproto.moretags) = "yaml:\"parent_client_id\""]; // empty for a completely new chain
```
jtremback, line 17: Whitespace?

x/ccv/child/keeper/genesis.go
AdityaSripal: Correct. This is assuming the genesis file just includes the `[]abci.ValidatorUpdate` which is exactly what is expected by Tendermint

x/ccv/child/keeper/genesis.go
AdityaSripal: You can do it for the initial child chain genesis on `ValidateGenesis` which I do.

For restart child chain genesis, we need to read the state of the IBC module. There isn't a way for ValidateGenesis to look at another module's genesis. So we need to look at the already initialized IBC state in `InitGenesis` to do our verification.

x/ccv/child/keeper/genesis.go
```diff
@@ -38,6 +43,31 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *types.GenesisState) {
...
+
+		// ensure that initial validator set is same as initial consensus state on provider client.
+		// this will be verified by provider module on channel handshake.
+		vals, err := tmtypes.PB2TM.ValidatorUpdates(state.InitialValSet)
+		if err != nil {
+			panic(err)
+		}
+		valSet := tmtypes.NewValidatorSet(vals)
+
+		if !bytes.Equal(tmConsState.NextValidatorsHash, valSet.Hash()) {
```
mpoke, line 65: This check requires that the initial validator set in the GenesisState matches the one from the latest consensus state stored in the provider client. This consensus state depends on the last ClientUpdate before the GenesisState was exported. How can an off-chain process identify the validator set of that consensus state?

PR #31, opened 2022-02-02
Gov proposal cli and genesis state query

jtremback:
This adds
- The boilerplate to make a governance proposal to start a consumer chain.
- Code to generate a (hopefully) valid genesis state for a consumer chain and query it.


x/ccv/parent/keeper/proposal.go
```diff
@@ -42,10 +46,49 @@ func (k Keeper) CreateChildClient(ctx sdk.Context, chainID string, initialHeight
...
+	if !ok {
+		return gen, sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "self consensus state not found for height: %s", height)
+	}
+
+	gen.Params.Enabled = true
+	gen.NewChain = true
+	gen.ParentClientState = clientState
+	gen.ParentConsensusState = consState.(*ibctmtypes.ConsensusState)
+
+	vals := k.stakingKeeper.GetBondedValidatorsByPower(ctx)
```
mpoke, line 80: You should use the `LastValidatorPower` index instead, see https://github.com/cosmos/cosmos-sdk/blob/master/x/staking/genesis.go#L193. This contains the last validator set provided to Tendermint. It was provided at the end of the previous block, which means it's in the header of the current block as `NextValidatorsHash`. 

The reason is that by the end of the previous block, the block was already decided, thus the header cannot be changed. As a result, we have these delay of one block: The validator updates provided to Tendermint at height `h`, will be visible in the header as `NextValidatorsHash` at height `h+1` and the updated validator set will validate at hheight `h+2`.

x/ccv/types/expected_keepers.go
```diff
@@ -26,6 +26,8 @@ type StakingKeeper interface {
 	// slash the validator and delegators of the validator, specifying offence height, offence power, and slash fraction
 	Jail(sdk.Context, sdk.ConsAddress) // jail a validator
 	Slash(sdk.Context, sdk.ConsAddress, int64, int64, sdk.Dec)
+	GetBondedValidatorsByPower(ctx sdk.Context) []types.Validator
+	PowerReduction(ctx sdk.Context) sdk.Int
```
mpoke, line 30: Replace with `IterateLastValidatorPowers`, see https://github.com/cosmos/cosmos-sdk/blob/master/x/staking/genesis.go#L193.

x/ccv/parent/client/proposal_handler.go
```diff
@@ -0,0 +1,168 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+	govclient "github.com/cosmos/cosmos-sdk/x/gov/client"
+	govrest "github.com/cosmos/cosmos-sdk/x/gov/client/rest"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	clienttypes "github.com/cosmos/ibc-go/modules/core/02-client/types"
+	"github.com/cosmos/interchain-security/x/ccv/parent/types"
+	"github.com/spf13/cobra"
+)
+
```
mpoke, line 20: The filename is the same as `../proposal_handler.go`. It gets confusing.

x/ccv/parent/client/proposal_handler.go
```diff
@@ -0,0 +1,168 @@
...
+	Proposer sdk.AccAddress `json:"proposer"`
+
+	Title         string             `json:"title"`
+	Description   string             `json:"description"`
+	ChainId       string             `json:"chainId"`
+	InitialHeight clienttypes.Height `json:"initialHeight"`
+	GenesisHash   []byte             `json:"genesisHash"`
+	BinaryHash    []byte             `json:"binaryHash"`
+	SpawnTime     time.Time          `json:"spawnTime"`
+	Deposit       sdk.Coins          `json:"deposit"`
```
mpoke, line 111: Use `CreateChildChainProposalJSON` instead.

mpoke (changes requested):
LGTM. See comments bellow.

x/ccv/parent/client/proposal_handler.go
jtremback: You mean, just embed it?

mpoke: Yeah, to not duplicate code. But it's minor.

AdityaSripal (approved):
Nice work!

One optimization: We should delete the child genesis once the CCV channel for the chainID is OPEN since we no longer need to store it on parent chain.

x/ccv/parent/client/proposal_handler.go
jtremback: They are actually different

x/ccv/parent/keeper/proposal.go
jtremback: Hey @mpoke I have used the method you suggested but there is one edge case. `LastPowers` includes the operator address, but not the consensus address. We need the consensus address to send to the consumer chain, so I use `stakingKeeper.GetValidator` to retrieve it. I am wondering whether it is possible under some scenarios (validator leaving the set the same block?), for this validator not to be found.

Also, we use `stakingKeeper.GetValidatorUpdates` to send the updates during CCV operation- why not use it here? I'm guessing those updates are from the wrong block, right?

x/ccv/parent/keeper/proposal.go
```diff
@@ -6,10 +6,15 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	clienttypes "github.com/cosmos/ibc-go/modules/core/02-client/types"
 	commitmenttypes "github.com/cosmos/ibc-go/modules/core/23-commitment/types"
 	ibctmtypes "github.com/cosmos/ibc-go/modules/light-clients/07-tendermint/types"
 	"github.com/cosmos/interchain-security/x/ccv/parent/types"
+	tmtypes "github.com/tendermint/tendermint/abci/types"
```
mpoke, line 15: For consistency with the rest of the implementation, use `abci` instead of `tmtypes` (see parent/relayer.go). It's also `abci` throughout cosmos-sdk https://github.com/cosmos/cosmos-sdk/blob/056db1295768cbf4675b8025b66d83bbaadd923b/x/staking/genesis.go#L6

x/ccv/parent/keeper/proposal.go
mpoke: That's actually very interesting. In the function `CreateChildClient()` (see above), the proposer creates a consensus state for the consumer chain (i.e., a root of trust):
```go
consensusState := ibctmtypes.NewConsensusState(ctx.BlockTime(), commitmenttypes.NewMerkleRoot([]byte(ibctmtypes.SentinelRoot)), ctx.BlockHeader().NextValidatorsHash)
```
In my understanding,  `ctx.BlockHeader().NextValidatorsHash` MUST be the initial validator set on the consumer chain. Thus, if the current height is `h`, then `NextValidatorsHash` is the hash of the validators validating on the provider chain at height `h+1`. 

Now, in `MakeChildGenesis()`, the provider creates the consumer genesis state, which MUST contain the initial validator set, i.e., the one validating on the provider at `h+1`. This is the set updated by the staking module at height `h-1`. Thus, it is part of the  `LastValidatorPowers` as long as `MakeChildGenesis()` is **_called before_** the staking module `EndBlock()` for the current height `h`. 

If I'm not mistaken, `CreateChildChainProposal` (which ends up calling `MakeChildGenesis`) is called in the `EndBlock()` of the gov module, i.e., https://github.com/cosmos/cosmos-sdk/blob/056db1295768cbf4675b8025b66d83bbaadd923b/x/gov/abci.go#L73. I'm not sure though which `EndBlock` has higher priority.

PR #32, opened 2022-02-05
Simple Distribution

rigelrozanski:
Simple Distribution which also splits a parameterized fraction of the tokens to a consumer-side module account for further use (governance, DAO etc.) and sends the remainder to the provider chain through ibc


x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+		//}
+
+		err := k.ibcTransferKeeper.SendTransfer(ctx,
+			port,
+			channel,
+			token,
+			consumerFeePoolAddr,
+			providerFeePoolAddrStr,
+			clienttypes.Height{0, 0}, // XXX timeoutHeight
+			0,                        // XXX timeoutTimestamp
```
rigelrozanski, line 42: @AdityaSripal is it appropriate to set these to 0 here? Also can I set the height as I have?

x/ccv/child/module.go
```diff
@@ -249,6 +278,32 @@ func (am AppModule) OnChanOpenInit(
...
+
+	transferPortID := "transfer"
+	am.keeper.SetDistributionTransmissionPort(ctx, transferPortID)
+
+	distChan := channeltypes.Channel{
+		State:          channeltypes.UNINITIALIZED,
+		Ordering:       channeltypes.UNORDERED,
+		Counterparty:   counterparty,          // XXX is it correct to reuse
+		ConnectionHops: connectionHops,        // XXX is it correct to reused
+		Version:        transfertypes.Version, // XXX verify that this is correct
```
rigelrozanski, line 294: @AdityaSripal have I set `Counterparty`, `ConnectionHops`, and `Version` correctly here?

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+	ltbh := &types.LastTransmissionBlockHeight{}
+	ltbh.Unmarshal(bz)
+	bpdt := k.GetBlocksPerDistributionTransmission(ctx)
+	curHeight := ctx.BlockHeight()
+	return (curHeight - ltbh.Height) >= bpdt
+}
+
+func (k Keeper) ChannelOpenInit(ctx sdk.Context, msg *channeltypes.MsgChannelOpenInit) (
+	*channeltypes.MsgChannelOpenInitResponse, error) {
+	return k.ibcCoreKeeper.ChannelOpenInit(ctx, msg)
```
rigelrozanski, line 68: @AdityaSripal I'm not sure if we're allowed to use `context.Context` and `sdk.Context` interchangeably here or if this is just a dep issue? I get a compilation error here: 
```
x/ccv/child/keeper/distribution.go:68:40: cannot use ctx (type "github.com/cosmos/cosmos-sdk/types".Context) as type context.Context in argument to k.ibcCoreKeeper.ChannelOpenInit:
	"github.com/cosmos/cosmos-sdk/types".Context does not implement context.Context (missing Deadline method)
```
Thoughts on resolving this error?

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+	providerFeePoolAddrStr := k.GetProviderFeePoolAddrStr(ctx)
+
+	tokens := k.bankKeeper.GetAllBalances(ctx, consumerFeePoolAddr)
+	for _, token := range tokens {
+
+		// XXX delete or uncomment this block before merge
+		//if !strings.HasPrefix(token.Denom, "ibc/") {
+		//    denomTrace := ibctypes.ParseDenomTrace(token.Denom)
+		//    token.Denom = denomTrace.IBCDenom()
+		//}
```
rigelrozanski, line 33: @AdityaSripal any ideas if I need this block that I've left commented out here?

x/ccv/child/keeper/distribution.go
AdityaSripal: No, don't set the height you have since timeout height is the height of the counterparty.

Use `ctx.BlockTime()` plus a relatively large timeoutDelta to set the absolute timeoutTimestamp for the packet

x/ccv/child/module.go
```diff
@@ -249,6 +278,32 @@ func (am AppModule) OnChanOpenInit(
...
+	}
+	distrTransferMsg := channeltypes.MsgChannelOpenInit{
+		PortId:  transferPortID,
+		Channel: distChan,
+		Signer:  "", // signer unused
```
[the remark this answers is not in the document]
AdityaSripal: Yes you can

x/ccv/child/keeper/distribution.go
AdityaSripal: They are different types.

You need to wrap sdk.Context: https://github.com/cosmos/cosmos-sdk/blob/a0a1197c2786647139bab92fcc6f9f3ee0d867d5/types/context.go#L278

x/ccv/child/module.go
```diff
@@ -249,6 +278,32 @@ func (am AppModule) OnChanOpenInit(
...
+	// NOTE the handshake for this channel is handled by
+	//      the ibc-go/transfer module
+
+	transferPortID := "transfer"
+	am.keeper.SetDistributionTransmissionPort(ctx, transferPortID)
+
+	distChan := channeltypes.Channel{
+		State:          channeltypes.UNINITIALIZED,
+		Ordering:       channeltypes.UNORDERED,
+		Counterparty:   counterparty,          // XXX is it correct to reuse
```
AdityaSripal, line 292:
```suggestion
		Counterparty:   channeltypes.NewCounterparty(transfertypes.Version, ""),
```

x/ccv/child/module.go
AdityaSripal: ConnectionHops and Version is correct, Counterparty is wrong. see above comment

x/ccv/child/keeper/distribution.go
AdityaSripal: Ahh yes you will need this if you expect that people will pay fees on the consumer chain with tokens they got from somewhere else.  this is because the internal representation of IBC coins in bank is different than the on-the-wire representation of the coin in IBC packets.

See: https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md

x/ccv/child/module.go
AdityaSripal: we're basically asking for a counterparty channel with port transfer. ChannelID is left blank, since it will be autogenerated on counterparty and returned to us later in the handshake

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// Simple model, donate tokens to the fee pool of the provider validator set
+// reference: cosmos/ibc-go/modules/apps/transfer/keeper/msg_server.go
+func (k Keeper) DonateToProviderValidatorSet(ctx sdk.Context) error {
+	if !k.shouldTransmit(ctx) {
+		return nil
+	}
+
+	port := k.GetDistributionTransmissionPort(ctx)
```
AdityaSripal, line 19: Shouldn't this just be a constant: `transfer`?

Is this to future-proof the portID changing?

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+	channel := k.GetDistributionTransmissionChannel(ctx)
+
+	// work around to reuse the IBC token transfer logic
+	consumerFeePoolAddr := k.accountKeeper.GetModuleAccount(ctx, k.feeCollectorName).GetAddress()
+	providerFeePoolAddrStr := k.GetProviderFeePoolAddrStr(ctx)
```
AdityaSripal, line 24: We need to document very clearly that `providerFeePoolAddress` is not included as a blocked address in bank keeper. Otherwise, any transfer to it will be blocked

x/ccv/child/module.go
```diff
@@ -249,6 +278,32 @@ func (am AppModule) OnChanOpenInit(
 		return err
 	}
 
+	///////////////////////////////////////////////////
+	// Initialize distribution token transfer channel
+	// NOTE the handshake for this channel is handled by
+	//      the ibc-go/transfer module
+
+	transferPortID := "transfer"
```
AdityaSripal, line 286: nit: Use the constant available in transfer/types in ibc-go

x/ccv/child/module.go
```diff
@@ -249,6 +278,32 @@ func (am AppModule) OnChanOpenInit(
...
+		Counterparty:   counterparty,          // XXX is it correct to reuse
+		ConnectionHops: connectionHops,        // XXX is it correct to reused
+		Version:        transfertypes.Version, // XXX verify that this is correct
+	}
+	distrTransferMsg := channeltypes.MsgChannelOpenInit{
+		PortId:  transferPortID,
+		Channel: distChan,
+		Signer:  "", // signer unused
+	}
+	resp, err := am.keeper.ChannelOpenInit(ctx, &distrTransferMsg)
```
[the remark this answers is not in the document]
AdityaSripal: This is a good point, cc: @rigelrozanski

x/ccv/child/keeper/distribution.go
okwme: in a normal transfer, it would be good UX for a transaction that hasn't gotten relayed to revert. A user may want to try something else in the face of a timed out transaction. in our scenario, what happens if it times out? ideally a relayer submits the timeout information and the tokens are put back into the consumerfeepooladdr. from there they could be included in a subsequent relayed transaction. is this any better than continuing to wait for the original unrelayed transaction? i'd say yes, because i make the assumption that the original relayed transaction was missed by a relayer but a subsequent attempt might not be—implying that it is easier for a relayer to do the basic action of relaying a new message, rather than recovering a lost message (unsubstantiated). if that is in fact the case then we want unrelayed messages to timeout (this may all be moot of course because the timeout needs to also be submitted by a relayer, right? or do timeouts undo themselves automatically?)

x/ccv/child/keeper/distribution.go
rigelrozanski: sweet. Thanks for the clarification. uncommented this block.

x/ccv/child/module.go
rigelrozanski: Thanks for the clarification. I've included this logic in the newest commit with some more comments reflecting your explanation.

x/ccv/child/keeper/distribution.go
rigelrozanski: Oh yeah sweet. Thanks

x/ccv/child/keeper/distribution.go
rigelrozanski: I've changed this code to identical to the link @AdityaSripal provided, so the delta is 4 weeks. Seems a bit long I was going to make it an hour.

docs/ccv-distribution/ccv-distribution.md
```diff
@@ -0,0 +1,89 @@
...
+          │   │validate into a ProviderPoolWeights object│         │            
+          │   └───────────┬───────────┬──────────────────┘         │            
+          │               │           │    ┌───────────────────────┴─────┐    
+          │     ┌─────────┴─────────┐ └────┤truncate ProviderPoolTokens  │ 
+          │     │ProviderPoolWeights│      │converting: DecCoins -> Coins│
+          │     └─────────┬─────────┘      └─────────────┬───────────────┘  
+          │               │                              │                
+          └─              ▼                              ▼  
+          ┌─    transmit to provider            transmit to provider                            
+  Relayer─┤     chain on CCV ordered            chain on unordered
```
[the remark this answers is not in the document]
rigelrozanski: Leaving in here for now, it's still accurate for itself, but it's not the simple distr. I'll probably make another simple distr diagram

x/ccv/child/module.go
```diff
@@ -160,6 +163,10 @@ func (am AppModule) EndBlock(ctx sdk.Context, req abci.RequestEndBlock) []abci.V
 	}
 	am.keeper.DeletePendingChanges(ctx)
 	am.keeper.UnbondMaturePackets(ctx)
+
+	// distribution transmission
+	am.keeper.DonateToProviderValidatorSet(ctx)
```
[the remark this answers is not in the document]
rigelrozanski: Yup that's fine, changed

proto/interchain_security/ccv/child/v1/child.proto
```diff
@@ -1,10 +1,24 @@
...
+    // distribution params
+    int64  BlocksPerDistributionTransmission = 2; 
+    string ProviderFeePoolAddrStr = 3; 
+    string DistributionTransmissionChannel = 4; 
+    string DistributionTransmissionPort = 5; 
+}
+
+// LastTransmissionBlockHeight is the last time validator holding
+// pools were transmitted to the provider chain
+message LastTransmissionBlockHeight {
```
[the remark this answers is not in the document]
rigelrozanski: don't think I'm up to speed. I'm not sure what you are referring to when you say `revision_number, revision_height`... 

Is this just with regards to the maybe non-specific naming here? Maybe this should be called `LastDistrTransmissionBlockHeight` or something to that effect?

if this is wrt updates sent from the provider to the consumer then this is more or less not applicable. Right now the distribution events are triggered only by the time, we could also easily trigger them on validator set updates if we wanted which would require a tweaks - if we did this all those heights would be embedded into the height of `LastTransmissionBlockHeight`

x/ccv/child/module.go
```diff
@@ -219,6 +226,23 @@ func ValidateChildChannelParams(
 }
 
 // OnChanOpenInit implements the IBCModule interface
+// this function is called by the relayer.
```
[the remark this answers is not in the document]
rigelrozanski: They were just my working notes, removed them

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,69 @@
...
+	clienttypes "github.com/cosmos/ibc-go/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	"github.com/cosmos/interchain-security/x/ccv/child/types"
+	// XXX delete or uncomment this block before merge
+	//ibctypes "github.com/cosmos/ibc-go/v3/modules/apps/transfer/types"
+)
+
+// Simple model, donate tokens to the fee pool of the provider validator set
+// reference: cosmos/ibc-go/modules/apps/transfer/keeper/msg_server.go
+func (k Keeper) DonateToProviderValidatorSet(ctx sdk.Context) error {
```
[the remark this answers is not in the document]
rigelrozanski: Changed to @jtremback 's comment of `DistributeToProviderValidatorSet`

x/ccv/child/module.go
rigelrozanski: updated (removed this variable altogether)

x/ccv/child/keeper/distribution.go
rigelrozanski: Okay yeah good point, haven't dealt with much of this blocked address business. Blocked addresses are configured in the `app.go` if I'm not mistaken. Leaving this comment open for now

x/ccv/child/keeper/distribution.go
rigelrozanski: Yeah I mean we _could_ just keep it as a constant, it is also a bit more future proof to keep it in the params as I've done, but maybe that's not such a huge concern as updates to the transfer port would also likely come with hard-updates to the code base. 

I guess I also like the flow of information to be coupled, as in we get the channel from the param store, so why not also keep a reference to the port-id there? 

Kind of on the fence, I don't think it matters so much either way. Maybe we should cut it to reduce code... Alright I'm going to cut it out

x/ccv/child/module.go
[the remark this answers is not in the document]
rigelrozanski: Okay I've moved the code to the Ack - a small issue is introduced now, we need to get the `connectionHops` from somewhere. Previously, when the handshake was initiated in `OnChanOpenInit` the connection hops was passed in through the function signature. There are two good options as I see it: 
 1) pass back connectionHops from provider OnChanOpenTry through version-metadata (easy). However this can only work if the connectionHops string is the same on the provider chain (is it? @AdityaSripal  @mpoke) 
 2)  store the connectionHops from OnChanOpenInit on the consumer chain then retrieve it during OnChanOpenAck (little bit more annoying because we need to setup a key/do all the store boilerplate code... but not a huge deal)

(side thought any good documentation for what the connection hops is kicking around?)

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// work around to reuse the IBC token transfer logic
+	consumerFeePoolAddr := k.accountKeeper.GetModuleAccount(ctx, k.feeCollectorName).GetAddress()
+	tokens := k.bankKeeper.GetAllBalances(ctx, consumerFeePoolAddr)
+	for _, token := range tokens {
+
+		if !strings.HasPrefix(token.Denom, "ibc/") {
+			denomTrace := ibctypes.ParseDenomTrace(token.Denom)
+			token.Denom = denomTrace.IBCDenom()
+		}
```
AdityaSripal, line 30: Ahh sorry, I this this block should actually just be removed. Also this conditional should be `if strings.HasPrefix...` anyway

We already reconstruct the denom within `SendTransfer`: 

https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L97
https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L407

AdityaSripal: We should also have a test for this

x/ccv/child/module.go
```diff
@@ -263,25 +272,71 @@ func (am AppModule) OnChanOpenTry(
...
+
+	///////////////////////////////////////////////////
+	// Initialize distribution token transfer channel
+	// NOTE the handshake for this channel is handled by
+	//      the ibc-go/transfer module
+
+	// here NewCounterparty asks for a counterparty channel with the port
+	// transfer. ChannelID (second input) is left blank, since it will be
+	// autogenerated by the counterparty and returned to us when calling
+	// ChannelOpenInit
```
AdityaSripal, line 314:
```suggestion
	// autogenerated by the counterparty and returned to us on ChanOpenAck
```

x/ccv/child/keeper/keeper.go
```diff
@@ -22,36 +22,52 @@ import (
...
-	connectionKeeper ccv.ConnectionKeeper
-	clientKeeper     ccv.ClientKeeper
+	storeKey          sdk.StoreKey
+	cdc               codec.BinaryCodec
+	paramSpace        paramtypes.Subspace
+	scopedKeeper      capabilitykeeper.ScopedKeeper
+	channelKeeper     ccv.ChannelKeeper
+	portKeeper        ccv.PortKeeper
+	connectionKeeper  ccv.ConnectionKeeper
+	clientKeeper      ccv.ClientKeeper
```
AdityaSripal, line 32: You don't really need this since they're all accessible by `ibcCoreKeeper`

x/ccv/child/module.go
```diff
@@ -263,25 +272,71 @@ func (am AppModule) OnChanOpenTry(
...
+		Counterparty: transferCounterparty,
+
+		// XXX can either pass back from provider OnChanOpenTry through
+		// version-metadata (easy) or store on the consumer chain in
+		// OnChanOpenInit (little bit more difficult but also do-able)
+		ConnectionHops: "", //connectionHops,
+
+		Version: transfertypes.Version,
 	}
+	distrTransferMsg := channeltypes.MsgChannelOpenInit{
```
AdityaSripal, line 329: This is incorrect. Use the constructor function: https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/msgs.go#L19

ConnectionHops should not be empty. Use the connection hops that this channel has.

Get connectionhops here by:

```go
// pseudocode
channel := am.keeper.GetChannel(ctx, portID, channelID)
connectionHops := channel.ConnectionHops
```

AdityaSripal (changes requested):
Nice work! Creation of ChanOpenInit msg for transfer is still incorrect

x/ccv/child/keeper/distribution.go
rigelrozanski: Alright remove the block, I'm going to leave this "unresolved" as a reminder about adding a test for other ibc denoms.

x/ccv/child/keeper/keeper.go
rigelrozanski: The clients are? makes sense, however that logic is actually outside of the scope of this PR (I only modified further lines below - just weird github highlighting). I'll make a PR removing the client keeper in a separate PR in the future (or feel free to do the same)

x/ccv/child/module.go
rigelrozanski: Thank you. I've updated the code to using the constructor function (didn't know it was there). Let me know what you think, I think it's correct now

x/ccv/child/module.go
```diff
@@ -270,16 +279,51 @@ func (am AppModule) OnChanOpenAck(
...
+		connHops,
+		transfertypes.PortID,
+		"", // signer unused
+	)
+
+	resp, err := am.keeper.ChannelOpenInit(ctx, distrTransferMsg)
+	if err != nil {
+		return err
 	}
+	am.keeper.SetDistributionTransmissionChannel(ctx, resp.ChannelID)
```
AdityaSripal, line 325: One thing interesting to note is that this logic will be an issue if the transfer handshake fails at a later point. I don't think that's possible to happen currently. But worth noting

AdityaSripal (approved):
Nice work! Handshake and transmission logic looks good pending tests

x/ccv/child/module.go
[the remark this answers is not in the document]
mpoke: https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#definitions
> The `connectionHops` stores the list of connection identifiers, in order, along which packets sent on this channel will travel. At the moment this list must be of length 1. In the future multi-hop channels may be supported.

I'd assume that the cleaner option is to store the connectionHops from OnChanOpenInit on the consumer chain. I'm not sure that the connectionHops on the provider will contain the same connections identifiers. 

@AdityaSripal  How is the relayer creating `MsgChannelOpenInit`? We can do it in the same way since we already know the connection identifier.

x/ccv/child/module.go
rigelrozanski: Okay - so if I understand this correctly, the handshake could fail and we wouldn't know it here. The solution would be to verify that the transfer channel was opened successfully before using it, if it wasn't opened successfully, we could reinitiate it at that later time... Do you think we want this feature or no? Having trouble right now with verifying that the transfer channel handshake actually succeeds (see next comment).

x/ccv/parent/parent_test.go
```diff
@@ -501,3 +506,111 @@ func (s *ParentTestSuite) TestHandleConsumerDowntime() {
...
+	s.Require().NoError(err)
+	bpdt = cKeep.GetBlocksPerDistributionTransmission(ctx)
+	curHeight := ctx.BlockHeight()
+	fmt.Printf(
+		"before distribution:\nltbh:%v\nbpdt:%v\ncurrHeight:%v\n(curHeight - ltbh.Height) < bpdt:%v\n",
+		ltbh.Height, bpdt, curHeight, (curHeight-ltbh.Height) < bpdt)
+
+	// Verify that the destinationChannel exists
+	// XXX if this doesn't exist then the transfer logic will fail when
+	// a the distribution transfer is invoked in the next block.
```
rigelrozanski, line 585: @AdityaSripal The transfer `destinationChannel` doesn't exist at this point within my test (as you'll see a few lines down... This snipped from of lines 586-592 basically replicates the details of failing code that occurs when I actually try to submit a transfer (which is triggered by distribution once the next block is committed) - hence if I can figure out why the destinationChannel doesn't exist here then failing transfer may succeed (provided no new errors arise)... I'm not totally sure if I've setup the test properly, or if the error lies in how the connection is setup during `ConnAck`... Appreciate your insight!

AdityaSripal: So the handshake could not have completed by this point, because the CCV channel setup will only do the INIT step of the transfer channel.

In the real world, a relayer will then have to complete the rest of the handshake. So you have to mock the completion of the transfer channel handshake in your test setup before you get to this line.

It is completely expected destination channel doesn't exist yet, because INIT will not populate destination channel. That gets created on TRY, and sent back to the sender on ACK.

rigelrozanski: Great, thanks Adiyta, your fixes make total sense to me.. didn't understand the limitations of the test suite

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Send the remainder to the Provider fee pool over ibc
+	remainingTokens := fpTokens.Sub(consRedistrTokens)
+	ch := k.GetDistributionTransmissionChannel(ctx)
+	providerAddr := k.GetProviderFeePoolAddrStr(ctx)
+	timeoutHeight := clienttypes.Height{0, 0}
+	timeoutTimestamp := uint64(ccv.GetTimeoutTimestamp(ctx.BlockTime()).UnixNano())
```
mpoke, line 53: We should use a different timeout for token transfer than for CCV packets. For CCV packets we use large timeouts to ensure that the CCV channel will not timeout and close (since it's an ordered channel). In the case a token transfer times out, the transfer channel is not closed and the funds are returned to `consumerFeePoolAddr`. 

These funds will be sent again with the next transfer. However, these funds shouldn't be multiplied again with `ConsumerRedistributeFrac`.  This means that the `consumerFeePool` should collect only the funds that need to be transferred to the provider.

x/ccv/child/types/params.go
```diff
@@ -16,15 +22,26 @@ func ParamKeyTable() paramtypes.KeyTable {
...
 
 // DefaultParams is the default params for the child module
 func DefaultParams() Params {
-	return NewParams(false)
+	return NewParams(
+		false,
+		1000, // about 2 hr at 7.6 seconds per blocks
+		"",
+		"",
+		"0.5",
```
mpoke, line 43: I'd set the default to 0 (all rewards go to the provider) since it's the simplest case. The alternative entails the consumer to figure out a mechanism to distribute its share of the rewards.

proto/interchain_security/ccv/child/v1/child.proto
```diff
@@ -3,15 +3,27 @@ package interchain_security.ccv.child.v1;
...
 // Params defines the parameters for CCV child module
 message Params {
     bool Enabled = 1;
-}
 
+    // distribution params
+    int64  BlocksPerDistributionTransmission = 2; 
+    string DistributionTransmissionChannel = 3; 
+    string ProviderFeePoolAddrStr = 4; 
+    string ConsumerRedistributeFrac = 5;  // ex: "0.5"
```
mpoke, line 22: Add a comment to each one explaining what it is. This would be useful especially for `ConsumerRedistributeFrac`.

x/ccv/parent/module.go
```diff
@@ -246,13 +246,16 @@ func (am AppModule) OnChanOpenTry(
 	chanCap *capabilitytypes.Capability,
 	counterparty channeltypes.Counterparty,
 	counterpartyVersion string,
-) (version string, err error) {
-	if err := ValidateParentChannelParams(ctx, am.keeper, order, portID, channelID, ccv.Version); err != nil {
-		return ccv.Version, err
+) (metadata string, err error) {
+
+	if err := ValidateParentChannelParams(
```
mpoke, line 251: Update `ValidateParentChannelParams` to not check the version. Currently it checks whether `ccv.Version == ccv.Version`.

mpoke (changes requested):
Great work. The only concern I have is the case when the token transfer does not succeed and the tokens are refunded (see my comment bellow).

x/ccv/child/types/keys.go
```diff
@@ -51,11 +51,15 @@ const (
 
 	// CrossChainValidatorPrefix is the key prefix that will store cross-chain validators by consensus address
 	CrossChainValidatorPrefix = "crosschainvalidator"
+
+	// ConsumerRedistributeName the root string for the consumer-redistribution account address
+	ConsumerRedistributeName = "cons_redistribute"
```
jtremback, line 56: Why isn't this a param? Just curious, not saying it should be

x/ccv/parent/parent_test.go
```diff
@@ -722,3 +763,157 @@ func (s *ParentTestSuite) UpdateChildHistInfo(changes []abci.ValidatorUpdate) {
...
+	sourceChannel := cKeep.GetDistributionTransmissionChannel(ctx)
+	sourceChannelEnd, found := cApp.IBCKeeper.ChannelKeeper.GetChannel(ctx, sourcePort, sourceChannel)
+	s.Require().True(found)
+	destinationChannel := sourceChannelEnd.GetCounterparty().GetChannelID()
+	s.Require().True(len(destinationChannel) > 0)
+
+	// commit 1 more block (which should invoke a distribution event)
+	rspEB, _, _ := s.coordinator.CommitBlockGetResponses(cChain)
+
+	// get the packet from the endblock events
```
jtremback, line 861: Nice!

jtremback (approved):

x/ccv/child/keeper/distribution.go
jtremback: What is the benefit of this channel timing out? For the regular transfer use case, when a human is trying to transfer, it is so that they can do something else with their money if the transfer doesn't work. For this, it seems like it would be ok for packets to stay valid as long as possible.

mpoke: Yeah, but what if nobody relays that transfer ever? The transfer channel is unordered, which means that a transfer can be skipped. A timeout fixes that.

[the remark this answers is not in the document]
mpoke: No, it is set to the hardcoded value that we use also for CCV packets, see https://github.com/cosmos/interchain-security/blob/76f97477b9c9c9705701dff387fea565897102ee/x/ccv/types/utils.go#L6

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,114 @@
...
+	fpTokens := k.bankKeeper.GetAllBalances(ctx, consumerFeePoolAddr)
+
+	// split the fee pool, send the consumer's fraction to the consumer redistribution address
+	fracStr := k.GetConsumerRedistributeFrac(ctx)
+	frac, err := sdk.NewDecFromStr(fracStr)
+	if err != nil {
+		return err
+	}
+	decFPTokens := sdk.NewDecCoinsFromCoins(fpTokens...)
+	consRedistrTokens, _ := decFPTokens.MulDec(frac).TruncateDecimal()
```
AdityaSripal, line 42: Why are we doing this logic here? It seems simpler to me that the child chain application decides how much of the fee goes into their community pool, and how much goes into the ccv account. And then the full ccv account gets sent to the provider.

It also reduces the code inside ccv

x/ccv/child/keeper/distribution.go
AdityaSripal: Agreed. We can use a smaller timeout. And since the money is refunded to the same address, we effectively retry the transfer in the next epoch (along with the new accumulated funds).

By the way, this is an additional reason to do the splitting of funds elsewhere, and send full account balance to provider

x/ccv/parent/parent_test.go
```diff
@@ -722,3 +763,157 @@ func (s *ParentTestSuite) UpdateChildHistInfo(changes []abci.ValidatorUpdate) {
...
+	ctx := cChain.GetContext()
+	sourcePort := transfertypes.PortID
+	sourceChannel := cKeep.GetDistributionTransmissionChannel(ctx)
+	sourceChannelEnd, found := cApp.IBCKeeper.ChannelKeeper.GetChannel(ctx, sourcePort, sourceChannel)
+	s.Require().True(found)
+	destinationChannel := sourceChannelEnd.GetCounterparty().GetChannelID()
+	s.Require().True(len(destinationChannel) > 0)
+
+	// commit 1 more block (which should invoke a distribution event)
+	rspEB, _, _ := s.coordinator.CommitBlockGetResponses(cChain)
```
AdityaSripal, line 859: Nice we should upstream any event parsing you've added to the test suite

AdityaSripal (changes requested):
Nice work! I think my main suggestion is that we remove the redistribution logic. I think it's cleaner to move that logic outside the scope of CCV, and then the timeout logic for transfer is cleaner and we have less code to maintain

x/ccv/child/types/params.go
rigelrozanski: Sounds good. Done

x/ccv/parent/module.go
rigelrozanski: yup, done

app/app.go
```diff
@@ -244,7 +250,7 @@ type App struct { // nolint: golint
 	ScopedIBCParentKeeper capabilitykeeper.ScopedKeeper
 
 	// the module manager
-	mm *module.Manager
+	MM *module.Manager
```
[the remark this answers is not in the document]
rigelrozanski: I needed this for the purposes of disabling distribution during the test suite. This can be returned to "unexposed once we split the consumer and provider apps"

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,114 @@
...
+	if (curHeight - ltbh.Height) < bpdt {
+		// not enough blocks have passed for  a transmission to occur
+		return nil
+	}
+
+	// work around to reuse the IBC token transfer logic
+	consumerFeePoolAddr := k.accountKeeper.GetModuleAccount(ctx, k.feeCollectorName).GetAddress()
+	fpTokens := k.bankKeeper.GetAllBalances(ctx, consumerFeePoolAddr)
+
+	// split the fee pool, send the consumer's fraction to the consumer redistribution address
```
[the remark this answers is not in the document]
rigelrozanski: It is (provided that the standard distribution module is excluded from the consumer chain)

x/ccv/parent/parent_test.go
rigelrozanski: Yeah, I've added this to a ibc branch already... will open a PR before resolving this conversion

x/ccv/child/types/keys.go
rigelrozanski: Because it doesn't need to be a param, really any further logic should be hardcoded to this use the consumer redistribution address, for instant a "send to the developers multisig" could easily be an automatic transaction from this address. This is similar to the fees address which is hardcoded for the application. Closing issue, feel free to reopen if you want to discuss more

docs/ccv-distribution/ccv-distribution.md
```diff
@@ -0,0 +1,89 @@
+# CCV Distribution
```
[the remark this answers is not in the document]
rigelrozanski: For the expanded future design approach yes. (this is not the simple distribution)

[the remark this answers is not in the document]
rigelrozanski: Added a clarifying comment to the top of the file.

x/ccv/child/keeper/distribution.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	bpdt := k.GetBlocksPerDistributionTransmission(ctx)
+	curHeight := ctx.BlockHeight()
+
+	if (curHeight - ltbh.Height) < bpdt {
+		// not enough blocks have passed for  a transmission to occur
+		return nil
+	}
+
+	// work around to reuse the IBC token transfer logic
```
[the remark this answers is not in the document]
rigelrozanski: There actually is no real work around, so I just removed this comment. We just need module the address (next line), normally with module accounts you just send the string name of the account not the address, but because ibc-token transfer was designed for non-module accounts it expects the address... but it's simple to determine, so I wouldn't really even call this a workaround.

x/ccv/child/keeper/distribution.go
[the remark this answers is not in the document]
rigelrozanski: That's incorrect, I'm adding a comment to clarify this. The truncated amount is included in the "remainder" which is sent to the fee pool. If you wanted 100% of the tokens to go to the ConsumerRedistribute then a factor of "1" would do this with no remainder to be truncated.

x/ccv/child/keeper/distribution.go
rigelrozanski: First note: I've updated the timeout to a hardcoded calculation set to 1 week (instead of 4 weeks as it was before). @mpoke if you think this should be a param or some other value, just lmn. 

Secondly: Great catch for the rejected transfer situation. The solution is really easy, which is to send the tokens to a buffer address on the consumer chain side. If the transfer times out, it just gets sent back to the buffer where it _will not_ have a second redistribution split applied to it when it is sent out with the next distribution event. Personally I don't see this as a maintenance item. It's literally 9 lines of code right now + a few comments. I think it's valuable to potential consumer chains to have this boilerplate split written out, I imagine most chains will want to have this split functionality whereby they can choose. 

Re: @AdityaSripal (replying to your comment in this thread) 

> Why are we doing this logic here? It seems simpler to me that the child chain application decides how much of the fee goes into their community pool, and how much goes into the ccv account. And then the full ccv account gets sent to the provider. It also reduces the code inside ccv

So this location _is_ the first point in which fees/minted-tokens are first processes and where the fundamental split is made between consumer tokens and provider tokens. The design decision here is that all further use/distribution of funds on the consumer chains should be made _from_ the consumer's redistribute address, whether that be providing the tokens to a consumer chain community pool, developer multi-sig or whatever, I don't think any of those splits should happen before the "major" split between what's staying on the chain and what's going to the provider chain. In this way, this logic is actually consistent with your comment "that the child chain application decides how much of the fee goes into their community pool" We're not writing logic here about how a consumer chain should use its funds, just providing the infrastructure to make a split at this fundamental level. I also don't see it as a maintenance item given the simplicity of the solution. If we don't provide this basic split functionality in one way or another, the ccv distribution feels a bit "half-baked" to me.

x/ccv/child/keeper/distribution.go
rigelrozanski: I've replied to your comment in this thread (let's continue discussion there). https://github.com/cosmos/interchain-security/pull/32#discussion_r859101977

x/ccv/child/module.go
rigelrozanski: Added a comment to note this

x/ccv/child/keeper/distribution.go
rigelrozanski: Added comments in app.go and provider module.go

x/ccv/child/module.go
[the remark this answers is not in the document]
rigelrozanski: connection hops are saved in the channel. Resolved

x/ccv/parent/parent_test.go
rigelrozanski: Alright - here is the PR (still in DNM status)... needs to be rebased looks like, still good to have this PR open for reference purposes. https://github.com/cosmos/ibc-go/pull/1304

rigelrozanski:
Rebase looks up to date with master in my command line. Just need to come to consensus on the redistribution, then we ought to merge ASAP.

x/ccv/child/keeper/distribution.go
jtremback: I agree with Rigel here. The nice thing about the way he's done things is that it requires no changes to Cosmos-SDK, just app.go. I think having more code in CCV is a good thing because it is easier for us to work with here and we are the one building the feature. It is also good for the user because the functionality is closely related to IS, so it keeps all the params and stuff related to it in one place.

I would recommend that he do the "buffer account" to avoid the double splitting and keep the current splitting logic in CCV.

mpoke: I agree. The buffer account will solve the issue. Also, having the split happing in the CCV module seems to me cleaner since it's part of the CCV protocol. 

Regarding the transfer timeout, I'm not sure whether 1 week is still too long. By [default](https://github.com/cosmos/interchain-security/blob/ab153fceccac12b7957d73e5401785f7f758b3e1/x/ccv/child/types/params.go#L37), the consumer is sending rewards every 2h. Keeping this interval small reduces the inaccuracy of the simple distribution (i.e., the rewards are distributed to the validators that validate at the time the transfer is received). For the same reason, the timeout should also be smaller. Having it too small though will lead to a lot of timeout packets that need to be relay. So maybe 1 day would be decent. What do you say @AdityaSripal? 

In my view, the only reason timeouts will be common is if the amount of tokens transferred in each packet is very small, i.e., the value of the packet is too low to even bother relaying it. In that case, relayers would ignore these packets and wait for more rewards to accumulate. However, timing out packets also requires relaying TimeoutMsgs, so I'm not sure if that would make sense financially.

AdityaSripal (approved):
The fix for redistribution looks good! I think the timeout can be even lower but this is not an issue and doesn't need to block this or. Great work @rigelrozanski !

mpoke (approved):
Great work @rigelrozanski

PR #45, opened 2022-02-14
Version upgrade

rigelrozanski:
version upgrade, 
 - uses cosmos/cosmos-sdk `interchain-security-rebase` branch
 - needs: https://github.com/cosmos/ibc-go/pull/917
 - needed for: https://github.com/cosmos/interchain-security/pull/32

PR #48, opened 2022-02-22
Fix: Patch Consumer initiated slashing #28

sainoe:
- Increment `ValsetUpdateID` once per provider chain block; fix the impacted unit-tests

- Use a reverse iterator to get last  the `ValsetUpdateID` on consumer chains

- Check that slashing packets are sent in the slashing hook unit-tests

- Add acknowledgement boilerplates to the provider chain

- Allow slashing with 0 jail time

- Change the consumer `SendPacket` API to `SendPenalties`

TODO: slashing round-trip tests require a [custom IBC-GO](https://github.com/sainoe/ibc-go/tree/sainoe/valset-update-fixes). This will be sorted out after the #45 merge.


x/ccv/child/keeper/hooks.go
```diff
@@ -20,10 +20,16 @@ func (k Keeper) AfterValidatorDowntime(ctx sdk.Context, consAddr sdk.ConsAddress
...
 	}
 
+	// get last the valset update id
+	valsetUpdate, err := k.GetLastUnbondingPacketData(ctx)
+	if err != nil {
+		return
+	}
+
 	// increase jail time
 	signInfo.JailedUntil = ctx.BlockHeader().Time.Add(k.slashingKeeper.DowntimeJailDuration(ctx))
```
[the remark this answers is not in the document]
mpoke: Why does the consumer Slashing module need the `JailedUntil` to be set? AFAIK, it is only used to check whether a validator can unjail itself (i.e., handling a `MsgUnjail`). However, the consumer chain will have `MsgUnjail` disabled. 

If we need a variable `JailedUntil` for the CCV logic, then I'd keep it internal to CCV.

x/ccv/child/keeper/relay.go
```diff
@@ -42,7 +42,7 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, newCha
 	} else {
 		pendingChanges = utils.AccumulateChanges(currentChanges.ValidatorUpdates, newChanges.ValidatorUpdates)
 	}
-	k.SetPendingChanges(ctx, ccv.ValidatorSetChangePacketData{ValidatorUpdates: pendingChanges})
+	k.SetPendingChanges(ctx, ccv.ValidatorSetChangePacketData{ValidatorUpdates: pendingChanges, ValsetUpdateId: newChanges.ValsetUpdateId})
```
mpoke, line 45: I don't think we need the `ValsetUpdateId` for the pending changes. It's sufficient to just store a list of validator updates.

x/ccv/parent/keeper/relay.go
```diff
@@ -112,23 +112,23 @@ func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, dat
 // in a packet over the CCV channel.
 func (k Keeper) EndBlockCallback(ctx sdk.Context) {
 	valUpdateID := k.GetValidatorSetUpdateId(ctx)
-	// k.IncrementValidatorSetUpdateId(ctx) TODO: move this here
+	k.IncrementValidatorSetUpdateId(ctx)
```
mpoke, line 115: This needs to be moved after the loop. 

**Reason**:
Let `ValidatorSetUpdateId` be equal to `x`. During `DeliverTX`, the Staking module receives `Undelegate` messages, and as a result, it maps unbonding delegation entries to `x`. During `EndBlock`, the CCV module increments  `ValidatorSetUpdateId` to `x+1` and sends packets to the consumer chains. When the provider receives ACKs for these packets, it tries to match `x+1` to the unbonding delegation entries that were mapped to `x`.

x/ccv/parent/keeper/relay.go
```diff
@@ -112,23 +112,23 @@ func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, dat
...
 	k.IterateBabyChains(ctx, func(ctx sdk.Context, chainID string) (stop bool) {
 		valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)
 		if len(valUpdates) != 0 {
+			fmt.Println(valUpdateID)
 			k.SendPacket(ctx, chainID, valUpdates, valUpdateID)
-			k.IncrementValidatorSetUpdateId(ctx) // TODO: this needs to be moved out of this scope
 			k.SetValsetUpdateBlockHeight(ctx, valUpdateID, uint64(ctx.BlockHeight()))
```
mpoke, line 121: This also needs to be moved outside of the loop.

x/ccv/parent/keeper/relay.go
```diff
@@ -112,23 +112,23 @@ func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, dat
...
 }
 
-// OnRecvPacket slahes and jails the given validator in the packet data
+// OnRecvPacket slashes and jails the given validator in the packet data
 func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data ccv.ValidatorDowntimePacketData) exported.Acknowledgement {
-	if childChannel, ok := k.GetChannelToChain(ctx, packet.DestinationChannel); !ok && childChannel != packet.DestinationChannel {
+	if chainID, ok := k.GetChannelToChain(ctx, packet.DestinationChannel); !ok {
 		ack := channeltypes.NewErrorAcknowledgement(
-			fmt.Sprintf("packet sent on a channel %s other than the established child channel %s", packet.DestinationChannel, childChannel),
+			fmt.Sprintf("chain ID doesn't exist for channel ID: %s", chainID),
```
mpoke, line 131: If `ChannelToChain` is not set for a particular channel, it doesn't mean that the chain doesn't exist, but rather that the channel was not established, i.e., https://github.com/cosmos/interchain-security/blob/fc59e104ef25203f92d411d7b293d7c9a980d711/x/ccv/parent/keeper/keeper.go#L243.

x/ccv/parent/keeper/relay.go
```diff
@@ -112,23 +112,23 @@ func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, dat
...
 		)
 		chanCap, _ := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(packet.DestinationPort, packet.DestinationChannel))
 		k.channelKeeper.ChanCloseInit(ctx, packet.DestinationPort, packet.DestinationChannel, chanCap)
```
mpoke, line 134: The channel shouldn't be close if the provider chain's channel end is not established yet.

mpoke (changes requested):
Thanks for applying the changes. I'd ask @AdityaSripal and @jtremback for a review. See some extra comments bellow.

x/ccv/child/module.go
```diff
@@ -350,16 +350,49 @@ func (am AppModule) OnRecvPacket(
...
+			),
+		)
 	}
-	return nil, sdkerrors.Wrap(ccv.ErrInvalidChannelFlow, "cannot receive acknowledgement on child port, child chain does not send packet over channel.")
+	return &sdk.Result{
+		Events: ctx.EventManager().Events().ToABCIEvents(),
+	}, nil
 }
 
 // OnTimeoutPacket implements the IBCModule interface
```
mpoke, line 398: The error message bellow, i.e., 
> cannot timeout packet on child port, child chain does not send packet over channel

is no longer accurate.

x/ccv/child/keeper/hooks.go
```diff
@@ -32,15 +38,18 @@ func (k Keeper) AfterValidatorDowntime(ctx sdk.Context, consAddr sdk.ConsAddress
...
 		abci.Validator{
 			Address: consAddr.Bytes(),
 			Power:   power,
 		},
+		valsetUpdate.ValsetUpdateId,
 		k.slashingKeeper.SlashFractionDowntime(ctx).TruncateInt64(),
 		signInfo.JailedUntil.UnixNano(),
 	)
+
+	panic(err)
```
AdityaSripal, line 52: Why are you panicing here?

x/ccv/child/keeper/keeper_test.go
```diff
@@ -372,8 +372,13 @@ func TestKeeperTestSuite(t *testing.T) {
...
 	// initial setup
-	app := suite.childChain.App.(*app.App)
-	ctx := suite.childChain.GetContext()
+	suite.SetupCCVChannel()
+	app, ctx := suite.childChain.App.(*app.App), suite.ctx
+	channelID := suite.path.EndpointA.ChannelID
+
+	// add parent channel to childkeeper store
+	app.ChildKeeper.SetParentChannel(ctx, channelID)
+	app.IBCKeeper.ChannelKeeper.SetChannel(ctx, types.PortID, channelID, suite.path.EndpointB.GetChannel())
```
AdityaSripal, line 381: This line doesn't make sense. Why are you doing this here?

x/ccv/child/keeper/hooks.go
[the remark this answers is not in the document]
AdityaSripal: Agreed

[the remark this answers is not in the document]
sainoe: > Why does the consumer Slashing module need the JailedUntil to be set?

To know when it has already sent a slashing packet for a distinct validator. Otherwise, if the validator is still down during the consecutive blocks, the consumer would spam the provider with slashing packets. What would be useless and expensive.

x/ccv/parent/keeper/relay.go
sainoe: Good call thanks!

x/ccv/child/keeper/hooks.go
sainoe: Oh I missed to remove that one. It was for testing purpose. Thanks

x/ccv/child/keeper/hooks.go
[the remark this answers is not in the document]
mpoke: But the slashing module doesn't send slashing packets. The CCV module does.

x/ccv/child/keeper/keeper_test.go
sainoe: Agreed, but it works to not have my tests panicing with:
`test panicked: parent channel not set: channel not found`

I'm still digging  for the issue into the consumer suite. I would be glad to have your view on this.

x/ccv/child/keeper/hooks.go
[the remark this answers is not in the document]
jtremback: It seems to me that it would be pretty complicated and involve redundant state to have the CCV module keep track of the missed blocks and jailedUntil. I think that we should keep modifying this state, but do it in a neater way and not use the SetValidatorSigningInfo setter directly, instead having an method on slashing to ResetMissedBlocks that would handle the internal logic of the missed block array and associated variables. Then addressing JailedUntil, there already is a JailUntil method that sets that directly that can be used.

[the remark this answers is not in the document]
mpoke: >It seems to me that it would be pretty complicated and involve redundant state to have the CCV module keep track of the missed blocks and jailedUntil.

I don't think we should do that either. The consumer slashing module should keep doing the slashing logic (i.e., count the missing blocks and trigger a slash). Triggering a slash entails telling the consumer CCV module to send a SlashPacket to the provider CCV module. 

The `jailedUntil` state on the slashing module is not used on the **_consumer_** chain. It's used in the single-chain case to decide whether a validator can be unjailed. The `jailedUntil` that we use on the consumer chain is part of the CCV module logic, i.e., to avoid sending multiple SlashPackets for the same infraction (in case of downtime). What I'm saying is that **_if it is CCV module logic then it should be part of the CCV module state_**.

Issue #46, opened 2022-02-15
Provider genesis state validation 


mpoke:
The validation of a child state (within a provider chain genesis state) validates the chain ID as a client ID, see https://github.com/cosmos/interchain-security/blob/c4ce132b75087818873e769a828ccb4db36df74e/x/ccv/parent/types/genesis.go#L34

jtremback:
Could use some more context on this one

mpoke:
It's a minor thing. I would expect `ClientIdentifierValidator` should be used for validating client IDs and not chain IDs.

PR #73, opened 2022-04-28
Fix provider genesis state validation

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/46


mpoke:
Is there a better way to check for the validity of a chain ID besides checking whether the ID is an empty string? Is there a SDK function?

mpoke:
Closed and replaced by https://github.com/cosmos/interchain-security/pull/85

PR #76, opened 2022-04-29
Create single send function

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/57

PR #78, opened 2022-04-29
Fix error running the tests

sainoe:
Fix error caused by importing  ibc-go/v3/modules/core/24-host in ccv/utils.go


mpoke:
Thanks @sainoe. I manage to fix it in another way though. The problem was that I was not importing the correct version of ibc-go for the host, see https://github.com/cosmos/interchain-security/pull/76/commits/ea6e485871b41e5695afd91d8051afb86f950163.

Issue #57, opened 2022-04-21
Create single send function


mpoke:
The code needed to send a packet is duplicated throughout the implementation, e.g., [here](https://github.com/cosmos/interchain-security/blob/df6f8c567a4a62e91b031a7e308db06a094f5aaf/x/ccv/parent/keeper/relay.go#L27) and [here](https://github.com/cosmos/interchain-security/blob/df6f8c567a4a62e91b031a7e308db06a094f5aaf/x/ccv/child/keeper/relay.go#L111). We should have one single function for sending packets, e.g.,
```go
func (k Keeper) SendPacket(ctx sdk.Context, channelID string, packetData []byte) error {
  channel, ok := k.channelKeeper.GetChannel(ctx, types.PortID, channelID)
  if !ok {
    return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "channel not found for channel ID: %s", channelID)
  }
  channelCap, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(types.PortID, channelID))
  if !ok {
    return sdkerrors.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
  }

  // get the next sequence
  sequence, found := k.channelKeeper.GetNextSequenceSend(ctx, types.PortID, channelID)
  if !found {
    return sdkerrors.Wrapf(
	channeltypes.ErrSequenceSendNotFound,
	"source port: %s, source channel: %s", types.PortID, channelID,
    )
  }
  packet := channeltypes.NewPacket(
    packetData, sequence,
    types.PortID, channelID,
    channel.Counterparty.PortId, channel.Counterparty.ChannelId,
    clienttypes.Height{}, uint64(ccv.GetTimeoutTimestamp(ctx.BlockTime()).UnixNano()),
  )
  if err := k.channelKeeper.SendPacket(ctx, channelCap, packet); err != nil {
    return err
  }
  return nil
```

mpoke:
[...]

PR #75, opened 2022-04-28
Naming Updates

rigelrozanski:
child/baby->consumer, parent->provider renames

PR #86, opened 2022-05-04
Create single send function

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/57

Instead of https://github.com/cosmos/interchain-security/pull/76 due to the renaming from https://github.com/cosmos/interchain-security/pull/75


danwt (approved):
LGTM

Issue #66, opened 2022-04-26
Create consumer governance suite


jtremback:

jtremback:
@rigelrozanski has done some work already:

> I'm thinking that ccv governance is as simple as disabling the endblock in staking (which is what provides that tendermint validator power updates), and changing the ("normal") distribution address from the feepool address to the consumer-redistribute address (or another dev-specified address depending on the blockchain) 

> We really just need to separate out the consumer/provider apps at this point though, because they absolutely can't be the same for governance then you'd obviously want to reparamaterize staking too

>Also we should probably remove the slashing hooks from staking (app.go)

> all possible from app.go (although a bit hacky to remove the staking endblocker from the module manager in the current design... it can be done)

zmanian:
Does the democracy module contain the idea of consumer chain governance tokens?

Does it handle minting or locking of consumer chain governance tokens?

okwme:
notes from the call last week:

    * Should the “democracy module” be part of the spec? I find it outside of the scope of CCV and very cosmos-sdk related. 
    * 3 areas
        * Remove ABCI communication from staking (InitGenesis & EndBlock)
        * Name collision?
        * Overlap with slashing / ccv slashing? Slashing dependency?
        * Bez’s past modified behavior module
            * Imported the module directly, create new module that has the original module embedded with overridden methods
            * Methods overridden must be public
            * End blocker needed to be a no-op, import staking and have it do nothing
            * Private things can’t be overridden
        * Don’t want to overwrite the module name so it has consistent
        * New repo that is a new module that imports and overwrites InitGenesis & EndBlock
            * Has go.mod
            * Actually make it a submodule in interchain security

alexanderbez:
My summary for what consider this "feature complete" is as follows:

- Create `x/ccv/staking` package/module that a consumer chain will use
- This `x/ccv/staking` module will wrap (i.e. embed) the SDK's `x/staking` module with one major difference, it's `EndBlock` is a no-op. This makes it so that's it only about a few lines of code.

okwme:
and the initGenesis also no longer communicates with Tendermint

mpoke:
Finally, it should be possible, when deploying a consumer chain, to set the `ConsumerRedistributeName` (the account that holds the consumer part of the rewards for producing blocks) to the distribution module account. @rigelrozanski Is `ConsumerRedistributeName` a parameter or is it hardcoded?

alexanderbez:
[...]

PR #89, opened 2022-05-05
feature: support "democracy" module

alexanderbez:
- Update min supported go version to 1.17
- Introduce `x/ccv/staking` module that wraps the SDK's core `x/staking` module
  - `x/ccv/staking` performs identical to the SDK's `x/staking` in that it directly embeds the module, except it overrides `InitGenesis` and `EndBlock` to become no-ops.
- Update `app.go` (as if it were the consumer chain)

TODO:

- [ ] Determine if we want `BeginBlock`  to be a no-op as well?

closes: #66


x/ccv/staking/module.go
```diff
@@ -0,0 +1,53 @@
...
+// AppModule embeds the Cosmos SDK's x/staking AppModule where we only override
+// specific methods.
+type AppModule struct {
+	// embed the Cosmos SDK's x/staking AppModule
+	staking.AppModule
+}
+
+// NewAppModule creates a new AppModule object using the native x/staking module
+// AppModule constructor.
+func NewAppModule(cdc codec.Codec, keeper keeper.Keeper, ak types.AccountKeeper, bk types.BankKeeper) AppModule {
```
marbar3778, line 35: if possible could you override the module keyname here?

alexanderbez: The module name? I _think_ so, but why would you want to anyway?

alexanderbez: A serious problem would arise if **_other_** modules depend on the module name, e.g. `bk.SendFrom*ToModule(..., staking.ModuleName)`.

x/ccv/staking/doc.go
```diff
@@ -0,0 +1,12 @@
...
+x/staking module. In other words, it provides the exact same functionality as
+the native module in that it simply embeds the native module. However, it
+overrides two core methods, `InitGenesis` and `EndBlock`. Specifically, these
+methods perform no-ops and return no validator set updates, as validator sets
+are tracked by the consumer chain's x/ccv/consumer module.
+
+The consumer chain should utilize the x/ccv/staking module to perform democratic
+actions such as participating and voting within the chain's governance system.
+*/
+package staking
```
marbar3778, line 12: should we rename this? its as confusing as tendermint the algorithm,  tendermint the software, tendermint the company

alexanderbez: I thought about this, but see my comment: https://github.com/cosmos/interchain-security/pull/89#discussion_r866852714

x/ccv/staking/module.go
```diff
@@ -0,0 +1,53 @@
...
+	stakingAppMod := staking.NewAppModule(cdc, keeper, ak, bk)
+	return AppModule{stakingAppMod}
+}
+
+// InitGenesis performs a no-op in that it returns no validator set updates.
+// Recall, validators are tracked via the consumer chain's x/cvv/consumer module
+// and so this module is not responsible for returning the initial validator set.
+func (am AppModule) InitGenesis(ctx sdk.Context, cdc codec.JSONCodec, data json.RawMessage) []abci.ValidatorUpdate {
```
mpoke, line 43: I think `InitGenesis` should still execute the logic from the original staking module (x/staking), since that initializes the internal state based on the genesis state. Once the logic is executed, the function should indeed return an empty validator set.

x/ccv/staking/module.go
```diff
@@ -0,0 +1,53 @@
...
+	return []abci.ValidatorUpdate{}
+}
+
+// EndBlock performs a no-op in that it returns no validator set updates. Recall,
+// validators are tracked via the consumer chain's x/cvv/consumer module. Validators
+// contained within this module act as "pseudo" validators, in that they may be
+// used in the consumer chain's governance system.
+func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
```
mpoke, line 51: Similarly here, `EndBlock` should execute parts of the logic from the original staking module, i.e., the completion of unbonding operations (see [here](https://github.com/cosmos/cosmos-sdk/blob/92f7df3c9c91c3998558bd72e0fd3c9cb750f09f/x/staking/keeper/val_state_change.go#L17)).  

I'm not sure whether `ApplyAndReturnValidatorSetUpdates` is necessary.  We do not need a validator set, but the logic to move validators through different states (i.e., bonded, unbonding, unbonded) is still necessary. To my understanding, the democracy module will enable delegators of the native token to bond their tokens in order to get the right to participate in  governance. In addition, all validators with bonded tokens can participate in governance, not only the first e.g., `175`.

x/ccv/staking/module.go
mpoke: I'm wondering whether this name collision could be an issue on the consumer chain, since the slashing module replaces the staking module with the CCV module, while other modules, like governance, with still use the actual staking module (or more precisely, this modified version of it). 

@sainoe What do you think?

mpoke (changes requested):
Nice work @alexanderbez. See my comments below. Also, we should probably rebase to incorporate the changes with the two app.go files.

mpoke:
> Determine if we want BeginBlock to be a no-op as well?

Do we need to track the historical info? I'm not sure for is this used in the SDK. But I don't think it's needed since it consist of a list of headers with the corresponding validator set. Do we need to know who was able to vote at a given height?

x/ccv/staking/module.go
alexanderbez: Sounds good. I suppose I had a misunderstanding in what needed to be done here.

alexanderbez:
> Do we need to track the historical info? I'm not sure for is this used in the SDK. But I don't think it's needed since it consist of a list of headers with the corresponding validator set. Do we need to know who was able to vote at a given height?

@AdityaSripal do you think we need historical info? IIRC, this is for IBC related things, no?

x/ccv/staking/module.go
```diff
@@ -0,0 +1,56 @@
...
+func (am AppModule) InitGenesis(ctx sdk.Context, cdc codec.JSONCodec, data json.RawMessage) []abci.ValidatorUpdate {
+	return []abci.ValidatorUpdate{}
+}
+
+// EndBlock performs a no-op in that it returns no validator set updates. Recall,
+// validators are tracked via the consumer chain's x/cvv/consumer module. Validators
+// contained within this module act as "pseudo" validators, in that they may be
+// used in the consumer chain's governance system.
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	_ = am.keeper.BlockValidatorUpdates(ctx)
```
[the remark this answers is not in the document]
mpoke: If we set `MaxValidators` to max int, the following loop
```
for count := 0; iterator.Valid() && count < int(maxValidators); iterator.Next() {
}
```
will exit once it goes through all available validator, right?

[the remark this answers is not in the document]
mpoke: And `MaxValidators` is a chain param, so we might not have to do anything, right?

mpoke (approved):
Great work.

x/ccv/staking/module.go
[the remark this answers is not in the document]
alexanderbez: 1. Yes, `MaxValidators` is an on-chain param, modify-able by governance.
2. Yes, you can set it to `MaxInt`.

mpoke:
> > Do we need to track the historical info? I'm not sure for is this used in the SDK. But I don't think it's needed since it consist of a list of headers with the corresponding validator set. Do we need to know who was able to vote at a given height?
> 
> @AdityaSripal do you think we need historical info? IIRC, this is for IBC related things, no?

I think we should make the BeginBlock also a no-op. The IBC module on the consumer chain needs to get the historical info from the CCV module that keeps track of the validator set that produces blocks on the consumer. The "validators" in this staking module are not producing blocks and IBC should not have anything to do with them.

rigelrozanski:
Just added some quick non-compiling code to `frog/wrapped-distr` branch of what the distribution should probably look like. Should the wrapped distribution module also go in this PR? Feel free to snag the code in that branch (see x/distribution)

mpoke:
> Just added some quick non-compiling code to `frog/wrapped-distr` branch of what the distribution should probably look like -pretty straight forward, just need to modify the Allocation function to ignore everything proposer reward related. Should the wrapped distribution module also go in this PR? Feel free to snag the code in that branch (see x/distribution)

@rigelrozanski I think a separate PR will be better. Could you please open one with the changes from `frog/wrapped-distr`?

Issue #65, opened 2022-04-26
Double sign slashing


mpoke:
Implement Consumer Initiated Slashing for double-signing evidence (see [spec](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#consumer-initiated-slashing)).

zmanian:
Has a decision been made that double sign slashing evidence will be initiated by the consumer chain and evaluated by the consumer's tendermint instance?

The alternative design would be to extend the IBC client misbehavior handling mechanism to enable handling proposal, prevote and precommit misbehavior so that evidence is evaluated on the provider chain

PR #84, opened 2022-05-03
Support different app.go's for consumer and provider

danwt:
## Issues

Mostly completes

- https://github.com/cosmos/interchain-security/issues/58

Related to

- https://github.com/cosmos/interchain-security/issues/59
- https://github.com/cosmos/interchain-security/issues/77

## Overview

There is a goal to split provider/consumer apps so that people can build from a minimal viable consumer chain. This PR is paired with [another PR in informalsystems/ibc-go](https://github.com/informalsystems/ibc-go/pull/1). The two PR's together

1. Split the app.go into two app.go's. One for parent and child.
2. Modify ibc-go testing to be able to use two different app.go's.

Therefore this PR should be merged together with 

- https://github.com/informalsystems/ibc-go/pull/1

Note: there are two app.go's but they are _identical_.

PR #97, opened 2022-05-12
Add double-sign slashing

sainoe:
**Close** #65

* Add double-sign slashing to CCV following the [spec](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#consumer-initiated-slashing)

* Execute slashing and jailing operations using the provider chain parameters

**Implementations details:**
* Evidence and Slashing modules pass an `InfractionType` enum to CCV in ways it can handles the double-sign and downtime slashing logics

* Evidence module instantiated using the consumer keeper in [/consumer/app.go](https://github.com/cosmos/interchain-security/blob/ea1c6d5fb05a368cd7dfc2bbcc75cc15a86196b1/app/consumer/app.go#L540)


app/provider/app.go
```diff
@@ -477,18 +477,6 @@ func New(
 		authtypes.FeeCollectorName,
 	)
 	providerModule := ibcprovider.NewAppModule(&app.ProviderKeeper)
-
```
sainoe, line 480: @danwt I should have catch this while reviewing #84

danwt: Hey, I'm not exactly sure what you mean. Can you elaborate a little?

mpoke (approved):
Great work @sainoe. Just one comment regarding jailing on the provider.

okwme (approved):

x/ccv/provider/keeper/relay.go
```diff
@@ -147,47 +149,73 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data c
...
+	if !found || validator.IsUnbonded() {
 		return fmt.Errorf("should not be slashing unbonded validator: %s", validator.GetOperator())
 	}
 
-	// slash and jail the validator
-	k.stakingKeeper.Slash(ctx, consAddr, int64(infractionHeight), downtimeData.Validator.Power, sdk.NewDec(1).QuoInt64(downtimeData.SlashFraction))
-	if !validator.IsJailed() {
-		k.stakingKeeper.Jail(ctx, consAddr)
+	// spare jailed and/or tombstoned validator preventing to slash it again
+	if validator.IsJailed() || k.slashingKeeper.IsTombstoned(ctx, consAddr) {
```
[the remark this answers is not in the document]
okwme: good catch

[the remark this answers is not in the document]
sainoe: Ok thanks. I will change that once I get all reviews :)

x/ccv/consumer/keeper/relay.go
```diff
@@ -155,8 +157,10 @@ func (k Keeper) SendSlashPacket(ctx sdk.Context, validator abci.Validator, valse
 		panic(err)
```
danwt, line 157: Can it be replaced with sendIBCPacket util introduced in https://github.com/cosmos/interchain-security/pull/86?

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -197,3 +205,243 @@ func (suite *KeeperTestSuite) TestInitHeight() {
...
 	}
 }
+
+func (suite *KeeperTestSuite) TestHandleSlashPacketDistribution() {
+	providerStakingKeeper := suite.providerChain.App.GetStakingKeeper()
+	providerSlashingKeeper := suite.providerChain.App.(*appProvider.App).SlashingKeeper
+	ProviderKeeper := suite.providerChain.App.(*appProvider.App).ProviderKeeper
+
+	// bonded amount
+	bondAmt := sdk.NewInt(1000000)
```
danwt, line 215: This is fine, but overall in the tests we should try to keep magic numbers a bit under control. For example I think this is the DefaultPowerReduction so better to use that. 

I made an issue as this is a widespread phenomena in the existing code. I'm not singling this PR, or you, out.

x/ccv/provider/keeper/relay.go
```diff
@@ -147,47 +149,73 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data c
...
+		slashFraction = k.slashingKeeper.SlashFractionDowntime(ctx)
+		jailTime = ctx.BlockTime().Add(k.slashingKeeper.DowntimeJailDuration(ctx))
+		k.AppendSlashAck(ctx, chainID, consAddr.String())
+	// set double-signing slash fraction and infinite jail duration
+	// then tombstone the validator
+	case stakingtypes.DoubleSign:
+		slashFraction = k.slashingKeeper.SlashFractionDoubleSign(ctx)
+		jailTime = evidencetypes.DoubleSignJailEndTime
+		k.slashingKeeper.Tombstone(ctx, consAddr)
+	default:
```
danwt, line 202: I'm probably misunderstanding the protocol, but do you need to also send the ack in case of DoubleSign slash?

danwt:
Great work!

x/ccv/consumer/keeper/relay.go
sainoe: Yea sure. Thanks for pointing this out.

danwt (approved):

x/ccv/provider/keeper/relay.go
sainoe: `SlashAck`s prevent that a validator gets slashed multiple times on the provider chain for having downtime. This scenario can happen in CCV as pointed out in this [issue](https://github.com/cosmos/ibc/issues/685). You can also check the `SendSlashRequest`'s logic in the [ICS 26 spec],(https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#ccv-pcf-rcvslash1) where the consumer chains use `outstandingDowntime` to record and track the expected slash acknowledgments.

@mpoke [onRecvSlashPacket](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#ccv-pcf-rcvslash1) needs to be updated in that regard.

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -157,20 +165,20 @@ func (suite *KeeperTestSuite) TestSlashAcks() {
 	suite.Require().Len(chainsAcks, len(chains))
 }
 
-func (suite *KeeperTestSuite) TestAppendslashingAck() {
+func (suite *KeeperTestSuite) TestAppendSlashgAck() {
```
AdityaSripal, line 168: misspelled

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -197,3 +205,243 @@ func (suite *KeeperTestSuite) TestInitHeight() {
...
+
+	// save valset update ID mapping the next block height
+	valseUpdateID1 := ProviderKeeper.GetValidatorSetUpdateId(suite.providerChain.GetContext())
+
+	// get valset update ID mapping the current block height
+	valseUpdateID0 := valseUpdateID1 - 1
+
+	// create first undelegation entry
+	ubdBalance := ubdAmount.Mul(bondAmt.ToDec()).TruncateInt()
+	ubd := unboundHalf()
```
AdityaSripal, line 251: misspelled

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -197,3 +205,243 @@ func (suite *KeeperTestSuite) TestInitHeight() {
 		suite.Require().EqualValues(t.expected, height)
 	}
 }
+
+func (suite *KeeperTestSuite) TestHandleSlashPacketDistribution() {
```
AdityaSripal, line 209: generally in favor of more scoped module table-driven tests

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -197,3 +205,243 @@ func (suite *KeeperTestSuite) TestInitHeight() {
...
+		slashingPkt.ValsetUpdateId = t.valsetUpdateID
+
+		// slash
+		err := ProviderKeeper.HandleSlashPacket(suite.providerChain.GetContext(), suite.consumerChain.ChainID, slashingPkt)
+		suite.Require().NoError(err)
+
+		// check that second undelegation was slashed
+		ubd = undel()
+
+		suite.Require().EqualValues(t.expBalances[0], ubd.Entries[0].Balance)
```
AdityaSripal, line 320: can we also check that someone who is still bonded also gets slashed

x/ccv/consumer/keeper/relay.go
```diff
@@ -103,24 +104,25 @@ func (k Keeper) UnbondMaturePackets(ctx sdk.Context) error {
...
+	packetData := ccv.NewSlashPacketData(validator, valsetUpdateID, infraction)
 
 	// check that provider channel is established
 	// if not, append slashing packet to pending slash requests
 	channelID, ok := k.GetProviderChannel(ctx)
 	if !ok {
 		k.AppendPendingSlashRequests(ctx, types.SlashRequest{
 			Packet:   &packetData,
-			Downtime: true}, // TODO Simon: add double-signing check
+			Downtime: downtime},
```
AdityaSripal, line 125: It's odd here to use a boolean to determine what type of slash it is, where everywhere else you use an enum.

I suggest that this gets changed to be the enum value as well.

AdityaSripal (commented):
In general looks good, see comments

x/ccv/provider/keeper/keeper_test.go
danwt: Just curious, what do you mean by 'more scoped module'? (The table part I got).

x/ccv/provider/keeper/relay.go
mpoke: > onRecvSlashPacket needs to be updated in that regard.
@sainoe Do you mean calling `AppendSlashAck` only for downtime slashing?

mpoke: > You should still be sending acks even if you don't perform logic on them so sending chain can clean up their state

@AdityaSripal These SlashAcks here are not IBC Packet Acknowlegdments, but rather some info sent by the provider to the consumer (in the VSC packet) notifying it that the Slash requests for downtime were processed.

sainoe: > @sainoe Do you mean calling AppendSlashAck only for downtime slashing?`

@mpoke Yes!

x/ccv/provider/keeper/keeper_test.go
AdityaSripal: Sorry typo. Should just be `more scoped`, so more unit-testy is preferred

x/ccv/provider/keeper/relay.go
mpoke: > Ahh ok, so double sign packets still get an acknowledgement?

Yes. 

> I assume the default acknowledgment is used?

Cannot use them because of the order with VSC packets.

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -390,7 +394,54 @@ func (suite *KeeperTestSuite) TestValidatorDowntime() {
...
+	// save next sequence before sending a slash packet
+	seq, ok := app.GetIBCKeeper().ChannelKeeper.GetNextSequenceSend(ctx, types.PortID, channelID)
+	suite.Require().True(ok)
+
+	// handles double-signing evidence
+	app.EvidenceKeeper.HandleEquivocationEvidence(ctx, e)
+
+	// check that slash packet is sent
+	commit := app.IBCKeeper.ChannelKeeper.GetPacketCommitment(ctx, types.PortID, channelID, seq)
+	suite.NotNil(commit)
```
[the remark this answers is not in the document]
sainoe: Agree. Now downtime and double-sign tests check the commit value using a reconstructed packet's hash.

x/ccv/consumer/keeper/relay.go
sainoe: Agree. `PendingSlashRequest` msg definition updated accordingly.

Issue #58, opened 2022-04-21
Create example of app.go for users, test in real configuration, fix IBC-Go test suite to handle it


mpoke:
- [x] branch ibc-go and make necessary changes
- [x] fork out the ibc-go test framework and make necessary changes
  - postponed
- [x] create non-minimal example app.go for consumer
- [x] create non-minimal example app.go for provider
- [x] create minimal example app.go for consumer
  - fixed by https://github.com/cosmos/interchain-security/pull/98
- [x] create minimal example app.go for provider
  - fixed by https://github.com/cosmos/interchain-security/pull/98
- [x] make sure the apps are tested in the unit tests using ibc-go/testing
- [x] test the two apps using the docker integration tests
- [x] add comment in app.go addressing the order of `EndBlock` (see https://github.com/cosmos/interchain-security/issues/51#issuecomment-1125143988)
  - moved to https://github.com/cosmos/interchain-security/pull/98

jtremback:
In the short term, there will be 3 different users of our `app.go` example. 

- First, the Cosmos Hub team at IG will need to install the CCV provider module into the Cosmos Hub's `app.go`.
- Then the Ethernal team will need to install the CCV consumer module into the CosmWasm consumer chain. 
- Also, teams with custom consumer chains will need to install the CCV consumer module as well.

I think it will be important have separate app.go files for provider and consumer chains. This is because the functioning of the consumer chain depends on certain modules being disabled. Even if we fixed this somehow, we would still want separate `app.go`s because the Hub will not want to pull in the consumer module and increase the attack surface.

For this to happen some changes will need to be made to the IBC-Go test suite to allow it to test two chains with separate `app.go`s.

As I just discussed with @danwt I think the time has come for us to fork the IBC-Go test framework into it's own repo. Dealing with a test framework that needs frequent modifications but also is bundled with one of our core dependencies is a pain.

danwt:
> fork the IBC-Go test framework into it's own repo.

I think it's a good idea. The ultimate goal would be to have a general tool for testing ibc apps across heterogeneous chains so the name could be something like 'ibc-tester'. It must be built for our needs first though, without premature generalization.

danwt:
- [ ] fork ibc-go test framework and make necessary changes
- [ ] create example app.go for consumer
- [ ] create example app.go for provider
- [ ] test the two chains in a real configuration using the improved ibc-go testing

danwt:
[...]

PR #98, opened 2022-05-12
remove unnecessary parts of app.go in consumer and provider

okwme:
In removing unnecessary parts of the two app.go files it became apparent that there was still a dependence in the Consumer Chain on the staking module. Consumer chain should not have a staking module (at least not until democracy is added back with the `ccv/staking` module in issue #89 ). It uses the Consumer Module to handle all staking module duties. As such it fulfills the staking module interface and works with slashing. However there are other modules on a consumer chain that consume the staking module and are not happy when you swap in the consumer module instead.

- [x] `IBCKeeper` can't accept `app.ConsumerKeeper` in place of `app.StakingKeeper` because it's missing the method `GetHistoricalInfo`
- [x] `app.ConsumerKeeper` is missing `ApplyAndReturnValidatorSetUpdates` in order to be used by `genutil`
  - `genutil` module should be removed from the consumer chain; the init val set comes always from the CCV module (see `InitGenesis`)
- [x] fails when adding to `slashing.NewAppModule()` without reason listed
- [x] fails when adding to `distr.NewAppModule` without reason listed
- [ ] fails as return value in `GetStakingKeeper`
- [x] fails within `prepForZeroHeightGenesis` while trying to use the `GetAllDelegations`, `IterateRedelegations`, `SetRedelegation`, `IterateUnbondingDelegations`,  `SetUnbondingDelegation`, `GetValidator`, `SetValidator` and `ApplyAndReturnValidatorSetUpdates` methods of `stakingKeeper` (these may be unnecessary, do we need to export genesis on a child chain?)

Some of these changes are already present in @sainoe 's #97 (which relies on https://github.com/cosmos/cosmos-sdk/pull/11921).

this also removes the liquidity module from the child chain as it should not be considered part of a standard application. It could stay on the provider chain since gaia has it but should also probably just be removed.

Closes https://github.com/cosmos/interchain-security/issues/58


mpoke:
> `IBCKeeper` can't accept `app.ConsumerKeeper` in place of `app.StakingKeeper` because it's missing the method `GetHistoricalInfo`

I think we can add the `TrackHistoricalInfo()` method from [staking](https://github.com/cosmos/cosmos-sdk/blob/7a31a28e243c61c42d344d42ed70361f0349e91b/x/staking/keeper/historical_info.go#L68) to the consumer ccv module. This would enable the `IBCKeeper` to use the  `app.ConsumerKeeper` for calling the `GetHistoricalInfo` method.

mpoke:
> `app.ConsumerKeeper` is missing `ApplyAndReturnValidatorSetUpdates` in order to be used by `genutil`

Do we need the `genutil` module on the consumer? I don't know exactly what's the purpose of it.

okwme:
@mpoke , genutil is used in creating genesis files for the blockchain afaiu.
The answer for whether we can remove it would be the same for whether we need to make it work with `prepForZeroHeightGenesis` as part of the genesis export.

I think it's probably important to keep. It's necessary for a certain type of network upgrade and is also valuable for data since you can essentially dump the entire state of the blockchain with that command. We should try to preserve it.

mpoke:
> @mpoke , genutil is used in creating genesis files for the blockchain afaiu. The answer for whether we can remove it would be the same for whether we need to make it work with `prepForZeroHeightGenesis` as part of the genesis export.
> 
> I think it's probably important to keep. It's necessary for a certain type of network upgrade and is also valuable for data since you can essentially dump the entire state of the blockchain with that command. We should try to preserve it.

Strange though, this means that the `InitGenesis` of both `genutil` and `staking` return a validator set (see [here](https://github.com/cosmos/cosmos-sdk/blob/7a31a28e243c61c42d344d42ed70361f0349e91b/x/genutil/module.go#L99) and [here](https://github.com/cosmos/cosmos-sdk/blob/7a31a28e243c61c42d344d42ed70361f0349e91b/x/staking/module.go#L152)). Both of them are calling `ApplyAndReturnValidatorSetUpdates`. In the case of the consumer chain, the `InitGenesis` of the CCV module sends the initial validator set to Tendermint.

Still I don't know why `DeliverGenTxs` must return the validator set to Tendermint, i.e.,
```
// DeliverGenTxs iterates over all genesis txs, decodes each into a Tx and
// invokes the provided deliverTxfn with the decoded Tx. It returns the result
// of the staking module's ApplyAndReturnValidatorSetUpdates.
```

rigelrozanski:
Overall this PR looks good to me. Only thing I'll mention is that we should also remove this testing function `DisableConsumerDistribution ` (https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/provider_test.go#L614) 

It's nice to keep the ModuleManager (MM) exposed within the app.go on the provider chain so we can disable distribution on the provider chain to check consumer funds received on the provider chain.

okwme:
ok so the current problem has to do with making all the places in the consumer module that looks for `stakingKeeper` become replaced with `consumerKeeper` since that is where our staking is coming from. I was able to add new methods, some of them that needed to actually do something and some that were just no-ops to satisfy most of them.

The issue now is that the IBC testing framework expects a real functioning staking keeper in order to run.

One solution was to try implementing all of the methods within the real `stakingKeeper` as part of `consumerKeeper`. I added all of them that were missing to https://github.com/cosmos/interchain-security/blob/okwme/app-specific/x/ccv/consumer/keeper/no-ops.go. In order to confirm whether the `consumerKeeper` could be used as a `stakingKeeper` I made this test: https://github.com/cosmos/interchain-security/blob/okwme/app-specific/x/ccv/consumer/keeper/no-ops_test.go. This threw an error though, I'm not sure if it's possible or what's missing to make the `consumerKeeper` interchangeable with the `stakingKeeper`. 

So instead I went back and realized that the part of the `app.go` that wants a real `stakingKeeper` could just be ignored like:
https://github.com/cosmos/interchain-security/blob/okwme/app-specific/app/consumer/app.go#L830-L833
```
// GetStakingKeeper implements the TestingApp interface.
func (app *App) GetStakingKeeper() stakingkeeper.Keeper {
	return stakingkeeper.Keeper{}
}
```
since `GetStakingKeeper` isn't called anywhere in the consumer chain.

However this breaks all of the IBC test framework we're using for testing Interchain Security because the IBC test framework wants to generate a chain with a real validator set.


Seems like even if we could get `stakingKeeper` and `consumerKeeper` interchangeable we still won't be able to use the IBC test framework because it has no concept of a Consumer Chain with this limited staking functionality.

Not sure how to proceed, anyone have ideas?
@alexanderbez @rigelrozanski @jtremback ?

x/ccv/consumer/keeper/keeper.go
```diff
@@ -26,7 +30,7 @@ import (
 type Keeper struct {
 	storeKey          sdk.StoreKey
 	cdc               codec.BinaryCodec
-	paramSpace        paramtypes.Subspace
+	paramstore        paramtypes.Subspace
```
okwme, line 33: `paramSpace` and `accountKeeper` were renamed to match the stakingKeeper's naming conventions

x/ccv/consumer/keeper/keeper.go
```diff
@@ -359,6 +369,72 @@ func (k Keeper) GetHeightValsetUpdateID(ctx sdk.Context, height uint64) uint64 {
 	return binary.BigEndian.Uint64(bz)
 }
 
+// TrackHistoricalInfo saves the latest historical-info and deletes the oldest
+// heights that are below pruning height
+func (k Keeper) TrackHistoricalInfo(ctx sdk.Context) {
+	entryNum := 4 * 7 * 24 * 60 * 60 // 4 weeks (TODO: confirm safe limit)
```
okwme, line 375: should this use `types.UnbondingTime` ?

x/ccv/consumer/module.go
```diff
@@ -155,6 +155,8 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 	blockHeight := uint64(ctx.BlockHeight())
 	vID := am.keeper.GetHeightValsetUpdateID(ctx, blockHeight)
 	am.keeper.SetHeightValsetUpdateID(ctx, blockHeight+1, vID)
+	am.keeper.TrackHistoricalInfo(ctx)
```
okwme, line 158: this allows us to satisfy the requirements of IBC in opening new connections from the consumer chain. We're now storing the val set at each block as well as the val set updates.

okwme:
also should note that I merged in @sainoe 's PR (https://github.com/cosmos/interchain-security/pull/97) to this one to resolve some of the other interface issues.

his work is mostly present in this PR within `x/ccv/consumer/keeper/keeper_test.go`, `x/ccv/consumer/keeper/relay.go`, `x/ccv/consumer/keeper/relay_test.go`, `x/ccv/consumer/keeper/validators.go`,  `x/ccv/provider/keeper/keeper.go`, `x/ccv/provider/keeper/keeper_test.go`, `x/ccv/provider/keeper/relay.go`, `x/ccv/provider/provider_test.go`, `x/ccv/types/ccv.go`, `x/ccv/types/expected_keepers.go` and should mostly be ignored as part of this PR.

marbar3778:
with https://github.com/cosmos/cosmos-sdk/pull/12181 being merged I think we can close this PR since the reliance on staking the struct is no longer there.

Issue #54, opened 2022-04-04
Create doc with info on how to start a consumer chain


mpoke:
For both CosmWasm & Cosmos-SDK chains

jtremback:
@stana-ethernal and @dusan-ethernal have accomplished both of these things: 

https://github.com/informalsystems/consumer-chain-tool
https://github.com/informalsystems/wasmd

PR #102, opened 2022-05-13
ConsumerRedistributeFrac now hardcoded

### LINKED ISSUE #81: Decide whether the consumer can update ConsumerRedistributeFrac
Currently, `ConsumerRedistributeFrac`, i.e., the fraction of tokens allocated to the consumer redistribution address during distribution events, which means it could be modified by the consumer through governance. I'd argue that the consumer should not be able to change this parameter without the consent of the provider chain. Thus, for now, we should add this as a field of the `CreateChildChainProposal`. Later we could allow this to be change by submitting a TX to the provider or through an IBC packet.  

After discussions, the preferred solution is to keep the `ConsumerRedistributeFrac` hardcoded in the consumer chain. As a result, the validators on the provider chain can check the value when auditing the consumer binary, and the consumer chain cannot update it through governance.

rigelrozanski:
Closes #81


mpoke (approved):
LGTM. Nice work @rigelrozanski

x/ccv/consumer/keeper/distribution.go
```diff
@@ -16,6 +16,11 @@ import (
 
 const TransferTimeDelay = 1 * 7 * 24 * time.Hour // 1 weeks
 
+// The fraction of tokens allocated to the consumer redistribution address
+// during distribution events. The fraction is a string representing a
+// decimal number. For example "0.5" would represent 50%.
+const ConsumerRedistributeFrac = "0.5"
```
mpoke, line 22: @jtremback Do you think we should set this to 0.75 as the default value?

rigelrozanski: So one thing to mention is that because this is a const it's "testing unfriendly" basically we have to use whatever is in here as the testing value (50% is simple to reason about). This being said, previously we'd set the default to 0% but we need a value higher than 0% to test functionality properly... so yeah, we should probably just have notes about parameterizing this value in some form of "master" guide for those who want to use CCV

mpoke: I agree. We need documentation on how to bootstrap a consumer chain (there is an open issue for this https://github.com/cosmos/interchain-security/issues/54).

mpoke:
```suggestion
const ConsumerRedistributeFrac = "0.75"
```

mpoke, line 22:
```suggestion
// decimal number. For example "0.75" would represent 75%.
 const ConsumerRedistributeFrac = "0.75"
```

AdityaSripal (approved):

PR #108, opened 2022-05-19
make proto generation work with buf

marbar3778:


marbar3778:
the failing test is fixed in #109

mpoke (approved):
LGTM. Thanks @marbar3778

Issue #74, opened 2022-04-28
Enable user txs only after CCV channel enabled


mpoke:
For safety, the consumer chain should not enable user transactions until the CCV channel is established. At this point the consumer is secured by the provider. The CCV channel is established on the consumer chain when the first VSC packet is received, see [spec](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#ccv-ccf-rcvvsc1).

jackzampolin:
So architecture wise this would be an antehandler that rejects all non-CCV messages prior to a param being flipped?

mpoke:
> So architecture wise this would be an antehandler that rejects all non-CCV messages prior to a param being flipped?

Yes, something like this. I think it shouldn't be too hard to implement, I just don't know where this needs to be done and how to get access to the CCV flag (probably through a ccvKeeper).

alexanderbez:
Assign this to me, I'll knock it out.

alexanderbez:
@mpoke I was going to introduce a new parameter to the `x/ccv/consumer` module, but I noticed there already is a `Enabled` parameter. What is this parameter for? Can we use this parameter for the purposes of this feature?

alexanderbez:
Right, but we need to add a parameter for that. I was going to actually add a `Enabled` or `TxsEnabled` parameter and then I saw `Enabled` so hence my question. It's a bit confusing. So we need to add a 2nd parameter to gate tx flow?

sainoe:
This `Enabled` parameter is set when creating the consumer genesis in the provider chain
https://github.com/cosmos/interchain-security/blob/f95925c2d04ae155d8877efc2115d3791a20a9a3/x/ccv/provider/keeper/proposal.go#L78

And is mandatory in the consumer `InitGenesis` 
https://github.com/cosmos/interchain-security/blob/f95925c2d04ae155d8877efc2115d3791a20a9a3/x/ccv/consumer/keeper/genesis.go#L18-L21 

So I guess we're good to go with a second parameter like `TxsEnabled`

PR #113, opened 2022-05-25
feature: CCV Channel-based Message Filtering

alexanderbez:
Add an AnteHandler that does message filtering based on if the CCV channel has been established or not. If it has been established, the AnteHandler decorator performs a no-op, i.e. all messages are OK. Otherwise, only certain IBC messages are supported.

closes: #74


app/consumer/msg_filter_ante.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func hasValidMsgs(msgs []sdk.Msg) bool {
+	for _, msg := range msgs {
+		msgType := sdk.MsgTypeURL(msg)
+		// TODO: Perform message filtering that only allows certain IBC messages
```
alexanderbez, line 40: @mpoke do you know what message types we need to support when the CCV channel has not been established yet?

mpoke: @alexanderbez All IBC messages. The client needs to be updated, the connection and the channel need to be set up, and then a normal IBC packet will arrive with valset updates, which will mark the CCV channel as established.

Besides, I don't see any other message that is necessary during channel initialization.

mpoke:
@alexanderbez Issue https://github.com/cosmos/interchain-security/issues/115 is related

app/consumer/ante/msg_filter_ante.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func hasValidMsgsPreCCV(msgs []sdk.Msg) bool {
+	for _, msg := range msgs {
+		msgType := sdk.MsgTypeURL(msg)
+
+		// Only accept IBC messages prior to the CCV channel being established.
+		// Note, rather than listing out all possible IBC message types, we assume
+		// all IBC message types have a correct and canonical prefix -- /ibc.*
+		if !strings.HasPrefix(msgType, "/ibc.") {
```
AdityaSripal, line 52: This seems good to me. I don't think we need to filter specific core IBC messages out per se. What do you think @mpoke ?

AdityaSripal (approved):
Thanks for the addition @alexanderbez !

app/consumer/ante/msg_filter_ante.go
mpoke: I agree. I think all IBC messages should be enabled at all times.

app/consumer/ante/msg_filter_ante.go
```diff
@@ -0,0 +1,58 @@
...
+func NewMsgFilterDecorator(k ConsumerKeeper) MsgFilterDecorator {
+	return MsgFilterDecorator{
+		ConsumerKeeper: k,
+	}
+}
+
+func (mfd MsgFilterDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	currHeight := ctx.BlockHeight()
+
+	// If the CCV channel has yet been established, then we must only allow certain
```
mpoke, line 34:
```suggestion
	// If the CCV channel has not yet been established, then we must only allow certain
```

mpoke (approved):
LGTM. Nice work @alexanderbez

mpoke:
@alexanderbez Is this ready to merge or there is something else required? The TODO in the PR's description seems to be completed.

PR #116, opened 2022-05-30
Fix proto-gen

mpoke:
Fix `make proto-gen`:
- get `staking.proto` from cosmos/cosmos-sdk/interchain-security-rebase 
- get `cosmos.proto` needed by `staking.proto`
- update confio version
- remove `gogo.proto` from `make proto-update-deps`


sainoe (approved):
Great work @mpoke! Thanks for fixing this issue.

PR #118, opened 2022-05-31
Unbonding ops are put on hold even w/o consumer chains

### LINKED ISSUE #117: Unbonding ops are put on hold even w/o consumer chains
When an unbonding operation is initiated in the staking module, the `AfterUnbondingInitiated` hook is called, which results in calling the `PutUnbondingOnHold` function of the staking module, i.e., https://github.com/cosmos/interchain-security/blob/5acf1eeaeeed6cb921051526f7724562cfea9cac/x/ccv/provider/keeper/keeper.go#L453 

This happens even when there are no consumer chains, in which case, the unbonding operation never completes. 

**Fix:** 
```go
func (h StakingHooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) {
	var consumerChainIDS []string

	h.k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID string) (stop bool) {
		consumerChainIDS = append(consumerChainIDS, chainID)
		return false
	})
        if len(consumerChainIDS) > 0 {
	    valsetUpdateID := h.k.GetValidatorSetUpdateId(ctx)
	    unbondingOp := ccv.UnbondingOp{
		    Id:                      ID,
		    UnbondingConsumerChains: consumerChainIDS,
	    }
    
	    // Add to indexes
	    for _, consumerChainID := range consumerChainIDS {
		    index, _ := h.k.GetUnbodingOpIndex(ctx, consumerChainID, valsetUpdateID)
		    index = append(index, ID)
		    h.k.SetUnbondingOpIndex(ctx, consumerChainID, valsetUpdateID, index)
	    }
    
	    // Set unbondingOp
	    h.k.SetUnbondingOp(ctx, unbondingOp)
    
	    // Call back into staking to tell it to stop this op from unbonding when the unbonding period is complete
	    h.k.stakingKeeper.PutUnbondingOnHold(ctx, ID)
      }
}
```

mpoke:
Closes #117

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -433,26 +433,28 @@ func (h StakingHooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) {
 		consumerChainIDS = append(consumerChainIDS, chainID)
 		return false
 	})
-	if len(consumerChainIDS) > 0 {
-		valsetUpdateID := h.k.GetValidatorSetUpdateId(ctx)
-		unbondingOp := ccv.UnbondingOp{
-			Id:                      ID,
-			UnbondingConsumerChains: consumerChainIDS,
-		}
-
-		// Add to indexes
-		for _, consumerChainID := range consumerChainIDS {
-			index, _ := h.k.GetUnbodingOpIndex(ctx, consumerChainID, valsetUpdateID)
-			index = append(index, ID)
-			h.k.SetUnbondingOpIndex(ctx, consumerChainID, valsetUpdateID, index)
-		}
-
-		// Set unbondingOp
-		h.k.SetUnbondingOp(ctx, unbondingOp)
-
-		// Call back into staking to tell it to stop this op from unbonding when the unbonding period is complete
-		h.k.stakingKeeper.PutUnbondingOnHold(ctx, ID)
+	if len(consumerChainIDS) == 0 {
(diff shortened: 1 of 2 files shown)


x/ccv/provider/keeper/keeper.go
```diff
@@ -433,24 +433,26 @@ func (h StakingHooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) {
...
 
-	// Set unbondingOp
-	h.k.SetUnbondingOp(ctx, unbondingOp)
+		// Set unbondingOp
+		h.k.SetUnbondingOp(ctx, unbondingOp)
 
-	// Call back into staking to tell it to stop this op from unbonding when the unbonding period is complete
-	h.k.stakingKeeper.PutUnbondingOnHold(ctx, ID)
+		// Call back into staking to tell it to stop this op from unbonding when the unbonding period is complete
+		h.k.stakingKeeper.PutUnbondingOnHold(ctx, ID)
```
AdityaSripal, line 454: what happens theoretically if the unbonding is put on hold while child chains exist, but then child chain gets removed with gov proposal before hold is lifted?

mpoke: Nice catch @AdityaSripal

sainoe (approved):
Nice catch @mpoke ! LGTM

PR #120, opened 2022-06-01
separate consumer from provider in app.go

marbar3778:
spin out of #98 

- remove consumer dependency in provider app.go


marbar3778:
was one app.go being used for two different apps?

app/provider/app.go
```diff
@@ -518,15 +491,13 @@ func New(
...
 	ibcRouter.AddRoute(ibctransfertypes.ModuleName, ibcmodule)
-	ibcRouter.AddRoute(ibcconsumertypes.ModuleName, consumerModule)
 	ibcRouter.AddRoute(ibcprovidertypes.ModuleName, providerModule)
 	app.IBCKeeper.SetRouter(ibcRouter)
 
 	// create evidence keeper with router
 	evidenceKeeper := evidencekeeper.NewKeeper(
 		appCodec,
 		keys[evidencetypes.StoreKey],
-		app.ConsumerKeeper,
```
marbar3778, line 529: this should be the provider, but the provider does not met the interface for evidence. anyone have an idea of what the design here should be?

x/ccv/provider/keeper/genesis_test.go
```diff
@@ -17,14 +16,14 @@ func (suite *KeeperTestSuite) TestGenesis() {
...
-	suite.consumerChain.App.(*appConsumer.App).ProviderKeeper.InitGenesis(suite.consumerChain.GetContext(), genState)
+	suite.consumerChain.App.(*appProvider.App).ProviderKeeper.InitGenesis(suite.consumerChain.GetContext(), genState)
 
 	ctx = suite.consumerChain.GetContext()
 	for i := 0; i < 4; i++ {
 		expectedChainId := fmt.Sprintf("chainid-%d", i)
 		expectedChannelId := fmt.Sprintf("channelid-%d", i)
-		channelID, channelOk := suite.consumerChain.App.(*appConsumer.App).ProviderKeeper.GetChainToChannel(ctx, expectedChainId)
-		chainID, chainOk := suite.consumerChain.App.(*appConsumer.App).ProviderKeeper.GetChannelToChain(ctx, expectedChannelId)
+		channelID, channelOk := suite.consumerChain.App.(*appProvider.App).ProviderKeeper.GetChainToChannel(ctx, expectedChainId)
```
marbar3778, line 25: this left me confused. why is there a need for a consumer to have the provider keeper?

sainoe: This test is definitely outdated.

Is it still required @AdityaSripal ?

AdityaSripal: It should not be. Originally the tests were written with simapp having both consumer and provider hooked up to it. This may be a holdover

danwt: Yeah the simapps are now seperate too so it should be possible to delete this.

marbar3778: ah okay thank you

mpoke (approved):
LGTM. Nice work @marbar3778

Issue #35, opened 2022-02-07
Enable the removal of a consumer chain from the provider


mpoke:
The provider chain should be able to remove a consumer chain, i.e., clean up the state and enable the completion of unbonding operations that were waiting for that consumer chain.

IMO this removal should be triggered by only two events,
 - either a governance proposal (on the provider) to stop providing security to a consumer chain
- or the channel timeouts, which we assume it doesn’t, but we still need to do something when this assumption fails.

In my understanding, anyone (that knows the port and the channel ID) can send a `ChannelCloseInit` to the provider CCV module to close a CCV channel, right? The only way to implement some kind of access control is in `OnChanCloseInit`, right? I’m not sure though how to implement such access control, i.e., who can close a channel? Thus, I’d leave the logic inside `OnChanCloseInit` to abort any attempt to close the channel.

Instead, I suggest creating a handler of a new governance proposal (e.g., `StopConsumerChainProposal`) where we add all the cleanup logic. The same logic we add also in the `OnTimeoutPacket` handler on the provider.

What happens if the consumer chain’s channel end timeouts? This means that channel end moves to CLOSED. How will the provider find out about this? ACKs are no longer send back to the provider. Eventually the VSCPackets sent by the provider will timeout (since nobody will receive them). However, this requires for the timeout period to expire, which is very large to avoid timeouts :)

mpoke:
See also discussion on https://github.com/cosmos/ibc/issues/669

mpoke:
The PR https://github.com/cosmos/ibc/pull/707 specifies how safely remove a consumer chain.

Issue #123, opened 2022-06-03
Add ChanCloseConfirm to IBC-GO testing


sainoe:
Enable the full simulation of a channel closing handshake according to [ICS4](https://github.com/cosmos/ibc/tree/marius/ccv/spec/core/ics-004-channel-and-packet-semantics#closing-handshake). 

* An IBC testing endpoint should complete the handshake by calling `ChanCloseConfirm` in respond to a counter-party endpoint closing its channel's end with `ChanCloseInit` which is  already implemented in the IBC-GO testing suite.

* Required to test the removal of a consumer chain

smarshall-spitzbart:
We have [this test](https://github.com/cosmos/interchain-security/blob/30084e034ed88f09193c2faf156a53e32c97e905/x/ccv/consumer/ibc_module_test.go#L392) merged, and both the consumer/providers `OnChanCloseConfirm` simply returns a nil error. @sainoe may know better if there's still action to take

mpoke:
Closing as it is low priority.

Issue #18, opened 2021-10-25
Variable length string prefixes could lead to injection attack


jtremback:
I'm not 100% sure about this, but I think that the way we are using variable length string prefixes in the CCV module (for example: https://github.com/cosmos/interchain-security/blob/main/x/ccv/child/types/keys.go) could lead to a situation where an injection attack is possible.

Think about the (contrived) scenario where you have one set of record with keys made like this: `"bond" + <ID>` and a set of records with keys made like this `"bonded" + <ID>`. If the user has some kind of control over the ID on the "bond" records, they could create a situation where the first bytes of the ID are the same as the letters "ed". This could allow them to insert records into the "bonded" set.

I think we should use byte prefixes as is done in cosmos-sdk modules, or use iota to generate byte prefixes (we did this in gravity). Another option is to take string prefixes but hash or pad them.

PR #124, opened 2022-06-03
Remove consumer chain from provider

sainoe:
Close https://github.com/cosmos/interchain-security/issues/35

- Stop a consumer chain either in the event of a channel timeout or due to an accepted `StopConsumerChainProposal` on the provider chain (see [ICS28 spec](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md)) .

- Use a `LockUnbondingOnTimeout` flag to decide whether the funds corresponding to the outstanding unbonding operations are to be released in case of a timeout. 
 
Note that the tests simulating consumer chain shutdowns will be updated once https://github.com/cosmos/interchain-security/issues/123 is solved.


x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,74 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
...
+	k.DeleteLockUnbondingOnTimeout(ctx, chainID)
+
+	// close channel and delete the mappings between chain ID and channel ID
+	if channelID, found := k.GetChainToChannel(ctx, chainID); found {
+		k.chanCloseInit(ctx, channelID)
+		k.DeleteChainToChannel(ctx, chainID)
+		k.DeleteChannelToChain(ctx, channelID)
+	}
+
+	k.DeleteInitChainHeight(ctx, chainID)
```
mpoke, line 59:
```suggestion
        // TODO remove pending VSC packets once https://github.com/cosmos/interchain-security/issues/27 is fixed
	k.DeleteInitChainHeight(ctx, chainID)
```

x/ccv/provider/keeper/keeper.go
```diff
@@ -306,8 +318,38 @@ func (k Keeper) SetUnbondingOpIndex(ctx sdk.Context, chainID string, valsetUpdat
...
+func (k Keeper) IterateOverUnbondingOpIndex(ctx sdk.Context, chainID string, cb func(vscID uint64, ubdIndex []uint64) bool) {
+	store := ctx.KVStore(k.storeKey)
+	prefix := append(types.HashString(types.UnbondingOpIndexPrefix), types.HashString(chainID)...)
+	iterator := sdk.KVStorePrefixIterator(store, prefix)
+
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		// parse key to get the current VSC ID
+		var vscID uint64
+		vscBytes := iterator.Key()[64:]
```
mpoke, line 331: Why `[64:]`? The key's format should be `UnbondingOpIndexPrefix | chainID | vscID`. Shouldn't this be `iterator.Key()[len(iterator.Key())-64:]` instead. Is this because `HashString()` returns a 32 byte hash? We should make sure that https://github.com/cosmos/interchain-security/issues/18 is not going to break this logic.

x/ccv/provider/keeper/proposal.go
```diff
@@ -55,6 +124,11 @@ func (k Keeper) CreateConsumerClient(ctx sdk.Context, chainID string, initialHei
 	}
 
 	k.SetConsumerGenesis(ctx, chainID, consumerGen)
+
+	// store LockUnbondingOnTimeout flag
+	if consumerGen.GetParams().LockUnbondingOnTimeout {
+		k.SetLockUnbondingOnTimeout(ctx, chainID)
+	}
```
mpoke, line 131: ditto, i.e., we don't need `LockUnbondingOnTimeout` on the consumer.

x/ccv/provider/keeper/proposal.go
```diff
@@ -116,19 +190,25 @@ func (k Keeper) MakeConsumerGenesis(ctx sdk.Context) (gen consumertypes.GenesisS
 	return gen, nil
 }
 
-// SetConsumerClient sets the clientID for the given chainID
+// SetConsumerClient sets the client ID for the given chain ID
```
mpoke, line 193: `SetConsumerClient`, `GetConsumerClient`, and `DeleteConsumerClient` should be move to `keeper.go`. It will make it easier to keep track of the consumer state.

x/ccv/provider/keeper/proposal.go
```diff
@@ -179,3 +259,72 @@ func (k Keeper) IteratePendingClientInfo(ctx sdk.Context) {
...
+		// splitKey contains the bigendian time in the first element and the chainID in the second element
+		splitKey := strings.Split(string(suffixKey), "/")
+
+		timeNano := binary.BigEndian.Uint64([]byte(splitKey[0]))
+		stopTime := time.Unix(0, int64(timeNano))
+		chainID := string([]byte(splitKey[1]))
+
+		if ctx.BlockTime().After(stopTime) {
+			k.StopConsumerChain(ctx, chainID, false)
+			k.DeletePendingStopProposal(ctx, chainID, stopTime)
```
mpoke, line 305: Shouldn't we delete also the `PendingClientInfo`s when iterating through them (see `IteratePendingClientInfo`)?

mpoke (changes requested):
Nice work @sainoe. See my comments below. The only concern is that `LockUnbondingOnTimeout` should not be a param on the consumer.

proto/interchain_security/ccv/consumer/v1/consumer.proto
```diff
@@ -25,6 +25,11 @@ message Params {
   // provider handshake procedure.
   string distribution_transmission_channel = 3;
   string provider_fee_pool_addr_str        = 4;
+
+  // Indicates whether the outstanding unbonding operations should be released
+  // in case of a channel time-outs. When set to true, a governance proposal
+  // on the provider chain would be necessary to release the locked funds.
+  bool lock_unbonding_on_timeout = 6;
```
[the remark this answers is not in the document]
sainoe: Thanks for pointing this out @mpoke. I misread that part of the spec.

x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,74 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
...
 }
 
+// StopConsumerChainProposal stops a consumer chain and released the outstanding unbonding operations.
+// If the stop time hasn't already passed, it stores the proposal as a pending proposal.
+func (k Keeper) StopConsumerChainProposal(ctx sdk.Context, p *types.StopConsumerChainProposal) error {
+	if ctx.BlockTime().After(p.StopTime) {
+		return k.StopConsumerChain(ctx, p.ChainId, false)
+	}
+
+	k.Hooks().k.SetPendingStopProposal(ctx, p.ChainId, p.StopTime)
```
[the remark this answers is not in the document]
sainoe: Nice catch. It's a useless code completion from my IDE.

x/ccv/provider/keeper/keeper.go
sainoe: The three key's parts in `UnbondingOpIndexPrefix | chainID | vscID` are hashed into 32 bytes fixed output.
https://github.com/cosmos/interchain-security/blob/579db08d6bb34ea2f5ad793841f66f464935f6bb/x/ccv/provider/types/keys.go#L117

So `iterator.Key()[64:]` returns the key's tail containing a `vscID` digest
 while `iterator.Key()[len(iterator.Key())-64:]` would return the aggregated `chainID | vscID` hashes.

I believe we're safe regarding #18 since we use static length keys, yet its worth having a second opinion on this.

x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,75 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
...
+// it completes the outstanding unbonding operations lock by the consumer chain.
+func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, lockUbd bool) (err error) {
+
+	// clean up states
+	k.DeleteConsumerClient(ctx, chainID)
+	k.DeleteLockUnbondingOnTimeout(ctx, chainID)
+
+	// close channel and delete the mappings between chain ID and channel ID
+	if channelID, found := k.GetChainToChannel(ctx, chainID); found {
+		k.chanCloseInit(ctx, channelID)
```
mpoke, line 54: I'm not sure whether this is correct (I know that it's following the spec, but it may be wrong there also). I'm thinking of the logic of the closing handshake. This function is called by the provider as a result of either a gov proposal or a timeout. 

- In the case of a gov proposal, clearly we want to close this channel end (by calling `chanCloseInit`), so a relayer can relay a `MsgChannelCloseConfirm` to the consumer. 
- In the case of a timeout, this function is called from `OnTimeoutPacket()`, which is called on receiving either `MsgTimeout` (i.e., a packet sent by the provider timed out) or `MsgTimeoutOnClose` (i.e., a packet sent by the provider cannot be received since the consumer's channel end is closed). In both cases, the provider's channel end will be closed once either of the packets is handled (see [here](https://github.com/cosmos/ibc-go/blob/b2ca1932849cc3f05fe9e6ec30049aa4d08e6fc6/modules/core/04-channel/keeper/timeout.go#L157)).  Thus, we shouldn't call `chanCloseInit` in this case. 

@AdityaSripal Do you agree with this?

The consumer could also call `chanCloseInit` when a packet send to the provider times out. This will solve the following edge case, which is very unlikely:
- a packet send to the provider times out
- the consumer's channel end closes 
- the provider has no validator updates to send to the consumer, thus there is no packet for which a relayer can send a `MsgTimeoutOnClose`, which means that the provider never remove the consumer chain. 

Not sure whether it's worth the time to cover this edge case, since eventually there will be a validator set update on the provider, which will timeout  and lead to the call of `StopConsumerChain`.

x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,74 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
...
+
+				// If unbonding op is completely unbonded from all relevant consumer chains
+				if len(unbondingOp.UnbondingConsumerChains) == 0 {
+					// Attempt to complete unbonding in staking module
+					err = k.stakingKeeper.UnbondingCanComplete(ctx, unbondingOp.Id)
+					if err != nil {
+						return false
+					}
+					// Delete unbonding op
+					k.DeleteUnbondingOp(ctx, unbondingOp.Id)
```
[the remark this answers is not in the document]
sainoe: Correct.

x/ccv/provider/keeper/proposal.go
sainoe: Agree. Should we create a separate PR for the changes touching the `PendingClientInfos` states?

x/ccv/consumer/keeper/relay.go
```diff
@@ -205,3 +205,12 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
 func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, data ccv.SlashPacketData) error {
 	return nil
 }
+
+// IsChannelClosed returns a boolean whether a given channel is in the CLOSED state
+func (k Keeper) IsChannelClosed(ctx sdk.Context, channelID string) bool {
+	channel, found := k.channelKeeper.GetChannel(ctx, types.PortID, channelID)
+	if !found || channel.State == channeltypes.CLOSED {
```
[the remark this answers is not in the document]
AdityaSripal: Where and why would this be used?

x/ccv/consumer/module.go
```diff
@@ -151,7 +151,18 @@ func (AppModule) ConsensusVersion() uint64 { return 1 }
...
+// Panic if the provider's channel was established and then closed
 func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
+	channelID, found := am.keeper.GetProviderChannel(ctx)
+	if found && am.keeper.IsChannelClosed(ctx, channelID) {
+		// cleanup state
+		am.keeper.DeleteProviderChannel(ctx)
+
+		channelClosedMsg := fmt.Sprintf("CCV channel %q was closed - shutdown consumer chain since it is not secured anymore", channelID)
+		ctx.Logger().Error(channelClosedMsg)
+		panic(channelClosedMsg)
```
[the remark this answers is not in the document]
AdityaSripal: Yes this is how we do it for upgrades:

https://github.com/cosmos/cosmos-sdk/blob/main/x/upgrade/abci.go#L78

x/ccv/provider/keeper/keeper.go
AdityaSripal: Let's not hardcode the `64` here. Use helper function and the length of the variable. See the standard way to do this in ibc-go:

https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/types/keys.go

x/ccv/provider/keeper/proposal.go
AdityaSripal: I think in general we should not be deleting while iterating

x/ccv/provider/keeper/proposal.go
mpoke: Also, as a sanity check, we should just check if the channel is already closed before calling `chanCloseInit`

x/ccv/consumer/keeper/params_test.go
```diff
@@ -7,12 +7,12 @@ import (
 
 func (suite *KeeperTestSuite) TestParams() {
 	// suite setup initializes genesis
-	expParams := types.NewParams(true, 1000, "", "") // these are the default params
+	expParams := types.NewParams(true, 1000, "", "", false) // these are the default params
```
danwt, line 10: default should be lockUnbondingOnTimeout=true right?

https://github.com/cosmos/ibc/issues/669#issuecomment-1087320240

I know this is a test so maybe it's just the comment that's misleading

x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,75 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
 	return nil
 }
 
+// StopConsumerChainProposal stops a consumer chain and released the outstanding unbonding operations.
+// If the stop time hasn't already passed, it stores the proposal as a pending proposal.
+func (k Keeper) StopConsumerChainProposal(ctx sdk.Context, p *types.StopConsumerChainProposal) error {
+	if ctx.BlockTime().After(p.StopTime) {
```
danwt, line 36: The `After` could make it a PITA in the ass to test. Would having `!Before` cause a mismatch with the spec?

(cc @mpoke )

x/ccv/provider/stop_consumer_test.go
```diff
@@ -0,0 +1,294 @@
...
+	// 	- delegate the total bond amount to the chosed validator
+	// 	- undelegate the shares in four consecutive blocks evenly; create UnbondigOp and UnbondingOpIndex entries for the consumer chain ID
+	// 	- set SlashAck and LockUnbondingOnTimeout states for the consumer chain ID
+	setupOperations := []struct {
+		fn func(suite *ProviderTestSuite) error
+	}{
+		{func(suite *ProviderTestSuite) error {
+			suite.SetupCCVChannel()
+			return nil
+		},
```
danwt, line 54: Formatter might have missed this

danwt (commented):
It seems pretty nice, great job!

x/ccv/consumer/module.go
```diff
@@ -151,7 +151,21 @@ func (AppModule) ConsensusVersion() uint64 { return 1 }
 
 // BeginBlock implements the AppModule interface
 // Set the VSC ID for the subsequent block to the same value as the current block
+// Panic if the provider's channel was established and then closed
 func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
```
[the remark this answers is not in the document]
mpoke: As far as I can tell, the BeginBlock is idempotent, except for the case when the established CCV channel is closed. In this case we need to shut down the chain anyway, so I guess is shouldn't matter. @danwt do you see this as a problem for testing?

x/ccv/consumer/keeper/relay.go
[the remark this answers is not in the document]
mpoke: > Where and why would this be used?

In BeginBlock to check whether the CCV channel was closed.

x/ccv/provider/keeper/proposal.go
```diff
@@ -29,6 +30,75 @@ func (k Keeper) CreateConsumerChainProposal(ctx sdk.Context, p *types.CreateCons
...
+						return false
+					}
+					// Delete unbonding op
+					k.DeleteUnbondingOp(ctx, unbondingOp.Id)
+				} else {
+					k.SetUnbondingOp(ctx, unbondingOp)
+				}
+			}
+			// clean up index
+			k.DeleteUnbondingOpIndex(ctx, chainID, vscID)
```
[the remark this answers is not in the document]
danwt: Damn but I feel like I've seen this kind of thing in many places

see

https://github.com/cosmos/cosmos-sdk/blob/aba9bdc24cb6a7b9a85e6cad617f7b55d6dcdcec/x/staking/keeper/delegation.go#L360-L386

(line 382) `store.Delete(unbondingTimesliceIterator.Key())`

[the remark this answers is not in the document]
danwt: Or could it be this: 'Exceptionally allowed for cachekv.Store, safe to write in the modules.'

https://github.com/cosmos/cosmos-sdk/blob/aba9bdc24cb6a7b9a85e6cad617f7b55d6dcdcec/store/types/store.go#L217

The modules _are_ using cachekv.Store, and the actual KVStore is only written on commit in EndBlock right?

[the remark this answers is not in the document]
AdityaSripal: Ahh yes, I believe you are correct.

x/ccv/consumer/module.go
```diff
@@ -151,7 +151,18 @@ func (AppModule) ConsensusVersion() uint64 { return 1 }
...
+	channelID, found := am.keeper.GetProviderChannel(ctx)
+	if found && am.keeper.IsChannelClosed(ctx, channelID) {
+		// cleanup state
+		am.keeper.DeleteProviderChannel(ctx)
```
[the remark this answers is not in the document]
AdityaSripal: It seems unnecessary here if we're just going to stop the consumer chain?

It will require off-chain coordination to restart the chain anyway. So we may as well leave it to off-chain coordination to decide what the new genesis should look like

x/ccv/provider/keeper/proposal.go
```diff
@@ -21,17 +24,90 @@ import (
...
+	// clean up states
+	k.DeleteConsumerClient(ctx, chainID)
+	k.DeleteLockUnbondingOnTimeout(ctx, chainID)
+
+	// close channel and delete the mappings between chain ID and channel ID
+	if channelID, found := k.GetChainToChannel(ctx, chainID); found {
+		if closeChan {
+			k.CloseChannel(ctx, channelID)
+		}
+		k.chanCloseInit(ctx, channelID)
```
AdityaSripal, line 60: Shouldn't this line be inside the conditional as well?

AdityaSripal (commented):
Minor point on CloseChannel that still needs clarifying. Everything else looks good!

x/ccv/provider/keeper/proposal.go
sainoe: Definitely. Thanks!

x/ccv/consumer/module.go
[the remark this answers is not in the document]
mpoke: @AdityaSripal Yeah, but we should be cleaning the state as well as we can to facilitate the restart. Is there a reason not to do it?

x/ccv/provider/keeper/proposal.go
```diff
@@ -55,6 +130,11 @@ func (k Keeper) CreateConsumerClient(ctx sdk.Context, chainID string, initialHei
 	}
 
 	k.SetConsumerGenesis(ctx, chainID, consumerGen)
+
+	// store LockUnbondingOnTimeout flag
+	if lockUbdOnTimeout {
```
mpoke, line 135: Why store `LockUnbondingOnTimeout` flag only if it's true?

mpoke (approved):
LGTM. Only one comment.

Issue #90, opened 2022-05-06
Modify distribution module on consumer


mpoke:
This issue is related to the "democracy module" issue (#66) and come up in the discussion here https://github.com/cosmos/interchain-security/issues/66#issuecomment-1119459734, i.e.,
> the distribution module on the consumer needs to be modified to not distribute rewards (inflated native tokens) based on `LastCommitInfo` from Tendermint, but rather based on the stake from the staking module. In other words, in the `BeginBlock` of the distribution module, all stakers of native tokens should receive rewards (i.e., as if everyone has a vote in `LastCommitInfo`) and there is no special reward for block proposers.

PR #130, opened 2022-06-08
gov-distribution module

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/90

Closes https://github.com/cosmos/interchain-security/issues/141


okwme:
removed ccvstaking and ccvdistribution from the consumer/app.go
we need to make a new consumer app like `consumer-democracy` that imports these things

mpoke:
> we need to make a new consumer app like `consumer-democracy` that imports these things

Yes, there is an open issue https://github.com/cosmos/interchain-security/issues/141

okwme (approved):
nice job with the distribution e2e test @dusan-ethernal and @stana-ethernal !

mpoke:
> how are we handling coverage reports from sonarCloud @mpoke ?
> this pr fails because it doesn't have enough unit tests to keep the coverage up. I think it may be difficult to test this in unit tests though?

@danwt can provide more context re. SonarCloud.

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// TODO this is Tendermint-dependent
+	// ref https://github.com/cosmos/cosmos-sdk/issues/3095
+	if ctx.BlockHeight() > 1 {
+		am.AllocateTokens(ctx)
+	}
+
+	// record the proposer for when we payout on the next block
+	consAddr := sdk.ConsAddress(req.Header.ProposerAddress)
+	am.keeper.SetPreviousProposerConsAddr(ctx, consAddr)
```
jtremback, line 75: Is this used for anything any more?

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+// AllocateTokens handles distribution of the collected fees
+// bondedVotes is a list of (validator address, validator voted on last block flag) for all
+// validators in the bonded set.
+func (am AppModule) AllocateTokens(
+	ctx sdk.Context,
+) {
+
+	// fetch and clear the collected fees for distribution, since this is
+	// called in BeginBlock, collected fees will be from the previous block
+	// (and distributed to the previous proposer)
```
jtremback, line 87: Is this comment correct? There's no proposer any more

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+	if totalPower.IsZero() {
+		feePool.CommunityPool = feePool.CommunityPool.Add(feesCollected...)
+		am.keeper.SetFeePool(ctx, feePool)
+		return
+	}
+
+	// calculate fraction allocated to validators
+	remaining := feesCollected
+	communityTax := am.keeper.GetCommunityTax(ctx)
+	voteMultiplier := sdk.OneDec().Sub(communityTax)
```
jtremback, line 112: Can you write a comment about what is happening here?

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+	err := am.bankKeeper.SendCoinsFromModuleToModule(ctx, consumertypes.ConsumerRedistributeName, distrtypes.ModuleName, feesCollectedInt)
+	if err != nil {
+		panic(err)
+	}
+
+	// temporary workaround to keep CanWithdrawInvariant happy
+	// general discussions here: https://github.com/cosmos/cosmos-sdk/issues/2906#issuecomment-441867634
+	feePool := am.keeper.GetFeePool(ctx)
+	vs := am.stakingKeeper.GetValidatorSet()
+	totalPower := vs.TotalBondedTokens(ctx)
```
jtremback, line 102: This should be called totalBondedTokens or something

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+	voteMultiplier := sdk.OneDec().Sub(communityTax)
+
+	// allocate tokens proportionally to voting power
+	// TODO consider parallelizing later, ref https://github.com/cosmos/cosmos-sdk/pull/3099#discussion_r246276376
+
+	vs.IterateBondedValidatorsByPower(ctx, func(_ int64, validator stakingtypes.ValidatorI) bool {
+
+		// TODO consider microslashing for missing votes.
+		// ref https://github.com/cosmos/cosmos-sdk/issues/2525#issuecomment-430838701
+		powerFraction := sdk.NewDecFromInt(validator.GetTokens()).QuoTruncate(sdk.NewDecFromInt(totalPower))
```
jtremback, line 121: Comment: "we get this validator's percentage of the total power by dividing their tokens by the total bonded tokens"

jtremback: Comment: "we truncate here, which means that the validator gets slightly less than they would be entitled to. e.g. the sum of all validator's powerFraction will be less than 100%"

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+		reward := feesCollected.MulDecTruncate(voteMultiplier).MulDecTruncate(powerFraction)
```
jtremback, line 122: Comment: "we truncate here again, which means that the reward will be slightly lower than it should be"

x/ccv/democracy/distribution/module.go
```diff
@@ -0,0 +1,132 @@
...
+		am.keeper.AllocateTokensToValidator(ctx, validator, reward)
+		remaining = remaining.Sub(reward)
+
+		return false
+	})
+
+	// allocate community funding
+	feePool.CommunityPool = feePool.CommunityPool.Add(remaining...)
```
jtremback, line 130: Comment" "due to the 3 truncations above, `remaining` sent to the community pool will be slightly more than it should be. This is OK"

tests/e2e/democracy_test.go
```diff
@@ -0,0 +1,257 @@
...
+	representativeDifference := map[string]sdk.Dec{}
+	consumerRedistributeDifference := communityPoolDifference
+
+	for key, currentReward := range currentRepresentativesRewards {
+		representativeDifference[key] = nextRepresentativesRewards[key].Sub(currentReward)
+		consumerRedistributeDifference = consumerRedistributeDifference.Add(representativeDifference[key])
+	}
+
+	s.Require().Equal(distrModuleDifference, consumerRedistributeDifference)
+	s.Require().Equal(communityPoolDifference.Quo(consumerRedistributeDifference), distrKeeper.GetCommunityTax(s.consumerCtx()))
```
jtremback, line 239: Comment: "confirm that the percentage given to the community pool is equal to the configured community tax percentage."

tests/e2e/democracy_test.go
```diff
@@ -0,0 +1,257 @@
...
+	s.Require().InEpsilon(distrModuleDifference.Quo(providerDifference.Add(distrModuleDifference)).MustFloat64(), consumerFraction.MustFloat64(), float64(0.0001))
```
jtremback, line 240: Comment: "check that the fraction actually kept by the consumer is the correct fraction. using InEpsilon because the math code uses truncations"

tests/e2e/democracy_test.go
```diff
@@ -0,0 +1,257 @@
...
+	s.Require().InEpsilon(providerDifference.Quo(providerDifference.Add(distrModuleDifference)).MustFloat64(), sdk.NewDec(1).Sub(consumerFraction).MustFloat64(), float64(0.0001))
```
jtremback, line 241: Comment: "inverse of above check"

tests/e2e/democracy_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	totalRepresentativePower := stakingKeeper.GetValidatorSet().TotalBondedTokens(s.consumerCtx())
+
+	for key, representativeTokens := range representativesTokens {
+		powerFraction := sdk.NewDecFromInt(representativeTokens).QuoTruncate(sdk.NewDecFromInt(totalRepresentativePower))
+		s.Require().Equal(powerFraction, representativeDifference[key].Quo(consumerRedistributeDifference.Sub(communityPoolDifference)))
```
jtremback, line 247: Comment: "check that each representative has gotten the correct amount of rewards"

jtremback (approved):
Approved, with my comments

x/ccv/democracy/distribution/module.go
stana-ethernal: It can be safely removed since it is not used in democracy distribution module. In original core democracy module, it is used only in AllocateTokens function to calculate the rewards based on proposers but in overridden democracy distribution module,  AllocateTokens does not use it since the logic for rewards distribution is based on representatives and its powers. Also, it is used in distribution keeper ExportGenesis  function which export this key and in the case of democracy consumer chain this key will be empty (export does not fail, it is successfully done)

x/ccv/democracy/distribution/module.go
stana-ethernal: updated comment is pushed

Issue #80, opened 2022-04-30
Set unbonding period on the consumer


mpoke:
Currently, the unbonding period on the consumer is hardcoded to 4 weeks, i.e.,  https://github.com/cosmos/interchain-security/blob/aaa9a3b69c285bce51e07729eb7d429cbc04b76c/x/ccv/child/types/keys.go#L44. We should use the unbonding period of the provider chain instead.

mpoke:
One option would be for the provider to pass the unbonding period in every VSCPacket. 

@AdityaSripal Could we get this value from the provider client stored in the consumer CCV module? I'd assume that if the provider updates it's unbonding period, then it needs to update all its clients too, right?  This would also fit the logic [here](https://github.com/cosmos/interchain-security/blob/aaa9a3b69c285bce51e07729eb7d429cbc04b76c/x/ccv/parent/keeper/proposal.go#L42) where we use the provider unbonding period to create a client to the consumer.

jtremback:
IMO, the unbonding time should be several days less than the provider chain. This is so that users will not perceive variation in unbonding times because of relaying delays.

mpoke:
> IMO, the unbonding time should be several days less than the provider chain. This is so that users will not perceive variation in unbonding times because of relaying delays.

I agree. Once we get the unbonding period from the provider, we can subtract a few days from it and use it as the unbonding period on the consumer.

PR #144, opened 2022-06-09
Set unbonding period on the consumer

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/80

Closes https://github.com/cosmos/interchain-security/issues/153


x/ccv/utils/utils.go
```diff
@@ -78,3 +80,23 @@ func SendIBCPacket(
 	}
 	return nil
 }
+
+// ComputeConsumerUnbondingPeriod computes the unbonding period on the consumer
+// from the unbonding period on the provider (providerUnbondingPeriod)
+func ComputeConsumerUnbondingPeriod(providerUnbondingPeriod time.Duration) time.Duration {
```
AdityaSripal, line 86: Why are we doing this instead of just using the same unbonding period?

mpoke: @jtremback was saying that it will help with the user experience on the provider to not have to wait longer for an undelegation to complete, especially since we cannot provide the exact info on how long users should wait. Most likely, the one day is sufficient to account for relaying delays, which means that in general the unbonding operations will complete first on the consumer chains.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -138,6 +139,30 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
 	return k.scopedKeeper.ClaimCapability(ctx, cap, name)
 }
 
+// SetPacketMaturityTime sets the unbonding period of the consumer chain
```
danwt, line 142: Comment doesnt match fun name

mpoke, line 142:
```suggestion
// SetUnbondingTime sets the unbonding period of the consumer chain
```

mpoke:
https://github.com/cosmos/interchain-security/pull/144/commits/9d3adeaa0f72da09efd364dde8d7412c8f15936f fixes https://github.com/cosmos/interchain-security/issues/153

sainoe (approved):
Great work @mpoke :rocket: 
Just minor details to check otherwise LGTM. See my comments below.

x/ccv/utils/utils.go
AdityaSripal: We should document this reasoning in the function go-doc

x/ccv/provider/keeper/proposal.go
```diff
@@ -123,9 +126,13 @@ func (k Keeper) SetConsumerClient(ctx sdk.Context, chainID, clientID string) {
 }
 
 // GetConsumerClient returns the clientID for the given chainID.
-func (k Keeper) GetConsumerClient(ctx sdk.Context, chainID string) string {
+func (k Keeper) GetConsumerClient(ctx sdk.Context, chainID string) (string, bool) {
```
AdityaSripal, line 129: Can we change these functions to `SetConsumerClientId` and `GetConsumerClientId`? Was confused why the return value was string for a second

AdityaSripal (approved):
Nice work!

Issue #106, opened 2022-05-19
Ensure VSCMature packets are sent even if no ValidatorUpdates are received


danwt:
In the [spec at a6bc8d1](https://github.com/cosmos/ibc/commit/a6bc8d14dc1fc45cdcd052e6da8c6765389b7d61) it is possible for Maturity packets to never be sent. If the early return is taken here

```go
  if pendingChanges.IsEmpty() {
    // do nothing
    return []
  }
 ..
  if providerChannel != "" {
    UnbondMaturePackets()
  }
```

we need to make sure this doesn't happen in the implementation.

Relevant issue in spec:

- https://github.com/cosmos/ibc/issues/739

PR #146, opened 2022-06-10
Ensure VSCMature packets are sent even if no ValidatorUpdates are received

danwt:
Fixes

- https://github.com/cosmos/interchain-security/issues/106


mpoke (approved):

sainoe (approved):
Nice catch @danwt

PR #150, opened 2022-06-14
Add Historical Info to consumer 

sainoe:
Implement Historical info in consumer to work with IBC
  
  * Close [#142](https://github.com/cosmos/interchain-security/issues/142)
  
  * Add validator public key to consumer chain states

  * Call TrackHistoricalInfo in consumer BeginBlocker

Issue #139, opened 2022-06-08
Create app.go for minimum viable consumer chain


mpoke:
The default consumer `app.go`, i.e., `app/consumer/app.go`, should contain only the modules necessary for the minimum viable consumer chain:
- Staking module is disabled
- Governance module is disabled
- Distribution module is disabled
- Mint module is disabled
- Genutil module is disabled
- `x/ccv/consumer` is enabled and acts as Staking keeper for any other modules that rely on state from the Staking module
  - [ ] https://github.com/cosmos/interchain-security/issues/142

For more details, see https://github.com/cosmos/interchain-security/issues/59

mpoke:
See https://github.com/cosmos/interchain-security/pull/98 for more details on how to fix this issue.

Issue #141, opened 2022-06-08
Create app.go for gov-enabled consumer chain


mpoke:
This issue refers to the so called democracy package (see https://github.com/cosmos/interchain-security/issues/66 for details). 

`app/gov-consumer/app.go`, should contain the modules necessary for the minimum viable consumer chain (see [`app/consumer/app.go`](https://github.com/cosmos/interchain-security/issues/139)) and in addition the modules necessary for the gov-enabled consumer chain:
- `x/ccv/staking` module is enabled (see https://github.com/cosmos/interchain-security/pull/89)
- Governance module is enabled and connected to `x/ccv/staking`
- Mint module is enabled and connected to `x/ccv/staking`
- `x/ccv/distribution` module is enabled (see https://github.com/cosmos/interchain-security/pull/130) and connected to `x/ccv/staking`

okwme:
@hxrts and I created a voting matrix with a first pass about which chain should be in charge of which votes. The bulk of this sheet is param changes which could be debated. 

https://docs.google.com/spreadsheets/d/1s7IGkxIjvebFu0guC5iFC4gWaEOJLBWx2QQHik9WmYo/edit?usp=sharing


The primary proposal type that imo should be performed by ATOM token holders is the upgrade proposal. Our messaging is that ATOM validators don't need to worry about gov on these consumer chains because they have their own gov tokens. This is true except for factors that impact the validator directly. Upgrades need to be prepared and performed by validators so I think it's important they happen on the hub where validators are still meant to be active.

This would be possible by leveraging the adminmodule inside of the composer repo: https://github.com/cosmos/composer/tree/master/x/adminmodule
If the Cosmos Hub gov module account were listed in the whitelist then the cosmos hub gov module could send out proposal + vote msgs using interchain accounts. That way it is voted on the hub and if it passes it is submitted to the consumer chain and immediately passed.

jtremback:
The ultimate objective of this issue is to create a chain in the `/app/gov-consumer` folder which implements the "democracy suite". The "democracy suite" is collection of modifications to certain modules and `app.go` which allows several things to take place on a consumer chain:

- The consumer chain has a native token much like the native staking token on a normal Cosmos chain.
- This token can be delegated to "representatives". Representatives are just like validators on a normal Cosmos chain, but they do not have any responsibility to run chain binaries (obviously, these are run by the provider chain validators).
[...]

PR #152, opened 2022-06-15
MVCC part A

sainoe:
Partially solve #139 and contain all the changes that can be done before #150 is merged. 

Disable the following modules in the consumer:
 * Distribution
 * Governance
 * Mint
 * Genutil
 * Liquidity

Note that TestDistribution in `provider_test.go` started failing after removing the mint module
and is now commented.

### DIFF
--- a/app/consumer/export.go
+++ b/app/consumer/export.go
@@ -75,50 +75,50 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []str
 	/* Handle fee distribution state. */
 
 	// withdraw all validator commission
-	app.StakingKeeper.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) (stop bool) {
-		_, err := app.DistrKeeper.WithdrawValidatorCommission(ctx, val.GetOperator())
-		if err != nil {
-			panic(err)
-		}
-		return false
-	})
+	// app.StakingKeeper.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) (stop bool) {
+	// 	_, err := app.DistrKeeper.WithdrawValidatorCommission(ctx, val.GetOperator())
+	// 	if err != nil {
+	// 		panic(err)
+	// 	}
+	// 	return false
+	// })
 
 	// withdraw all delegator rewards
-	dels := app.StakingKeeper.GetAllDelegations(ctx)
-	for _, delegation := range dels {
-		_, err := app.DistrKeeper.WithdrawDelegationRewards(ctx, delegation.GetDelegatorAddr(), delegation.GetValidatorAddr())
-		if err != nil {
-			panic(err)
-		}
-	}
+	// dels := app.StakingKeeper.GetAllDelegations(ctx)
+	// for _, delegation := range dels {
+	// 	_, err := app.DistrKeeper.WithdrawDelegationRewards(ctx, delegation.GetDelegatorAddr(), delegation.GetValidatorAddr())
+	// 	if err != nil {
+	// 		panic(err)
+	// 	}
+	// }
 
 	// clear validator slash events
-	app.DistrKeeper.DeleteAllValidatorSlashEvents(ctx)
+	// app.DistrKeeper.DeleteAllValidatorSlashEvents(ctx)
 
 	// clear validator historical rewards
-	app.DistrKeeper.DeleteAllValidatorHistoricalRewards(ctx)
+	// app.DistrKeeper.DeleteAllValidatorHistoricalRewards(ctx)
 
 	// set context height to zero
 	height := ctx.BlockHeight()
 	ctx = ctx.WithBlockHeight(0)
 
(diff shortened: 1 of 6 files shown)


x/ccv/provider/provider_test.go
```diff
@@ -574,122 +573,122 @@ func (s *ProviderTestSuite) ReenableProviderDistribution() {
...
-	s.coordinator.CommitNBlocks(pChain, 1)
-	balance = pApp.BankKeeper.GetBalance(pChain.GetContext(), providerFeePoolAddr,
-		"ibc/3C3D7B3BE4ECC85A0E5B52A3AEC3B7DFC2AA9CA47C37821E57020D6807043BE9")
-	s.Assert().Equal(balance.Amount.Int64(), int64(0))
-	communityPool := pApp.DistrKeeper.GetFeePool(pChain.GetContext()).CommunityPool
-	balanceI64 := communityPool.AmountOf(
-		"ibc/3C3D7B3BE4ECC85A0E5B52A3AEC3B7DFC2AA9CA47C37821E57020D6807043BE9").RoundInt64()
-	s.Assert().Equal(balanceI64, expProviderAmt)
-}
+// func (s *ProviderTestSuite) TestDistribution() {
```
sainoe, line 576: @mpoke
Could you ping @rigelrozanski to debug this :) ?
I can hop on call with him to help

app/consumer/export.go
```diff
@@ -75,50 +75,50 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs []str
...
 
 	// reinitialize all delegations
-	for _, del := range dels {
-		app.DistrKeeper.Hooks().BeforeDelegationCreated(ctx, del.GetDelegatorAddr(), del.GetValidatorAddr())
-		app.DistrKeeper.Hooks().AfterDelegationModified(ctx, del.GetDelegatorAddr(), del.GetValidatorAddr())
-	}
+	// for _, del := range dels {
+	// 	app.DistrKeeper.Hooks().BeforeDelegationCreated(ctx, del.GetDelegatorAddr(), del.GetValidatorAddr())
+	// 	app.DistrKeeper.Hooks().AfterDelegationModified(ctx, del.GetDelegatorAddr(), del.GetValidatorAddr())
+	// }
```
danwt, line 121: I'm not fan of commenting out code, it might be nicer to just delete it and put a comment with a permalink to the commit with the deleted code. It's just my preference. Ultimately it's up to you

x/ccv/consumer/module_test.go
```diff
@@ -83,12 +83,14 @@ func (suite *ConsumerTestSuite) SetupTest() {
...
+
 	// set consumer endpoint's clientID
 	path.EndpointA.ClientID = providerClient
 
-	// TODO: No idea why or how this works, but it seems that it needs to be done.
+	// set chains sender account number
+	// TODO: to be fixed in #151
 	path.EndpointB.Chain.SenderAccount.SetAccountNumber(6)
-	path.EndpointA.Chain.SenderAccount.SetAccountNumber(6)
+	path.EndpointA.Chain.SenderAccount.SetAccountNumber(3)
```
danwt, line 93: 🚀

danwt (approved):
Nice job!

app/consumer/export.go
rigelrozanski: I usually leave commented code in until the system is pretty much finalized, then delete code all the code still commented at that point

x/ccv/provider/provider_test.go
rigelrozanski: Yo. Okay so the distribution tests are failing _because_ mint has been removed from `app/consumer/app.go`. If there is nothing to distribute then there will be nothing to test. This test could be modified to artificially add tokens to the FeePool (basically simulating mint, or fees being spent on the consumer chain) but that doesn't seem like a great use of time, the test would need to be largely rewritten. So if the consensus is to remove mint, then I'd recommend adding a bool in the consumer app.go that _optionally_ removes mint so it can still be used for testing. OR alternatively if mint will be re-added at some later point, this test can just remain commented out for the time being.

I guess I don't really understand _why_ mint isn't a part of the MVCC? Isn't the whole idea that a consumer chain has a token which it uses for something? If it doesn't have mint, it doesn't have a token to send back... I guess just the consumer would simply send back any fees collected... which may work I suppose (thinking out loud here).

mpoke: @rigelrozanski We plan to have multiple flavors of consumer chains, each one with its own app.go. For the minimum viable consumer chain (MVCC), the mint module is disable since there is no native token. For the governance-enabled consumer chain (see https://github.com/cosmos/interchain-security/issues/141), the mint module is required. 

Regarding testing, can't we just add in the unit test some tokens to `feeCollectorName` account on the consumer chain? Does that require a lot of work?

mpoke (approved):
LGTM. Nice work @sainoe. Let's open another issue to deal with `TestDistribution`.

Issue #27, opened 2021-12-24
Ensure unbonding ops cannot complete during initialisation


mpoke:
During the initialisation period, the consumer chain has the same initial validator set `V1`. To keep the security model of Tendermint, **none of the validators in `V1` should be able to unbond any amount of their stake during this period**. In other words, no unbonding delegation would complete during the CCV initialisation protocol. 

This feature would enable CCV to guarantee the following invariant that would preserve the security model of Tendermint on the consumer chains:
- Let `t` be an instance of time and `power(T)` be the function that calculates the voting power granted to some validators given an amount `T` of tokens bonded by these validators on the provider chain. Then, the voting power granted to validators on any consumer chain at time `t` MUST NOT exceed `power(T_t)`, where `T_t` is the amount of tokens bonded by these validators on the provider chain at time `t`.

This feature can be easily added to the current implementation.
- First, in `UnbondingDelegationEntryCreated` we need to iterate over all consumer chains, not just the ones with established channels, i.e., https://github.com/cosmos/interchain-security/blob/ddc91d30b042f1c2e793758bca52d192c6250b12/x/ccv/parent/keeper/keeper.go#L409 This will block the completion of any unbonding delegation until ACKs are received from all consumer chains. 
- Second, once the CCV channel is set to VALIDATING, the consumer chain still has the initial validator set `V1`. In the meantime, the provider chain may have gone through multiple updates, e.g., `V1` &rightarrow; `V2` &rightarrow; ... &rightarrow; `V10`. Thus, the first packet sent to the consumer chain should contain all the validator set changes e.g., from `V1` to `V10`. As in the previous step, we need to iterate over all consumer chains, i.e., https://github.com/cosmos/interchain-security/blob/ddc91d30b042f1c2e793758bca52d192c6250b12/x/ccv/parent/module.go#L158 and if there is no established channel, then accumulate all validator updates; otherwise, send a `ValidatorSetChange` packet with all accumulated validator updates. \
Note that this step needs to be done regardless of the completion of unbonding operations. 
[...]

PR #163, opened 2022-06-23
Ensure unbonding ops cannot complete during initialization

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/27


sainoe (approved):
Super work @mpoke !
Just one suggestion.

x/ccv/provider/keeper/keeper.go
```diff
@@ -621,6 +621,69 @@ func (k Keeper) DeleteInitChainHeight(ctx sdk.Context, chainID string) {
...
+	packets = append(packets, packet)
+	k.SetPendingVSCs(ctx, chainID, packets)
+}
+
+// EmptyPendingVSC empties and returns the list of pending ValidatorSetChange packets for chain ID
+func (k Keeper) EmptyPendingVSC(ctx sdk.Context, chainID string) (packets []ccv.ValidatorSetChangePacketData) {
+	packets = k.GetPendingVSCs(ctx, chainID)
+	if len(packets) < 1 {
+		return
+	}
```
danwt, line 681: What's the purpose of the function? I don't understand the reason for early returning in the `len(packets) < 1` case.

x/ccv/provider/keeper/keeper.go
```diff
@@ -621,6 +621,69 @@ func (k Keeper) DeleteInitChainHeight(ctx sdk.Context, chainID string) {
...
+
+// AppendPendingVSC adds the given ValidatorSetChange packet to the list
+// of pending ValidatorSetChange packets stored under chain ID
+func (k Keeper) AppendPendingVSC(ctx sdk.Context, chainID string, packet ccv.ValidatorSetChangePacketData) {
+	packets := k.GetPendingVSCs(ctx, chainID)
+	packets = append(packets, packet)
+	k.SetPendingVSCs(ctx, chainID, packets)
+}
+
+// EmptyPendingVSC empties and returns the list of pending ValidatorSetChange packets for chain ID
```
danwt, line 676: The comment doesn't accurately describe the functionality, because sometimes the list is deleted and sometimes it is found empty but not deleted. Can the comment be more precise?

danwt (approved):
Great work, just a couple of nits.

x/ccv/provider/keeper/keeper.go
```diff
@@ -136,11 +136,11 @@ func (k Keeper) DeleteChainToChannel(ctx sdk.Context, chainID string) {
 }
 
 // IterateConsumerChains iterates over all of the consumer chains that the provider module controls.
-// It calls the provided callback function which takes in a chainID and channelID and returns
+// It calls the provided callback function which takes in a chainID and returns
 // a stop boolean which will stop the iteration.
 func (k Keeper) IterateConsumerChains(ctx sdk.Context, cb func(ctx sdk.Context, chainID string) (stop bool)) {
 	store := ctx.KVStore(k.storeKey)
-	iterator := sdk.KVStorePrefixIterator(store, []byte(types.ChainToChannelKeyPrefix+"/"))
+	iterator := sdk.KVStorePrefixIterator(store, []byte(types.ChainToClientKeyPrefix+"/"))
```
AdityaSripal, line 143: nit: make a variable here and reuse in line 152

x/ccv/provider/keeper/keeper.go
```diff
@@ -621,6 +621,70 @@ func (k Keeper) DeleteInitChainHeight(ctx sdk.Context, chainID string) {
...
+		}
+		packets = append(packets, p)
+	}
+
+	return packets, true
+}
+
+// AppendPendingVSC adds the given ValidatorSetChange packet to the list
+// of pending ValidatorSetChange packets stored under chain ID
+func (k Keeper) AppendPendingVSC(ctx sdk.Context, chainID string, packet ccv.ValidatorSetChangePacketData) {
```
AdityaSripal, line 669: nit: this function seems like it can be merged with `SetPendingVSC`, if you take in a list and always append to existing list if it exists

x/ccv/provider/keeper/keeper.go
```diff
@@ -621,6 +621,70 @@ func (k Keeper) DeleteInitChainHeight(ctx sdk.Context, chainID string) {
 	store.Delete(types.InitChainHeightKey(chainID))
 }
 
+// SetPendingVSCs sets in store the list of pending ValidatorSetChange packets under chain ID
+func (k Keeper) SetPendingVSCs(ctx sdk.Context, chainID string, packets []ccv.ValidatorSetChangePacketData) {
```
AdityaSripal, line 625: Clarification: For CCV v1, why is there a list here when all mature child chains get the same updates? Is it because chains that have just finished initializing may have a different update?

x/ccv/provider/keeper/relay.go
```diff
@@ -112,14 +100,33 @@ func (k Keeper) EndBlockCallback(ctx sdk.Context) {
...
+			// construct validator set change packet data
+			packetData := ccv.NewValidatorSetChangePacketData(valUpdates, valUpdateID, k.EmptySlashAcks(ctx, chainID))
+
+			// check whether there is an established CCV channel to this consumer chain
+			if channelID, found := k.GetChainToChannel(ctx, chainID); found {
+				// send this validator set change packet data to the consumer chain
+				k.SendValidatorSetChangePacket(ctx, channelID, packetData)
+			} else {
+				// store the packet data to be sent once the CCV channel is established
+				k.AppendPendingVSC(ctx, chainID, packetData)
```
AdityaSripal, line 126: Seems like we can greatly simplify this code if we resolve this issue: https://github.com/cosmos/ibc/issues/645

I can take that up.

AdityaSripal (approved):
Looks like we can entirely remove the concept of pendingVSCs by using Optimistic Sends, so long as we finally fix optimistic packets. Since there's now an actual need for it, I can take a look into it

mpoke:
> Looks like we can entirely remove the concept of pendingVSCs by using Optimistic Sends, so long as we finally fix optimistic packets. Since there's now an actual need for it, I can take a look into it

Is this something that will be fixed in ibc-go before IS release (e.g., in Q3)?

Also, not sure what will happen in the following scenario:
- a consumer chain starts the handshake for the CCV channel
- VSCs are sent optimistically on that CCV channel
- the handshake fails (need to check how we handle that)
- the consumer starts a new handshake with a new channel ID 
- all the VSCs that were sent optimistically will be lost, right?

x/ccv/provider/keeper/keeper.go
mpoke: > Is it because chains that have just finished initializing may have a different update?

Yes. The CCV channels to different consumer chains get established async. This means that for each consumer chain there will be a list of VSC packets with validator updates that occurred during the initialization of that consumer chain.

x/ccv/provider/keeper/relay.go
mpoke: See my comment below https://github.com/cosmos/interchain-security/pull/163#issuecomment-1168389102

PR #177, opened 2022-06-28
Fix #175 delete outdated parts of the readme

### LINKED ISSUE #175: Delete outdated readme content
Delete the readme content related to startport as it is outdated

danwt:
Solves

- https://github.com/cosmos/interchain-security/issues/175


mpoke:
@danwt Shouldn't we update the outdated parts instead of just deleting them?

danwt:
@mpoke sure but that will more work it's and I don't think it makes sense to do it now. So might as well remove the misleading parts.

mpoke (approved):

Issue #39, opened 2022-02-09
Use independent packet type for unbonding information instead of ack, and put in same ordered channel as slashing packets


jtremback:
This will prevent slashing and unbonding packets getting out of order which could stop necessary slashings from happening.
It will also mean that we don’t need to store every packet for the whole unbonding period.

PR #188, opened 2022-06-29
Add VSCMatured packets instead of acks

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/39


x/ccv/consumer/keeper/relay.go
```diff
@@ -83,23 +82,26 @@ func (k Keeper) UnbondMaturePackets(ctx sdk.Context) error {
...
 	}
-	channelCap, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(types.PortID, channelID))
-	if !ok {
-		return sdkerrors.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
-	}
 
 	for maturityIterator.Valid() {
-		sequence := types.GetSequenceFromPacketMaturityTimeKey(maturityIterator.Key())
+		vscId := types.GetIdFromPacketMaturityTimeKey(maturityIterator.Key())
 		if currentTime > binary.BigEndian.Uint64(maturityIterator.Value()) {
```
danwt, line 88: The inequality doesn't match the spec. It won't matter in practice but will break automated tests

https://github.com/cosmos/ibc/blob/5ea902172b20d53e209f8c1fda9b11d005d9c110/spec/app/ics-028-cross-chain-validation/methods.md?plain=1#L1602-L1621

x/ccv/provider/keeper/relay.go
```diff
@@ -60,7 +73,8 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
 			// Attempt to complete unbonding in staking module
 			err := k.stakingKeeper.UnbondingCanComplete(ctx, unbondingOp.Id)
 			if err != nil {
-				return err
+				nack := channeltypes.NewErrorAcknowledgement(err.Error())
+				return &nack
 			}
 			// Delete unbonding op
 			k.DeleteUnbondingOp(ctx, unbondingOp.Id)
```
danwt, line 80: `DeleteUnbondingOp` is not called if `UnbondingCanComplete` returns an error. Is this intentional?

x/ccv/provider/unbonding_test.go
```diff
@@ -435,71 +290,68 @@ func undelegate(s *ProviderTestSuite, delAddr sdk.AccAddress, valAddr sdk.ValAdd
...
-// incrementTimeByProviderUnbondingPeriod increments the overall time by
-// the unbonding period on the provider; note that it is expected for this
-// period to be larger than the unbonding period on the consumer
-func incrementTimeByProviderUnbondingPeriod(s *ProviderTestSuite) {
+// incrementTimeByUnbondingPeriod increments the overall time by
+// 	- if provider == true, the unbonding period on the provider;
+//	- otherwise, the unbonding period on the consumer.
+// Note that it is expected for the provider unbonding period
+// to be one day larger than the consumer unbonding period.
+func incrementTimeByUnbondingPeriod(s *ProviderTestSuite, provider bool) {
```
danwt, line 298: I think it's better to use an enum or another way to differentiate cases, instead of having `provider bool`. I think we should have an enum or just two constant strings in the tests `const P = "provider"; const C = "consumer"; and differentiate with a function param `chain string`

```
if chain == P{
...
}
```

AdityaSripal (changes requested):
I believe the ack logic on the provider side needs to be revisited. Everything else is nits. Nice work!!

x/ccv/provider/keeper/keeper.go
```diff
@@ -392,6 +392,52 @@ func (k Keeper) GetUnbondingOpsFromIndex(ctx sdk.Context, chainID string, valset
...
+	}
+	err = json.Unmarshal(bz, &ids)
+	if err != nil {
+		return nil, err
+	}
+	return ids, nil
+}
+
+// AppendeMaturedUnbondingOps adds a list of ids to the list of matured unbonding operation ids
+func (k Keeper) AppendeMaturedUnbondingOps(ctx sdk.Context, ids []uint64) error {
```
danwt, line 410: Spelling mistake Appende -> Append

x/ccv/provider/keeper/relay.go
```diff
@@ -43,58 +26,95 @@ func removeStringFromSlice(slice []string, x string) (newSlice []string, numRemo
...
 
+	ack := channeltypes.NewResultAcknowledgement([]byte{byte(1)})
+	return ack
+}
+
+// CompleteMaturedUnbondingOps attempts to complete all matured unbonding operations
+func (k Keeper) CompleteMaturedUnbondingOps(ctx sdk.Context) {
+	ids, err := k.EmptyMaturedUnbondingOps(ctx)
+	if err != nil {
+		panic(fmt.Sprintf("could not get the list of matured unbonding ops: %s", err.Error()))
```
danwt, line 72: This is a panic in end block right. I guess it is a very bad situation, so it does make sense.

danwt (commented):
Wow, excellent work. This really simplifies a lot of things! Well done

danwt (approved):

danwt (approved):

x/ccv/provider/keeper/relay.go
mpoke: Yes. Both panics in this function should never happen. If they do happen, we should shut down and check what's going on. k

x/ccv/consumer/keeper/relay.go
```diff
@@ -194,17 +189,28 @@ func (k Keeper) SendPendingSlashRequests(ctx sdk.Context) {
...
 	if err := ack.GetError(); err != "" {
-		// packet was sent to a non-established channel
-		if err != sdkerrors.Wrap(
-			channeltypes.ErrInvalidChannelState,
-			packet.DestinationChannel,
-		).Error() {
-			return fmt.Errorf(err)
+		// Reasons for ErrorAcknowledgment
+		//  - packet data could not be successfully decoded
+		//	- packet sent on a non-established channel
```
AdityaSripal, line 197:
```suggestion
		//  - packet sent on a non-established channel
```

x/ccv/consumer/keeper/relay.go
```diff
@@ -194,17 +189,28 @@ func (k Keeper) SendPendingSlashRequests(ctx sdk.Context) {
...
+			return fmt.Errorf("ChanCloseInit(%s) failed: %s", packet.SourceChannel, err.Error())
+		}
+		// check if there is an established CCV channel
+		channelID, found := k.GetProviderChannel(ctx)
+		if !found {
+			return sdkerrors.Wrapf(types.ErrNoProposerChannelId, "recv ErrorAcknowledgement on non-established channel %s", packet.SourceChannel)
+		}
+		if channelID != packet.SourceChannel {
+			// Close the established CCV channel as well
+			return k.ChanCloseInit(ctx, types.PortID, channelID)
```
AdityaSripal, line 211: Why are we closing the established channel here as well?

x/ccv/consumer/keeper/relay.go
```diff
@@ -194,17 +189,28 @@ func (k Keeper) SendPendingSlashRequests(ctx sdk.Context) {
...
-		if err != sdkerrors.Wrap(
-			channeltypes.ErrInvalidChannelState,
-			packet.DestinationChannel,
-		).Error() {
-			return fmt.Errorf(err)
+		// Reasons for ErrorAcknowledgment
+		//  - packet data could not be successfully decoded
+		//	- packet sent on a non-established channel
+		//  - the Slash packet was ill-formed (errors while handling it)
+		// None of these should ever happen.
```
AdityaSripal, line 199: In this case, it would be good to add some logging or maybe panic at the end of the code so that information is provided to node operator that this is not normal and should be reported

x/ccv/provider/keeper/keeper.go
```diff
@@ -392,6 +392,52 @@ func (k Keeper) GetUnbondingOpsFromIndex(ctx sdk.Context, chainID string, valset
...
+	}
+	existingIds, err := k.GetMaturedUnbondingOps(ctx)
+	if err != nil {
+		return err
+	}
+	// append works also on a nil list
+	existingIds = append(existingIds, ids...)
+
+	store := ctx.KVStore(k.storeKey)
+	bz, err := json.Marshal(existingIds)
```
AdityaSripal, line 422: nit: more standard to marshal/unmarshal in stores with protobuf and wrap a list in a proto type

x/ccv/provider/keeper/relay.go
AdityaSripal: Noting that this will only shut down the node if it is called in Endblock, which it is

AdityaSripal (commented):
Partial review

x/ccv/provider/keeper/relay.go
```diff
@@ -43,58 +26,95 @@ func removeStringFromSlice(slice []string, x string) (newSlice []string, numRemo
...
-			}
+			// Store id of matured unbonding op for later completion of unbonding in staking module
+			maturedIds = append(maturedIds, unbondingOp.Id)
 			// Delete unbonding op
 			k.DeleteUnbondingOp(ctx, unbondingOp.Id)
 		} else {
 			k.SetUnbondingOp(ctx, unbondingOp)
 		}
 	}
+	k.AppendMaturedUnbondingOps(ctx, maturedIds)
```
danwt, line 59: Deferring completion to EndBlock is a divergence from the spec, we shouldn't forget that. 

- https://github.com/cosmos/ibc/issues/781

x/ccv/consumer/keeper/relay.go
mpoke: If we get here, the consumer chain should be shut down. The only way to do that is to close the CCV channel and then wait for the next BeginBlock where the code will panic (see [here](https://github.com/cosmos/interchain-security/blob/61153cd6084b0ddb6deae06c3a1882a5d3a481ad/x/ccv/consumer/module.go#L157)).

x/ccv/provider/keeper/keeper.go
mpoke: @AdityaSripal Not sure how to do that. Do you have an example (from cosmos-sdk or ibc-go)?

AdityaSripal: https://github.com/cosmos/cosmos-sdk/blob/76bbcd2ce3/proto/cosmos/distribution/v1beta1/distribution.proto#L93

https://github.com/cosmos/cosmos-sdk/blob/76bbcd2ce38fd0b56d999da41ba9ea630f5ef1b4/x/distribution/keeper/store.go#L47

Not a huge deal, just a note

x/ccv/provider/keeper/relay.go
```diff
@@ -43,58 +26,95 @@ func removeStringFromSlice(slice []string, x string) (newSlice []string, numRemo
...
+		return sdkerrors.Wrapf(types.ErrUnknownConsumerChannelId, "recv ErrorAcknowledgement on unknown channel %s", packet.SourceChannel)
+	}
 	return nil
 }
 
 // OnTimeoutPacket aborts the transaction if no chain exists for the destination channel,
 // otherwise it stops the chain
 func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet) error {
-	chainID, found := k.GetChannelToChain(ctx, packet.DestinationChannel)
+	chainID, found := k.GetChannelToChain(ctx, packet.SourceChannel)
```
AdityaSripal, line 103: nice catch

x/ccv/provider/keeper/relay.go
```diff
@@ -132,34 +167,27 @@ func (k Keeper) EndBlockCallback(ctx sdk.Context) {
...
 	}
 
 	ack := channeltypes.NewResultAcknowledgement([]byte{byte(1)})
 	return ack
 }
 
-// HandleSlashPacket slash and jail a wrong doing validator according the infraction height and type
-func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.SlashPacketData) error {
+// HandleSlashPacket slash and jail a misbehaving validator according the infraction type
+func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.SlashPacketData) (success bool, err error) {
```
AdityaSripal, line 190: where does the success bool get used if its ignored above?

AdityaSripal (approved):
The error ack handling seems a bit overly paranoid to me, but I suppose there's nothing wrong with that

x/ccv/provider/keeper/relay.go
mpoke: In tests. For example, there was a test that expected `HandleSlashPacket` to return error if the validator is tombstoned. However, in case of error while executing `HandleSlashPacket()`, `OnRecvSlashPacket()` returns an error ack and if the validator is already tombstoned, there should be no error ack. i.e., https://github.com/cosmos/interchain-security/blob/345fed1ffe11b27c74ef1c5dbbdf1969e75ac0da/x/ccv/provider/keeper/keeper_test.go#L405

x/ccv/provider/keeper/keeper.go
danwt: Yeah I do 👍

Issue #230, opened 2022-07-13
Consumer sends invalid validator updates to tendermint, resulting in `tm.verifyRemovals` fail


danwt:
This tendermint code checks if the validator set updates returned in EndBlock are valid

https://github.com/tendermint/tendermint/blob/85870def7b628effad73af942e638bbddf2ba8fd/types/validator_set.go#L538-L544

```go
func verifyRemovals(deletes []*Validator, vals *ValidatorSet) (votingPower int64, err error) {
	removedVotingPower := int64(0)
	for _, valUpdate := range deletes {
		address := valUpdate.Address
		_, val := vals.GetByAddress(address)
		if val == nil {
			return removedVotingPower, fmt.Errorf("failed to find validator %X to remove", address)
```

The check can fail on consumer because the passed updates are not filtered correctly.

https://github.com/cosmos/interchain-security/blob/657aaea0a6ed97f3a04d906186d353c56bef8d5e/x/ccv/consumer/module.go#L187-L195

The data needs to be modified in here

https://github.com/cosmos/interchain-security/blob/657aaea0a6ed97f3a04d906186d353c56bef8d5e/x/ccv/consumer/keeper/validators.go#L16-L52

See the very closely related spec issue that was fixed already

- https://github.com/cosmos/ibc/issues/743

I can make a PR for this soon.

Found with diff testing.

PR #231, opened 2022-07-13
Filter tendermint updates in consumer endblock #230

danwt:
Fixes https://github.com/cosmos/interchain-security/issues/230

- https://github.com/cosmos/interchain-security/issues/230

### DIFF
--- a/x/ccv/consumer/keeper/validators.go
+++ b/x/ccv/consumer/keeper/validators.go
@@ -13,42 +13,42 @@ import (
 )
 
 // ApplyCCValidatorChanges applies the given changes to the cross-chain validators states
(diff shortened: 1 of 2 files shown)


x/ccv/consumer/keeper/validators.go
```diff
@@ -13,42 +13,42 @@ import (
 )
 
 // ApplyCCValidatorChanges applies the given changes to the cross-chain validators states
```
sainoe, line 15: It should be commented that changes are filtered from an edge case. See my comment in
https://github.com/cosmos/interchain-security/issues/230#issuecomment-1186929234

danwt: Good point, I'll add this.

PR #235, opened 2022-07-14
Fix golangci-lint errors

smarshall-spitzbart:
Fixes the ```golangci-lint``` errors that exist in the repo. One error was [suppressed using nolint](https://github.com/cosmos/interchain-security/pull/235/commits/a50f4452c608ee379a51a870e053af892536f9b4), and tracked as an issue in #236


x/ccv/provider/keeper/proposal_test.go
```diff
@@ -20,10 +20,11 @@ func (suite *KeeperTestSuite) TestMakeConsumerGenesis() {
...
...distribution_transmission":1000, "lock_unbonding_on_timeout": false},"new_chain":true,"provider_client_state":{"chain_id":"testchain1","trust_level":{"numerator":1,"denominator":3},"trusting_period":907200000000000,"unbonding_period":1814400000000000,"max_clock_drift":10000000000,"frozen_height":{},"latest_height":{"revision_height":5},"proof_specs":[{"leaf_spec":{"hash":1,"prehash_value":1,"length":1,"prefix":"AA=="},"inner_spec":{"child_order":[0,1],"child_size":33,"min_prefix_length":4,"max_prefix_length":12,"hash":1}},{"leaf_spec":{"hash":1,"prehash_value":1,"length":1,"prefix":"AA=="},"inner_spec":{"child_order":[0,1],"child_size":32,"min_prefix_length":1,"max_prefix_length":1,"hash":1}}],"upgrade_path":["upgrade","upgradedIBCState"],"allow_update_after_expiry":true,"allow_update_after_misbehaviour":true},"provider_consensus_state":{"timestamp":"2020-01-02T00:00:10Z","root":{"hash":"LpGpeyQVLUo9HpdsgJr12NP2eCICspcULiWa5u9udOA="},"next_validators_hash":"E30CE736441FB9101FADDAF7E578ABBE6DFDB67207112350A9A904D554E1F5BE"},"unbonding_sequences":null,"initial_val_set":[{"pub_key":{"type":"tendermint/PubKeyEd25519","value":"dcASx5/LIKZqagJWN0frOlFtcvz91frYmj/zmoZRWro="},"power":1}]}`
```
smarshall-spitzbart, line 23: It was necessary to change the hardcoded json string here to prevent test failure. Previously this bug was undetected, since the unmarshaling error was not being checked below. 

The public key for the single validator in the val set at genesis needed to be changed to a schema like 
```
"pub_key": {
   "type": "tendermint/PubKeyEd25519",
   "value": "0kNlxBMpm+5WtfHIG1xsWatOXTKPLtmSqn3EiEIDZeI="
}
```
instead of 
```
"pub_key":{
    "Sum":{"ed25519":"dcASx5/LIKZqagJWN0frOlFtcvz91frYmj/zmoZRWro="}
}
```

I got inspiration from [this issue](https://github.com/tendermint/tendermint/issues/8583). Tbh it's still not super clear to me where the ```type``` and ```value``` fields are defined, but this works! Explanations welcome :)

PR #241, opened 2022-07-18
Remove blocked out code

danwt:
I was restless while building docker.


mpoke:
@danwt Are we sure that this code is not needed and was commented just temporarily to make things work?

danwt:
I would prefer for a v0.1 release that we clear this.  Commented out code is never needed IMO. It's a useful thing in short term but now it's been a while. The code still lives in git if anyone wants to see it.

mpoke (approved):

jtremback (commented):
Were you able to see if it seemed that the functionality of the blocked out code was implemented nearby? Especially thinking of the distribution test. Not that that should stop us from deleting it, but it makes me want to investigate

PR #242, opened 2022-07-20
Constant Length Key Prefixes

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/18

Prefixes and singular keys are now defined as byte constants, assigned with iota. Corresponding helpers returning ```[]byte``` are defined in the same file as the byte constants. 

Comments: 

1. This PR enforces the semantics of a "key"/"singular key" and a "key prefix" in some places. The former being a key associated to a single value (or a single sequence of values marshaled/unmarshaled together), while the latter is a prefix where further bytes are appended to the key to retrieve values.

2. It may be worth standardizing on where/if we use separators at all. **Update after discussions** - Separators (```"/"```) were removed wherever possible from existing key functions.

### DIFF
--- a/x/ccv/consumer/types/keys.go
+++ b/x/ccv/consumer/types/keys.go
@@ -74,12 +74,12 @@ const (
 	CrossChainValidatorBytePrefix
 )
 
-// Returns the key to the port ID in the store
+// PortKey returns the key to the port ID in the store
 func PortKey() []byte {
 	return []byte{PortByteKey}
 }
 
-// Returns the key to the last distribution transmission in the store
+// LastDistributionTransmissionKey returns the key to the last distribution transmission in the store
 func LastDistributionTransmissionKey() []byte {
 	return []byte{LastDistributionTransmissionByteKey}
 }
@@ -104,33 +104,33 @@ func PendingChangesKey() []byte {
 	return []byte{PendingChangesByteKey}
 }
 
-// Returns the historical info key prefix for historical info on each height
+// HistoricalInfoPrefix returns the historical info key prefix for historical info on each height
 func HistoricalInfoPrefix() []byte {
 	return []byte{HistoricalInfoBytePrefix}
 }
 
-// Returns the key prefix of maturity times stored for each received VSC packet
+// PacketMaturityTimePrefix returns the key prefix of maturity times stored for each received VSC packet
 func PacketMaturityTimePrefix() []byte {
 	return []byte{PacketMaturityTimeBytePrefix}
 }
 
-// Returns the key prefix for the mapping from block height to valset update ID
+// HeightValsetUpdateIDPrefix returns the key prefix for the mapping from block height to valset update ID
 func HeightValsetUpdateIDPrefix() []byte {
 	return []byte{HeightValsetUpdateIDBytePrefix}
 }
 
-// Returns the key prefix for validators' outstanding downtime by consensus address
+// OutstandingDowntimePrefix returns the key prefix for validators' outstanding downtime by consensus address
 func OutstandingDowntimePrefix() []byte {
 	return []byte{OutstandingDowntimeBytePrefix}
 }
 
-// Returns the key prefix for a list of slash request that must be sent
+// PendingSlashRequestsPrefix returns the key prefix for a list of slash request that must be sent
 // to the provider chain once the CCV channel is established
 func PendingSlashRequestsPrefix() []byte {
 	return []byte{PendingSlashRequestsBytePrefix}
 }
 
-// Returns the key prefix for cross-chain validators by consensus address
+// CrossChainValidatorPrefix returns the key prefix for cross-chain validators by consensus address
 func CrossChainValidatorPrefix() []byte {
 	return []byte{CrossChainValidatorBytePrefix}
 }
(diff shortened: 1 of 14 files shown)


x/ccv/types/keys.go
```diff
@@ -8,13 +8,20 @@ const (
 	// module supports
 	Version = "1"
 
-	// ChannelStatusKeyPrefix is the key prefix for storing the validation status of the CCV channel
-	ChannelStatusKeyPrefix = "channelstatus"
-
 	RouterKey = ModuleName
 )
 
+// Iota generated keys/byte prefixes (as a byte), supports 256 possible values
```
sainoe, line 14: You can remove the `ChannelStatus` key and prefix, they aren't needed anymore.

x/ccv/consumer/types/keys.go
```diff
@@ -23,101 +23,150 @@ const (
...
+	return []byte{ProviderChannelByteKey}
 }
 
 // PendingChangesKey returns the key for storing pending validator set changes
 func PendingChangesKey() []byte {
-	return []byte(PendingChangesKeyString)
+	return []byte{PendingChangesByteKey}
+}
+
+// Returns the historical info key prefix for historical info on each height
```
sainoe, line 107: Comments should begin with the element's name they describe.

```suggestion
// HistoricalInfoPrefix returns the historical info key prefix for historical info on each height
```

sainoe:
Good work @smarshall-spitzbart ! Just a few details.

I agree that we should decide if we use separators in keys.
I don't think that they're still needed since the prefixes and single keys have constant length now.

Any thoughts on this @danwt  @jtremback  ?

danwt:
I think we should avoid separators if we can as they can just lead to bugs.

smarshall-spitzbart:
All key separators have been removed as of latest commit

x/ccv/provider/types/keys.go
```diff
@@ -28,113 +28,181 @@ const (
...
-	ChannelToChainKeyPrefix = "channeltochain"
+	ChannelToChainBytePrefix
 
-	// ChainToClientKeyPrefix is the key prefix for storing the consumer chainID for a given consumer clientid.
-	ChainToClientKeyPrefix = "chaintoclient"
+	// ChainToClientBytePrefix is the byte prefix for storing the consumer chainID for a given consumer clientid.
+	ChainToClientBytePrefix
 
-	// PendingClientKeyPrefix is the key prefix for storing the pending identified consumer chain client before the spawn time occurs.
+	// PendingClientBytePrefix is the byte prefix for storing the pending identified consumer chain client before the spawn time occurs.
```
mpoke, line 58: Could you change the name to `PendingCreateProposalBytePrefix` and make it consistent throughout the code.

x/ccv/provider/keeper/genesis.go
```diff
@@ -33,7 +33,8 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
 
 func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 	store := ctx.KVStore(k.storeKey)
-	iterator := sdk.KVStorePrefixIterator(store, []byte(providertypes.ChannelToChainKeyPrefix+"/"))
+	keyPrefix := string(types.ChannelToChainPrefix())
+	iterator := sdk.KVStorePrefixIterator(store, []byte(keyPrefix))
```
mpoke, line 37: It's a bit weird that we convert from []byte to string and then back to []byte.

x/ccv/provider/keeper/keeper.go
```diff
@@ -186,15 +186,18 @@ func (k Keeper) DeleteChannelToChain(ctx sdk.Context, channelID string) {
 // or the callback returns stop=true
 func (k Keeper) IterateChannelToChain(ctx sdk.Context, cb func(ctx sdk.Context, channelID, chainID string) (stop bool)) {
 	store := ctx.KVStore(k.storeKey)
-	iterator := sdk.KVStorePrefixIterator(store, []byte(types.ChannelToChainKeyPrefix+"/"))
+	keyPrefix := string(types.ChannelToChainPrefix())
+	iterator := sdk.KVStorePrefixIterator(store, []byte(keyPrefix))
```
mpoke, line 190: ditto: why convert from []byte to string and then back to []byte?

mpoke (approved):
Great work @smarshall-spitzbart. See my comments below.

x/ccv/provider/types/keys.go
smarshall-spitzbart: makes sense 👍, addressed in https://github.com/cosmos/interchain-security/pull/242/commits/853150edf824880f16a4e3aa364f665179e08891

x/ccv/provider/keeper/genesis.go
smarshall-spitzbart: nice catch, this and other cleanups are addressed in https://github.com/cosmos/interchain-security/pull/242/commits/c944ad13197ce410e4fdb861f5f413e201dd162e

x/ccv/consumer/types/keys.go
```diff
@@ -23,101 +23,150 @@ const (
...
 }
 
 // PendingChangesKey returns the key for storing pending validator set changes
 func PendingChangesKey() []byte {
-	return []byte(PendingChangesKeyString)
+	return []byte{PendingChangesByteKey}
+}
+
+// HistoricalInfoPrefix returns the historical info key prefix for historical info on each height
+func HistoricalInfoPrefix() []byte {
```
jtremback, line 108: Changed my mind about this from when we talked this morning. Can you get rid of these boilerplate methods and just make people use e.g. `[]byte{HistoricalInfoBytePrefix}` directly in code that uses these prefixes?

smarshall-spitzbart: Addressed in https://github.com/cosmos/interchain-security/pull/242/commits/10d1ae2cbc8dd2906cb1ddcc94a6a7703570bd95

jtremback (approved):

Issue #244, opened 2022-07-28
Calling UnbondMaturePackets Accesses State Twice


smarshall-spitzbart:
By adding the existence of the provider channel as a precondition of calling UnbondMaturePackets in [this commit](https://github.com/cosmos/interchain-security/commit/bf14d648e302aefede469a740f1ed6d09a09860b) as a part of [this PR](https://github.com/cosmos/interchain-security/pull/225), most of the times the code will access the state twice, i,e., once to check whether the provider channel was set and the other time to actually retrieve the channelId

PR #245, opened 2022-07-28
Fix UnbondMaturePackets

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/244

There's not much to gain from accessing the provider channel both inside and outside the UnbondMaturePackets method


mpoke (approved):
Nice work

Issue #168, opened 2022-06-26
Integration test fails at 80d146a


danwt:
Fails for 80d146a02b0b7631a01cd40f026e6609ccb2f3d9

Debugging, the first `sendTokens` on the consumer that occurs after starting the consumer chain but before relaying any packets between consumer and provider does not work as expected

https://github.com/cosmos/interchain-security/blob/80d146a02b0b7631a01cd40f026e6609ccb2f3d9/integration-tests/steps.go#L128-L143

Leads to

```bash
main.SendTokensAction{chain:0x1, from:0x0, to:0x1, amount:0x1}
"actual state"
{1: {ValBalances: {0: 10000000000,
                   1: 10000000000},
     Proposals:   nil,
     ValPowers:   nil}}
"model state"
{1: {ValBalances: {0: 9999999999,
                   1: 10000000001},
     Proposals:   nil,
     ValPowers:   nil}}
2022/06/26 14:21:47 actual state (-) not equal to model state (+):  {
  1: {
   ValBalances: {
-   0: 10000000000,
-   1: 10000000000,
+   0: 9999999999,
+   1: 10000000001,
   },
   Proposals: nil,
   ValPowers: nil,
  },
 }
exit status 1
```

raised here

https://github.com/cosmos/interchain-security/blob/80d146a02b0b7631a01cd40f026e6609ccb2f3d9/integration-tests/main.go#L63

Running the transactions by hand inside the running docker containers gives

```bash
[root@7f1ea8b614c0 /]# interchain-security-cd query bank balances cosmos1dkas8mu4kyhl5jrh4nzvm65qz588hy9qcz08la --chain-id consumer --home /consumer/validator0 --node tcp://7.7.8.0:26658
balances:
- amount: "10000000000"
  denom: stake
pagination:
  next_key: null
  total: "0"
[root@7f1ea8b614c0 /]# interchain-security-cd query bank balances cosmos19pe9pg5dv9k5fzgzmsrgnw9rl9asf7ddwhu7lm --chain-id consumer --home /consumer/validator0 --node tcp://7.7.8.0:26658
balances:
- amount: "10000000000"
  denom: stake
pagination:
  next_key: null
  total: "0"
[...]

PR #248, opened 2022-08-02
Fix integration tests

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/168

Removes a couple auxiliary smoke tests to allow integration tests to pass ASAP. Also corrects an outdated cli command for hermes. Future work will check that no tokens can be sent on the consumer chain before the ICS channel is setup, but can be sent after the channel is setup (need to confirm this is the expected behavior)

Green check woo! Cred to Jehan

### DIFF
--- a/integration-tests/actions.go
+++ b/integration-tests/actions.go
@@ -463,10 +463,10 @@ func (s System) addIbcChannel(
 		"create", "channel",
 		"--order", action.order,
 		"--channel-version", s.containerConfig.ccvVersion,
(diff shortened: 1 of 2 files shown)


mpoke (approved):
LGTM. Great work @smarshall-spitzbart . 🚀

sainoe (approved):
Great work @smarshall-spitzbart !!

In order to be consistent with the IS-testnet. I suggest to update the IBC-relayer binary and commands to the new Hermes release (v1.0rc), as a future improvement.

PR #250, opened 2022-08-03
Add additional transaction checks to integration tests

smarshall-spitzbart:
Adds back the check that was removed in https://github.com/cosmos/interchain-security/pull/248, which confirms that a "send tokens" action _will not_ execute on a consumer chain (ie. the tx is filtered out by ante handler) before the ICS channel is fully initialized, and that the "send tokens" action _will_ execute properly on a consumer chain after the ICS channel is fully initialized.

### DIFF
--- a/integration-tests/steps.go
+++ b/integration-tests/steps.go
@@ -192,7 +192,7 @@ var happyPathSteps = []Step{
 		},
 		state: State{
 			1: ChainState{
-				// Tx should not go through, ICS channel is not setup until token delegation has been relayed to consumer
+				// Tx should not go through, ICS channel is not setup until first VSC packet has been relayed to consumer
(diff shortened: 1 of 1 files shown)


integration-tests/steps.go
```diff
@@ -174,6 +183,23 @@ var happyPathSteps = []Step{
...
+	{
+		action: SendTokensAction{
+			chain:  1,
+			from:   0,
+			to:     1,
+			amount: 1,
+		},
+		state: State{
+			1: ChainState{
+				// Tx should not go through, ICS channel is not setup until token delegation has been relayed to consumer
```
mpoke, line 195: by "token delegation" you mean "validator set updates"?

mpoke (approved):
Nice work

PR #252, opened 2022-08-05
Integration Test Addition - Unbonding

### LINKED ISSUE #94: Add undelegate functionality to integration tests
Support an unbonding action, and verify expected state in integration tests

smarshall-spitzbart:
Closes #94, Adds relevant action and state check to test basic unbonding functionality.

While implementing this PR, the error mentioned [here](https://github.com/informalsystems/ibc-rs/issues/2536) presented itself from hermes CLI. The solution was to adjust the non-default tendermint consensus parameters used in integration tests. Before finding that solution, I did the work to migrate the integration test driver from using hermes ```v0.15.0``` to ```v1.0.0-rc.1```. These changes are included in this PR since it doesn't hurt to use a newer version of hermes. More detail on the differences [here](https://github.com/informalsystems/ibc-rs/blob/v1.0.0-rc.1/UPGRADING.md).

### DIFF
--- a/integration-tests/steps.go
+++ b/integration-tests/steps.go
@@ -233,4 +233,47 @@ var happyPathSteps = []Step{
 			},
 		},
 	},
+	{
+		action: UnbondTokensAction{
+			chain:      0,
+			unbondFrom: 0,
+			sender:     0,
+			amount:     11111111,
+		},
+		state: State{
+			0: ChainState{
+				ValPowers: &map[uint]uint{
+					0: 500,
+					1: 500,
+					2: 500,
+				},
+			},
+			1: ChainState{
+				ValPowers: &map[uint]uint{
+					// Voting power on consumer should not be affected yet
+					0: 511,
+					1: 500,
+					2: 500,
+				},
+			},
+		},
+	},
+	// This final relay action seems to be failing non-deterministically :/
+	{
+		action: RelayPacketsAction{
+			chain:   0,
+			port:    "provider",
+			channel: 0,
+		},
+		state: State{
+			1: ChainState{
+				ValPowers: &map[uint]uint{
+					0: 500,
+					1: 500,
+					2: 500,
+				},
+			},
+		},
+	},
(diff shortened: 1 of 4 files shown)


jtremback (approved):
LGTM

integration-tests/steps.go
```diff
@@ -233,4 +233,46 @@ var happyPathSteps = []Step{
 			},
 		},
 	},
+	{
+		action: UnbondTokensAction{
+			chain:      0,
+			unbondFrom: 0,
+			sender:     0,
+			amount:     11111111,
```
mpoke, line 241: nit: Why unbonding `11111111` and not `11000000`?

integration-tests/steps.go
```diff
@@ -233,4 +233,46 @@ var happyPathSteps = []Step{
...
+			1: ChainState{
+				ValPowers: &map[uint]uint{
+					0: 500,
+					1: 500,
+					2: 500,
+				},
+			},
+		},
+	},
+	// TODO: Test full unbonding functionality, considering liquidity after unbonding period, etc.
```
mpoke, line 277: Shouldn't this be part of this integration test?

mpoke (commented):
Nice work @smarshall-spitzbart. See my comments below?

integration-tests/steps.go
smarshall-spitzbart: Yes, definitely. I'm having trouble manipulating the chains in such a way that would allow this (more extended) test to function tho, so I made it a TODO for now. Essentially, I'm setting custom genesis parameters that don't seem to be registered by the chains that are spawned up in the integration test docker container. I will be looking more into it today

integration-tests/steps.go
smarshall-spitzbart: Originally this was pattern matched from the delegate tokens action above, but ```10000000``` does indeed work with the arithmetic. I Changed each to your number 👍

jtremback (approved):
LGTM

PR #257, opened 2022-08-08
Test Validator Downtime - Part 1

### LINKED ISSUE #256: Test Validator Downtime in Integration Tests
We need to integration test the voting power of validators when a downtime has occurred. This should be tested for both a downtime observed on a consumer chain, and a downtime observed on a provider chain.

smarshall-spitzbart:
A solution to #256 was difficult to implement, and this PR will serve as a way to merge my initial progress. Part two will implement downtime in such a way that a non-operating node can easily become responsive/operational again. See #256 for more detailed comments.

This PR is built off #262


integration-tests/steps.go
```diff
@@ -238,12 +238,12 @@ var happyPathSteps = []Step{
 			chain:      providerChainId,
 			unbondFrom: 0,
 			sender:     0,
-			amount:     11000000,
+			amount:     1000000,
```
smarshall-spitzbart, line 241: This amount was made smaller to ensure that non-faulty validators maintain more than 2/3 validating power below

mpoke (approved):
Nice work @smarshall-spitzbart

PR #254, opened 2022-08-08
Drop duplicate spawn/stop proposals

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/155

Issue #160, opened 2022-06-21
Make sure setters and getters cover edge cases


mpoke:
The methods used to access the CCV state (both on provider and consumer) should be callable with all kind of params without panic.

danwt:
It's not always a good idea to program extra defensively like this IMO. Only the cases which can actually be called with nil, empty ect params should contain the checks.

danwt:
I think we should discuss this

- https://github.com/cosmos/interchain-security/issues/174

mpoke:
In general I think that 
- we should panic in case of programmer errors that shouldn't happen, e.g., using `MustMarshal` in setters;
- we should return errors that are handled at a higher layer if it's something that can be triggered by data (i.e., user TXs), e.g., returning an error or false when looking for the channel ID for a consumer chain that was not registered. 

The idea of this issue is to make sure that the code within setters / getters does not panic without us wanting that behavior, e.g., `store.Get` Panics on nil key (see https://github.com/cosmos/cosmos-sdk/blob/b6f867d0b674d62e56b27aa4d00f5b6042ebac9e/store/types/store.go#L197).

smarshall-spitzbart:
Got it, and thank you for the clarification of intention on this issue. Agreed on the points made here

jtremback:
All of the panics (with the exception below) on getters and setters in consumer and provider keeper.go are related to database corruption- i.e. there is stuff in the database that should not be in there, and there is no way for our program to recover from it. However, we are inconsistent. Some database corruption errors in some getters and setters are returned. We should probably at least be consistent about it.

Exception: This panic is in `AfterUnbondingInitiated` which is called by the staking module and then calls back into the staking module. It is unclear to me what the proper response would be here. Panic may still be the best option, because it is due to the staking module malfunctioning somehow and we cannot recover.

```
	// Call back into staking to tell it to stop this op from unbonding when the unbonding period is complete
	if err := h.k.stakingKeeper.PutUnbondingOnHold(ctx, ID); err != nil {
		panic(fmt.Errorf("unbonding could not be put on hold: %w", err))
	}
```

danwt:
Agree ^: we should panic iff the situation is not recoverable.

danwt:
Closing as there is no defined closing criteria

PR #258, opened 2022-08-08
small test fix for #254

smarshall-spitzbart:
@mpoke something like this should fix your branch. The changes in #254 will throw an error if two ```createConsumerProposal```s are set for a single chain id. Alternatively, we can allow multiple proposals to be submitted for a single chain id, and ignore duplicates without throwing error. 

Related to #160, have we come to an agreement on best design patterns for this issue?


jtremback (approved):
LGTM

mpoke:
> Related to https://github.com/cosmos/interchain-security/issues/160, have we come to an agreement on best design patterns for this issue?

Let's discuss it on the issue itself.

mpoke (approved):

Issue #219, opened 2022-07-07
Move all end-to-end tests to special folder


mpoke:

smarshall-spitzbart:
Started working on this today. 

Proposed hierarchy of testing:
- Unit tests for simple standalone functionality, and CRUD operations on consumer/provider keeper (native go testing package)
- Unit tests with mocked external keepers for a little more complex functionality, but still no interchain functionality (uses native go testing package, and testify/mock)
- e2e tests in their own directory for testing interchain functionality, things where advancing blocks would be useful, handshakes, packet relays, etc. (uses IBC testing suite)
- integration tests for testing highest level of functionality (real nodes, queries, txs, docker)
- diff testing?

#267 is still WIP but, but gives a good idea of what the first two bullet points would look like.

danwt:
Thanks @smarshall-spitzbart the above is very clear. We should probably put the info in the top level README (or CONTRIBUTING.md) when the PRs are done

PR #267, opened 2022-08-12
Provider Unit Test Cleanup

smarshall-spitzbart:
Adds helper functions for in-memory, bare bones unit testing of the provider keeper, and migrates existing unit tests to use the new helper functions. Also adds a couple unit tests to methods I refactored. This is a part of completing #219, in being able to distinguish between e2e tests and unit tests in an organized way.

Notes: 
- This PR is only for the provider ccv module, consumer to come
- Sonar cloud analysis fails, but is n/a unless I test my tests :)

Related tasks:
- Implement these same changes for the ```ccv/consumer``` folder
- Full reorganization of where e2e testing files live, move all tests using the IBC testing suite to designated folder,```e2e-test-split``` branch
- Updated readme on test organization


x/ccv/provider/keeper/proposal.go
```diff
@@ -249,17 +249,34 @@ func (k Keeper) PendingCreateProposalIterator(ctx sdk.Context) sdk.Iterator {
...
+// spawn time has passed.
+//
+// Note: this method is split out from IteratePendingCreateProposal to be easily unit tested.
+func (k Keeper) CreateProposalsToExecute(ctx sdk.Context) []types.CreateConsumerChainProposal {
 
 	iterator := k.PendingCreateProposalIterator(ctx)
 	defer iterator.Close()
 
+	// store the (to be) executed proposals in order
+	propsToExecute := []types.CreateConsumerChainProposal{}
```
mpoke, line 275: nit: the assignment to iterator and the check for validity should be together

x/ccv/provider/keeper/proposal.go
```diff
@@ -268,23 +285,22 @@ func (k Keeper) IteratePendingCreateProposal(ctx sdk.Context) {
 			panic(fmt.Errorf("failed to parse pending client key: %w", err))
 		}
 
-		var clientInfo types.CreateConsumerChainProposal
-		k.cdc.MustUnmarshal(iterator.Value(), &clientInfo)
+		var prop types.CreateConsumerChainProposal
+		k.cdc.MustUnmarshal(iterator.Value(), &prop)
+		prop.ChainId = chainID
+		prop.SpawnTime = spawnTime
```
mpoke, line 291: Why do we need these two lines?

x/ccv/provider/keeper/proposal.go
```diff
@@ -327,37 +343,54 @@ func (k Keeper) PendingStopProposalIterator(ctx sdk.Context) sdk.Iterator {
...
+// stop time has passed.
+//
+// Note: this method is split out from IteratePendingCreateProposal to be easily unit tested.
+func (k Keeper) StopProposalsToExecute(ctx sdk.Context) []types.StopConsumerChainProposal {
+
 	iterator := k.PendingStopProposalIterator(ctx)
 	defer iterator.Close()
 
+	// store the (to be) executed stop proposals in order
+	propsToExecute := []types.StopConsumerChainProposal{}
```
mpoke, line 367: ditto

x/ccv/provider/keeper/proposal.go
```diff
@@ -327,37 +343,54 @@ func (k Keeper) PendingStopProposalIterator(ctx sdk.Context) sdk.Iterator {
...
 			panic(fmt.Errorf("failed to parse pending stop proposal key: %w", err))
 		}
 
 		if !ctx.BlockTime().Before(stopTime) {
-			err = k.StopConsumerChain(ctx, chainID, false, true)
-			if err != nil {
-				panic(fmt.Errorf("consumer chain failed to stop: %w", err))
-			}
-			execProposals = append(execProposals,
+			// No more proposals to check, since they're stored/ordered by timestamp.
```
mpoke, line 382: This comment should be on the else branch.

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -277,17 +271,17 @@ func (suite *KeeperTestSuite) TestInitHeight() {
...
-		suite.Require().EqualValues(t.expected, height)
+	for _, tc := range tc {
+		height := providerKeeper.GetInitChainHeight(ctx, tc.chainID)
+		require.Equal(t, tc.expected, height)
 	}
 }
 
 func (suite *KeeperTestSuite) TestHandleSlashPacketDoubleSigning() {
-	ProviderKeeper := suite.providerChain.App.(*appProvider.App).ProviderKeeper
+	providerKeeper := suite.providerChain.App.(*appProvider.App).ProviderKeeper
```
mpoke, line 284: Why not use the `GetProviderKeeperAndCtx` function? Similarly, why do we need to use `suite.ctx` in the slashing tests?

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -508,44 +502,44 @@ func (suite *KeeperTestSuite) TestHandleSlashPacketDistribution() {
...
 		return true
 	})
-	suite.Require().Equal(len(unbondingOpIndex), i)
+	require.Equal(t, len(unbondingOpIndex), i)
 }
 
-func (suite *KeeperTestSuite) TestMaturedUnbondingOps() {
-	providerKeeper := suite.providerChain.App.(*appProvider.App).ProviderKeeper
+func TestMaturedUnbondingOps(t *testing.T) {
+	keeper, ctx := testkeeper.GetProviderKeeperAndCtx(t)
```
mpoke, line 529: nit: rename to `providerKeeper` for clarity

mpoke (approved):
Great work @smarshall-spitzbart. See my comments bellow.

x/ccv/provider/keeper/keeper_test.go
smarshall-spitzbart: Valid question, in this particular case I was able to make an equivalent test function using native golang testing and testify mocking (see [new test](https://github.com/cosmos/interchain-security/blob/56d61919c9c3d8cd548798f0e3205ee731aaac70/x/ccv/provider/keeper/keeper_test.go#L327)). However, I think it'd be best to still keep the e2e test version, as the e2e test validates the interaction between keepers at a higher level.

Looking at the bigger picture, the decision of unit tests vs e2e tests will have to be made on a case by case basis. In order to test this specific 
```TestHandleSlashPacketDoubleSigning``` functionality with [mock](https://pkg.go.dev/github.com/stretchr/testify/mock), there was a decent amount of [function call expectations I had to specify](https://github.com/cosmos/interchain-security/blob/56d61919c9c3d8cd548798f0e3205ee731aaac70/x/ccv/provider/keeper/keeper_test.go#L342). But I was able to test the slash packet functionality at a more granular level.

**Review by smarshall-spitzbart — COMMENTED** [review]
Thanks @mpoke, here's some changes/responses addressing your comments

**Comment by smarshall-spitzbart** [comment]
Update: I migrated this PR away from [testify's mock package](https://pkg.go.dev/github.com/stretchr/testify/mock) to [gomock](https://pkg.go.dev/github.com/golang/mock/gomock). The former was not allowing this PR to pass lint checks, along with gomock seeming to be far more simple. No functional changes are associated with this change, but there is now a large auto generated code file with mock definitions,```testutils/keeper/mocks.go```

File: x/ccv/provider/keeper/proposal.go (hunk 1 above)
Comment (mpoke) [reply x/ccv/provider/keeper/proposal.go:278]: I was suggesting sth like this:
```
	// store the (to be) executed proposals in order
	propsToExecute := []types.CreateConsumerChainProposal{}

	iterator := k.PendingCreateProposalIterator(ctx)
	defer iterator.Close()

	if !iterator.Valid() {
		return propsToExecute
	}
```

File: x/ccv/provider/keeper/proposal.go (hunk 2 above)
Comment (mpoke) [reply x/ccv/provider/keeper/proposal.go:295]: > Previously we were creating a consumer client directly from this method, meaning we didn't have to populate prop.ChainId and instead used the local variable chainID. Note that ChainId is stored in the key accessing these proposals, so it's not unmarshaled in k.cdc.MustUnmarshal(iterator.Value(), &prop) on L289

Why is `ChainId` not unmarshaled? AFAIK, it is part of `CreateConsumerChainProposal` which is set in `SetPendingCreateProposal`, i.e., 
```golang
// SetPendingCreateProposal stores a pending proposal to create a consumer chain client
func (k Keeper) SetPendingCreateProposal(ctx sdk.Context, clientInfo *types.CreateConsumerChainProposal) error {
	bz, err := k.cdc.Marshal(clientInfo)
	if err != nil {
		return err
	}

	store.Set(types.PendingCreateProposalKey(clientInfo.SpawnTime, clientInfo.ChainId), bz)
	return nil
}
```

x/ccv/provider/keeper/keeper_test.go
mpoke: That makes sense. I think it is useful to have both -- e2e tests that are checking that e.g. consumer initiated slashing works and unit tests that (as you said) are checking functionalities at a more granular level.

x/ccv/provider/keeper/proposal.go
smarshall-spitzbart: Sorry my last comment was a bit incorrect. ```ChainID``` is marshaled/unmarshaled to the store in association with every ```CreateConsumerChainProposal```. However, ```ChainID``` is a field that is stored in the key itself (along with spawn time). So it is not a part of ```bz``` in your code snippet.

This means we must explicitly set ```ChainID``` and ```SpawnTime``` Within ```GetPendingCreateProposal``` to have each ```CreateConsumerChainProposal``` be populated with the fields stored in the key.
```
func (k Keeper) GetPendingCreateProposal(ctx sdk.Context, spawnTime time.Time, chainID string) types.CreateConsumerChainProposal {
	bz := store.Get(types.PendingCreateProposalKey(spawnTime, chainID))
	if len(bz) == 0 {
		return types.CreateConsumerChainProposal{}
	}
	var clientInfo types.CreateConsumerChainProposal
	k.cdc.MustUnmarshal(bz, &clientInfo)
	// Populate info that's only stored in key, not store value
	clientInfo.SpawnTime = spawnTime
	clientInfo.ChainId = chainID

	return clientInfo
}
```

Your original comment was asking about the new ```CreateProposalsToExecute``` method, which uses an iterator, bypassing ```GetPendingCreateProposal```. The same consideration must apply here
```
		key := iterator.Key()
		spawnTime, chainID, err := types.ParsePendingCreateProposalKey(key)
		if err != nil {
			panic(fmt.Errorf("failed to parse pending client key: %w", err))
		}

		var prop types.CreateConsumerChainProposal
		k.cdc.MustUnmarshal(iterator.Value(), &prop)
		// Populate info that's only stored in key, not store value
		prop.ChainId = chainID
		prop.SpawnTime = spawnTime
       ...
```
ie. ```iterator.Key()``` has some of the fields we care about, and ```iterator.Value()``` has the rest.

x/ccv/provider/keeper/proposal.go
smarshall-spitzbart: Ah a lot more simple than I assumed :) sounds good

x/ccv/provider/keeper/proposal.go
```diff
@@ -231,14 +231,17 @@ func (k Keeper) SetPendingCreateProposal(ctx sdk.Context, clientInfo *types.Crea
...
+func (k Keeper) GetPendingCreateProposal(ctx sdk.Context, spawnTime time.Time, chainID string) types.CreateConsumerChainProposal {
 	store := ctx.KVStore(k.storeKey)
-	bz := store.Get(types.PendingCreateProposalKey(timestamp, chainID))
+	bz := store.Get(types.PendingCreateProposalKey(spawnTime, chainID))
 	if len(bz) == 0 {
 		return types.CreateConsumerChainProposal{}
 	}
 	var clientInfo types.CreateConsumerChainProposal
 	k.cdc.MustUnmarshal(bz, &clientInfo)
+	// Populate info that's only stored in key, not store value
```
jtremback, line 242: This is uneccesary

x/ccv/provider/keeper/proposal.go
```diff
@@ -249,17 +252,34 @@ func (k Keeper) PendingCreateProposalIterator(ctx sdk.Context) sdk.Iterator {
...
+			panic(fmt.Errorf("consumer client could not be created: %w", err))
+		}
+	}
+}
+
+// CreateProposalsToExecute iterates over the pending proposals and returns an ordered list of proposals to be executed,
+// ie. consumer clients to be created. A prop is included in the returned list (and is deleted from store) if its proposed
+// spawn time has passed.
+//
+// Note: this method is split out from IteratePendingCreateProposal to be easily unit tested.
```
jtremback, line 271: Make sure that the only change here is this refactor

x/ccv/provider/keeper/proposal.go
```diff
@@ -268,23 +288,23 @@ func (k Keeper) IteratePendingCreateProposal(ctx sdk.Context) {
...
 		} else {
+			// No more proposals to check, since they're stored/ordered by timestamp.
 			break
 		}
 	}
 
-	// delete the proposals executed
-	k.DeletePendingCreateProposal(ctx, execProposals...)
+	// delete the proposals to be executed
+	k.DeletePendingCreateProposal(ctx, propsToExecute...)
```
jtremback, line 306: Move this out to IteratePendingCreateProposal so that everything happens in the same order. As is, your code calls DeletePendingCreateProposal before CreateConsumerClient, which is different than how it used to be.

x/ccv/provider/keeper/proposal.go
```diff
@@ -327,37 +347,54 @@ func (k Keeper) PendingStopProposalIterator(ctx sdk.Context) sdk.Iterator {
...
 		} else {
+			// No more proposals to check, since they're stored/ordered by timestamp.
 			break
 		}
 	}
 
-	// delete the proposals executed
-	k.DeletePendingStopProposals(ctx, execProposals...)
+	// delete the proposals to be executed
+	k.DeletePendingStopProposals(ctx, propsToExecute...)
```
jtremback, line 395: Same thing as above with the order of DeletePendingStopProposals and StopConsumerChain

jtremback (approved):
LGTM

PR #268, opened 2022-08-12
Cleanup states after consumer chain removal

### LINKED ISSUE #125: Cleanup the consumer chain state on removal
The consumer chain state (i.e., see all the setters in `keeper/keeper.go`) must be cleaned up when the consumer chain halts (see `BeginBlock()` on consumer chain). This will enable the halted consumer chain to be restarted either as a consumer chain (by sending another gov proposal to the provider chain) or as a normal chain. 

- [x] #39 should be fixed first to decide on the state needed for `UnbondingPacket`s

sainoe:
Close #125


x/ccv/consumer/module.go
```diff
@@ -158,8 +158,9 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 		// the CCV channel was established, but it was then closed;
 		// the consumer chain is no longer safe
 
-		// cleanup state
+		// clean up states
 		am.keeper.DeleteProviderChannel(ctx)
+		am.keeper.DeleteProviderClient(ctx)
```
mpoke, line 163: What about the rest of the state? i.e., see setters, e.g., `SetPort`, `SetUnbondingTime`, `SetPendingChanges`, `SetPacketMaturityTime`, `SetHeightValsetUpdateID`, `SetOutstandingDowntime`, `SetCCValidator`, `SetPendingSlashRequests`

mpoke (approved):
Nice work @sainoe.

x/ccv/consumer/module.go
jtremback: Seems like almost all the state created by the module should be deleted really

jtremback (commented):
Lots of good stuff here, would like to review it on a call though

sainoe:
Thanks for yours reviews @mpoke @jtremback. 
I've extended the cleanup to all the consumer states before it shutdowns.

I've also noticed that my UT assertions needed to be defered in order to be really checked after a `panic` call.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -271,12 +283,25 @@ func (k Keeper) GetPacketMaturityTime(ctx sdk.Context, vscId uint64) uint64 {
...
 
-// DeletePacketMaturityTime deletes the the maturity time for a given received VSC packet id
+// DeletePacketMaturityTime deletes the packet maturity time for a given received VSC packet id
 func (k Keeper) DeletePacketMaturityTime(ctx sdk.Context, vscId uint64) {
 	store := ctx.KVStore(k.storeKey)
 	store.Delete(types.PacketMaturityTimeKey(vscId))
 }
 
+// ClearPacketMaturityTime deletes all the packet maturity time in the store
+func (k Keeper) ClearPacketMaturityTimes(ctx sdk.Context) {
```
mpoke, line 293: nit: "Clear..." is ambiguous since it suggests that the function is reseting the times. Why not use "Delete..." instead?

x/ccv/consumer/keeper/keeper.go
```diff
@@ -318,6 +343,27 @@ func (k Keeper) GetHeightValsetUpdateID(ctx sdk.Context, height uint64) uint64 {
...
+		if cb(height) {
+			break
+		}
+	}
+}
+
+// ClearHeightValsetUpdateIDs clears the valset update ids in the store
+func (k Keeper) ClearHeightValsetUpdateIDs(ctx sdk.Context) {
+	k.IterateHeightValsetUpdateID(ctx, func(height uint64) bool {
+		k.DeleteHeightValsetUpdateID(ctx, height)
```
mpoke, line 362: Is it okay to delete while iterating? Both `ClearPacketMaturityTimes` and `ClearOutstandingDowntimes` delete entries after the iteration.

x/ccv/consumer/keeper/relay.go
```diff
@@ -61,7 +61,7 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
 
 	// set outstanding slashing flags to false
```
mpoke, line 62: The comment is misleading. Can you expend why it's okay to remove the outstanding downtime flag.

x/ccv/consumer/keeper/validators.go
```diff
@@ -202,3 +203,30 @@ func (k Keeper) TrackHistoricalInfo(ctx sdk.Context) {
...
+		height := binary.BigEndian.Uint64(iterator.Key()[1:])
+
+		if cb(height) {
+			break
+		}
+	}
+}
+
+// ClearHistoricalInfo clears the historical info in stores
+func (k Keeper) ClearHistoricalInfo(ctx sdk.Context) {
```
mpoke, line 223: ditto: "Clear..." -> "Delete..."

x/ccv/consumer/module.go
```diff
@@ -160,6 +160,17 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
...
+		am.keeper.DeletePort(ctx)
+		am.keeper.ClearCCValidators(ctx)
+		am.keeper.DeletePendingChanges(ctx)
+		am.keeper.DeleteUnbondingTime(ctx)
+		am.keeper.ClearPacketMaturityTimes(ctx)
+		am.keeper.ClearOutstandingDowntimes(ctx)
+		am.keeper.ClearHistoricalInfo(ctx)
+		am.keeper.ClearHeightValsetUpdateIDs(ctx)
+		am.keeper.ClearPendingSlashRequests(ctx)
+		am.keeper.SetParams(ctx, types.NewParams(false, 0, "", ""))
```
mpoke, line 173: Why do we need to reset the params?

mpoke (approved):
LGTM. Just some minor comments.

x/ccv/consumer/module.go
sainoe: Like most of the consumer states, the params will be overwritten during `InitGenesis` if the chain is restarted as a consumer chain. So it isn't required to delete them but it shows that they won't be used anymore.

danwt:
In an ideal world we would get the SonarCloud gate to pass. It requires 80% coverage on new code, currently the figure is 72%

>  71.4% Coverage on New Code (is less than 80%)

(Just an optional suggestion)

danwt (commented):
Nice work!

I'm wondering if it would have been possible to take the approach of deleting everything in store prefixed with some prefix (or some similar solution). For example for provider side cleanup ideally we should just delete the store indexed with prefix `provider/<chainID>` and for consumer we could do for just `consumer/`.

I'm just concerned that the way of deleting each piece individually is very error prone and a lot of work to maintain if state ever gets added/removed from the protocol.

Nice work though!

x/ccv/consumer/module.go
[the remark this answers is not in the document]
mpoke: > I'm also curious why we do this in BeginBlock and not EndBlock? AFAIU the consumer will only become aware that the channel has closed if receives a packet telling it so in a TX.

When handling a timeout (i.e., https://github.com/cosmos/ibc-go/blob/397b88cf98214e62437a94db1a26c935d4d153fd/modules/core/keeper/msg_server.go#L449) the channel state is moved to CLOSED. If the consumer chain shuts down at the end of the same block, then the CLOSED state is not persisted (i.e., the timeout never happened).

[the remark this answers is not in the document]
danwt: Ok I see so it simplifies to keep it in BeginBlock. If you shutdown in EndBlock you don't care if it's persisted or not though?

Or is there a situation where the provider can only reach CLOSED state IF the consumer CLOSED state is already persisted? If so, then I agree it should be in BeginBlock.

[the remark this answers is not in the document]
mpoke: See https://github.com/cosmos/interchain-security/issues/283#issuecomment-1224248917

jtremback:
Discussed with @sainoe this morning:

There is no reason to clean up consumer state.

The scenario in which one would want to do this not clearly defined. Presumably the consumer chain has been stopped and will be restarted as a sovereign chain, or have another consumer governance proposal created to re-launch it.

If it restarted from genesis, then the ccvconsumer section will be completely removed anyway, so the state deletion is irrelevant. 

If it's done with a non-genesis upgrade (keeping the existing database state in the home folder), then perhaps this functionality could be useful. However, it is probably better done in a migration script, since new staking module stake or new ccvconsumer state will have to be added as well anyway. Stopping the chain and leaving all state intact to be dealt with in whatever way later is probably the cleaner option.

Also, the way it is written now, the state is deleted, and then the chain is stopped, in beginblocker. A block will never be produced, and it's unclear whether this will even have any effect. 

So we should probably not add anything to erase all state on the consumer at all, and @sainoe will be removing all that stuff from this PR.

Provider side is completely different story, and that stuff is looking great.

mpoke:
@jtremback The only problem I may see is the halted consumer wanting to apply again to be a consumer chain (via a gov proposal). Due to the polluted state, checks like [this](https://github.com/cosmos/interchain-security/blob/b6aff82c471b4d86b990c80aeee4401442ae7671/x/ccv/consumer/ibc_module.go#L33) will fail.

jtremback:
> @jtremback The only problem I may see is the halted consumer wanting to apply again to be a consumer chain (via a gov proposal). Due to the polluted state, checks like [this](https://github.com/cosmos/interchain-security/blob/b6aff82c471b4d86b990c80aeee4401442ae7671/x/ccv/consumer/ibc_module.go#L33) will fail.

Yea, worse things could happen as well if there are other collisions in the store. My point is that some migration of the state will likely be required anyway (especially since there was apparently some reason to stop it), so they should just clean it there.

Issue #26, opened 2021-09-14
Validator key assignment ('key delegation')


AdityaSripal:
Allow Validators to set a different consensus key for each baby chain

- [ ] Create Msg type
- [ ] Implement setting handler (should always override previous value)
- [ ] Default to same consensus key as parent chain
- [ ] In Endblocker, make the transformation from parent keys to baby keys in the change set before sending the packet

mpoke:
This will be a user tx on the provider CCV module, i.e., CreateKey(valAddr, consumerChainId).

jtremback:
Jack's comment- a validator set change of more than 1/3 will break all the IBC light client

mpoke:
I'm not sure I understand. I thought that the validators can choose their consensus keys for a given chain before the chain is initialized. Can validators do this while validating on the consumer chain? What would be the benefit of that?

jackzampolin:
Key rotation is a nice to have feature but could present problems. Setting the key prior to consumer chain start should be fine.

mpoke:
@okwme Is this feature something we need for V1 or it's a "nice to have"?

mohammedpatla:
> Key rotation is a nice to have feature but could present problems. Setting the key prior to consumer chain start should be fine.

Honestly waiting for key rotation option.
If this was available it would be more open to reusing priv key across child chains. For now we spin a new one for every chain.

marbar3778:
We are thinking of providing the expected_keeper interface for the ccv module as such:

```
type LookUpModule interface {
 GetLookupList() []entry
 GetUserEntry(address sdk.Address) entry 
}
```

Does this make sense from the ccv point of view?


(names aren't final)

marbar3778:
QQ: If a validator joins the set how long would it take the update to propagate to N chains? The issue arises that once a validator joins the active set how long do they have to start spinning up on consumer chains, this would be the deciding factor for how this gets implemented for me

okwme:
If a validator joins the set, this should trigger an update to the child chain so it could be quite fast. Faster than the val might have to submit a new key for the child chain. However, maybe it shoudl be possible to associate keys before joining the val set? Also probably shouldn't try to validate whether a key is or is not currently a validator for the same reason. Would this improve the scenario you were worried about @marbar3778 ?

marbar3778:
[...]

PR #269, opened 2022-08-15
[draft] Validator key delegation

jtremback:
This should do it. Needs some cli methods and need to figure how to serialize keys correctly.

Closes https://github.com/cosmos/interchain-security/issues/26

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -26,6 +26,7 @@ import (
 	ccv "github.com/cosmos/interchain-security/x/ccv/types"
 
 	"github.com/tendermint/tendermint/libs/log"
+	tmcrypto "github.com/tendermint/tendermint/proto/tendermint/crypto"
 )
 
 // Keeper defines the Cross-Chain Validation Provider Keeper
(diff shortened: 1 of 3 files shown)


jtremback:
Passing this off to @danwt. 

The way I am doing it here is not going to work. Since what is sent over is a list of changes, not a list of all current powers, just substituting out the key like I do here is going to result in nonsensical updates.

There might be a few alternate approaches:

1. Synthesize validator changes: When a key is delegated, generate a valset update that makes it look to the consumer chain like the old key left the set and the new key was brought in with the same power. The consumer chain will not know that they are the same validator. After that, the keys can simply be substituted in updates like they are here. To make slashing work, you would probably need to switch the delegated key for the real key in slashing packets.

2. Delegate on the consumer: When someone delegates a key, send a packet with the key delegation to the consumer chain. Then the consumer chain will do the substitution. More code, and I'm not exactly clear on what the benefit would be over the option above.

mpoke:
A third option would be to expand the `abci.ValidatorUpdate` data structure. Currently it contains a `PubKey` and the `Power` as `int64`. We could also add the consensus address `Address` of the validator  on the provider chain. In case the validator doesn't use another key for a consumer chain, then `Address  = hash(PubKey)`. Otherwise, the `PubKey` is the key used on the consumer chain. 

Consequently, `VSCPacket`s contain power updates and validator key updates. `SlashPacket`s require the consensus address on the provider (i.e., `Address`), which the CCV module can retrieve from the mapping kept internally.

Note that the mapping of old validator keys must be kept to avoid a validator misbehaving on the consumer and then changing its address to not get slashed.

danwt:
@mpoke I don't have a strong opinion on your question.

jtremback:
Closing this for now, until it is more finished

PR #274, opened 2022-08-18
Consumer Unit Test Cleanup

smarshall-spitzbart:
Builds off #267 for the consumer directory. SonarCloud analysis failure is again N/A. 

Details:
- Migrated consumer related tests (that I deemed to be unit tests) away from the IBC testing suite, to now use native go testing and sometimes mocked keepers
- Made changes in logic for some tests where intention was unclear or didn't make sense
- Refactored the consumer and provider keepers to reference a new ```ScopedKeeper``` interface, instead of the scoped keeper struct itself from the capabilities sdk module. This enables necessary mocking of that keeper
- Added auto generated mock code for the scoped keeper
- Modified ```testutil/keeper/unit_test_helpers.go``` to aid in new or modified unit tests

Next up: Move remaining IBC test suite functions to a dedicated folder


x/ccv/provider/keeper/keeper_test.go
```diff
@@ -340,38 +340,44 @@ func TestHandleSlashPacketDoubleSigning(t *testing.T) {
...
 	mockSlashingKeeper := testkeeper.NewMockSlashingKeeper(ctrl)
-	mockSlashingKeeper.EXPECT().SlashFractionDoubleSign(ctx).Return(sdk.NewDec(1)).Times(1)
-	mockSlashingKeeper.EXPECT().JailUntil(ctx, sdk.ConsAddress(slashPacket.Validator.Address),
-		evidencetypes.DoubleSignJailEndTime).Times(1)
-	mockSlashingKeeper.EXPECT().IsTombstoned(ctx, sdk.ConsAddress(slashPacket.Validator.Address)).Return(false).Times(1)
-	mockSlashingKeeper.EXPECT().Tombstone(ctx, sdk.ConsAddress(slashPacket.Validator.Address)).Times(1)
 
-	providerKeeper := testkeeper.GetProviderKeeperWithMocks(t,
+	// Setup expected mock calls
+	gomock.InOrder(
```
smarshall-spitzbart, line 347: This diff simply adds an expected order of mock calls to the test

x/ccv/consumer/keeper/relay_test.go
```diff
@@ -109,44 +115,87 @@ func (suite *KeeperTestSuite) TestOnRecvVSCPacket() {
...
+
 	for _, tc := range testCases {
-		ack := suite.consumerChain.App.(*appConsumer.App).ConsumerKeeper.OnRecvVSCPacket(suite.ctx, tc.packet, tc.newChanges)
-		suite.Require().NotNil(ack, "invalid test case: %s did not return ack", tc.name)
+		ack := consumerKeeper.OnRecvVSCPacket(ctx, tc.packet, tc.newChanges)
 
-		if tc.expErrorAck {
-			suite.Require().False(ack.Success(), "invalid test case: %s did not return an Error Acknowledgment", tc.name)
+		if tc.noAck {
+			// if closing channel was successful, no acknowledgement returned, see OnRecvPacketOnUnknownChannel
```
smarshall-spitzbart, line 172: This is slightly different than previous expected behavior, an err ack should only be returned if something went wrong with ```ChanCloseInit``` for the ```"invalid packet: different destination channel than provider channel"``` test case

x/ccv/consumer/keeper/relay_test.go
```diff
@@ -263,22 +312,93 @@ func incrementTimeBy(s *KeeperTestSuite, jumpPeriod time.Duration) {
...
+	)
+
+	// Set an established provider channel for later in test
+	consumerKeeper.SetProviderChannel(ctx, channelIDToProvider)
+
 	packetData := types.NewSlashPacketData(
 		abci.Validator{Address: bytes.HexBytes{}, Power: int64(1)}, uint64(1), stakingtypes.Downtime,
 	)
 
-	packet := channeltypes.NewPacket(packetData.GetBytes(), 1, providertypes.PortID, suite.path.EndpointB.ChannelID,
```
smarshall-spitzbart, line 271: This packet construction logic was changed. per [spec](https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#processing-acknowledgements), it seems like ```acknowledgePacket``` executes on an ack of a packet originally sourced from this (consumer) chain. The provider chain's port/channel should not be "source" as existed here previously. Lmk if I'm misunderstanding the spec, or the original intention of this test

x/ccv/consumer/keeper/validators_test.go
```diff
@@ -93,60 +97,79 @@ func (k KeeperTestSuite) TestApplyCCValidatorChanges() {
 			changes: []abci.ValidatorUpdate{
 				{PubKey: changes[0].PubKey, Power: changes[0].Power + 1},
 				{PubKey: changes[1].PubKey, Power: changes[1].Power + 2},
-				{PubKey: changes[2].PubKey, Power: changes[1].Power + 3},
-				{PubKey: changes[3].PubKey, Power: changes[1].Power + 4},
+				{PubKey: changes[2].PubKey, Power: changes[2].Power + 3},
```
smarshall-spitzbart, line 100: Here's another change in testing logic. Previously the IBC testing suite had all validator powers start equal. So ```changes[1].Power``` vs ```changes[2].Power``` didn't matter. 

In the new test, 4 validators start out with powers of 1,2,3, and 4 respectively. So indexing here now matters

x/ccv/consumer/keeper/relay_test.go
jtremback: @mpoke I don't understand the code this is testing (OnRecvPacketOnUnknownChannel)

mpoke: See `utils.OnRecvPacketOnUnknownChannel`.
```
// OnRecvPacketOnUnknownChannel handles a received packet that was sent
// not on an established CCV channel
```
There was duplicated code on both consumer and provider and I decided to create a helper function.

mpoke (approved):
LGTM. Nice work @smarshall-spitzbart. Could you please add also comments describing what each UT is doing?

smarshall-spitzbart:
> LGTM. Nice work @smarshall-spitzbart. Could you please add also comments describing what each UT is doing?

Thanks for the review, comments on various unit/e2e tests related to this PR have been added in https://github.com/cosmos/interchain-security/pull/274/commits/02b3deeb7c20e05524e3bb06965a01794d2e2b38

x/ccv/consumer/keeper/relay_test.go
jtremback: I don't understand why this scenario would ever be exposed to our application logic, given how many layers of abstraction exist in IBC. Why would the IBC module ever give this packet to us? This seems like if you had to write web server code to manually reject all packets with the wrong IP on them instead of that happening in the kernel.

x/ccv/provider/keeper/keeper_test.go
jtremback: Nice

jtremback: Is this whole string of calls done by `providerKeeper.HandleSlashPacket`?

jtremback (approved):
LGTM

x/ccv/consumer/keeper/relay_test.go
mpoke: So do we assume that for example this condition is never triggered? https://github.com/cosmos/interchain-security/blob/eaa7f6e7ef98fa16b38e6213a194506dc9d50f26/x/ccv/provider/keeper/relay.go#L37

In other words, there is no chance of receiving a packet on an unexpected channel?

jtremback: I'm honestly not sure, but I'd be disappointed if we had to handle stuff like this ourselves.

x/ccv/provider/keeper/keeper_test.go
smarshall-spitzbart: Yep! All within that method

x/ccv/consumer/keeper/relay_test.go
mpoke: See https://github.com/cosmos/interchain-security/issues/292 for a more detail description of these scenarios.

PR #275, opened 2022-08-18
Differential testing for Interchain Security core protocol - (version 1 of diff testing)

danwt:
## Overview

Version 1 of a differential testing approach to testing the 'core' aspects of the Interchain Security system.

This PR contains two major components

1. A Typescript project containing modelling code and code for generating traces for differential testing
2. A golang test driver which is called from a unit test, which runs traces read from a json file, and uses them to test the system

All of the included files are necessary for a full understanding, however, some files contain the most information.

1. Please find a [README file](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/diff-test/core/README.md) which is contained within the Typescript project. This readme file explains the intentions of the work, and gives usage instructions for adapting the model, and generating traces using it.

2. [`model.ts`](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/diff-test/core/src/model.ts) contains the majority of the semantic logic used for modelling the system.

3. [`main.ts`](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/diff-test/core/src/main.ts) contains the entry point for interacting with the model and generating traces.

4. [`properties.ts`](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/diff-test/core/src/properties.ts) contains the code used to check some properties of the protocol, most importantly the `bond based consumer voting power` property.

5. [`diff_test.go`](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/x/ccv/difftest/diff_test.go) contains the core driver code used to read json traces and execute them against real instances of the system, simulated using ibc-go/testing. The pivotal unit test is [here](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/x/ccv/difftest/diff_test.go#L786-L814). There is a second unit test [here](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/x/ccv/difftest/diff_test.go#L842-L981) which serves only to check that the driver is written correctly.

I suggest to read the README first, and then decide which to read next based on your interest.

## Gathering feedback

All feedback and especially criticism is welcome!

I am aware of a few issues already. The most immediate problem with this work, as I see it, is the very complicated setup needed in the test driver, to get the SUT to a state where the trace can begin being executed. For example the [`zeroState.go`](https://github.com/cosmos/interchain-security/blob/e34060fe58/x/ccv/difftest/zeroState.go) file is dedicated to this setup, and is very long, it was adapted from an existing function in ibc-go/testing. I am also aware that the [test suite setup in `diff_test.go`](https://github.com/cosmos/interchain-security/blob/e34060fe58e438f6ff3faf35e98c0f7a06b36ea8/x/ccv/difftest/diff_test.go#L209-L354) is also very complex and long. I do think it is valuable work to simplify all of this setup, however, right now I think my time might be better spent elsewhere. I'm also aware of the large number of magic constants in the code. However, this is very hard to remove before improving the aforementioned overall complexity of the setup code. 

I'm happy to present the work to the team in a 1-many session or in individual sessions as needed, and of course please send me a slack message for any questions.


diff-test/core/.github/FUNDING.yml
```diff
@@ -0,0 +1,12 @@
+# These are supported funding model platforms
```
jtremback, line 1: What is this lol?

diff-test/core/__tests__/gen.test.ts
```diff
@@ -0,0 +1,8 @@
+import { gen } from '../src/main.js';
```
jtremback, line 1: What's this?

diff-test/core/README_typescript_boilerplate.md
```diff
@@ -0,0 +1,64 @@
+# node-typescript-boilerplate
```
jtremback, line 1: Can we get a real readme of the diff testing in here, or at least delete this file?

diff-test/core/src/sanity.ts
```diff
@@ -0,0 +1,51 @@
...
+ * For example, we should not ever
+ *    - allow an empty validator set
+ *    - cause a tendermint light client to expire
+ */
+class Sanity {
+  // the timestamp contained in the latest trusted header
+  tLastTrustedHeader = _.object([
+    [P, 0],
+    [C, 0],
+  ]) as { provider: number; consumer: number };
```
jtremback, line 17: Why not just write `{P: 0, C: 0}`?

diff-test/core/src/sanity.ts
```diff
@@ -0,0 +1,51 @@
...
+  // the timestamp of the last committed block
+  tLastCommit = _.object([
```
jtremback, line 19: Ditto

jtremback:
Just a comment on the typescript. You have used a .tsconfig file that turns off most of typescript's type checking. I've recommended some changes on that file to turn typescript back on. Once you do that, you will need to go through and correct a very large number of compile errors. This should actually be pretty quick.

But if you want, I can recommend another option which will make the code even better. Use Deno (https://deno.land/), which is a js runtime that has typescript built in. Then you will no longer need the build steps in the readme, and you will have a standard typescript config without needing a .tsconfig file.

diff-test/core/src/main.ts
```diff
@@ -0,0 +1,437 @@
...
+  infractionHeight: number;
+  isDowntime: number;
+};
+
+/**
+ * Takes an object where values are probabilities and returns a random
+ * key based on the distribution.
+ */
+function weightedRandomKey(distr) {
+  const scalar = _.reduce(_.values(distr), (sum, y) => sum + y, 0);
```
jtremback, line 68: JS has a built in reduce. In general, you probably shouldn't need to import underscore anywhere in this project. It is a relic from the days (10+ years ago) when the js stdlib was very minimal.

jtremback: Also Object.values()

diff-test/core/src/main.ts
```diff
@@ -0,0 +1,437 @@
...
+  const x = Math.random() * scalar;
+  const pairs = _.pairs(distr);
```
jtremback, line 70: Object.entries()

diff-test/core/src/main.ts
```diff
@@ -0,0 +1,437 @@
...
+  };
+
+  // Fill out template for a selected Delegate action
+  selectDelegate = (a): Delegate => {
+    return { ...a, amt: _.random(1, 5) * TOKEN_SCALAR };
+  };
+
+  // Return templates for possible undelegate actions
+  candidateUndelegate = (): Action[] => {
+    return _.range(NUM_VALIDATORS).map((i) => {
```
jtremback, line 158: [...Array(NUM_VALIDATORS).keys()].map((i) => {...})

diff-test/core/src/main.ts
```diff
@@ -0,0 +1,437 @@
...
+      return {
+        kind: 'Delegate',
+        val: i,
+      };
+    });
+  };
+
+  // Fill out template for a selected Delegate action
+  selectDelegate = (a): Delegate => {
+    return { ...a, amt: _.random(1, 5) * TOKEN_SCALAR };
```
jtremback, line 153: 1 + (Math.random() * 4)

diff-test/core/src/main.ts
```diff
@@ -0,0 +1,437 @@
...
+      });
+  };
+
+  // Fill out a template for a selected Consumer initiated slash action
+  selectConsumerSlash = (a): ConsumerSlash => {
+    this.didSlash[a.val] = true;
+    return {
+      ...a,
+      infractionHeight: Math.floor(Math.random() * this.model.h[C]),
+      isDowntime: _.sample([true, false]),
```
jtremback, line 195: Math.random() > 0.5

diff-test/core/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "alwaysStrict": true,
+    "sourceMap": true,
+    "forceConsistentCasingInFileNames": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": true,
+    "noImplicitAny": false,
+    "noImplicitThis": false,
+    "strictNullChecks": false,
```
jtremback, line 22: "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "strictNullChecks": true,

diff-test/core/src/model.ts
```diff
@@ -0,0 +1,734 @@
...
+  /**
+   * Compute the new set of active validators
+   */
+  newVals = () => {
+    const valid = (i): boolean =>
+      1 <= this.tokens[i] && this.jailed[i] === undefined;
+    let vals = _.range(NUM_VALIDATORS);
+    // stable sort => breaks ties based on validator
+    // address numerical value. This mimics staking module.
+    vals = _.sortBy(vals, (i) => -this.tokens[i]);
```
jtremback, line 214: Why not use the built in `.sort`?

diff-test/core/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+The actual functionality has entrypoint in [src/main.ts](./src/main.ts). Please see the file for details. The available functionalities are
+
+```bash
+# generate traces
+yarn start gen <num minutes>
+# create a subset of traces
+yarn start subset <output file abs path> <num event instances (optional)>
+# replay a trace from a file (for debugging)
+yarn start replay <filename> <trace index> <num actions>
```
jtremback, line 99: Just an idea, maybe it could be possible to put models like this into a cli to allow for interactive exploration and verfication of the model's behavior

jtremback:
Beyond the superficial JS comments, there's a huge amount to take in here. Before I approve, we should go over this on a call

diff-test/core/.github/FUNDING.yml
danwt: 👍

diff-test/core/__tests__/gen.test.ts
danwt: It's useful to have a test to check model coverage. I've added a comment now.

diff-test/core/README_typescript_boilerplate.md
danwt: There is one, I can also delete this file.

https://github.com/cosmos/interchain-security/blob/a774948f10c1d26c1afd4dc7ceaa6d6b2f8d6aa0/diff-test/core/README.md

diff-test/core/src/sanity.ts
danwt: This would create an object mapping the characters 'P' to 0 and 'C' to 0 but I want the words 'provider', 'consumer'.

diff-test/core/src/sanity.ts
danwt: ^^

danwt:
> ...
> But if you want, I can recommend another option which will make the code even better. Use Deno (https://deno.land/), which is a js runtime that has typescript built in. Then you will no longer need the build steps in the readme, and you will have a standard typescript config without needing a .tsconfig file.

Hmm Ok let's discuss Deno in a call. I was aware of it but haven't used it as AFIU it's still a bit non standard and nodejs is still 10000x more popular.

danwt:
> Just a comment on the typescript. You have used a .tsconfig file that turns off most of typescript's type checking. I've recommended some changes on that file to turn typescript back on. Once you do that, you will need to go through and correct a very large number of compile errors. This should actually be pretty quick.

Could you please be more precise? You are right that these

```
    "noUnusedLocals": false,
    "noUnusedParameters": true,
    "noImplicitAny": false,
    "noImplicitThis": false,
    "strictNullChecks": false,
```

should probably be true. Is there anything else?

Ill change the above.

diff-test/core/src/main.ts
danwt: > JS has a built in reduce. In general, you probably shouldn't need to import underscore anywhere in this project. It is a relic from the days (10+ years ago) when the js stdlib was very minimal.

Ok, I'll make a pass to see if I can do things in vanilla. I got the impression underscore is really quite helpful though.

danwt: Already found one example where underscore seems to save writing a lot of code

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_pick

danwt: I do take your point though, which is that I've probably overused it.

diff-test/core/src/main.ts
danwt: That's not an improvement IMO

diff-test/core/src/model.ts
danwt: That doesn't sort numbers

diff-test/core/src/main.ts
danwt: It would have to be `1 + Math.floor(Math.random() * 5)` which I'm not sure is bettter

diff-test/core/src/model.ts
danwt: Yep, good point. It's probably like this because I was using sort at one point, not realising it did string conversion and then when I fixed it didn't check if there was an optional lambda arg.

danwt: Actually, no, just remembered. I did it because underscore sort is guaranteed stable. I didn't see stable guarantee written anywhere in js version.

diff-test/core/README.md
danwt: I'm not sure I understand what you mean. Like submitting actions 1 by 1 via cli?

danwt:
> Beyond the superficial JS comments, there's a huge amount to take in here. Before I approve, we should go over this on a call

👍

danwt:
Closing to make somewhat substantial changes, will open new PR.

PR #277, opened 2022-08-23
Integration tests with local sdk version

smarshall-spitzbart:
Enables a user to specify a local path to their cosmos sdk repository when running integration tests. The validators spawned for integration tests will then reference the local sdk, allowing for better debugging. This implementation is not perfect, but it gets the job done, and unblocks some issues I've been having with debugging integration tests.


integration-tests/testnet-scripts/start-docker.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Delete old sdk directory if it exists 
+if [ -d "./cosmos-sdk" ]; then
+    rm -rf ./cosmos-sdk/
+fi 
+
+# Copy sdk directory to working dir if path was specified
+if [[ "$LOCAL_SDK_PATH" != "default" ]]
+then
+    cp -n -r "$LOCAL_SDK_PATH" ./cosmos-sdk
```
smarshall-spitzbart, line 24: This implementation copies the local sdk repo into the ICS repo, pipes it to docker, then deletes the copy. There are alternatives to this scheme, but the chosen solution seemed the simplest and standardizable.

We could also try building the docker image outside the ICS repo to allow ```ADD``` commands with external directories

danwt:
This will be so useful!

Dockerfile
```diff
@@ -13,8 +13,14 @@ WORKDIR /downloads
 # Copy in the repo under test
 ADD . /interchain-security
 
+WORKDIR /interchain-security
```
jtremback, line 16: Nice

integration-tests/testnet-scripts/start-docker.sh
danwt: love the `cp -n` 👍

danwt (approved):

Issue #281, opened 2022-08-23
Check counterparty port id during the channel handshake


mpoke:
Check counterparty port id during the channel handshake, i.e., https://github.com/cosmos/interchain-security/blob/eccbfaae9cdc39a769d94814eece02a2eafe333f/x/ccv/parent/module.go#L236 and https://github.com/cosmos/interchain-security/blob/eccbfaae9cdc39a769d94814eece02a2eafe333f/x/ccv/child/module.go#L221

See spec, i.e., https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-ccf-coinit1 and https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-cotry1

PR #287, opened 2022-08-23
Check counterparty port id during the channel handshake

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/281

Before merging: 
- [x] add UTs for the channel opening handshake on the provider
- [x] improve coverage of `ConsumerTestSuite.TestOnChanOpenAck`, see https://github.com/cosmos/interchain-security/blob/eaa7f6e7ef98fa16b38e6213a194506dc9d50f26/x/ccv/consumer/module.go#L305

### DIFF
--- a/x/ccv/types/keys.go
+++ b/x/ccv/types/keys.go
@@ -8,8 +8,11 @@ const (
 	// module supports
 	Version = "1"
 
-	// ProviderPortID is the default port id that provider CCV module binds to
+	// ProviderPortID is the default port id the provider CCV module binds to
 	ProviderPortID = "provider"
+
+	// ConsumerPortID is the default port id the consumer CCV module binds to
+	ConsumerPortID = "consumer"
 
 	RouterKey = ModuleName
 
--- a/x/ccv/types/keys.go
+++ b/x/ccv/types/keys.go
@@ -8,6 +8,9 @@ const (
 	// module supports
 	Version = "1"
 
+	// ProviderPortID is the default port id that provider CCV module binds to
+	ProviderPortID = "provider"
+
 	RouterKey = ModuleName
 
 	// StoreKey defines the primary module store key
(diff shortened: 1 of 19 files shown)


jtremback:
> Closes https://github.com/cosmos/interchain-security/issues/281

Before merging: 
- [ ] add UTs for the channel opening handshake on the provider
- [ ] improve coverage of `ConsumerTestSuite.TestOnChanOpenAck`, see https://github.com/cosmos/interchain-security/blob/eaa7f6e7ef98fa16b38e6213a194506dc9d50f26/x/ccv/consumer/module.go#L305 

Can you mark this draft?

x/ccv/types/keys.go
```diff
@@ -8,6 +8,12 @@ const (
 	// module supports
 	Version = "1"
 
+	// ProviderPortID is the default port id the provider CCV module binds to
```
smarshall-spitzbart, line 11: It may be a good idea to change the hardcoded string to ```"providerPort"``` and ```"consumerPort"``` or something similar, to avoid having the same value as [ModuleName](https://github.com/cosmos/interchain-security/blob/marius/281-cp-portid/x/ccv/provider/types/keys.go#L18) for example

smarshall-spitzbart:
I can take a stab at unit tests since I've touched that code a lot lately. I'll have to think about the best merging order between this PR and the final step of #219 (actually moving all the e2e tests). It seems there will be a lot of merge conflicts in either case :)

x/ccv/types/keys.go
mpoke: I don't think it's necessary. AFAIK, there is no risk of collision between IBC ports and module names. See the ibc-go implementation of ICS-20 https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/apps/transfer/types/keys.go#L12

mpoke:
@smarshall-spitzbart a source of inspiration for the channel opening handshake UTs could be found here https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/apps/transfer/ibc_module_test.go#L15

jtremback (approved):

smarshall-spitzbart:
See https://github.com/cosmos/interchain-security/pull/307

Issue #263, opened 2022-08-11
Implement full-node for integration test queries


smarshall-spitzbart:
There are a lot of places within integration test code that validators are queried for chain state [like so](https://github.com/cosmos/interchain-security/blob/test-validator-downtime/integration-tests/state.go#L79). Currently, the first validator in a config list fulfills these queries. However, as we add validator downtime testing, these queries can fail if the desired validator is not operational. 

Solution: a dedicated non validating, full node should be spawned up for integration tests. This node will never be taken down during the test, and can fulfill all queries

PR #302, opened 2022-08-26
Integration test refactors

smarshall-spitzbart:
Various refactors to integration tests that'll make future additions easier


integration-tests/testnet-scripts/start-chain.sh
```diff
@@ -9,7 +9,8 @@ BIN=$1
 #     mnemonic: "crackle snap pop ... etc",
 #     allocation: "10000000000stake,10000000000footoken",
 #     stake: "5000000000stake",
-#     number: "0",
+#     val_id: "1",
```
jtremback, line 12: "val_id" probably won't be a number in most cases. Something like "al" might be better to demonstrate.

integration-tests/config.go
```diff
@@ -6,23 +6,40 @@ import (
 	"time"
 )
 
-// Attributes that are unique to a validator. Allows us to map (part of) the set of uints to
-// a set of viable validators
+type chainID string
```
jtremback, line 9: matter of opinion, but let's not do the marker types for now

integration-tests/config.go
```diff
@@ -6,23 +6,40 @@ import (
...
+type validatorID string
+
+const (
+	// Strings mapped to validator configs, literals must be two characters long for network config
+	alice = validatorID("al")
```
jtremback, line 14: IMO you should just use the string "al" everywhere, not define a variable

integration-tests/config.go
smarshall-spitzbart, line 9: add comment about maybe removing these types in future

integration-tests/config.go
```diff
@@ -6,23 +6,40 @@ import (
 	"time"
 )
 
-// Attributes that are unique to a validator. Allows us to map (part of) the set of uints to
-// a set of viable validators
+type chainID string
+type validatorID string
+
+const (
+	// Strings mapped to validator configs, literals must be two characters long for network config
```
jtremback, line 13: This is a requirement imposed by the fact that i was naming interfaces veth-$CHAIN_ID-$VAL_ID-out etc. We can change this to $CHAIN_ID-$VAL_ID-out, freeing up 5 characters and making it so that each string literal can be 5 chars or less.

integration-tests/config.go
jtremback: Because otherwise it makes it harder to debug by looking at what is inside the system

integration-tests/config.go
smarshall-spitzbart: I'll also add a validation function to enforce this

integration-tests/config.go
```diff
@@ -34,69 +51,73 @@ type ContainerConfig struct {
...
 		},
-		chainConfigs: []ChainConfig{
-			{
-				chainId:        "provider",
+		chainConfigs: map[chainID]ChainConfig{
+			provider: {
+				chainId:        provider,
 				binaryName:     "interchain-security-pd",
-				ipPrefix:       "7.7.7",
+				ipPrefix:       "7.7.7", // TODO: Change this
```
jtremback, line 105: take away this todo

integration-tests/state.go
```diff
@@ -283,3 +283,21 @@ func (s System) getValPower(chain uint, validator uint) uint {
 
 	return 0
 }
+
+// TODO: Best solution for default validator fulfilling queries etc. is a dedicated, non validating, full node.
+// See https://github.com/cosmos/interchain-security/issues/263
+func (tr TestRun) getDefaultValidator(chain chainID) validatorID {
+	return alice
```
jtremback, line 290: This won't work

jtremback: Stick to the old way, or just implement https://github.com/cosmos/interchain-security/issues/263 in this PR

jtremback (approved):
Conditionally approved with my changes

PR #303, opened 2022-08-29
Move IBC handlers to ibc-module.go

mpoke:
Refactoring of IBC handlers similar to the ibc-go implementation if ICS-20, i.e., https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go.


smarshall-spitzbart (approved):
Looks good 👍 seems like there's no functional changes in this PR

jtremback (approved):

PR #305, opened 2022-08-29
Make it so that LockUnbondingOnTimeout can only equal false

jtremback:
LockUnbondingOnTimeout is a setting which dictates whether unbonding tokens should be locked if a consumer chain times out. Since this would be very disruptive to the Cosmos Hub, we need to write more tests to make sure it works correctly, and it is unclear whether it is a use case we would ever want to use on the Hub, I am disabling it.


mpoke (approved):
Could we open an issue to not forget to get back to this? 

In general, I still believe that we shouldn't take lightly the unbonding period. This is at the core of the PoS mechanism (which provides security). Setting `lockUbdOnTimeout` to false MAY violate the _Bond-Based Consumer Voting Power_ and _Slashable Consumer Misbehavior_ properties (see the [System Properties](https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md#system-properties) section). For more details, see the Note under the spec of the `StopConsumerChain`  method https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-stcc1.

danwt:
I think we could change the security model.

If we set `lockUnbondingOnTimeout === false (always)` then a timeout will free up provider unbondings, and as @mpoke says, I think this will violate the properties for the consumer. This is OK if we change our guarantees to something like

'When the consumer has not timed out, it benefits from properties _Bond-Based Consumer Voting Power_ ect, but once it has timed out, the consumer can have arbitrary behavior.'

danwt (approved):
Agree we should track this carefully

Issue #310, opened 2022-08-30
Integration tests should allow multiple test runs


smarshall-spitzbart:
As-is, integration tests run a single "TestRun". We need to expand integration tests to allow multiple test runs to execute (ideally concurrently, in their own docker containers).

TestRun: a set of actions and related state checks that run synchronously in a specific order

smarshall-spitzbart:
see #371

PR #313, opened 2022-08-30
Validator downtime integration tests

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/256 using learnings from #266 

Note: this PR increases the local runtime of integration tests from ```6m53.668667917s``` on my machine, compared to main's time of ```2m33.332940625s```. 

There are various constraints in this PR's implementation that cause so much wait time, example: ```slash_fraction_downtime``` must be at least 60sec. However, solving #310 would improve performance drastically by executing ```TestRun```s steps concurrently in independent docker containers


jtremback:
> slash_fraction_downtime must be at least 60sec

Why?

smarshall-spitzbart:
> > slash_fraction_downtime must be at least 60sec
> 
> Why?

Constrained [here](https://github.com/cosmos/cosmos-sdk/blob/interchain-security-rebase/x/slashing/types/genesis.go#L58), not sure on the intuition why that value was set. But here's a draft PR if it's ok to change https://github.com/cosmos/cosmos-sdk/pull/13106

integration-tests/steps.go
```diff
@@ -240,12 +240,14 @@ var happyPathSteps = []Step{
 			chain:      chainID("provi"),
 			unbondFrom: validatorID("alice"),
 			sender:     validatorID("alice"),
-			amount:     11000000,
+			// Leave alice with majority stake so non-faulty validators will maintain more than
+			// 2/3 voting power during downtime tests below.
```
jtremback, line 244: Add something about avoiding a chain halt in this comment

integration-tests/steps.go
```diff
@@ -269,12 +271,199 @@ var happyPathSteps = []Step{
...
+				ValPowers: &map[validatorID]uint{
+					validatorID("alice"): 510,
+					validatorID("bob"):   500,
+					validatorID("carol"): 500,
+				},
+			},
+			chainID("consu"): ChainState{
+				ValPowers: &map[validatorID]uint{
+					validatorID("alice"): 510,
+					validatorID("bob"):   500,
```
jtremback, line 300: @mpoke random question- i don't remember which way we ended up going on how the downtime slashing works, but just wanted to make sure that the power wasn't supposed to be reduced immediately as soon as the downtime happened on the consumer

integration-tests/steps.go
```diff
@@ -269,12 +271,199 @@ var happyPathSteps = []Step{
...
-					validatorID("alice"): 500,
+					validatorID("alice"): 510,
+					validatorID("bob"):   500,
+					validatorID("carol"): 500,
+				},
+			},
+		},
+	},
+	{
+		action: SlashAction{
```
jtremback, line 282: I would call this DowntimeSlashAction

integration-tests/actions.go
```diff
@@ -616,3 +616,83 @@ func (tr TestRun) unbondTokens(
...
+}
+
+type RestoreVotingPowerAction struct {
+	bringUpOn chainID
+	provider  chainID
+	toRestore validatorID
+}
+
+func (tr TestRun) RestoreVotingPower(action RestoreVotingPowerAction, verbose bool) {
+	tr.ToggleValidatorDowntime(action.bringUpOn, action.toRestore, false, verbose)
```
jtremback, line 638: I'd move this to InvokeSlash

integration-tests/actions.go
```diff
@@ -616,3 +616,83 @@ func (tr TestRun) unbondTokens(
...
+	// Wait the downtime_jail_duration set in config.go
+	time.Sleep(61 * time.Second)
```
jtremback, line 640: This too

integration-tests/main.go
```diff
@@ -54,6 +54,10 @@ func (tr TestRun) runStep(step Step, verbose bool) {
 		tr.delegateTokens(action, verbose)
 	case UnbondTokensAction:
 		tr.unbondTokens(action, verbose)
+	case SlashAction:
+		tr.InvokeSlash(action, verbose)
```
jtremback, line 58: In general, only export a method if necessary. i.e. this should probably be lower case

jtremback (approved):

integration-tests/main.go
smarshall-spitzbart: I went ahead and changed each integration test action and corresponding functions to be non-exported

integration-tests/actions.go
smarshall-spitzbart: I had to move this line back to the other action, ```unjailValidator```, because the wait time isn't effective until the validator jailing is actually processed by the provider (after a slash packet is relayed from consumer -> provider).

Alternatively I can relay a packet within ```downtimeSlashAction```, but that's not consistent with other actions

integration-tests/steps.go
mpoke: That's right. An infraction on the consumer causes a SlashRequest, which eventually causes a Slash on the provider, which eventually causes the power on the consumer to be reduced.

danwt:
> slash_fraction_downtime must be at least 60sec

This variable is referring to a value in [0,1], the multiplier for burning tokens. It's not a duration. Right?

smarshall-spitzbart:
> > slash_fraction_downtime must be at least 60sec
> 
> This variable is referring to a value in [0,1], the multiplier for burning tokens. It's not a duration. Right?

Oof yes, I was referring to ```downtime_jail_duration``` this whole time, not ```slash_fraction_downtime```, good catch 👍

smarshall-spitzbart:
Update: with https://github.com/cosmos/cosmos-sdk/pull/13106 being merged, and the addition of https://github.com/cosmos/interchain-security/pull/313/commits/7ec2f54369d2cd5349dbc47d01a538e44b09ffe4 to this PR, the integration tests now take ```4m14.531286792s``` on my machine. Saving us 2.5 minutes, woot!

jtremback (approved):

danwt:
@smarshall-spitzbart what's the situation with this?

>  Seems the sdk code owners are against changing the relevant constraint values

I see you changed the genesis minimums in our branch on the sdk. Is that change OK or do we still need another solution?

Re.

> Setup a package wrapper around sdk to overwrite these values
Use sed to edit referenced source code for the sdk (only for integration tests)
Wait for a solution to
https://github.com/cosmos/interchain-security/issues/310 to deal with the increase in test run time

I wonder if, generally speaking, there should exist a cheatcode module with full object capabilities. It would let you change any state/parameter and could be called via transactions. That way for integration tests we'd just send a tx to that module to change something. For release, you just wouldn't include the module.

integration-tests/actions.go
```diff
@@ -655,3 +655,84 @@ func (tr TestRun) redelegateTokens(action RedelegateTokensAction, verbose bool)
...
+	validator validatorID
+}
+
+func (tr TestRun) invokeDowntimeSlash(action downtimeSlashAction, verbose bool) {
+	// Bring validator down
+	tr.toggleValidatorDowntime(action.chain, action.validator, true, verbose)
+	// Wait appropriate amount of blocks for validator to be slashed
+	tr.waitBlocks(action.chain, 3, time.Minute)
+	// Bring validator back up
+	tr.toggleValidatorDowntime(action.chain, action.validator, false, verbose)
```
danwt, line 670: Toggle usually implies an on/off switch (where you don't need to pass the parameter). But here a parameter is passed so another word might work better, like 'set', for example.

integration-tests/testnet-scripts/start-docker.sh
```diff
@@ -34,5 +34,5 @@ docker build -t "$CONTAINER_NAME" .
 rm -rf ./cosmos-sdk/
 
 # Run new test container instance
-docker run --name "$INSTANCE_NAME" --cap-add=NET_ADMIN "$CONTAINER_NAME" /bin/bash /testnet-scripts/beacon.sh
-`
\ No newline at end of file
+docker run --name "$INSTANCE_NAME" --cap-add=NET_ADMIN --privileged "$CONTAINER_NAME" /bin/bash /testnet-scripts/beacon.sh
```
danwt, line 37: Nit: it would be nice to have a comment explaining why priveledged is there

danwt (approved):
Looks great! It's really nice to have consumer initiated slashing tested in integration tests. It's a bit hard for me to review the ip code properly because I don't know those APIs well, but if it works it works.

integration-tests/actions.go
smarshall-spitzbart: Yea this makes sense to me 👍, the ideal calling syntax would be that the method makes a call to ```ip link``` to determine whether the node is currently up or down, then toggle the node to it's opposite state. However, I don't want to add more complexity/runtime to the method for small syntax gains. "Set" it is

smarshall-spitzbart:
> @smarshall-spitzbart what's the situation with this?
> 
> > Seems the sdk code owners are against changing the relevant constraint values
> 
> I see you changed the genesis minimums in our branch on the sdk. Is that change OK or do we still need another solution?
> 
> Re.
> 
> > Setup a package wrapper around sdk to overwrite these values
> > Use sed to edit referenced source code for the sdk (only for integration tests)
> > Wait for a solution to
> > #310 to deal with the increase in test run time
> 
> I wonder if, generally speaking, there should exist a cheatcode module with full object capabilities. It would let you change any state/parameter and could be called via transactions. That way for integration tests we'd just send a tx to that module to change something. For release, you just wouldn't include the module.

On lowering the time it takes for tests to run, this was solved by https://github.com/cosmos/interchain-security/pull/313#issuecomment-1234776278 and the commit/PR that comment refers to

On your cheatcode module comment: This is an interesting idea! The reason I tried to stay away from wrapper modules is that imo, integration tests should run with production software/capabilities. Mocks, wrapped modules, and other "non-production" manipulations could be fine for unit/e2e/diff tests, but we should strive for integration tests being as close to the real thing as we can make it. That way integration tests serve a purpose that the other layers of testing cannot

danwt:
> integration tests being as close to the real thing as we can make it.

That's true. A cheatcode module would kind of go against that!

Issue #300, opened 2022-08-25
Redundant e2e test suite structs


smarshall-spitzbart:
Before https://github.com/cosmos/interchain-security/pull/297 we had e2e tests spread between multiple directories, in which 4 test suites were created with redundant code in setup methods, etc.

Now that all e2e tests will live within a single go package and directory, we can consolidate the 4 existing IBC test suite data structures into a single test suite struct. Note that some helper functions will need to be refactored as well.

PR #315, opened 2022-08-31
Difference testing for 'core' protocol using model generated (heuristic) traces

danwt:
## Overview

Version 3 of a differential testing approach to testing the 'core' aspects of the Interchain Security system.

This PR contains three major components

1. A 'simibc' testutil directory containing reusable components for testing ibc connected chains in a fine-grained way ([here](https://github.com/cosmos/interchain-security/tree/danwt%2Fdifftest-v3-reb0/testutil/simibc))
2. A Typescript project containing modelling code and code for generating traces for difference testing ([here](https://github.com/cosmos/interchain-security/tree/danwt%2Fdifftest-v3-reb0/diff-models/core))
3. A golang test driver which is called from a unit test, which runs traces read from a json file, and uses them to test the system ([here](https://github.com/cosmos/interchain-security/tree/danwt%2Fdifftest-v3-reb0/diff-tests/core))

All of the included files are necessary for a full understanding, however, some files contain the most information.

1. Please find a [README file](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-models/core/README.md) which is contained within the Typescript project. This readme file explains the intentions of the work, and gives usage instructions for adapting the model, and generating traces using it.

2. [`model.ts`](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-models/core/src/model.ts) contains the majority of the semantic logic used for modelling the system.

3. [`main.ts`](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-models/core/src/main.ts) contains the entry point for interacting with the model and generating traces.

4. [`properties.ts`](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-models/core/src/properties.ts) contains the code used to check some properties of the protocol, most importantly the `bond based consumer voting power` property.

5. [`core_test.go`](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-tests/core/core_test.go) contains the core driver code used to read json traces and execute them against real instances of the system, simulated using ibc-go/testing. The pivotal unit test is [here](https://github.com/cosmos/interchain-security/blob/f87496d109800ed4f766ebe7001a9ed4210976c6/diff-tests/core/core_test.go#L419-L443). There is a second unit test [here](https://github.com/cosmos/interchain-security/blob/f87496d109800ed4f766ebe7001a9ed4210976c6/diff-tests/core/core_test.go#L291-L417) which serves only to check that the driver is written correctly.

I suggest to read the README first, and then decide which to read next based on your interest.

## Gathering feedback

All feedback and especially criticism is welcome!

I'm aware of some major issues already but I think now is a sensible time to merge and to leave improvements to future work.

### Code quality

- diff-tests/core/setup.go is an absolute monster
- semi-duplicated efforts with simibc and the dependency ibc-go/testing
- model init state is hardcoded in the typescript and the go code, instead of read into the go code
- a few semi-duplicated helper methods between diff-tests/core/core_test.go and diff-tests/core/setup.go
- lack of comments

### Test capabilities

- Testing what happens when assumptions are broken
- Other, see [this text](https://github.com/cosmos/interchain-security/blob/danwt/difftest-v3-reb0/diff-models/core/README.md)


testutil/simibc/network_link.go
```diff
@@ -0,0 +1,93 @@
...
+// Packet represents a packet committed to block state
+type Packet struct {
+	Packet  channeltypes.Packet
+	Commits int
+}
+
+// NetworkLink contains outboxes of packets and acknowledgements and
+// allows fine-grained control over delivery of acks and packets
+// to mimic a real relaying relationship between two chains.
+type NetworkLink struct {
```
jtremback, line 21: Awkward naming IMO. This isn't simulating a network link, but rather an IBC link, right? This name makes it sound like it is simulating an ethernet device or something.

diff-models/core/__tests__/gen.test.ts
```diff
@@ -0,0 +1,12 @@
+import { gen } from '../src/main.js';
```
jtremback, line 1: Is the stuff in this folder necessary?

diff-models/core/src/common.ts
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Record a snapshot of both chain states at the height and timestamp
+ * of a committed block for a chain.
+ */
+interface CommittedBlock {
+  chain: Chain;
+  invariantSnapshot: InvariantSnapshot;
+}
+
+interface Consequence {
```
jtremback, line 125: I'd like to know more about what a Consequence is. Is it just the complete model state at any given time? Is it the parts of the model state that you care about checking, minus some "hidden" state?

diff-models/core/jest.config.js
```diff
@@ -0,0 +1,24 @@
+export default {
```
jtremback, line 1: Almost 100% certain this isn't needed. Can you remove all boilerplate that you don't have a specific purpose for from this folder?

diff-tests/core/core_test.go
```diff
@@ -0,0 +1,458 @@
...
+// Test a set of traces
+func (s *CoreSuite) TestTraces() {
+	s.traces = Traces{
+		Data: LoadTraces("traces.json"),
+	}
+	for i := range s.traces.Data {
+		s.Run(fmt.Sprintf("Trace num: %d", i), func() {
+			// Setup a new pair of chains for each trace
+			s.SetupTest()
+			s.traces.CurrentTraceIx = i
```
jtremback, line 428: This feels really awkward. Can't you just iterate the list and pass the current trace to executeTrace?

diff-tests/core/core_test.go
```diff
@@ -0,0 +1,458 @@
...
+	s.simibc.DeliverPackets(s.chainID(chain), numPackets)
+}
+
+func (s *CoreSuite) endAndBeginBlock(chain string) {
+	s.simibc.EndAndBeginBlock(s.chainID(chain), time.Second*6, func() {
+		s.matchState()
+	})
+}
+
+func (s *CoreSuite) matchState() {
```
jtremback, line 216: What does this do?

diff-tests/core/core_test.go
```diff
@@ -0,0 +1,458 @@
...
+				// and improve debugging experience.
+				if r := recover(); r != nil {
+					fmt.Println(s.traces.Diagnostic())
+					fmt.Println(r)
+					panic("Panic occurred during TestTraces")
+				}
+			}()
+			// Record information about the trace, for debugging
+			// diagnostics.
+			s.executeTrace()
```
jtremback, line 440: Where does the state after the action get checked (the "Consequence")? I see `matchState` being called in the `endAndBeginBlock`. Is that it? If it is, it seems odd to have your state being checked by one of your actions.

jtremback:
Looks good, excited to get this in. I'll approve it now, but can you get everything into one folder?

jtremback (approved):

diff-tests/core/core_test.go
danwt: We only compare committed block state because that's the only thing that matters at the end of the day. Anything that happens inside the block is implementation detail. I.e. as long as the same transactions result in the same committed block state we don't care what the go implementation actually does.

danwt: We could check after every action but IMO that would bring unnecessary restrictions and make it harder to make the model abstract and to make the implementation efficient

testutil/simibc/network_link.go
danwt: Hmm ok decent point.

diff-tests/core/core_test.go
danwt: It might be possible and I was just lazy. I'll check again. The thing is I want to keep the trace in 'global' state for diagnostics

diff-models/core/__tests__/gen.test.ts
danwt: The easiest way I know of to check the statement coverage of the trace generation on the model is to use `yarn jest --collect-coverage` which requires a jest test.
I don't think we should give this up

diff-models/core/src/common.ts
danwt: Good point, this is poorly documented. A consequence is a slice of model state relevant to the action that was last done. It's used for state comparison with the SUT and is helpful for understanding traces when reading through the actions. 
Previously I had much more state here, but it's too much and make the jsons huge and any problems really hard to debug.

diff-models/core/jest.config.js
danwt: Good point to remove unnecessary files, but this one is very much needed. Jest will not run without this.

diff-models/core/src/model.ts
```diff
@@ -0,0 +1,635 @@
...
+
+    {
+      // Check if 2/3 of power of old val set is included in new one
+      const lastPowerTheseVals = this.lastTokens.reduce(
+        (sum, x, i) => (vals.includes(i) ? sum + x : sum),
+        0,
+      );
+      const nowTotalPower = this.tokens.reduce((sum, x) => sum + x, 0);
+      if (lastPowerTheseVals < (2 / 3) * nowTotalPower) {
+        this.m.events.push(Event.MORE_THAN_ONE_THIRD_VAL_POWER_CHANGE);
```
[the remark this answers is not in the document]
danwt: I'll double check, I might have made a mistake here. It was a recent addition.

diff-models/core/src/model.ts
```diff
@@ -0,0 +1,635 @@
...
+    }
+    this.tokens[val] -= amt;
+    this.delegation[val] -= amt;
+    const und: Undelegation = {
+      val: val,
+      creationHeight: this.m.h[P],
+      completionTime: this.m.t[P] + UNBONDING_SECONDS_P,
+      balance: amt,
+      initialBalance: amt,
+      onHold: true,
```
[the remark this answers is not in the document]
danwt: Right. I'd say these are both extensions.

diff-models/core/src/model.ts
```diff
@@ -0,0 +1,635 @@
...
+  };
+}
+
+class Staking {
+  // Model handle
+  m;
+  // Validator delegations from the delegator
+  // A fixed descending order is used for the initial values to allow
+  // easy setup in the SUT.
+  delegation: number[];
```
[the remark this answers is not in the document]
danwt: Good question, I should definitely document this better.

Delegation is the delegation from the modelled delegator to the validator. Tokens is the total voting power of the validator. The voting power <-> token ratio is 1:1. There are more Tokens than Delegation because there is _ghost_ delegation. This is extra delegation that each validator has to stop it being deleted. 

This is needed because the SUT has a concept of minSelfDelegation. So this difference (tokens[i]-delegation[i]) is effectively minSelfDelegation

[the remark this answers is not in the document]
danwt: I agree it's not clear

diff-models/core/src/model.ts
```diff
@@ -0,0 +1,635 @@
...
+    this.delegation[val] -= amt;
+    const und: Undelegation = {
+      val: val,
+      creationHeight: this.m.h[P],
+      completionTime: this.m.t[P] + UNBONDING_SECONDS_P,
+      balance: amt,
+      initialBalance: amt,
+      onHold: true,
+      opID: this.opID,
+      expired: false,
```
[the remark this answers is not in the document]
danwt: I don't like matured because it is not clear if that means the unbonding maturations have been received from every consumer. Expired means it has been processed by the staking module EndBLock. I will change the name as such.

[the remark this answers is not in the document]
danwt: Updated code

https://github.com/cosmos/interchain-security/blob/0e3a9db8e7619f3d83f2bf8a711ee7c7d898258c/diff-tests/core/model/src/model.ts#L240-L248

diff-models/core/src/model.ts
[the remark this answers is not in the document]
danwt: Yeah I had fluffed it, can you check the new version?

https://github.com/cosmos/interchain-security/pull/315/commits/ef9046a72e2a1782fb66b1dcd6f0589825b6e1d7

testutil/simibc/relay_util.go
```diff
@@ -0,0 +1,166 @@
...
+	clienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
+	ibctmtypes "github.com/cosmos/ibc-go/v3/modules/light-clients/07-tendermint/types"
+	ibctesting "github.com/cosmos/ibc-go/v3/testing"
+	"github.com/cosmos/ibc-go/v3/testing/simapp"
+	"github.com/stretchr/testify/require"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
```
smarshall-spitzbart, line 15: These helpers are awesome!

smarshall-spitzbart (approved):
Very cool stuff!

diff-tests/core/driver/core_test.go
```diff
@@ -0,0 +1,462 @@
...
+	appProvider "github.com/cosmos/interchain-security/app/provider"
+
+	simibc "github.com/cosmos/interchain-security/testutil/simibc"
+
+	slashingkeeper "github.com/cosmos/cosmos-sdk/x/slashing/keeper"
+	consumerkeeper "github.com/cosmos/interchain-security/x/ccv/consumer/keeper"
+)
+
+type CoreSuite struct {
+	suite.Suite
```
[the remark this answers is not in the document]
danwt: Yeah I think it could work. I need to look into what suite actually offers more closely. I've used it so far as my original starting point was the old provider_test.go which used Suite. 

Would it be alright to merge as is and make the conversion to *testing.T an issue for future work?

diff-tests/core/driver/setup.go
```diff
@@ -0,0 +1,761 @@
...
+
+func (b *Builder) otherID(chainID string) string {
+	return map[string]string{ibctesting.GetChainID(0): ibctesting.GetChainID(1), ibctesting.GetChainID(1): ibctesting.GetChainID(0)}[chainID]
+}
+
+func (b *Builder) chain(chain string) *ibctesting.TestChain {
+	return map[string]*ibctesting.TestChain{P: b.providerChain(), C: b.consumerChain()}[chain]
+}
+
+func (b *Builder) providerChain() *ibctesting.TestChain {
```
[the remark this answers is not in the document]
danwt: I think so. I would prefer though to merge as is and track these improvements in new issues. The setup.go is particularly low quality, but it will take some time to improve it because it uses the ibctesting.Coordinator

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+    let infractionHeight = undefined;
+
+    if (data.vscID === 0) {
+      infractionHeight = this.initialHeight;
+    } else {
+      infractionHeight = this.vscIDtoH[data.vscID];
+    }
+
+    if (this.m.staking.status[data.val] === Status.UNBONDED) {
+      return;
```
mpoke, line 438: Add an event here, i.e., `RECEIVE_SLASH_REQUEST_FOR_UNBONDED`

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+    } else {
+      this.m.events.push(Event.RECEIVE_DOUBLE_SIGN_SLASH_REQUEST);
+    }
+
+    if (this.tombstoned[data.val]) {
+      return;
+    }
+
+    this.m.staking.slash(data.val, infractionHeight);
+    this.m.staking.jailUntil(data.val, this.m.t[P] + JAIL_SECONDS);
```
mpoke, line 452: A slash validator will be jailed forever. This is especially true since Unjail is not implemented. Until we fix this, we should increase the number of validators for the generated traces (currently `NUM_VALIDATORS = 4`).

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+  tombstoned: boolean[];
+  // unbonding operations to be completed in EndBlock
+  matureUnbondingOps: number[];
+
+  constructor(model: Model, { ccvP }: ModelInitState) {
+    this.m = model;
+    Object.assign(this, ccvP);
+  }
+
+  endBlock = () => {
```
mpoke, line 380: Split in two, similarly to the spec, i.e., `EndBlockCIS` and `EndBlockVSU`. Furthermore, I think it could be a nice idea to use the tags from the spec, e.g., `[CCV-PCF-EBLOCK-CIS.1]`.

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+    this.vscIDtoH[this.vscID] = this.m.h[P] + 1;
+    this.matureUnbondingOps.forEach((opID) => {
+      this.m.staking.unbondingCanComplete(opID);
+    });
+    this.matureUnbondingOps = [];
```
mpoke, line 385: Add some comments, e.g., "notify the Staking module that it can complete all matured unbondings"

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+    const valUpdates = this.m.staking.valUpdates();
+    const hasOps =
+      this.vscIDtoOpIDs.has(this.vscID) &&
+      0 < this.vscIDtoOpIDs.get(this.vscID)!.length;
```
mpoke, line 389: Is this check necessary?

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,671 @@
...
+      });
+      return ret;
+    })();
+
+    this.pendingChanges = [];
+
+    changes.forEach((power, val) => {
+      this.power[val] = undefined;
+      if (0 < power) {
+        this.m.events.push(Event.CONSUMER_UPDATE_POWER_POSITIVE);
```
mpoke, line 528: Add event for adding a new validator. Basically, there are three events: adding a new one, updating an existing one with a new power, removing am existing one.

mpoke (commented):
Second part of the review (still need to look over properties.ts). In general, the code should be commented better.

diff-tests/core/model/src/model.ts
danwt: > Until we fix this, we should increase the number of validators for the generated traces

Why?

diff-tests/core/model/src/model.ts
danwt: Don't think so, useful catch.

diff-tests/core/model/src/model.ts
mpoke: Any generated trace will have maximum 4 slash events. Moreover, maximum 2 slash events can happen at the same time. Seems little.

danwt: Hmm, what about keeping it as is for now and making a new issue to track this concern, and working on it in conjunction with

- https://github.com/cosmos/interchain-security/issues/319

diff-tests/core/model/src/properties.ts
```diff
@@ -0,0 +1,256 @@
...
+ * @param hist A history of blocks.
+ * @returns Is the property satisfied?
+ */
+function bondBasedConsumerVotingPower(hist: BlockHistory): boolean {
+  const partialOrder = hist.partialOrder;
+  const blocks = hist.blocks;
+  function powerProvider(block: CommittedBlock) {
+    return _.range(NUM_VALIDATORS).map(
+      (i) =>
+        block.invariantSnapshot.tokens[i] +
```
mpoke, line 187: These are all the tokens of validator `i` regardless of whether it is bonded, unbonding, or unbonded. For the property, only `bonded` and `unbonding` validators are relevant (plus the unbonding undelegations that are captured in the next lines).

mpoke (changes requested):
Final part of the review for `model.ts` and `properties.ts`. See my comments below. 

In general, amazing work @danwt 🚀

diff-tests/core/model/src/properties.ts
```diff
@@ -0,0 +1,256 @@
...
+    let bestV = -1;
+    let result = it.next();
+    while (!result.done) {
+      const h = result.value;
+      if (bestH < h && h <= height) {
+        bestH = h;
+        bestV = this.greatestPred[chain].get(h) as number;
+      }
+      result = it.next();
+    }
```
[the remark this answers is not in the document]
danwt: Fastest way is binary search, which I had implemented in the python ver. I guess I was in a rush when I converted to TS. I will update it to BS at some point, perhaps in conjunction with

- https://github.com/cosmos/interchain-security/issues/320

diff-tests/core/model/src/model.ts
```diff
@@ -0,0 +1,691 @@
...
+
+  delegate = (val: number, amt: number) => {
+    this.staking.delegate(val, amt);
+  };
+
+  undelegate = (val: number, amt: number) => {
+    this.staking.undelegate(val, amt);
+  };
+
+  consumerSlash = (
```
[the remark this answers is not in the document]
danwt: Right! Let's make an extension issue

diff-tests/core/model/src/common.ts
```diff
@@ -0,0 +1,206 @@
...
+  chain: Chain;
+};
+
+type InvariantSnapshot = {
+  h: Record<Chain, number>;
+  t: Record<Chain, number>;
+  tokens: number[];
+  undelegationQ: Undelegation[];
+  delegatorTokens: number;
+  power: (number | undefined)[];
```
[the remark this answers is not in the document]
danwt: Hmm, latest model.ts has all fields commented, did you see?

I'm not sure about also commenting these parts. Tbh there probably a much better way to write the typescript than I have done, I need to study the language a bit more.

diff-tests/core/model/src/properties.ts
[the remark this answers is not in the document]
mpoke: I don't think we need binary search. Just that now we look through all heights, including the ones larger than `height`. Also, we do not stop the iteration. We just need to go in reverse order starting from `height`.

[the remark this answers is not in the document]
danwt: https://github.com/cosmos/interchain-security/blob/680869e6b28c579fc4a7d6fe7ffa731979f52ae1/diff-model/diff/blocks.py#L30-L46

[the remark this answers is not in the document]
danwt: I mean there not much point discussing between linear vs linear when log is possible

diff-tests/core/model/src/properties.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+  function inner(hc: number): boolean {
+    const hp = partialOrder.getGreatestPred(C, hc);
+    assert(hp !== undefined, 'this should never happen.');
+    function getHC_() {
+      const tsHC = (blocks[C].get(hc) as CommittedBlock).invariantSnapshot
+        .t[C];
+      // Get earliest height on consumer
+      // that a VSC received at hc could mature
+      const heights = Array.from(blocks[C].keys()).sort((a, b) => a - b);
```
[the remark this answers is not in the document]
danwt: I'm putting my software engineer hat on to say that that would be premature optimization. We should improve the performance if and when needed using profile guided approach. My intuition would point towards using the right datastructure for greatestPred and leastSucc being the greatest win as that is just naive right now.

diff-tests/core/model/src/properties.ts
```diff
@@ -0,0 +1,256 @@
...
+          (blocks[C].get(hc_) as CommittedBlock).invariantSnapshot.t[C]
+        ) {
+          return hc_;
+        }
+      }
+      return undefined;
+    }
+    const hc_ = getHC_();
+    let hp_ = undefined;
+    if (hc_ !== undefined) {
```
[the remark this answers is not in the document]
danwt: Exactly, but we still have to check those blocks!

diff-tests/core/model/src/properties.ts
```diff
@@ -0,0 +1,256 @@
...
+      hp_ = partialOrder.getLeastSucc(C, hc_);
+    }
+    let limit = Math.max(...Array.from(blocks[P].keys())) + 1;
+    if (hp_ !== undefined) {
```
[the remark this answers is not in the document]
danwt: ditto^^

diff-tests/core/model/src/properties.ts
danwt: Useful catch!

https://github.com/cosmos/interchain-security/blob/07fa9a82c8ed9a93baa42e74a46c7df00349a42b/diff-tests/core/model/src/properties.ts#L189-L208

diff-tests/core/driver/core_test.go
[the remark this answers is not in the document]
smarshall-spitzbart: Yea seems reasonable to me. If the suite package offers value that native golang testing cannot, I'm all for it! We just wanna reduce the dependencies and complexity of testing where we can. Tracking this as an issue works 👍

diff-tests/core/driver/setup.go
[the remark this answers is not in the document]
smarshall-spitzbart: Yea this will become an issue similar to https://github.com/cosmos/interchain-security/issues/300, but agreed that it may not be the most important thing to focus on rn 👍

smarshall-spitzbart:
Could we also add the new files to ```tests/differential``` instead of ```diff-test``` to be compatible with https://github.com/cosmos/interchain-security/pull/325

mpoke (approved):
Nice work @danwt.

Issue #280, opened 2022-08-23
Replace hardcoded constants with params


mpoke:
Identify hardcoded constants, e.g., https://github.com/cosmos/interchain-security/blob/eaa7f6e7ef98fa16b38e6213a194506dc9d50f26/x/ccv/types/utils.go#L6

and remove unwanted ones (i.e., replace with params).

jtremback:
Should go into app.go actually, and be passed in NewKeeper or similar

MSalopek:
Should we also consider using a linter like `golangci-lint` to pick up on stuff like in the future?

jtremback:
These should all be replaced by params, pretty straightforward.

https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/distribution.go#L17
https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/distribution.go#L22
https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/types/keys.go#L24

These two:

https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/types/params.go#L16
https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/types/params.go#L19

are only used here:

https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/types/params.go#L66

and they seem to be placeholders that affect nothing. IMO I would move them inline to the function.

jtremback:
I can't believe that's it. @mpoke Do you think I'm missing any?

mpoke:
```bash
> ack "const " x

x/ccv/types/utils.go
6:const TimeDelay = 4 * 7 * 24 * time.Hour

x/ccv/provider/types/params.go
13:const (

x/ccv/utils/utils.go
18:const TrustingPeriodFraction = 2

x/ccv/consumer/types/params.go
10:const DefaultBlocksPerDistributionTransmission = 1000

x/ccv/consumer/types/keys.go
10:const (

x/ccv/consumer/keeper/distribution.go
17:const TransferTimeDelay = 1 * 7 * 24 * time.Hour // 1 weeks
22:const ConsumerRedistributeFrac = "0.75"
```

jtremback:
`const TrustingPeriodFraction = 2` should be made a param as well.

mpoke:
> `const TrustingPeriodFraction = 2` should be made a param as well.

[...]

PR #317, opened 2022-09-01
reduce usage of magic numbers and replace with constants

MSalopek:
This PR addresses isue #280.

Usage of magic numbers is reduced and replaced with defined constants.
Magic numbers remain in tests and testutils.

`golangci-lint` was used to discover the magic numbers.

Linter configuration can be found [here](https://gist.github.com/MSalopek/191fc6e493286aab33448506d2fecebe).

Comprehensive configuration options can be found [here](https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml)

It might be beneficial to setup `golangci-lint` to run as precommit hook in the future.

### DIFF
--- a/x/ccv/consumer/types/params.go
+++ b/x/ccv/consumer/types/params.go
@@ -6,6 +6,9 @@ import (
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 )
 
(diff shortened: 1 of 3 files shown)


danwt (approved):
Looks good! It's the little things that count!

mpoke (approved):

PR #321, opened 2022-09-02
Test for the reward distribution

stana-ethernal:
Added e2e test for distributing rewards from minimal viable consumer chain to the corresponding consumer account and provider chain.
Also, this includes a change in consumer's distribution logic that will transfer the rewards to the consumer and provider redistribute addresses in each block. This change is first introduced [here](https://github.com/cosmos/interchain-security/blob/4915a0a9c7150e1dc8516950dd52930785e427b5/x/ccv/consumer/keeper/distribution.go#L66) , but it is implemented here as well so that once the original [PR](https://github.com/cosmos/interchain-security/pull/130) is pushed, the test stays valid.


jtremback (approved):

mpoke (approved):
LGTM. Great work @stana-ethernal

Issue #294, opened 2022-08-24
Transfer channel is not opened between consumer and provider chain


stana-ethernal:
During opening a channel between consumer and provider, in the consumer module in OnChanOpenAck function there is a code for initialization of a transfer channel used for the rewards that should be sent to the provider chain.
The problem here is that only ChannelOpenInit is called, and transfer channel stays in INIT state (the handshake is never completed). The consumer module uses this channel to send a reward to the provider chain, but an error occurs with the following message:
"packet failed basic validation: invalid destination channel ID: identifier cannot be blank: invalid identifier"
This causes consumer chain halting.

mpoke:
> The problem here is that only ChannelOpenInit is called, and transfer channel stays in INIT state (the handshake is never completed).

A relayer should complete the handshake.

The `ibcTransferKeeper.SendTransfer` will not fail even if the channel is not OPEN yet (due to [optimistic sends](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md#sending-packets)). However, the packets cannot be received by the destination chain before the channel is OPEN, see https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/core/04-channel/keeper/packet.go#L148

stana-ethernal:
We got error in this [line](https://github.com/cosmos/interchain-security/blob/f3a996699756bc223a168cb0b8886a355204c19e/x/ccv/consumer/keeper/distribution.go#L77), and after debugging saw that [destinationPort](https://github.com/cosmos/ibc-go/blob/0fa2978b0edad338f995790b5ab3a517ff302351/modules/apps/transfer/keeper/relay.go#L76) is not set, and later when it try to [send packet](https://github.com/cosmos/ibc-go/blob/0fa2978b0edad338f995790b5ab3a517ff302351/modules/apps/transfer/keeper/relay.go#L166) it fails the basic validation because the counterparty port is not set. Also, we saw that the channel state at that moment was init. 

Is there a way to ensure that relayer will finish the handshake, because in our case this does not happen? Should we call something on relayer explicitly, except creating a connection between consumer and provider chains and opening a channel between provider and consumer ports?

mpoke:
[...]

PR #322, opened 2022-09-02
Check if transfer channel is opened (#294)

stana-ethernal:
Before sending rewards to provider chain, we now check if the transfer channel is in open state. If the channel is not opened, next try to send rewards will be after BlocksPerDistributionTransmission number of blocks.
It is the responsibility of the validator to complete the handshake for the transfer channel.


mpoke (approved):
LGTM

jtremback (approved):

PR #325, opened 2022-09-02
Common folder for tests

smarshall-spitzbart:
Follows pattern of [sdk's test folder](https://github.com/cosmos/cosmos-sdk/tree/main/tests).

```testutil/``` was left in the top level repo directory to be consistent with cosmos sdk, but moving it into ```tests/``` could make sense.


mpoke (approved):

danwt (approved):
Thanks for this!

Issue #292, opened 2022-08-24
Handling packets recv on unknown channels


mpoke:
Ref. https://github.com/cosmos/interchain-security/pull/274#discussion_r952965776

The discussion is about whether the CCV modules (both on provider and consumers) should handle the case of IBC packets received on other channels than the established CCV channel. Note that one of the properties guaranteed by Interchain Security is that the CCV channel is unique. 

Sending `MsgRecvPacket` messages to the core IBC core module is permissionless. However, there are several requirements for that packet (contained in the message) to be passed to the `OnRecvPacket` handler of an IBC application module. The relevant requirements for this discussions are:
- the module must have claimed the capability for the channel defined by `portId, channelId`, where `portId = msg.Packet.DestinationPort` and `channelId = msg.Packet.DestinationChannel`, see https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/core/keeper/msg_server.go#L380
- the channel must be OPEN, see https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/core/04-channel/keeper/packet.go#L148
- EDIT: the packet must have been sent by the counter party (i.e., the sender chain)

### Channel capabilities

The consumer CCV module claims a channel's capability in `OnChanOpenInit`, see https://github.com/cosmos/interchain-security/blob/f3a996699756bc223a168cb0b8886a355204c19e/x/ccv/consumer/module.go#L281 

Sending `MsgChannelOpenInit` messages to he core IBC core module is permissionless. Moreover, the only requirement (relevant for this discussion) for those messages to be passed to the `OnChanOpenInit` handler of an IBC application module is that the module claimed the port capability, see https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/core/keeper/msg_server.go#L170. 

[...]

PR #326, opened 2022-09-05
Handling packets recv on unknown channels

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/292


x/ccv/consumer/keeper/relay.go
```diff
@@ -25,8 +25,10 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
 	// get the provider channel
 	providerChannel, found := k.GetProviderChannel(ctx)
 	if found && providerChannel != packet.DestinationChannel {
-		// VSC packet was sent on a channel different than the provider channel
-		return utils.OnRecvPacketOnUnknownChannel(ctx, k.scopedKeeper, k.channelKeeper, packet)
+		// VSC packet was sent on a channel different than the provider channel;
+		// this should never happen
+		panic(fmt.Errorf("VSCPacket received on unknown channel %s; expected: %s",
+			packet.DestinationChannel, providerChannel))
```
danwt, line 31: I'm not sure I understand why this is a panic. From reading https://github.com/cosmos/interchain-security/issues/292 I understand that the consumer can receive packets from channels which arent OPEN on the provider. It sounds like recoverable behavior, not panic behavior.

danwt, line 31: I'm not sure I understand why this is a panic. From reading https://github.com/cosmos/interchain-security/issues/292 I understand that the consumer can receive packets from channels which arent OPEN on the provider. It sounds like recoverable behavior, not panic behavior.

x/ccv/provider/keeper/relay.go
```diff
@@ -87,10 +88,8 @@ func (k Keeper) CompleteMaturedUnbondingOps(ctx sdk.Context) {
 // OnAcknowledgementPacket handles acknowledgments for sent VSC packets
 func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
 	if err := ack.GetError(); err != "" {
-		// Either the VSC packet data could not be successfully decoded
-		// or the VSC packet was sent on a channel other than the established
-		// provider channel and ChanCloseInit failed.
-		// Neither of these should ever happen.
+		// The VSC packet data could not be successfully decoded.
+		// This should never happen.
```
danwt, line 92: I think it would be more precise to say that it should rarely happen.

danwt (approved):

x/ccv/consumer/keeper/relay.go
mpoke: See https://github.com/cosmos/interchain-security/issues/292#issuecomment-1231334565

x/ccv/provider/keeper/relay.go
mpoke: This is a programming error. Should never happen.

x/ccv/consumer/keeper/relay.go
danwt: I'm not sure if we should have this branch at all. Having it implies it can happen sometimes.

jtremback: Agree with Dan. Unless it can somehow happen, but it sounds like that would require an incorrect provider.

jtremback (approved):
Nice! Love to see stuff getting deleted

Issue #293, opened 2022-08-24
Closing non-CCV channels on provider


mpoke:
The provider CCV module should close channels attempting to complete the opening handshake if there is already an established CCV channel, see https://github.com/cosmos/interchain-security/blob/f3a996699756bc223a168cb0b8886a355204c19e/x/ccv/provider/keeper/keeper.go#L275

I see the following problem with the current implementation: `SetConsumerChain` is called by `OnChanOpenConfirm`, which means the error returned here https://github.com/cosmos/interchain-security/blob/f3a996699756bc223a168cb0b8886a355204c19e/x/ccv/provider/keeper/keeper.go#L279 will abort the execution of `ChannelOpenConfirm` here https://github.com/cosmos/ibc-go/blob/05c4bfbbebb20d1cc078ddf1d29dadc766d9251b/modules/core/keeper/msg_server.go#L306. As a result, all the changes done by `channelKeeper.ChanCloseInit` (i.e., moving the channel to `CLOSED`) will be reverted.

**_Potential solution:_** 
The calls to `channelKeeper.ChanCloseInit` should be defer to `EndBlock`.

mpoke:
@AdityaSripal do you agree?

jtremback:
Is this a bug or is it just garbage collection?

Issue #327, opened 2022-09-05
Close duplicate channels


mpoke:
Due to the way the channel initialization protocol is designed, it is possible to have multiple channels passing the ACK step on the consumer (i.e., move to OPEN on consumer). However, only one channel (the CCV channel) will pass the CONFIRM step (i.e., move to OPEN on provider). For cleaning up the state, the other channels that are opened on the consumer, but do not become the CCV channel, should be closed. 

**Suggestion:** Iterate over all channels on port `ConsumerPortId` when receiving the first VSC packet. See https://github.com/cosmos/ibc-go/blob/42240b54f23ae1d2f8f170f942e49e54ebb7588a/modules/core/04-channel/keeper/keeper.go#L370 for example of how to iterate over all channels.

jtremback:
I'm usually not that into garbage collecting state that people paid gas to add, but I wouldn't reject this. Certainly not a high priority IMO

mpoke:
This issue is more about dealing with particular state that will be difficult to clean due to the way the IS protocol is design, i.e., a channel that is open on one end and cannot open on the other end. I agree that is low priority ;)

jtremback:
Closing to reduce clutter. My rule of thumb is that if there is state that 

1. a user has paid to add and 
2. is not iterated on endblock

We shouldn't worry about cleaning it up.

PR #328, opened 2022-09-05
Closing non-CCV channels on provider

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/293

Calling `k.chanCloseInit` and then returning an error in `OnChanOpenConfirm()` will revert all the changes done by `chanCloseInit`. Thus, there is no reason to try to close the channel. Duplicate channels need to be closed in a different way (see https://github.com/cosmos/interchain-security/issues/327).


danwt (approved):
LG!

jtremback (approved):

PR #337, opened 2022-09-07
Bugfix and refactor NewCreateConsumerChainProposal

smarshall-spitzbart:
The ```NewCreateConsumerChainProposal``` function previously returned a useless error that would always be nil. The function only constructs an in-mem type without any marshaling, so it's safe to remove the returned error.

This PR makes existing and incoming unit tests simpler


jtremback (approved):
lol nice catch

danwt (approved):
Nice!

Issue #344, opened 2022-09-08
Coverage configuration for SonarCloud is wrong


danwt:
## Problem

Coverage doesn't make sense: some things which are obviously tested are showing as uncovered.

## Closing criteria

Sonarcloud shows the correct coverage

## Detail

I think the problem is with the `go test` cmd that sonar is configured to use

https://github.com/cosmos/interchain-security/blob/5e3ff16033d045551f71e76425659f0989a81833/.github/workflows/build.yml#L23-L24

It might be failing due to testsuite running in parallel or something.

I can confirm the problem locally (incorrect coverage) without even using sonar.

PR #346, opened 2022-09-08
Fix coverage computation command

danwt:
Fixes https://github.com/cosmos/interchain-security/issues/344

- https://github.com/cosmos/interchain-security/issues/344


danwt:
It turns out that by default go coverage, for a given test t in package p, will only compute coverage for statements in p. So another param was needed to include all src files in the scope. 

The _test.go are still excluded by sonarcloud.

PR #357, opened 2022-09-09
Reduce Unit Testing Boilerplate

smarshall-spitzbart:


testutil/keeper/unit_test_helpers.go
```diff
@@ -22,169 +23,173 @@ import (
...
+}
 
-	cdc, storeKey, paramsSubspace, ctx := SetupInMemKeeper(t)
+// NewInMemKeeperParams instantiates in-memory keeper params with default values
+func NewInMemKeeperParams(t testing.TB) InMemKeeperParams {
+	storeKey := sdk.NewKVStoreKey(types.StoreKey)
+	memStoreKey := storetypes.NewMemoryStoreKey(types.MemStoreKey)
+
+	db := tmdb.NewMemDB()
+	stateStore := store.NewCommitMultiStore(db)
```
danwt, line 47: Here a new store is created, so each paramspace returned (in successive calls to this function) will have a different underlying store object. Does that work out ok?

I like it.

testutil/keeper/unit_test_helpers.go
```diff
@@ -22,169 +23,173 @@ import (
...
+	*MockPortKeeper
+	*MockConnectionKeeper
+	*MockClientKeeper
+	*MockStakingKeeper
+	*MockSlashingKeeper
+	*MockAccountKeeper
+	*MockBankKeeper
+	*MockIBCTransferKeeper
+	*MockIBCCoreKeeper
+}
```
danwt, line 84: Could these concrete pointer types be replaced with their respective keeper interfaces?

testutil/keeper/unit_test_helpers.go
```diff
@@ -22,169 +23,173 @@ import (
...
-		&MockStakingKeeper{},
-		&MockSlashingKeeper{},
-		&MockAccountKeeper{},
-		"",
+		memStoreKey,
+		paramstypes.ModuleName,
 	)
-	return k, ctx
-}
+	ctx := sdk.NewContext(stateStore, tmproto.Header{}, false, log.NewNopLogger())
```
danwt, line 61: I think it would be useful to have the ctx creation be a separate function, taking stateStore, as argument. It may also be useful to be able to customize the header a bit (but we can leave that for a later PR). 

Then it would be good to make the user call the two parts separately

```
params := NewInMemKeeperParams(t)
ctx := NewCtx(params.statestore ) // would need to add handle to this to params
```

Maybe I'm overthinking it. I need to double check how ctx works. The thing is you want one ctx for the whole app (for a given block) AFAIK. So not sure what the effect is of having a different ctx for each keeper. It might be limiting.

testutil/keeper/unit_test_helpers.go
```diff
@@ -22,169 +23,173 @@ import (
...
+	mocks := NewMockedKeepers(ctrl)
+	return NewInMemConsumerKeeper(params, mocks), params.Ctx, ctrl
 }
 
-func SetupInMemKeeper(t testing.TB) (*codec.ProtoCodec, *storetypes.KVStoreKey, paramstypes.Subspace, sdk.Context) {
-	storeKey := sdk.NewKVStoreKey(types.StoreKey)
-	memStoreKey := storetypes.NewMemoryStoreKey(types.MemStoreKey)
+// Sets a template client state for a params subspace so that the provider's
+// GetTemplateClient method will be satisfied.
+func SetTemplateClientState(ctx sdk.Context, subspace *paramstypes.Subspace) {
```
danwt, line 167: Cool glad to have this!

danwt (commented):
Nice changes. I'm just curious about the flexibility potential. Maybe I'm thinking about it wrong though, since unit tests are supposed to be really simple. So maybe it doesn't need to be particularly flexible.

testutil/keeper/unit_test_helpers.go
smarshall-spitzbart: Interesting question, I don't think there's a use case for creating multiple keepers instantiations in a single unit test (that would exist at the same time). The assumption here is that each test case will make a new lightweight, in-mem keeper that will be cleaned up by the garbage collector at the end of the test case. This pattern has worked thus far, but I can't claim that I know every intricacy of how the in-mem keeper works. This design was initially based off the [cosmos developer tutorial source code](https://github.com/cosmos/b9-checkers-academy-draft/blob/b79a43c6e1db1c1df42290fbd037fa87192902a3/x/checkers/keeper/keeper_test.go).

testutil/keeper/unit_test_helpers.go
smarshall-spitzbart: Useful input here, this is making me really think through my design!

I don't really see a use case in which a user would want to pass in their own ```params.statestore``` to a context object. They'd pretty much always be happy with the default in-mem store, if they're writing pure unit tests. More customization should be reserved for other forms of testing imo. I could be missing a use-case tho, maybe we just make this change in the future if it becomes needed?

On customizing the ctx header, etc. after it has been constructed, this is already possible. The caller just needs to use
```go
ctx = ctx.WithBlockHeader(tmproto.Header{})
ctx = ctx.WithBlockTime(time.Now())
...etc
``` 

Emphasis with the design of this PR is the ability to setup a unit test with minimal lines of code, but still allow a user to replace default values for things, _only if needed_

danwt: Hmm ok, fair point! I agree we can just change it if needed. YAGNI lol

testutil/keeper/unit_test_helpers.go
danwt: Good point! Fine to leave like this then

danwt (approved):

testutil/keeper/unit_test_helpers.go
danwt: Ok we can change later if needed 👍

sainoe (approved):

smarshall-spitzbart:
> Great work @smarshall-spitzbart ! Just a small question regarding the changes description.
> 
> > An added benefit to this PR is that non-mocked unit tests now have an extra validation that no external keeper methods were called during the test. Ie.
> 
> That doesn't include the **expected** calls to external keepers right ?

Yep! If you write ```Expect()``` calls for methods of external keepers, those calls will be allowed within the test. 

However if you use the default (least lines of code) way of instantiating a keeper:
```go
providerKeeper, ctx, ctrl := testkeeper.GetProviderKeeperAndCtx(t)
defer ctrl.Finish()
```
The test will fail if any calls to external keepers are made

Issue #284, opened 2022-08-23
Democracy governance whitelisting


jtremback:
The democracy module described [here](https://github.com/cosmos/interchain-security/issues/141) works. We now need to make sure that the consumer chain governance cannot pass proposals that would be dangerous for the provider chain.

Here is a matrix containing all the parameters that the consumer chain governance should be able to change: https://docs.google.com/spreadsheets/d/1s7IGkxIjvebFu0guC5iFC4gWaEOJLBWx2QQHik9WmYo/edit?pli=1#gid=0 (thanks @okwme and @Sam-Interchain)

Implement a mechanism to prevent the democracy module from passing certain proposals, and input the whitelist into this mechanism.

Note, you will likely need to allow/disallow passing of any proposal of a given type (eg chain upgrade proposals), as well as looking into proposals of a certain type and allow/disallowing them more granularly (eg parameter changes).

Conditions of satisfaction:
- It is possible to encode this whitelisting functionality either as code in the consumer chain binary (might be easier), or in a serialized form (nice, if it is possible)
- Demonstrate that some gov proposals are not possible to execute (easier) or even create (nice, if it is possible)
- Demonstrate that some gov proposals are still possible to execute

mpoke:
@jtremback @stana-ethernal  I think this issue was fixed by https://github.com/cosmos/interchain-security/pull/364, right? 

@jtremback Should we open an issue to enable upgrading the consumer through consumer governance as long as the upgrade doesn't touch specific modules (e.g., ccv, slashing, evidence).

stana-ethernal:
> @jtremback @stana-ethernal I think this issue was fixed by #364, right?
yes, it is fixed

PR #364, opened 2022-09-23
Democracy governance whitelisting #284

stana-ethernal:
Contains implementation for [democracy governance whitelisting](Democracy governance whitelisting #284 ).
Forbidden gov proposals cannot be:
1. created - antehadler is added which checks MsgSubmitProposal messages if they are whitelisted or not 
2. executed - wrapper for governance module is added which overrides EndBlock function which first checks if some proposals should be deleted and then calls base EndBlock function for core governance module. Custom logic for proposal deletion, iterates through active proposals queue. Then checks if the proposal is whitelisted and if that's not the case, it refunds the deposit tokens, delete the votes (if proposal was active) and delete the proposal itself (from the proposal storage and active/inactive queue). No need to delete proposals from the inactive queue, because they will be deleted in core module after deposit expiration or if they become active they will be deleted by the implemented code for deleting active forbidden proposals.

Both antehandler and wrapped gov module's endblocker logic, on their creation, receive 'IsProposalWhitelisted' function to check if proposal is whitelisted or not. For democracy enabled consumer chain app, this function implements whitelisting specified [here](https://docs.google.com/spreadsheets/d/1s7IGkxIjvebFu0guC5iFC4gWaEOJLBWx2QQHik9WmYo/edit?pli=1#gid=0), but any other application can define its own whitelisting logic by implemented this fn in a custom way.


stana-ethernal:
@okwme @jtremback Host_enabled and Allow_messages parameters from the Interchain Account Host are listed in the [doc](https://docs.google.com/spreadsheets/d/1s7IGkxIjvebFu0guC5iFC4gWaEOJLBWx2QQHik9WmYo/edit?pli=1#gid=0) as allowed for consumer, but I did not whitelisted them in the code since consumer app does not contain Interchain Accounts module.

jtremback (approved):
LGTM, thanks!

app/consumer-democracy/proposals_whitelisting.go
```diff
@@ -0,0 +1,68 @@
...
+	case *proposal.ParameterChangeProposal:
+		return isParamProposalWhitelisted(c.Changes)
+
+	default:
+		return false
+	}
+
+}
+
+func isParamProposalWhitelisted(paramChanges []proposal.ParamChange) bool {
```
smarshall-spitzbart, line 25: Might be more clear if this method is changed to `isParamChangeWhitelisted`, since it's not referring to a proposal (correct?)

tests/e2e/democracy_test.go
```diff
@@ -253,6 +263,119 @@ func (s *ConsumerDemocracyTestSuite) TestDemocracyRewarsDistribution() {
...
+	s.consumerChain.NextBlock()
+	currentAuthParamValue = authKeeper.GetParams(s.consumerCtx()).MaxMemoCharacters
+	//check that parameters are not changed, since the proposal contained forbidden changes
+	s.Assert().Equal(oldAuthParamValue, currentAuthParamValue)
+	s.Assert().NotEqual(newAuthParamValue, currentAuthParamValue)
+	//deposit is refunded
+	s.Assert().Equal(votersOldBalances, getAccountsBalances(s.consumerCtx(), bankKeeper, bondDenom, votingAccounts))
+}
+
+func (s *ConsumerDemocracyTestSuite) TestDemocracyGovernanceWhitelistingKeys() {
```
smarshall-spitzbart, line 339: This looks like a unit test, but is there an advantage to using `ConsumerDemocracyTestSuite`?

smarshall-spitzbart (approved):
Nice!

app/consumer-democracy/proposals_whitelisting.go
stana-ethernal: yes, that's right

tests/e2e/democracy_test.go
stana-ethernal: Thanks for pointing that out. There's no need for ConsumerDemocracyTestSuite, since we need only consumer app with the initialized param subspaces to check if the keys and subspace string exists. The test is moved from e2e folder.

Issue #361, opened 2022-09-21
More integration tests needed


jtremback:
# Problem

We don't have enough integration tests. We had a weird bug in the testnet that wasn't caught in the integration tests.

- [x] Basic power changing proof of concept (done)
- [x] Downtime slashing tests (partially done)
   - [x] Need downtime on multiple consumer chains
   - [x] Need downtime on one consumer chain, validator is taken off all consumer chains
- [x] Double signing slashing tests
  - [x] Double signing on multiple consumer chains
  - [x] Double signing on one consumer chain, validator is taken off all consumer chains
- [x] Full unbonding (Cosmos-SDK "staking hooks" code)
  - [x] Happy path
  - [ ] Get your tokens back (as discussed here: https://github.com/cosmos/interchain-security/issues/311)
  - [ ] 6 total variations
    - Consumer unbonding over before provider and vice versa
    - Redelegate, undelegate, and validator removal
- [x] Consumer chain stopping
  - [x] Gov proposal
  - [ ] Timeout
- [x] Lots of consumer chains
  - [x] Test some basic stuff

- [ ] The dream: Test everything at once (hook it up to Dan's diff testing model)

# Closing criteria

Complete the tests above

jtremback:
Good first task for @MSalopek: Separate out the current power changing and downtime tests in the current "happy path" steps.

smarshall-spitzbart:
https://github.com/cosmos/interchain-security/issues/263 is also relevant to this discussion

danwt:
From these

> Redelegate, undelegate, and validator removal

I would suggest not doing redelegate/undelegate and only doing validator removal. Redel/undel are already well tested in diff tests.

Could we close this issue and create new more specific ones for the remaining checkboxes?

PR #367, opened 2022-09-29
use dedicated query node for tx and query RPC

MSalopek:
Add RPC node to integration test configs.

Each chain now has a defined validator set and a dedicated non-validating full node used for tx and query broadcast.

Closes: #263

PR #371, opened 2022-10-04
integration tests - split steps; run in parallel

MSalopek:
Changes:
* allow running integration tests in parallel (1 docker container per run).
* de-couple default  and democracy test runs
* split steps into multiple files/functions
* allow easy combining of step sequences
* minor package refactoring

Integration tests can now be run in parallel.

This is achieved through TestRun configuration parameters.
* you can provide new container and instance names
* each test run will be executed within its own docker container

Related: #361


tests/integration/demo_vote_panic.txt
```diff
@@ -0,0 +1,29 @@
+main.voteGovProposalAction{chain:"democ", from:[]main.validatorID{"alice", "bob"}, vote:[]string{"yes", "no"}, propNumber:0x1}
```
jtremback, line 1: What is this file?

tests/integration/main.go
```diff
@@ -96,6 +100,22 @@ func (tr TestRun) runStep(step Step, verbose bool) {
 	pretty.Print(actualState)
```
jtremback, line 100: I won't block merging on this, but it might be time to take this out or do something more intelligent with it. Right now I believe the steps are being interleaved on the console which makes printing it useless

jtremback (approved):
LGTM with removal or explanation of that text file

smarshall-spitzbart:
This would close #310

tests/integration/config.go
```diff
@@ -54,6 +53,37 @@ type TestRun struct {
 	localSdkPath     string
 }
 
+// TODO: check races
+var defaultValidators = map[validatorID]ValidatorConfig{
```
smarshall-spitzbart, line 57: Is it possible to make this a function returning a variable? Not a fan of global mutable `var`s

tests/integration/main.go
```diff
@@ -2,40 +2,44 @@ package main
...
 	"github.com/kylelemons/godebug/pretty"
 )
 
 var verbose = false
+var localSdkPath = flag.String("local-sdk-path", "",
+	"path of a local sdk version to build and reference in integration tests")
 
 func main() {
-	fmt.Println("============================================ start happy path tests ============================================")
+	flag.Parse()
```
smarshall-spitzbart, line 21: I tried running a local sdk path myself and it did not work with this branch. Prob has to do with moving the flag parsing commands around? 

Try 
```bash
go run ./tests/integration/... --local-sdk-path "/Users/bob/Documents/cosmos-sdk/"
```

tests/integration/main.go
smarshall-spitzbart: Yea I wonder if a good way to print concurrent tests even exists 🤔

tests/integration/steps_downtime.go
```diff
@@ -0,0 +1,195 @@
+package main
+
+// stepsDowntime tests validator jailing and slashing.
+func stepsDowntime(consumerName string) []Step {
+	return []Step{
+		{
+			action: downtimeSlashAction{
+				chain: chainID(consumerName),
+				// TODO: First validator cannot be brought down until this issue is resolved:
```
smarshall-spitzbart, line 9: This TODO is done now right?

tests/integration/steps_start_consumer.go
```diff
@@ -0,0 +1,184 @@
+package main
+
+import (
+	clienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
+)
+
+// starts provider and single consumer chain
+// * genesisParams overrides consumer genesis params
+// * setupTransferChan creates a transfer channel between provider and consumer
+func stepsStartChains(consumerName, consumerGenesisParams string, setupTransferChan bool) []Step {
```
smarshall-spitzbart, line 10: Could we make this consistent with file name

tests/integration/steps.go
```diff
@@ -1,512 +1,28 @@
...
+	stepsStartChains("consu", "", false),
+	stepsDelegate("consu"),
+	stepsUnbondRedelegate("consu"),
+	stepsDowntime("consu"),
+)
 
-}
+var democracySteps = concatSteps(
+	// democracySteps requires a transfer channel and overrides genesis
+	stepsStartChains("democ", ".app_state.ccvconsumer.params.blocks_per_distribution_transmission = \"10\"", true),
```
smarshall-spitzbart, line 25: Imo there should be a single place that the genesis file is mutated, lets consolidate this with the way of mutating genesis in `config.go`

smarshall-spitzbart (changes requested):
Great stuff! Most important comment is that running integration tests with a local sdk path needs to work with this branch

tests/integration/main.go
jtremback: Ah thanks for checking this

tests/integration/config.go
MSalopek: Yes. In the making.

tests/integration/steps_downtime.go
MSalopek: Waiting on merge on #367

tests/integration/main.go
MSalopek: Fixed it. I was passing a string as param but referencing a global var in the body of the func.

tests/integration/steps.go
MSalopek: Agree, moved the string to config.go for democracy chain.

tests/integration/main.go
jtremback: No, don't worry about it. There isn't actually a good reason to print it at all since it will just match the steps exactly, and if it doesn't, it will throw an error. I just had this in there originally for smoke testing

tests/integration/actions.go
```diff
@@ -392,8 +391,9 @@ func (tr TestRun) startConsumerChain(
 	}
 
 	genesisChanges := ".app_state.ccvconsumer = " + string(bz)
```
smarshall-spitzbart, line 393: Do we need this `genesisChanges` anymore?

tests/integration/config.go
```diff
@@ -131,12 +168,11 @@ func DefaultTestRun() TestRun {
 	}
 }
 
-func (s *TestRun) ParseCLIFlags() {
-	localSdkPath := flag.String("local-sdk-path", "",
-		"path of a local sdk version to build and reference in integration tests")
-	flag.Parse()
-	s.localSdkPath = *localSdkPath
-	fmt.Println(s.localSdkPath)
+func (s *TestRun) SetLocalSDKPath(path string) {
```
smarshall-spitzbart, line 171: Still not working for me locally, maybe we just keep the old method which we know works?

tests/integration/actions.go
MSalopek, line 393: Rename variable to `genesis` or sth similar.

tests/integration/config.go
MSalopek: Check now. There was a race condition in `start-docker.sh` when running in parallel.

The script is run once for each container, so while the first container was using the dir (copying into it), another container would attempt to delete the folder, causing a race condition.

tests/integration/main.go
```diff
@@ -2,44 +2,52 @@ package main
...
 	"os/exec"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/kylelemons/godebug/pretty"
 )
 
-var verbose = false
+var verbose = flag.Bool("verbose", false, "turn verbose logging on/off")
```
smarshall-spitzbart, line 16: Thanks for making this a flag 👍

smarshall-spitzbart (approved):
Great work!

PR #372, opened 2022-10-04
Remove redundant e2e test code

### LINKED ISSUE #300: Redundant e2e test suite structs

smarshall-spitzbart:
Closes #300


jtremback (approved):
Nice! LGTM

danwt (approved):
Beautiful

PR #373, opened 2022-10-04
close 339

### LINKED ISSUE #339: `params.Enabled` on the consumer is basically unused and should be deleted.
If `params.Enabled` is set to `false` in the genesis.json, then the consumer chain will not return a validator set on startup. Since the consumer chains have a disabled staking module, this will lead to a chain that is unable to start, making this a useless parameter.

If `params.Enabled` is set to `false` during running of the consumer chain, nothing will happen except that the consumer's CCV genesis state will no longer be correctly exported.

It is not useful and should be completely removed.

smarshall-spitzbart:
Closes #339 and adds TODOs in code which link to the issue. Jehan and I decided this issue is not important anytime soon


jtremback:
Force merging this because it's just some comments

PR #264, opened 2022-08-11
Update export and init genesis

sainoe:
Close: #121, #156, #158, #341

* Update the provider and consumer export/init genesis functions with the current CCV states

* Update UTs to cover the cases when a CCV channel is already established or not

At the moment the Hermes relayer doesn't support a genesis restart. This is due to IBC client counting on the block history to update the consensus states.
The IBC-rs team is currently working on a solution see [ibc-rs/issues/1152](https://github.com/informalsystems/ibc-rs/issues/1152)

PR #382, opened 2022-10-06
Update #264 - updates genesis and genesis tests

### LINKED ISSUE #380: TestInitGensis needs a second look
# Problem
While working on a pr for #340 I noticed that [TestInitGenesis](https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/genesis_test.go#L26) has some potentially misleading assertions, it's hard for me to understand what exactly the test is supposed to accomplish in certain spots

# Closing criteria
Fix these issues:
- Why is [this check](https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/genesis_test.go#L78) made? The  provider's unbonding period doesn't need to be equal to the consumer's
- When [this function](https://github.com/cosmos/interchain-security/blob/main/testutil/keeper/unit_test_helpers.go#L212) returns a reasonable (non-zero) value for unbonding period, TestInitGenesis fails
- What is the intention of [this line](https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/genesis_test.go#L108)? Seems similar to above

sainoe:
* Update some unit-tests and comments of #264 changes
* Close #380


smarshall-spitzbart (approved):
Looks great, thanks! Only feedback is that we can prob remove `GetClientState` and `GetConsensusState` from `unit_test_helpers.go` since they are now unused

x/ccv/consumer/keeper/genesis.go
```diff
@@ -120,7 +120,7 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) (genesis *consumertypes.GenesisSt
 	}
 
 	// export the current validator set
-	valset, err := k.GetValidatorUpdates(ctx)
+	valset, err := k.GetCurrentValidatorsAsABCIpdates(ctx)
```
jtremback, line 123: Typo

smarshall-spitzbart:
@sainoe just added a small fix to lint errors, hoping we can merge this today with a review from @jtremback. 

Appreciate you fixing the prob shit ton of merge conflicts on this one Simon!

proto/interchain_security/ccv/consumer/v1/genesis.proto
```diff
@@ -13,28 +13,31 @@ import "tendermint/abci/types.proto";
...
       [ (gogoproto.nullable) = false ];
-  // OutstandingDowntimes nil on new chain, filled on restart.
+  // OutstandingDowntimes nil on new chain, filled  in on restart.
   repeated OutstandingDowntime outstanding_downtime_slashing = 10
       [ (gogoproto.nullable) = false ];
-  // PendingSlashRequests filled in on new chain, nil on restart.
+  // PendingSlashRequests nil on new chain, filled in on restart.
  interchain_security.ccv.consumer.v1.SlashRequests pending_slash_requests = 11
  [ (gogoproto.nullable) = false ];
+  // LastTransmissionBlockHeight nil on new chain, filled in on restart.
```
jtremback, line 38: Is this being used for a zero height restart or a restart containing the current block height? Or both?

jtremback: If this genesis restore was being used on a zero height restart, then that might result in it taking a really long time to send out a distribution... not sure

tests/e2e/channel_init.go
```diff
@@ -21,11 +21,6 @@ func (suite *CCVTestSuite) TestConsumerGenesis() {
 
 	genesis := consumerKeeper.ExportGenesis(suite.consumerChain.GetContext())
 
-	// Confirm that client and cons state are exported from consumer keeper properly
```
jtremback, line 24: Was this deletion intentional?

x/ccv/consumer/keeper/genesis.go
```diff
@@ -37,39 +42,51 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
...
-		// verify that latest consensus state on provider client matches the initial validator set of restarted chain
-		// thus, IBC genesis MUST run before CCV consumer genesis
-		consState, ok := k.clientKeeper.GetLatestClientConsensusState(ctx, state.ProviderClientId)
-		if !ok {
-			panic("consensus state for provider client not found. MUST run IBC genesis before CCV consumer genesis")
-		}
-		tmConsState, ok := consState.(*ibctmtypes.ConsensusState)
-		if !ok {
-			panic(fmt.Sprintf("consensus state has wrong type. expected: %T, got: %T", &ibctmtypes.ConsensusState{}, consState))
-		}
```
jtremback, line 55: Why the switch? Were they not working in the other order?

x/ccv/consumer/keeper/genesis.go
```diff
@@ -37,39 +42,51 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
...
-		// ensure that initial validator set is same as initial consensus state on provider client.
-		// this will be verified by provider module on channel handshake.
-		vals, err := tmtypes.PB2TM.ValidatorUpdates(state.InitialValSet)
-		if err != nil {
+		// set default value for valset update ID
+		k.SetHeightValsetUpdateID(ctx, uint64(ctx.BlockHeight()), uint64(0))
+	} else {
+		// verify genesis initial valset against the latest consensus state
+		// IBC genesis MUST run before CCV consumer genesis
+		if err := k.verifyGenesisInitValset(ctx, state); err != nil {
```
jtremback, line 61: You just moved this all into a helper?

x/ccv/consumer/keeper/genesis.go
```diff
@@ -37,39 +42,51 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
...
 			panic(err)
 		}
-		valSet := tmtypes.NewValidatorSet(vals)
+		// chain restarts without the CCV channel established
+		if state.ProviderChannelId == "" {
+			k.SetPendingSlashRequests(ctx, state.PendingSlashRequests)
+
+			// chain restarts with the CCV channel established
+		} else {
```
jtremback, line 69: So i guess you don't even have to deal with the case where the handshake is in progress because it will just take care of itself?

x/ccv/consumer/keeper/genesis.go
```diff
@@ -37,39 +42,51 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
...
+			// set outstanding downtime slashing requests
+			for _, od := range state.OutstandingDowntimeSlashing {
+				consAddr, err := sdk.ConsAddressFromBech32(od.ValidatorConsensusAddress)
+				if err != nil {
+					panic(err)
+				}
+				k.SetOutstandingDowntime(ctx, consAddr)
+			}
+			// set last transmission block height
+			err := k.SetLastTransmissionBlockHeight(ctx, state.LastTransmissionBlockHeight)
```
jtremback, line 85: Does this code work for a zero height restart? Does it even need to work for a zero height restart? note sure...

x/ccv/consumer/types/genesis.go
```diff
@@ -104,9 +109,6 @@ func (gs GenesisState) Validate() error {
 		if gs.ProviderClientId == "" {
 			return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "provider client id must be set for a restarting consumer genesis state")
 		}
-		if gs.ProviderChannelId == "" {
-			return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "provider channel id must be set for a restarting consumer genesis state")
-		}
 		if gs.ProviderClientState != nil || gs.ProviderConsensusState != nil {
```
jtremback, line 112: If this is a restart of a running chain where the handshake did not complete, then how is it ever going to get restarted?

x/ccv/consumer/types/genesis.go
```diff
@@ -88,6 +90,9 @@ func (gs GenesisState) Validate() error {
 		if len(gs.MaturingPackets) != 0 {
 			return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "maturing packets must be empty for new chain")
 		}
+		if gs.LastTransmissionBlockHeight.Height != 0 {
```
jtremback, line 93: Leaving comment here because github doesn't let you leave comments outside of diffs lol:

This entire function feels weird to me. Why are there so many checks on a new chain genesis state, and so few the restart? What about a restart where the handshake didn't complete? Isn't that basically like a new chain restart?

x/ccv/consumer/types/genesis_test.go
```diff
@@ -265,17 +282,18 @@ func TestValidateRestartGenesisState(t *testing.T) {
...
 		{
 			"invalid restart consumer genesis state: nil initial validator set",
-			types.NewRestartGenesisState("ccvclient", "ccvchannel", nil, nil, nil, nil, params),
+			types.NewRestartGenesisState("ccvclient", "ccvchannel", nil, nil, nil, types.SlashRequests{}, nil, types.LastTransmissionBlockHeight{}, params),
 			true,
 		},
 		{
 			"invalid restart consumer genesis state: invalid params",
-			types.NewRestartGenesisState("ccvclient", "ccvchannel", nil, valUpdates, nil, nil,
+			types.NewRestartGenesisState("ccvclient", "ccvchannel", nil, valUpdates, nil, types.SlashRequests{}, nil, types.LastTransmissionBlockHeight{},
```
jtremback, line 296: Are you doing enough testing of the restart genesis state where the handshake hasn't completed?

smarshall-spitzbart:
Sweet, thought this would have merge conflicts with #444 but I guess not!

proto/interchain_security/ccv/consumer/v1/genesis.proto
sainoe: You're right. A zero height restart could take a problematic long time due to this line:
https://github.com/cosmos/interchain-security/blob/b10e13207a56dafcbfe81ccfb273211e738de334/x/ccv/consumer/keeper/distribution.go#L58
Actually `LastTransmissionBlockHeight` should never be greater than the current block height.

x/ccv/consumer/keeper/genesis.go
sainoe: Wdym?

x/ccv/consumer/keeper/genesis.go
sainoe: This case is detected when the channelID is empty and only these states are getting populated:
https://github.com/cosmos/interchain-security/blob/26ec9e85c6861795ff967a94e10ee1069d43158a/x/ccv/consumer/keeper/genesis.go#L64-L67
https://github.com/cosmos/interchain-security/blob/26ec9e85c6861795ff967a94e10ee1069d43158a/x/ccv/consumer/keeper/genesis.go#L92-L103
Note that the client to the provider is already created and imported by the IBC module in this case.

x/ccv/consumer/types/genesis.go
sainoe: Most of theses checks are validating that a "new chain" has all the required IBC sates. When a chain restarts, even when the handshake isn't completed, these states are handled by the IBC module.

x/ccv/consumer/types/genesis_test.go
sainoe: Agree. These genesis validation tests need some adjustments as well.

x/ccv/consumer/types/genesis.go
```diff
@@ -109,6 +110,19 @@ func (gs GenesisState) Validate() error {
 		if gs.ProviderClientId == "" {
 			return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "provider client id must be set for a restarting consumer genesis state")
 		}
+		// handshake is still in progress
+		if gs.ProviderChannelId == "" && (len(gs.MaturingPackets) != 0 || len(gs.OutstandingDowntimeSlashing) != 0 || gs.LastTransmissionBlockHeight.Height != 0) {
```
danwt, line 114: Nit: could we extract a variable `handshakeInProgress := ...` and do `if handshakeInProgress {` . It might be more readable.

danwt (commented):
It would probably be good to document somewhere in comments (maybe in types/genesis.go) exactly what is expected for the cases where the consumer is 1) added from new, 2) restarted ect.

MSalopek:
Consumer genesis is affected by changes in #448.

If it is not a problem, we would need to persist all unsent packets intended for the provider.
* https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/keeper.go#L407
* this mostly impacts chains that are performing a restart from a genesis file, new chains should not have any packets in store
* stored packets include matured VSCs and slash requests

x/ccv/consumer/types/genesis.go
```diff
@@ -104,11 +130,37 @@ func (gs GenesisState) Validate() error {
...
+					ccv.ErrInvalidGenesis, "outstanding downtime must be empty when handshake isn't completed")
+			}
+			if gs.LastTransmissionBlockHeight.Height != 0 {
+				return sdkerrors.Wrap(
+					ccv.ErrInvalidGenesis, "last transmission block height must be zero when handshake isn't completed")
+			}
+			if len(gs.PendingConsumerPackets.List) != 0 {
+				for _, packet := range gs.PendingConsumerPackets.List {
+					if packet.Type == VscMaturedPacket {
+						return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "pending maturing packets must be empty for new chain")
```
mpoke, line 151: nit:
```suggestion
						return sdkerrors.Wrap(ccv.ErrInvalidGenesis, "pending maturing packets must be empty when handshake isn't completed")
```

x/ccv/consumer/keeper/keeper.go
```diff
@@ -428,8 +429,36 @@ func (k Keeper) DeletePendingDataPackets(ctx sdk.Context) {
...
+// GetHeightToValsetUpdateIDs returns all height to valset update id mappings in store
+func (k Keeper) GetHeightToValsetUpdateIDs(ctx sdk.Context) []types.HeightToValsetUpdateID {
+	heightToVCIDs := []types.HeightToValsetUpdateID{}
+	k.IterateHeightToValsetUpdateID(ctx, func(height, vscID uint64) bool {
+		hv := types.HeightToValsetUpdateID{
+			Height:         height,
+			ValsetUpdateId: vscID,
+		}
+		heightToVCIDs = append(heightToVCIDs, hv)
+		return true
```
mpoke, line 447: This will need to be updated after https://github.com/cosmos/interchain-security/pull/459 is merged.

x/ccv/consumer/keeper/genesis_test.go
```diff
@@ -236,35 +361,35 @@ func TestExportGenesis(t *testing.T) {
...
-		ctx, host.PortPath(ccv.ConsumerPortID),
-	).Return(nil, true).Times(1)
+// assert that the given input match the height to valset update ID mappings in the store
+func assertHeightValsetUpdateIDs(t *testing.T, ctx sdk.Context, ck *consumerkeeper.Keeper, heighValsetUpdateIDs []consumertypes.HeightToValsetUpdateID) {
+	ctr := 0
+	ck.IterateHeightToValsetUpdateID(ctx, func(height uint64, vscID uint64) bool {
+		require.Equal(t, heighValsetUpdateIDs[ctr].Height, height)
+		require.Equal(t, heighValsetUpdateIDs[ctr].ValsetUpdateId, vscID)
+		ctr++
+		return true
```
mpoke, line 393: ditto: https://github.com/cosmos/interchain-security/pull/459

x/ccv/consumer/keeper/genesis.go
```diff
@@ -114,70 +127,90 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) (genesis *consumertypes.GenesisSt
...
-
-		outstandingDowntimes := []types.OutstandingDowntime{}
+		outstandingDowntimes := []consumertypes.OutstandingDowntime{}
 		k.IterateOutstandingDowntime(ctx, func(addr string) bool {
-			od := types.OutstandingDowntime{
+			od := consumertypes.OutstandingDowntime{
 				ValidatorConsensusAddress: addr,
 			}
 			outstandingDowntimes = append(outstandingDowntimes, od)
 			return true
```
mpoke, line 146: ditto: https://github.com/cosmos/interchain-security/pull/459

x/ccv/consumer/keeper/keeper.go
```diff
@@ -428,8 +429,36 @@ func (k Keeper) DeletePendingDataPackets(ctx sdk.Context) {
...
+
+// GetOutstandingDowntimes returns all outstanding downtimes in store
+func (k Keeper) GetOutstandingDowntimes(ctx sdk.Context) []consumertypes.OutstandingDowntime {
+	outstandingDowntimes := []types.OutstandingDowntime{}
+	k.IterateOutstandingDowntime(ctx, func(addr string) bool {
+		od := types.OutstandingDowntime{
+			ValidatorConsensusAddress: addr,
+		}
+		outstandingDowntimes = append(outstandingDowntimes, od)
+		return false
```
mpoke, line 461: ditto: https://github.com/cosmos/interchain-security/pull/459

PR #384, opened 2022-10-06
Consumer unbonding as param

### LINKED ISSUE #340: Replace `ComputeConsumerUnbondingPeriod` with a parameter.
There's no reason to compute the consumer's unbonding period on the fly. Provider validators should inspect a consumer's `genesis.json` to see if the unbonding period is satisfactory. Make a parameter and replace all uses of `ComputeConsumerUnbondingPeriod` with that parameter.

smarshall-spitzbart:
Closes #340 
- Base branch is `ccv-timeout-param` for now since that PR needs to be merged before this one
- Need to fix diff tests that don't pass
- Does this PR need to tie into recent param change proposals? See changes from https://github.com/cosmos/interchain-security/pull/364


smarshall-spitzbart:
TODO: Add unit test cases to `params_test` similar to #280

smarshall-spitzbart:
Update, do not set the new param in the consumer addition prop, just have it as default where people can change later

smarshall-spitzbart:
Closing in favor of a more up to date branch

PR #393, opened 2022-10-13
Replace hardcoded constants with params

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/280


x/ccv/consumer/keeper/params.go
```diff
@@ -62,9 +65,28 @@ func (k Keeper) SetProviderFeePoolAddrStr(ctx sdk.Context, addr string) {
...
 
-// GetCCVTimeoutPeriod returns the timeout period for sent ibc packets
+// GetCCVTimeoutPeriod returns the timeout period for sent ccv related ibc packets
 func (k Keeper) GetCCVTimeoutPeriod(ctx sdk.Context) time.Duration {
 	var p time.Duration
 	k.paramStore.Get(ctx, ccvtypes.KeyCCVTimeoutPeriod, &p)
 	return p
 }
+
+// GetTransferTimeoutPeriod returns the timeout period for sent transfer related ibc packets
```
smarshall-spitzbart, line 75: Should setters exist for all of these parameters? Or only the ability to set their value on genesis?

If setters should exist, I'll need to add to https://github.com/cosmos/interchain-security/blob/more-params-rebase/x/ccv/consumer/keeper/params_test.go#L15

jtremback (approved):
LGTM

x/ccv/provider/types/params.go
```diff
@@ -21,11 +21,16 @@ const (
 
 	// DefaultMaxClockDrift defines how much new (untrusted) header's Time can drift into the future.
 	DefaultMaxClockDrift = 10 * time.Second
+
+	// DefaultTrustingPeriodFraction is the default fraction used to compute TrustingPeriod
+	// as UnbondingPeriod / TrustingPeriodFraction
+	DefaultTrustingPeriodFraction = 2
```
mpoke, line 27: We should change this to 3/2, cf. https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-026-ibc-client-recovery-mechanisms.md, i.e., 
>  we recommend that the trusting period for the 07-tendermint client be set to 2/3 of the UnbondingPeriod.

x/ccv/provider/types/params.go
```diff
@@ -57,24 +65,31 @@ func (p Params) Validate() error {
...
-	if ccvtypes.ValidateDuration(p.CcvTimeoutPeriod) != nil {
-		return fmt.Errorf("ccv timeout period is invalid")
+	if err := validateTemplateClient(*p.TemplateClient); err != nil {
+		return err
+	}
+	if err := ccvtypes.ValidateDuration(p.CcvTimeoutPeriod); err != nil {
+		return err
 	}
-	return validateTemplateClient(*p.TemplateClient)
+	if err := ccvtypes.ValidatePositiveInt64(p.TrustingPeriodFraction); err != nil {
```
mpoke, line 74: `TrustingPeriodFraction` is not an integer since it's a fraction. We could use `ibctmtypes.Fraction`. See  `ibctmtypes.DefaultTrustLevel`.

x/ccv/provider/types/params.go
```diff
@@ -57,24 +65,31 @@ func (p Params) Validate() error {
...
+	}
+	return nil
 }
 
 // ParamSetPairs implements params.ParamSet
 func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 	return paramtypes.ParamSetPairs{
 		paramtypes.NewParamSetPair(KeyTemplateClient, p.TemplateClient, validateTemplateClient),
 		paramtypes.NewParamSetPair(ccvtypes.KeyCCVTimeoutPeriod, p.CcvTimeoutPeriod, ccvtypes.ValidateDuration),
+		paramtypes.NewParamSetPair(KeyTrustingPeriodFraction, p.TrustingPeriodFraction, ccvtypes.ValidatePositiveInt64),
```
mpoke, line 85: ditto: not an integer

x/ccv/provider/keeper/params.go
```diff
@@ -25,11 +25,20 @@ func (k Keeper) GetCCVTimeoutPeriod(ctx sdk.Context) time.Duration {
 	return p
 }
 
+// GetTrustingPeriodFraction returns a TrustingPeriodFraction
+// used to compute the provider IBC client's TrustingPeriod as UnbondingPeriod / TrustingPeriodFraction
+func (k Keeper) GetTrustingPeriodFraction(ctx sdk.Context) int64 {
```
mpoke, line 30: ditto: not an int64

x/ccv/consumer/types/params.go
```diff
@@ -1,15 +1,28 @@
...
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	ccvtypes "github.com/cosmos/interchain-security/x/ccv/types"
 )
 
 const (
 	// about 2 hr at 7.6 seconds per blocks
 	DefaultBlocksPerDistributionTransmission = 1000
+
+	// 1 week
+	DefaultTransferTimeoutPeriod = 1 * 7 * 24 * time.Hour
```
mpoke, line 17: `DefaultTransferTimeoutPeriod` should match `DefaultBlocksPerDistributionTransmission`, see https://github.com/cosmos/interchain-security/issues/396#issuecomment-1278611737. Since IBC token transfers do not have to be in order, it could be easier to reason about the distribution protocol if the previous reward times out before sending the next one. Notice that on timeout, the transferred funds are added back to the pool, so the next transfer will include them as well.

x/ccv/consumer/types/params.go
```diff
@@ -1,15 +1,28 @@
...
+
+	// The default fraction of tokens allocated to the consumer redistribution address
+	// during distribution events. The fraction is a string representing a
+	// decimal number. For example "0.75" would represent 75%.
+	DefaultConsumerRedistributeFrac = "0.75"
+
+	// Default number of historical info entries to persist in store
+	DefaultNumHistoricalEntries = 10000
```
mpoke, line 25: Could we use this as default: https://github.com/cosmos/cosmos-sdk/blob/df4d6d1a4cd9fa0247f9db9378db857d95a1c1cb/x/staking/types/params.go#L32

x/ccv/consumer/types/params.go
```diff
@@ -101,3 +139,21 @@ func validateProviderFeePoolAddrStr(i interface{}) error {
...
+func validateConsumerRedistributionFraction(i interface{}) error {
+	str, ok := i.(string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+	dec, err := sdk.NewDecFromStr(str)
+	if err != nil {
+		return err
+	}
+	if !dec.IsPositive() {
```
mpoke, line 152: From the name this function doesn't allow 0. If that is true, we should change that.

mpoke (changes requested):
Nice work @smarshall-spitzbart. See my comments below.

x/ccv/provider/types/params_test.go
```diff
@@ -22,16 +22,16 @@ func TestValidateParams(t *testing.T) {
...
 		{"blank client", types.NewParams(&ibctmtypes.ClientState{},
-			2*7*24*time.Hour), false},
-		{"nil client", types.NewParams(nil, 2*7*24*time.Hour), false},
+			2*7*24*time.Hour, 2), false},
+		{"nil client", types.NewParams(nil, 2*7*24*time.Hour, 2), false},
 		{"0 ccv timeout period", types.NewParams(ibctmtypes.NewClientState("", ibctmtypes.DefaultTrustLevel, 0, 0,
 			time.Second*40, clienttypes.Height{}, commitmenttypes.GetSDKSpecs(), []string{"ibc", "upgradedIBCState"}, true, false),
-			0), false},
+			0, 2), false},
 	}
```
mpoke, line 35: Add test for invalid `TrustingPeriodFraction`.

x/ccv/provider/types/params.go
```diff
@@ -21,11 +21,16 @@ const (
 
 	// DefaultMaxClockDrift defines how much new (untrusted) header's Time can drift into the future.
 	DefaultMaxClockDrift = 10 * time.Second
```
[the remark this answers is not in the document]
smarshall-spitzbart: Correct, the var you've linked is non exported and there's no `DefaultClientState` method

x/ccv/provider/types/params.go
smarshall-spitzbart: This variable is, and has been an integer. "fraction" is a bad previous naming for this variable, "denominator" would have been better. I've created [an issue](https://github.com/cosmos/interchain-security/issues/408) for this, but imo it'd be best to solve in a separate PR. As a parameter, it's not trivial to encode a concrete typed fraction into json, we could make two params for numerator and denominator maybe

x/ccv/provider/types/params.go
smarshall-spitzbart: See above

x/ccv/provider/types/params.go
smarshall-spitzbart: See above

x/ccv/provider/keeper/params.go
smarshall-spitzbart: See above

x/ccv/consumer/types/params.go
smarshall-spitzbart: All makes sense and nice catch, I've made the default transfer timeout 1 hour now

x/ccv/consumer/types/params.go
smarshall-spitzbart: Nice catch! Turns out there's an `IsNegative` method

smarshall-spitzbart:
Thanks for the thorough review @mpoke, you caught a good amount of intricacies I overlooked

All comments have been acted upon unless I've left them unresolved above

x/ccv/consumer/types/params.go
mpoke: @jtremback Do you see any problems with this?

mpoke (approved):

PR #394, opened 2022-10-13
Consistent params validation

smarshall-spitzbart:
Consumer genesis params were previously not validated on `ValidateGenesis`, now they are. This is consistent with how validation is done for provider genesis. These changes are needed for #393 

Note genesis param validation may be redundant with the checks that happen internally for `SetParams` in `InitGenesis`. An alternative to the current changes in this PR is removing genesis param validation altogether from consumer and provider. No matter what, we should be consistent between consumer and provider

### DIFF
--- a/x/ccv/consumer/types/params.go
+++ b/x/ccv/consumer/types/params.go
@@ -57,13 +57,13 @@ func (p Params) Validate() error {
 	if err := ccvtypes.ValidatePositiveInt64(p.BlocksPerDistributionTransmission); err != nil {
 		return err
 	}
-	if err := ccvtypes.ValidateString(p.DistributionTransmissionChannel); err != nil {
+	if err := validateDistributionTransmissionChannel(p.DistributionTransmissionChannel); err != nil {
 		return err
 	}
-	if err := ccvtypes.ValidateString(p.ProviderFeePoolAddrStr); err != nil {
+	if err := validateProviderFeePoolAddrStr(p.ProviderFeePoolAddrStr); err != nil {
 		return err
 	}
-	if err := ccvtypes.ValidateDuration(p.CcvTimeoutPeriod); err != nil {
+	if err := ccvtypes.ValidateCCVTimeoutPeriod(p.CcvTimeoutPeriod); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


x/ccv/consumer/types/params.go
```diff
@@ -52,41 +51,35 @@ func DefaultParams() Params {
 
 // Validate all ccv-consumer module parameters
 func (p Params) Validate() error {
+	if err := ccvtypes.ValidateBool(p.Enabled); err != nil {
+		return err
+	}
+	if err := ccvtypes.ValidatePositiveInt64(p.BlocksPerDistributionTransmission); err != nil {
+		return err
+	}
+	if err := ccvtypes.ValidateString(p.DistributionTransmissionChannel); err != nil {
```
mpoke, line 60: Maybe use `ChannelIdentifierValidator` https://github.com/cosmos/ibc-go/blob/f693f6c11053192f369b5908c9186bd494c3a33d/modules/core/24-host/validate.go#L79

x/ccv/consumer/types/params.go
```diff
@@ -52,41 +51,35 @@ func DefaultParams() Params {
...
+		return err
+	}
+	if err := ccvtypes.ValidateString(p.ProviderFeePoolAddrStr); err != nil {
```
mpoke, line 63: Isn't there a better way to validate an address?

x/ccv/consumer/types/params.go
```diff
@@ -52,41 +51,35 @@ func DefaultParams() Params {
...
+		return err
+	}
+	if err := ccvtypes.ValidateDuration(p.CcvTimeoutPeriod); err != nil {
```
mpoke, line 66: We should at least log a warning message if this value is too small.

x/ccv/provider/types/params.go
```diff
@@ -57,7 +57,7 @@ func (p Params) Validate() error {
 	if p.TemplateClient == nil {
 		return fmt.Errorf("template client is nil")
 	}
-	if ccvtypes.ValidateCCVTimeoutPeriod(p.CcvTimeoutPeriod) != nil {
+	if ccvtypes.ValidateDuration(p.CcvTimeoutPeriod) != nil {
```
mpoke, line 60: ditto

mpoke (approved):
LGTM. See my comments below.

x/ccv/consumer/types/params.go
smarshall-spitzbart: I agree that this is a better validator than just *is this a string?*

However, there are places all over tests that we set the value of `DistributionTransmissionChannel` to an empty string. In fact the[ default parameters ](https://github.com/cosmos/interchain-security/blob/params-validation/x/ccv/consumer/types/params.go#L42) for a consumer has that field as empty. I'm not entirely aware of the intricacies of this param, so is there a better default value we can use?

Btw same issue applies to your comment below and `ProviderFeePoolAddrStr`

smarshall-spitzbart: It might be the case that we only want to properly validate these parameters where they're properly set [here](https://github.com/cosmos/interchain-security/blob/params-validation/x/ccv/consumer/ibc_module.go#L102)

smarshall-spitzbart: See new `validateDistributionTransmissionChannel`

x/ccv/consumer/types/params.go
smarshall-spitzbart: Good call, see `validateProviderFeePoolAddrStr`

x/ccv/types/shared_params.go
```diff
@@ -15,12 +18,67 @@ var (
 )
 
 func ValidateCCVTimeoutPeriod(i interface{}) error {
+	if period, ok := i.(time.Duration); ok && period < 7*24*time.Hour {
+		fmt.Println("A CCV timeout period of less than 1 week is not recommended!")
```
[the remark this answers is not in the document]
smarshall-spitzbart: The validation function that gets called whenever the param is set must abide by [this type](https://github.com/cosmos/cosmos-sdk/blob/interchain-security-rebase/x/params/types/paramset.go#L4), so a context cannot be passed, meaning the sdk style of using `ctx.Logger` is not possible here. I can use `	log.Warn().Msg("message")` if that's better? I'm not too familiar with the pros/cons of different logging types

x/ccv/consumer/types/params_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	testCases := []struct {
+		name    string
+		params  consumertypes.Params
+		expPass bool
+	}{
+		{"default params", consumertypes.DefaultParams(), true},
+		{"custom valid params", consumertypes.NewParams(true, 5, "", "", 5), true},
+		{"custom invalid params, block per dist transmission", consumertypes.NewParams(true, -5, "", "", 5), false},
+		{"custom invalid params, dist transmission channel", consumertypes.NewParams(true, -5, "rubbishchannel", "", 5), false},
```
sainoe, line 22:
```suggestion
		{"custom invalid params, dist transmission channel", consumertypes.NewParams(true, 5, "badchannel", "", 5), false},
```

x/ccv/consumer/types/params_test.go
```diff
@@ -0,0 +1,35 @@
...
+		{"custom invalid params, provider fee pool addr string", consumertypes.NewParams(true, -5, "", "imabadaddress", 5), false},
```
sainoe, line 23:
```suggestion
		{"custom invalid params, provider fee pool addr string", consumertypes.NewParams(true, 5, "", "imabadaddress", 5), false},
```

sainoe (approved):
LGTM! Just two tiny adjustments in `x/ccv/consumer/types/params_test.go`

PR #405, opened 2022-10-19
Danwt/remove duplicate imports

danwt:
The linter was complaining.


x/ccv/provider/keeper/genesis_test.go
```diff
@@ -11,7 +11,6 @@ import (
 	consumertypes "github.com/cosmos/interchain-security/x/ccv/consumer/types"
 	"github.com/cosmos/interchain-security/x/ccv/provider/keeper"
 	"github.com/cosmos/interchain-security/x/ccv/provider/types"
-	providertypes "github.com/cosmos/interchain-security/x/ccv/provider/types"
```
mpoke, line 14: Personally, I would have prefer to keep `providertypes` than `types`, as it's more descriptive.

mpoke (approved):
LGTM. See my comment below.

x/ccv/provider/keeper/genesis_test.go
MSalopek: I agree with Marius on this one

MSalopek (approved):
Looks good.
A small nitpick would be to maybe use
`providertypes "github.com/cosmos/interchain-security/x/ccv/provider/types"` instead of
`"github.com/cosmos/interchain-security/x/ccv/provider/types"` since it is more apparent what pkg is used in the code.

danwt:
Out of date

Issue #278, opened 2022-08-23
Channel initialization should have a timeout


mpoke:
Once a gov proposal to add a new consumer chain passes on the provider, a new client to the consumer chain is created.  From this point on, all the unbonding operations (e.g., undelegations) cannot complete until the CCV channel is established, which requires both the connection and the channel opening handshakes to complete. Since there are no timeouts on these handshakes, the channel initialization could theoretically take forever. In practice, this can happen if there is no relayer to relay the connection and channel opening handshake messages (e.g., ChanOpenInit, ChanOpenTry).

To guarantee that all unbonding operation eventually complete, channel initialization should be time-bounded (via a timeout). For example, once the client is created on the provider, the CCV channel needs to be established within 1 week.

josef-widder:
I guess there are some details to be clarified, E.g., does the consumer chain need to shut down in case the channel is not open? Can it be the case that the channel is open on the consumer chain, but times out on the provider chain? All this can be solved, we just need to think about them.

In general, I don't think there is a way around adding some synchrony assumptions as you proposed here.

mpoke:
> does the consumer chain need to shut down in case the channel is not open?

That's a good point. We can let it happen outside the protocol, i.e., the validators will eventually stop running the consumer. Anyway the consumer chain has all user TXs disabled until the channel is opened, so there is not much utility running it. However, I agree that we need a mechanism of notifying the validators to stop running the consumer binary. 

> Can it be the case that the channel is open on the consumer chain, but times out on the provider chain?

No. The channel first becomes established (aka opened) on the provider during `OpenChanConfirm` and then on the consumer when receiving first VSC packet. If the timeout is reach on the provider, there is no VSC packet sent.

jtremback:
The only way I can see this happening IRL is if there is some coding issue with the consumer chain which causes it not to be able to complete the handshake. I would put this in the same bucket of arbitrarily bad behavior as a consumer chain that simply never sends VSCMaturedPackets, and it is just something that governance will need to catch in consumer chain code review.

Should this be left open? I will move it out of v1 release in any case.

mpoke:
Same argument as here https://github.com/cosmos/interchain-security/issues/283#issuecomment-1230930067. If the consumer goes down during channel initialization, all the unbondings started on the consumer during this time will not finish. The only way to deal with it is to pass a proposal to remove the non-responsive consumer chain. Seems much easier to have it implemented in the protocol.

Issue #283, opened 2022-08-23
VSCPackets should have timeout on provider


mpoke:
Currently, `VSCPacket`s (like any other IBC packets) have a timeout from the perspective of the consumer chains. This means that if a consumer chain is not live, the `VSCPacket`s sent to that consumer chain will be "frozen in time", i.e., they cannot time out, which means that the unbonding operations associated with these `VSCPacket`s cannot complete.

A solution would be to add a timeout period on the provider. Basically, for every `VSCPacket` sent to a consumer chain, unless the corresponding `MaturedVSCPacket` is received back within the timeout period, the consumer chain is removed. This would also provide a bound on the maximum time an unbonding operation takes on the provider chain. 

This issue is related to https://github.com/cosmos/interchain-security/issues/278

mpoke:
This would also fix the following issue:

- a packet sent by a consumer chain times out;
- the consumer receives a `MsgTimeout` message and, as a result, closes the CCV channel (see https://github.com/cosmos/ibc-go/blob/397b88cf98214e62437a94db1a26c935d4d153fd/modules/core/04-channel/keeper/timeout.go#L157);
- at the beginning of the next block, the consumer chain shuts down, see https://github.com/cosmos/interchain-security/blob/3f4e0cef2e1a217ddf6ee7d1802f7dd5ad708d9d/x/ccv/consumer/module.go#L157
- relayers can no longer create proofs for `MsgTimeoutOnClose` or `MsgTimeout` messages to send to the provider; 
- as a result, the unbonding operations waiting for that consumer chain can no longer complete.

smarshall-spitzbart:
Good stuff. As discussed the semantics of a timeout initiated on a consumer chain should be clearly distinguished from this new timeout initiated on the provider chain. The provider would (eventually) process either type of timeout

jtremback:
I think this is a great idea and mitigates a lot of the worries I have around eternally-frozen unbondings

mpoke:
This could also solve #278

danwt:
Is there a rough idea of what the timeout value will be?

danwt:
If the timeout value is determined by the provider, not on a chain-by-chain basis then also it strikes me as more efficient to store a list of timeouts once rather than per-chain and store a map `<chainId, greatestVSCIDMaturityReceived>`. The values of that map tell you the index in the list (or allow you to search via binary search), and the smallest value in the map corresponds to the first entry in the list.

mpoke:
[...]

PR #406, opened 2022-10-20
Channel initialization timeout

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/278


tests/e2e/common.go
```diff
@@ -1,23 +1,22 @@
...
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
-
 	"github.com/cosmos/interchain-security/testutil/e2e"
 	providertypes "github.com/cosmos/interchain-security/x/ccv/provider/types"
 	ccv "github.com/cosmos/interchain-security/x/ccv/types"
 	"github.com/cosmos/interchain-security/x/ccv/utils"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
 
-	"github.com/cosmos/ibc-go/modules/core/exported"
```
mpoke, line 17: Be careful to not introduce dependancies to multiple versions of ibc-go.

tests/e2e/unbonding.go
```diff
@@ -139,68 +140,114 @@ func (s *CCVTestSuite) TestUndelegationNoValsetChange() {
...
+				ts := s.providerCtx().BlockTime().Add(pUnbondingPeriod + 24*time.Hour)
+				pk.SetInitTimeoutTimestamp(s.providerCtx(), s.consumerChain.ChainID, uint64(ts.UnixNano()))
+			}, false,
+		},
+		{
+			"channel handshake times out before unbonding period", func(pk *providerkeeper.Keeper, pUnbondingPeriod time.Duration) {
+				// change the init timeout timestamp for this consumer chain
+				// to make sure the chain is removed before the unbonding period elapses
+				ts := s.providerCtx().BlockTime().Add(pUnbondingPeriod - 24*time.Hour)
+				pk.SetInitTimeoutTimestamp(s.providerCtx(), s.consumerChain.ChainID, uint64(ts.UnixNano()))
```
smarshall-spitzbart, line 163: Just curious how this works, you're overwriting the value that was set in `CreateConsumerClient` during setup correct?

x/ccv/provider/keeper/keeper.go
```diff
@@ -826,3 +828,44 @@ func (k Keeper) DeleteConsumerClientId(ctx sdk.Context, chainID string) {
...
+// SetInitTimeoutTimestamp sets the init timeout timestamp for the given chain ID
+func (k Keeper) SetInitTimeoutTimestamp(ctx sdk.Context, chainID string, ts uint64) {
+	store := ctx.KVStore(k.storeKey)
+	tsBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(tsBytes, ts)
+	store.Set(types.InitTimeoutTimestampKey(chainID), tsBytes)
+}
+
+// GetInitTimeoutTimestamp returns the init timeout timestamp for the given chain ID.
+func (k Keeper) GetInitTimeoutTimestamp(ctx sdk.Context, chainID string) (uint64, bool) {
```
smarshall-spitzbart, line 843: Some of these methods are purely test helpers, right? Might be useful to mark which ones

x/ccv/provider/keeper/relay.go
```diff
@@ -278,3 +296,28 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
...
+
+// EndBlockCIS contains the EndBlock logic needed for
+// the Consumer Chain Removal sub-protocol
+func (k Keeper) EndBlockCCR(ctx sdk.Context) {
+	currentTime := uint64(ctx.BlockTime().UnixNano())
+
+	// iterate over initTimeoutTimestamps
+	var removedChainIds []string
+	k.IterateInitTimeoutTimestamp(ctx, func(chainID string, ts uint64) bool {
+		if currentTime > ts {
```
smarshall-spitzbart, line 308: Would it be better to only store the timestamp when initialization started, and compute the timeout timestamp here with the most up to date `InitTimeoutPeriod` ? 

Maybe the desired functionality is that the new timeout period is effective immediately (even for in progress initializations), but I'm not sure

tests/e2e/unbonding.go
mpoke: Yes. :)

tests/e2e/channel_init.go
```diff
@@ -94,3 +94,100 @@ func (suite *CCVTestSuite) TestProviderClientMatches() {
...
+		// channel opening handshake
+		tc.handshake()
+
+		// call NextBlock
+		suite.providerChain.NextBlock()
+
+		// increment time
+		incrementTimeBy(suite, initTimeout)
+
+		// check whether the chain was removed
```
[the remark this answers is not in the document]
mpoke: I didn't want to change too much the `checkConsumerChainIsRemoved` utility function. Plus, checking that the chain was actually removed is not really necessary for this particular test. That's something we test in `TestStopConsumerChain`.

x/ccv/provider/keeper/relay.go
mpoke: Not sure that I follow. Do you have a specific concern in mind? What exactly do you mean by "when initialization started"? What initialization? IIUC what you're suggesting, it requires to compute the timeout timestamp in every EndBlock, right?

smarshall-spitzbart: The edge case I'm referring to is that a governance proposal changes the `InitTimeoutPeriod` param in the midst of one or more in-progress channel initializations. If you compute timeout timestamps during the endblock routine instead of precomputing the timeout timestamp as you've implemented here, you'd always be timing out channel initializations with the most recent `InitTimeoutPeriod`. 

Note that during the endblock, you do not have to compute a timeout timestamp for every in progress initialization. You can store the timestamps in order and stop iteration once you've reached the first chain that doesn't need a timeout

tests/e2e/channel_init.go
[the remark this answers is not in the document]
smarshall-spitzbart: All seems reasonable 👍

x/ccv/provider/keeper/relay.go
mpoke: I understand. It depends on the desired outcome. Do we want for a governance proposal to affect the init timeout timestamp of consumers with the channel handshake in progress? @jtremback 

The same question is relevant for the VSC timeout also (see https://github.com/cosmos/interchain-security/issues/283). Do we want a governance proposal to change how long we are waiting for the maturity of already sent VSCPackets?

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -58,11 +58,14 @@ message ConsumerAdditionProposal {
 // Params defines the parameters for CCV Provider module
 message Params {
   ibc.lightclients.tendermint.v1.ClientState template_client = 1;
+  // TrustingPeriodFraction is used to compute the consumer and provider IBC client's TrustingPeriod
```
AdityaSripal, line 61:
```suggestion
  // TrustingPeriodFraction is used to compute the consumer and provider IBC client's TrustingPeriod from the chain defined UnbondingPeriod
```

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -58,11 +58,14 @@ message ConsumerAdditionProposal {
...
+    int64 trusting_period_fraction = 2;
```
AdityaSripal, line 62: formatting

tests/e2e/channel_init.go
```diff
@@ -93,3 +93,100 @@ func (suite *CCVTestSuite) TestProviderClientMatches() {
...
+		_, found = providerKeeper.GetConsumerClientId(suite.providerCtx(), chainID)
+		suite.Require().Equal(!tc.removed, found, "unexpected outcome; test: %s", tc.name)
+
+		if tc.removed {
+			// check if the chain was properly removed
+			suite.checkConsumerChainIsRemoved(chainID, false, false)
```
AdityaSripal, line 184: Can we also check that the next step of the handshake can't complete as well

x/ccv/provider/keeper/keeper.go
```diff
@@ -282,10 +282,12 @@ func (k Keeper) SetConsumerChain(ctx sdk.Context, channelID string) error {
...
 	// - set channel mappings
-	k.SetChainToChannel(ctx, tmClient.ChainId, channelID)
-	k.SetChannelToChain(ctx, channelID, tmClient.ChainId)
+	k.SetChainToChannel(ctx, chainID, channelID)
+	k.SetChannelToChain(ctx, channelID, chainID)
 	// - set current block height for the consumer chain initialization
-	k.SetInitChainHeight(ctx, tmClient.ChainId, uint64(ctx.BlockHeight()))
+	k.SetInitChainHeight(ctx, chainID, uint64(ctx.BlockHeight()))
+	// - remove init timeout timestamp
+	k.DeleteInitTimeoutTimestamp(ctx, chainID)
```
AdityaSripal, line 290: Would be helpful to note in the godocs of these functions where it is called.

e.g. OnChanOpenConfirm

Since this is a complex application it would be easier for reviewers to quickly validate correctness, rather than having to find the caller themselves

AdityaSripal (approved):

tests/e2e/channel_init.go
mpoke: That's not that easy due to the `expPass` param of the `SignAndDeliver` method from the ibc-go testing framework (see https://github.com/cosmos/ibc-go/blob/a0e59b8e7a2e1305b7b168962e20516ca8c98fad/testing/simapp/test_helpers.go#L336). This method is called by `SendMsgs()` with `expPass = true` (see https://github.com/cosmos/ibc-go/blob/a0e59b8e7a2e1305b7b168962e20516ca8c98fad/testing/chain.go#L319). `SendMsgs()` is called by `ChanOpenTry()`. This means that unless I change the ibc-go testing framework, calling `ChanOpenTry()` after the consumer is removed will result in the test failing with the following error: 
```
failed to execute message; message index: 0: channel open try callback failed: cannot find client for consumer chain testchain2: client not found
```

@AdityaSripal is there another way to do this?

x/ccv/provider/types/params.go
```diff
@@ -25,12 +25,16 @@ const (
 	// DefaultTrustingPeriodFraction is the default fraction used to compute TrustingPeriod
 	// as UnbondingPeriod / TrustingPeriodFraction
 	DefaultTrustingPeriodFraction = 2
+
+	// DafaultInitTimeoutPeriod defines the init timeout period
+	DafaultInitTimeoutPeriod = 7 * 24 * time.Hour
```
jtremback, line 30: Spelling

tests/e2e/unbonding.go
```diff
@@ -139,68 +140,114 @@ func (s *CCVTestSuite) TestUndelegationNoValsetChange() {
...
+	testCases := []struct {
+		name                       string
+		updateInitTimeoutTimestamp func(*providerkeeper.Keeper, time.Duration)
+		removed                    bool
+	}{
+		{
+			"channel handshake completes before unbonding period", func(pk *providerkeeper.Keeper, pUnbondingPeriod time.Duration) {
+				// change the init timeout timestamp for this consumer chain
+				// to make sure the chain is not removed before the unbonding period elapses
+				ts := s.providerCtx().BlockTime().Add(pUnbondingPeriod + 24*time.Hour)
```
jtremback, line 154: Fix this by making it so channel handshake completes after unbonding period but before init timeout period

x/ccv/provider/keeper/relay.go
mpoke: After discussing with @jtremback we decided to store the timeout timestamp instead of the timestamp of starting the channel handshake. This will be consistent of how the `UnbondingPeriod` is used in the SDK (when computing the completion time of unbonding operations).

jtremback (approved):
LGTM

PR #407, opened 2022-10-20
add multiple consumer chains in integration tests

MSalopek:
Initial implementation of multiple consumer chains in integration tests.


tests/integration/actions.go
```diff
@@ -538,22 +550,38 @@ type addIbcChannelAction struct {
...
+		)
+	} else {
+		//#nosec G204 -- Bypass linter warning for spawning subprocess with cmd arguments.
+		cmd = exec.Command("docker", "exec", tr.containerConfig.instanceName, "hermes",
+			"create", "channel",
+			"--a-chain", string(tr.chainConfigs[action.chainA].chainId),
+			"--a-connection", "connection-"+fmt.Sprint(action.connectionA),
+			"--a-port", action.portA,
+			"--b-port", action.portB,
+			"--channel-version", tr.containerConfig.ccvVersion,
```
jtremback, line 581: Can we get rid of this version completely?

jtremback:
My comments from 11/2 code review:

- Get rid of transfer channels map in config
  - Config should work with steps in any order and combination, transfer channels map breaks this
- Pass transfer channel into the action itself
- Do not dynamically generate stuff in steps file, should basically be serialized
- Fill in correct transfer channel in action in the steps file by hand

tests/integration/config.go
```diff
@@ -206,6 +208,56 @@ func DemocracyTestRun() TestRun {
...
+				binaryName:     "interchain-security-cd",
+				ipPrefix:       "7.7.8",
+				votingWaitTime: 10,
+				genesisChanges: ".app_state.gov.voting_params.voting_period = \"10s\" | " +
+					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
+					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
+					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+			},
+			chainID("densu"): {
```
smarshall-spitzbart, line 246: `densu` doesn't stand for democracy consumer, does it?

tests/integration/steps_downtime.go
```diff
@@ -186,8 +186,5 @@ func stepsDowntime(consumerName string) []Step {
 				},
 			},
 		},
-
-		// TODO: Test full unbonding functionality, tracked as: https://github.com/cosmos/interchain-security/issues/311
```
smarshall-spitzbart, line 190: Is that issue solved now?

tests/integration/steps_multi_consumer_delegation.go
```diff
@@ -0,0 +1,305 @@
+package main
+
+// stepsDelegate tests basic delegation and resulting validator power changes.
+func stepsMultiConsumerDelegate(consumer1, consumer2 string) []Step {
```
smarshall-spitzbart, line 4: Could we make this a variadic function?

smarshall-spitzbart (commented):
The overall idea of this PR makes sense 👍 , I think some of the functions should be formatted in such a way that they could accept more than 2 consumers (variadic functions, consumer lists instead of "consumer1, consumer 2", etc.).

IMO switching from testing 2 consumers to 4 consumers for example should be pretty trivial. So maybe see if there's ways to implement the steps, and actions s.t. adding more consumers would only require adding more entries to the config file.

Lmk if you want me to take a closer look at any specific files, this is just my highest level feedback without getting too deep into the PR

tests/integration/config.go
MSalopek: There's `consu` and `densu` is just consu with the C replaced.

There's a constraint somewhere that chain names must be >= 5 chars long.

tests/integration/steps_multi_consumer_delegation.go
MSalopek: Let's leave it at just 2 consumers for now.

Each chain has it's own channel and consumer states can differ depending on when the relaying happens in the tests. I would not go there at this point.

tests/integration/actions.go
MSalopek: The functionality changed, this was for adding IBC connections with sovereign chains.

I don't think this can be removed yet.

MSalopek:
> My comments from 11/2 code review:
> 
> * Get rid of transfer channels map in config
>   
>   * Config should work with steps in any order and combination, transfer channels map breaks this
> * Pass transfer channel into the action itself
> * Do not dynamically generate stuff in steps file, should basically be serialized
> * Fill in correct transfer channel in action in the steps file by hand
> * You probably don't need to load and build gaiad in the Dockerfile, you should be able to just use interchain-security-pd

The PR was changed to add multiple consumer chains instead of a sovereign chain. Sovereign chains can be added later

tests/integration/steps_downtime.go
MSalopek: Not yet. Accidentally deleted the comment.

Comment was reintroduced

tests/integration/main.go
```diff
@@ -108,7 +108,8 @@ func (tr *TestRun) runStep(step Step, verbose bool) {
 
 	// Check state
 	if !reflect.DeepEqual(actualState, modelState) {
-		fmt.Println("action", reflect.TypeOf(step.action).Name())
+		fmt.Printf("=============== %s FAILED ===============\n", tr.name)
+		fmt.Println("FAILED action", reflect.TypeOf(step.action).Name())
```
smarshall-spitzbart, line 112: Nice

tests/integration/steps_double_sign.go
```diff
@@ -13,112 +14,204 @@ func stepsDoubleSign(consumerName, doubleSignChain, doubleSignValidator string)
...
-						validatorID("bob"):   0,
+						validatorID("alice"): 509,
+						validatorID("bob"):   0, // from 495 to 0
+						validatorID("carol"): 0,
+					},
+				},
+				chainID(consumer1): ChainState{
+					ValPowers: &map[validatorID]uint{
+						validatorID("alice"): 509,
+						validatorID("bob"):   495, // from 495 to 0
```
smarshall-spitzbart, line 149:
```suggestion
						validatorID("bob"):   495, // not slashed yet
```

tests/integration/steps_double_sign.go
```diff
@@ -13,112 +14,204 @@ func stepsDoubleSign(consumerName, doubleSignChain, doubleSignValidator string)
...
+					ValPowers: &map[validatorID]uint{
+						validatorID("alice"): 509,
+						validatorID("bob"):   495, // not slashed yet
+						validatorID("carol"): 0,
+					},
+				},
+			},
+		},
+		{
+			// consumer1 learns about the slash and powerchanges
```
smarshall-spitzbart, line 194:
```suggestion
			// consumer2 learns about the slash and powerchanges
```

tests/integration/steps_double_sign.go
```diff
@@ -13,112 +14,204 @@ func stepsDoubleSign(consumerName, doubleSignChain, doubleSignValidator string)
...
+					},
+				},
+			},
+		},
+		{
+			// relay power change to consumer1
+			action: relayPacketsAction{
+				chain:   chainID("provi"),
+				port:    "provider",
+				channel: 0, // consumer1 channel
```
smarshall-spitzbart, line 43: Neat that you can relay by channel, makes the tests very clear

tests/integration/steps_multi_consumer_delegation.go
```diff
@@ -0,0 +1,305 @@
...
+				chainID(consumer1): ChainState{
+					ValPowers: &map[validatorID]uint{
+						validatorID("alice"): 511,
+						validatorID("bob"):   500,
+						validatorID("carol"): 500,
+					},
+				},
+				chainID(consumer2): ChainState{
+					ValPowers: &map[validatorID]uint{
+						validatorID("alice"): 511, // unchanged
```
smarshall-spitzbart, line 93:
```suggestion
						validatorID("alice"): 511, // changed
```

tests/integration/steps_multi_consumer_downtime.go
```diff
@@ -0,0 +1,408 @@
+package main
+
+// stepsDowntime tests validator jailing and slashing.
+func stepsMultiConsumerDowntimeFromConsumer(consumer1, consumer2 string) []Step {
```
smarshall-spitzbart, line 4: Not the biggest fan of this name haha but I can't think of a better one. Maybe you could organize the multiple consumer steps into a file called `multi-consumer-steps.go` and avoid having "multi consumer" in the name?

Not that important tho, leaving as-is could be fine

tests/integration/steps_multi_consumer_downtime.go
```diff
@@ -0,0 +1,408 @@
...
+					},
+				},
+				chainID(consumer1): ChainState{
+					ValPowers: &map[validatorID]uint{
+						validatorID("alice"): 509,
+						validatorID("bob"):   495,
+						validatorID("carol"): 0,
+					},
+				},
+				// not relayed yet - powers unchanged
```
smarshall-spitzbart, line 306:
```suggestion
				// powers now changed
```

tests/integration/steps_start_chains.go
```diff
@@ -123,30 +108,12 @@ func stepsStartChains(consumerName string, setupTransferChan bool) []Step {
 				},
 			},
 		},
-		{
```
smarshall-spitzbart, line 126: Is there a reason this was removed? Seems like a useful check

smarshall-spitzbart (commented):
More comments after reading through steps logic

MSalopek:
> The overall idea of this PR makes sense 👍 , I think some of the functions should be formatted in such a way that they could accept more than 2 consumers (variadic functions, consumer lists instead of "consumer1, consumer 2", etc.).
> 
> IMO switching from testing 2 consumers to 4 consumers for example should be pretty trivial. So maybe see if there's ways to implement the steps, and actions s.t. adding more consumers would only require adding more entries to the config file.
> 
> Lmk if you want me to take a closer look at any specific files, this is just my highest level feedback without getting too deep into the PR

In some earlier talks with @jtremback we agreed that the optimal way to approach this is to treat all inputs (all steps) as if they are coming from some JSON dump. Ideally, we would use some sort of JSON trace dump generated by the diffetesting framework. That approach precludes the usage of variadic functions and consumer lists since all input is coming from an external source and contain the entire action and state regardless of the number of consumer chains.

I opted to adopt the stated approach, except for starting chains since that comes from a config file.

Having variadic functions looks nice, but is way more difficult to set up. Check some of the commits in the PR and you will see an attempt to use variadic funcs and the difficulties of reading such approach. In each function body you end up wriring a bunch of loops to populate the model state.

tests/integration/steps_start_chains.go
MSalopek: It made tracking validator balances more difficult. I later steps. I think I had some issues with democracy and shutting down a consumer

tests/integration/steps_multi_consumer_downtime.go
MSalopek: > Seems solid 👍, it's a bit unclear why a JSON dump of test traces couldn't include steps for multiple consumers, and be tied into a variadic function, since JSON has an array type. Regardless, there's more important things to worry about and this PR is a huge improvement already

Will update in the next iteration.

tests/integration/steps_double_sign.go
MSalopek: Done.

tests/integration/steps_double_sign.go
MSalopek: Done.

tests/integration/steps_double_sign.go
MSalopek: Done.

tests/integration/steps_multi_consumer_delegation.go
MSalopek: Done.

tests/integration/steps_multi_consumer_downtime.go
MSalopek: Done

tests/integration/actions.go
```diff
@@ -367,6 +367,7 @@ func (tr TestRun) submitParamChangeProposal(
 		`--chain-id`, string(tr.chainConfigs[action.chain].chainId),
 		`--home`, tr.getValidatorHome(action.chain, action.from),
 		`--node`, tr.getValidatorNode(action.chain, action.from),
+		`--gas`, "900000",
```
danwt, line 370: Is gas `auto` possible? What was the reason for 900000?

tests/integration/steps.go
```diff
@@ -14,21 +14,28 @@ func concatSteps(steps ...[]Step) []Step {
...
-	stepsStartChains("consu", false),
-	stepsDoubleSign("consu", "provi", "carol"),
+var multipleConsumers = concatSteps(
+	stepsStartChains([]string{"consu", "densu"}, false),
+	stepsMultiConsumerDelegate("consu", "densu"),
+	stepsMultiConsumerUnbond("consu", "densu"),
+	stepsMultiConsumerRedelegate("consu", "densu"),
+	stepsMultiConsumerDowntimeFromConsumer("consu", "densu"),
+	stepsMultiConsumerDowntimeFromProvider("consu", "densu"),
+	stepsDoubleSign("consu", "densu"), // double sign on one of the chains
```
smarshall-spitzbart, line 40: nit: consistency with the other names that incorporate multiple consumers
```suggestion
	stepsMultiConsumerDoubleSign("consu", "densu"), // double sign on one of the chains
```

danwt (approved):
Looks good! Nice job 👍

smarshall-spitzbart (approved):
🚀

tests/integration/actions.go
smarshall-spitzbart: @MSalopek can answer for this tx specifically, but I've previously needed to set a gas amount above the default for some integration test actions. Some of the on chain computations in these tests use a lot of gas

PR #410, opened 2022-10-20
Consumer Unbonding As Param

### LINKED ISSUE #340: Replace `ComputeConsumerUnbondingPeriod` with a parameter.

smarshall-spitzbart:
Closes #340 

Note that this implementation automatically sets the consumer unbonding period as default on genesis, where that param can later be changed through governance. The alternative is to make the consumer unbonding period a parameter in every consumer addition proposal, but that'd require a bigger refactor to integration tests, proposal JSONs, etc.


x/ccv/consumer/keeper/genesis.go
```diff
@@ -45,10 +44,6 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
 			panic(err)
 		}
 
-		// Set the unbonding period: use the unbonding period on the provider to
```
smarshall-spitzbart, line 48: Now this is set by SetParams above

MSalopek (approved):
LGTM. Seems like most of the work affects testing, from what I can tell.

> Note that this implementation automatically sets the consumer unbonding period as default on genesis, where that param can later be changed through governance. The alternative is to make the consumer unbonding period a parameter in every consumer addition proposal, but that'd require a bigger refactor to integration tests, proposal JSONs, etc.

Would changing a param through governance be a big issue for consumer chains? To me it seems like a minor inconvenience. If it's done through governance, do we need extra checks to make sure the unbonding period is less than the provider's period?

x/ccv/consumer/keeper/genesis.go
```diff
@@ -45,10 +44,6 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
...
-		// compute the unbonding period on the consumer
-		unbondingTime := utils.ComputeConsumerUnbondingPeriod(state.ProviderClientState.UnbondingPeriod)
```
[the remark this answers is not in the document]
smarshall-spitzbart: Great catch! I'm actually a dingus and the whole point of this PR (and making consumer unbonding a param) is to remove that method 😆, see 6c301ad2940885c5c2ad505c51b6b2938f199952

smarshall-spitzbart:
> LGTM. Seems like most of the work affects testing, from what I can tell.
> 
> > Note that this implementation automatically sets the consumer unbonding period as default on genesis, where that param can later be changed through governance. The alternative is to make the consumer unbonding period a parameter in every consumer addition proposal, but that'd require a bigger refactor to integration tests, proposal JSONs, etc.
> 
> Would changing a param through governance be a big issue for consumer chains? To me it seems like a minor inconvenience. If it's done through governance, do we need extra checks to make sure the unbonding period is less than the provider's period?

Good questions, it shouldn't be a big inconvenience to change the consumer unbonding period now that it's a "param". The governance module of cosmos sdk would have more specifics on that process if you're curious. 

Re extra checks, I believe this concern is addressed by the "democracy consumer". Although my knowledge is still fuzzy there, and I haven't yet found good documentation on the idea. @jtremback could prob answer better

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -17,19 +18,19 @@ import (
...
-	storedUnbondingPeriod, ok := consumerKeeper.GetUnbondingTime(ctx)
-	require.True(t, ok)
-	require.Equal(t, storedUnbondingPeriod, unbondingPeriod)
+	ubdTime := consumerKeeper.GetUnbondingPeriod(ctx)
+	require.Equal(t, consumertypes.DefaultConsumerUnbondingPeriod, ubdTime)
+
+	consumerKeeper.SetUnbondingPeriod(ctx, time.Hour*24*10)
+	storedUnbondingPeriod := consumerKeeper.GetUnbondingPeriod(ctx)
+	require.Equal(t, time.Hour*24*10, storedUnbondingPeriod)
 }
```
mpoke, line 34: I think we no longer need this test. This was needed for testing `SetUnbondingTime` and `GetUnbondingTime`, which were removed (moved to `param.go`). The testing should be part of `params_test.go`, see `TestParams`.

mpoke (approved):
Nice work @smarshall-spitzbart. Just one nit.

MSalopek:
> > LGTM. Seems like most of the work affects testing, from what I can tell.
> > > Note that this implementation automatically sets the consumer unbonding period as default on genesis, where that param can later be changed through governance. The alternative is to make the consumer unbonding period a parameter in every consumer addition proposal, but that'd require a bigger refactor to integration tests, proposal JSONs, etc.
> > 
> > 
> > Would changing a param through governance be a big issue for consumer chains? To me it seems like a minor inconvenience. If it's done through governance, do we need extra checks to make sure the unbonding period is less than the provider's period?
> 
> Good questions, it shouldn't be a big inconvenience to change the consumer unbonding period now that it's a "param". The governance module of cosmos sdk would have more specifics on that process if you're curious.
> 
> Re extra checks, I believe this concern is addressed by the "democracy consumer". Although my knowledge is still fuzzy there, and I haven't yet found good documentation on the idea. @jtremback could prob answer better

Talked to Marius, the protocol works regardless of the length of the consumer unbonding period. With that, I think no extra actions are necessary

x/ccv/consumer/keeper/keeper_test.go
smarshall-spitzbart: Nice catch 👍

PR #412, opened 2022-10-20
diff test patch

danwt:
@smarshall-spitzbart this should address the issue


smarshall-spitzbart:
🙏 🙏 Looks great and makes sense

PR #413, opened 2022-10-21
Importable integration

smarshall-spitzbart:


MSalopek:
Cleaning up PRs. Ideas from this PR are useful but needs a lot more work.

Will reopen if needed.

@smarshall-spitzbart FYI.

smarshall-spitzbart:
Yea sounds like we won't be sticking with "integration tests" forever. So we can build this functionality into future testing frameworks

PR #415, opened 2022-10-23
danwt/keymap milestone

danwt:
Hey @jtremback 

This is my latest milestone.

Link to my current notes (checkboxes don't mean anything):

https://github.com/cosmos/interchain-security/blob/505f74efd17be20e939e75880dc40e934eacddfc/x/ccv/provider/keeper/keymap.go#L17-L74

API doc (needs updating/finishing since a while ago but idea is there):

https://github.com/cosmos/interchain-security/tree/danwt/keymap-milestone/docs/keymap

Since slack comments on Friday things are now simpler and inline with the rest of the codebase.

The whole codebase unit tests and diff test are passing. Keymap core test are 100% coverage with a bunch of properties and invariants. I want to extend diff tests with some random key assignments, it shouldn't take long but that's my next step. The things that still need doing are

```
Major todos:
- query : boilerplate but I haven't done it in a while
- tx : boilerplate but I haven't done it before
- diff test with random key assignments : should be pretty quick. Hopefully it wont' find any bugs
- validator destruction GC : should be quick, need to hook into `OnDestroyValidator` or equivalent
- consumer chain removal : should  be very quick
```

Also I need to remove any debug prints, working comments, todos ect. 

Excuse the diff I need to figure out why it's got so many changes. I thought I was based off main/ .


danwt:
The diff seems to be broken I'll create a new branch.

PR #418, opened 2022-10-27
Remove unneeded const variables

smarshall-spitzbart:
https://github.com/cosmos/interchain-security/pull/317 added two const defaults that may have been relevant at the time, but are now are unneeded after recent param refactors. The uneeded const defaults are removed by this PR. 
- `DefaultTrustingPeriod` as it existed before can be derived from `DefaultConsumerUnbondingPeriod` and `DefaultTrustingPeriodFraction` which already exist as params, and the trusting period itself is not a provider param. 
- `DefaultUnbondingPeriod` as it existed before is confusing, because the provider uses the staking module's unbonding period default, and the consumer has its own default defined


MSalopek (approved):
Looks good! Awesome job!

danwt (approved):

PR #101, opened 2022-05-13
Staking changes that do not change the validator set

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/33

PR #422, opened 2022-11-01
VSCPackets should have timeout on provider

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/283


x/ccv/provider/types/keys.go
```diff
@@ -281,3 +245,83 @@ func HashString(x string) []byte {
 	hash := sha256.Sum256([]byte(x))
 	return hash[:]
 }
+
+// tsAndChainIdKey returns the key with the following format:
+// bytePrefix | len(timestamp) | timestamp | chainID
+func tsAndChainIdKey(prefix byte, timestamp time.Time, chainID string) []byte {
```
MSalopek, line 251: A small nitpick about this and other functions we use to create keys.

Sometimes using `copy` with array/slice contatenation can be difficult to read. To me it seems like there's a lot more going on with the code than there actually is.

In this function, all values are simply appended to a byte slice.

Maybe we could make the code a bit more readable and reduce the number of comments by doing something like this:

```golang
func tsAndChainIdKey(prefix byte, timestamp time.Time, chainID string) []byte {
	timeBytes := sdk.FormatTimeBytes(timestamp)
	lenTimeBytes := sdk.Uint64ToBigEndian(uint64(len(timeBytes)))
        lenKey := len([]byte{prefix})+8+len(timeBytes)+len(chainID)
	
        // allocates array but does not populate with zero vals
	bz := make([]byte, 0, lenKey)
	bz = append(bz, []byte{prefix}...)
	bz = append(bz, lenTimeBytes...)
	bz = append(bz, timeBytes...)
	bz = append(bz, chainID...)
	return bz
}
```

With this, it is immediately apparent what is going on and there's no question about the order of the copied values.

Working example:
https://go.dev/play/p/m8ZkY8t6LCE

mpoke: Fine by me. I just used the existing model. If nobody objects, I can do the change.

x/ccv/provider/keeper/relay.go
```diff
@@ -300,24 +317,63 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 // EndBlockCIS contains the EndBlock logic needed for
 // the Consumer Chain Removal sub-protocol
 func (k Keeper) EndBlockCCR(ctx sdk.Context) {
```
smarshall-spitzbart, line 319: Is this method tested? We should prob create an issue if not, since there's a good amount of logic here now

smarshall-spitzbart (approved):
Added some comments to address but overall looks great!

x/ccv/provider/keeper/keeper.go
```diff
@@ -825,3 +824,50 @@ func (k Keeper) IterateInitTimeoutTimestamp(ctx sdk.Context, cb func(chainID str
...
+
+	// Convert into bytes for storage
+	bz := make([]byte, 8)
+	binary.BigEndian.PutUint64(bz, vscID)
+
+	store.Set(types.VscTimeoutTimestampKey(chainID, timestamp), bz)
+}
+
+// DeleteVscTimeoutTimestamp removes from the store a specific vsc timeout timestamp for the given chainID.
+func (k Keeper) DeleteVscTimeoutTimestamp(ctx sdk.Context, chainID string, timestamp time.Time) {
```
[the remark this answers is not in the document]
smarshall-spitzbart: Nvm, the variadic function just moves the for-loop, either way works

tests/e2e/unbonding.go
```diff
@@ -9,102 +9,132 @@ import (
...
+				s.Require().Equal(expBalance, balance, "unexpected balance after consumer unbonding")
 
-	// relay 1 VSCMatured packet from consumer to provider
-	relayAllCommittedPackets(s, s.consumerChain, s.path, ccv.ConsumerPortID, s.path.EndpointA.ChannelID, 1)
+				// increment time so that the unbonding period ends on the provider
+				incrementTimeByUnbondingPeriod(s, Provider)
+			},
+		},
+		{
+			"no valset changes", 1, func(expBalance, balance sdk.Int) {
```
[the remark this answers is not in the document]
mpoke: It's testing a specific scenario that was fixed here https://github.com/cosmos/interchain-security/pull/101

x/ccv/provider/keeper/relay.go
mpoke: It's tested in the e2e tests. There is no UT though.

smarshall-spitzbart: Cool that should be fine 👍

x/ccv/provider/keeper/relay.go
```diff
@@ -66,6 +66,17 @@ func (k Keeper) OnRecvVSCMaturedPacket(
 	// clean up index
 	k.DeleteUnbondingOpIndex(ctx, chainID, data.ValsetUpdateId)
 
+	// remove the VSC timeout timestamp for this chainID and vscID
+	var timestamp time.Time
+	k.IterateVscTimeoutTimestamps(ctx, chainID, func(ts time.Time, vscID uint64) bool {
```
jtremback, line 71: @mpoke This `(chainID, timestamp) -> vscID` indexing scheme is unintuitive. I think that it could be a lot easier to understand with a different way of indexing, but I'm not sure what that would be right now.

x/ccv/provider/keeper/relay.go
```diff
@@ -66,6 +66,17 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
+		if vscID == data.ValsetUpdateId {
+			timestamp = ts
+			return false
+		}
+		return true
```
jtremback, line 76: Shouldn't this never happen if the channel is ordered? Shouldn't we panic here?

x/ccv/provider/keeper/relay.go
mpoke: It's the most efficient way I could think of to perform the following operations:
- add a VSC timeout timestamp for a VSCPacket with vscID
- timeout the VSCPackets, which requires iterating over timestamps
- remove a VSC timeout timestamp when receiving a VSCMaturedPacket with vscID
- remove all VSC timeout timestamp associated with a chainID when that consumer chain is removed

jtremback:
I feel very confused about the indexing scheme you have used. I think it is probably correct but it is weird. I want to hold off approving this until we've talked it through and possibly implemented something more understandable. 

We will put this into the v0.2 release though before then.

x/ccv/provider/keeper/relay.go
mpoke: Changed indexing from `(chainID, timestamp) -> vscID` to `(chainID, vscID) -> timestamp`.

mpoke:
After discussing with @jtremback, we decided to change the indexing key for VSC timeouts from `(chainID, timestamp) -> vscID` to `(chainID, vscID) -> timestamp`. 

Also, to make it easier to iterate through expired timestamps, we store the timestamp of sending a `VSCPacket` instead of the timestamp of timing out (this was previously suggested by @danwt). As a result, for a given `chainID` and any two entries `(chainID, vscID1) -> ts1` and `(chainID, vscID2) -> ts2`, if `vscID1 < vscID2` then `ts1 <= ts2`. The tread-off is that the timeout timestamp needs to be recomputed in every endblock.

x/ccv/provider/keeper/keeper.go
```diff
@@ -359,18 +360,15 @@ func (k Keeper) SetUnbondingOpIndex(ctx sdk.Context, chainID string, valsetUpdat
 // IterateOverUnbondingOpIndex iterates over the unbonding indexes for a given chain id.
 func (k Keeper) IterateOverUnbondingOpIndex(ctx sdk.Context, chainID string, cb func(vscID uint64, ubdIndex []uint64) bool) {
 	store := ctx.KVStore(k.storeKey)
-	iterationPrefix := append([]byte{types.UnbondingOpIndexBytePrefix}, types.HashString(chainID)...)
-	iterator := sdk.KVStorePrefixIterator(store, iterationPrefix)
-
+	iterator := sdk.KVStorePrefixIterator(store, types.ChainIdWithLenKey(types.UnbondingOpIndexBytePrefix, chainID))
```
jtremback, line 363: What's going on in this function? Did you just clean things up and start using a utility function?

x/ccv/provider/keeper/relay.go
```diff
@@ -300,24 +309,65 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 // EndBlockCIS contains the EndBlock logic needed for
 // the Consumer Chain Removal sub-protocol
 func (k Keeper) EndBlockCCR(ctx sdk.Context) {
-	currentTime := uint64(ctx.BlockTime().UnixNano())
+	currentTime := ctx.BlockTime()
+	currentTimeUint64 := uint64(currentTime.UnixNano())
```
jtremback, line 313: Why do you need the Uint64 version here? Can't you compare Go time stuff?

x/ccv/provider/keeper/relay.go
```diff
@@ -300,24 +309,65 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
...
 
 	// iterate over initTimeoutTimestamps
-	var removedChainIds []string
+	var chainIdsToRemove []string
 	k.IterateInitTimeoutTimestamp(ctx, func(chainID string, ts uint64) bool {
```
jtremback, line 317: Why is the time coming in as a uint64?

jtremback (approved):
LGTM

x/ccv/provider/keeper/keeper.go
mpoke: Yes

x/ccv/provider/keeper/relay.go
mpoke: The init timeout timestamp is saved as an uint64

x/ccv/provider/keeper/relay.go
mpoke: see https://github.com/cosmos/interchain-security/pull/422#discussion_r1014468716

Issue #433, opened 2022-11-04
Fix early returning iterator


danwt:
# Problem

https://github.com/cosmos/interchain-security/blob/c8a762c7b4c19d9b127abc9529395d2b3cff269d/x/ccv/provider/keeper/relay.go#L356-L357

I think it should return false, rather than true.

https://github.com/cosmos/interchain-security/blob/c8a762c7b4c19d9b127abc9529395d2b3cff269d/x/ccv/provider/keeper/keeper.go#L200-L202

# Closing criteria

Fix the iteration if necessary, _and add a regression test_

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

mpoke:
Nice catch @danwt. IMO the fix entails changing the if condition to 
```golang
if !cb(ctx, channelID, chainID)
```
in order to be consistent with the other iterators.

mpoke:
Could be worthwhile to also create a UT for `EndBlockCCR` to catch these types of errors. The e2e tests have only one consumer chain.

PR #434, opened 2022-11-06
Fix ChannelToChain iterator

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/433

PR #440, opened 2022-11-07
Use `doContinue` callback across all iterators

danwt:
All iterators now have a doContinue callback which should make things less error prone.

Avoids bugs like

- https://github.com/cosmos/interchain-security/issues/433

I noticed that the sdk, ibc ect use a 

```
if cb(..) {break}
``` 

pattern.

Where we do

```
if !cb(..) {break}
``` 

For 18/20 iterators, so I opted to bring the last 2/20 to our pattern to make the smaller PR.

But a part of me thinks we should go the other way.

I think I fixed 3 or so bugs here.

### DIFF
--- a/tests/difference/core/driver/core_test.go
+++ b/tests/difference/core/driver/core_test.go
@@ -388,7 +388,7 @@ func (s *CoreSuite) TestAssumptions() {
 	s.consumerKeeper().IteratePacketMaturityTime(s.ctx(C),
 		func(vscId uint64, timeNs uint64) bool {
 			s.T().Fatal(FAIL_MSG)
-			return false // Don't stop
+			return true // Don't stop
 		})
 
 	// Consumer power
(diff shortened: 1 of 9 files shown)


danwt:
> Maybe we should consider going the with the way sdk uses it 
Hmm yeah I agree to be honest but it's more work!

> I can help you if you want
That would be amazing, thank you!

smarshall-spitzbart:
Imo adding a new pattern `doContinue` just makes things more confusing. Probably best to use the sdk pattern everywhere, avoiding all inconsistencies

danwt:
I agree @smarshall-spitzbart .

My preference would be to merge this pr immediately as it brings fixes, and make a second PR to flip all the `doContinue` to `cb` like in the sdk. The 2nd PR is a lengthier piece of work and very error prone as it's clear we cannot rely on our current test suite. That's why I suggest we hotfix first.

smarshall-spitzbart:
Its unclear to me why the `doContinue` pattern is necessary to fix a bug. Is it possible to fix the bug in a smaller diff without any refactors, and then we can refactor the iterators in a separate PR?

Also, as @mpoke mentioned in [this PR](https://github.com/cosmos/interchain-security/pull/434), we should add regression tests to the fixes for #433 

Sorry, not trying to be nitpicky. I just think this close to release we should 
1. Minimize PR diffs where possible
2. Be granular about PR intention (ie. should code behavior change or not)
3. Add or update tests if code behavior should change

The above significantly reduces review effort and allows us to merge fast

danwt:
Hmm ok fair enough, you have me convinced hehe.

Issue #436, opened 2022-11-07
ClientConfig in e2e tests are reversed 


mpoke:
# Problem

IIUC the [UpdateClient](https://github.com/cosmos/ibc-go/blob/b1f494c643280c73f3696461abbb11a39303b190/testing/endpoint.go#L130) method, the [ClientConfig](https://github.com/cosmos/ibc-go/blob/b1f494c643280c73f3696461abbb11a39303b190/testing/endpoint.go#L30) field (of the `Endpoint` struct in the ibc-go testing framework) contains configuration for the client residing on this endpoint, hence a client to the counterparty chain. In the ICS e2e tests, `EndpointA` is the consumer and `EndpointB` is the provider (see code [here](https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/tests/e2e/setup.go#L121)). This means that the `EndpointA.ClientConfig.UnbondingPeriod`  should be set to `providerUnbondingPeriod`, however, we are doing it the other way around, see https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/tests/e2e/setup.go#L126

# Closing criteria

The `UnbondingPeriod` and `TrustingPeriod` in `CCVTestSuite. SetupTest` are correctly set. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

mpoke:
e2e that needs to pass:
```golang
func (s CCVTestSuite) TestEndpointsClientConfig() {
	consumerKeeper := s.consumerApp.GetConsumerKeeper()
	providerStakingKeeper := s.providerApp.GetStakingKeeper()

	consumerUnbondingPeriod := consumerKeeper.GetUnbondingPeriod(s.consumerCtx())
	cs, ok := s.providerApp.GetIBCKeeper().ClientKeeper.GetClientState(s.providerCtx(), s.path.EndpointB.ClientID)
	s.Require().True(ok)
	s.Require().Equal(
		consumerUnbondingPeriod,
		cs.(*ibctmtypes.ClientState).UnbondingPeriod,
		"unexpected unbonding period in consumer client state",
	)
	s.Require().Equal(
		consumerUnbondingPeriod,
		s.path.EndpointB.ClientConfig.(*ibctesting.TendermintConfig).UnbondingPeriod,
		"unexpected unbonding period in provider's endpoint ClientConfig",
	)

[...]

PR #442, opened 2022-11-07
E2e setup cleanup

smarshall-spitzbart:
Closes https://github.com/cosmos/interchain-security/issues/436.

This PR removes the code that is confusing and effectively dead. IBC client states are actually setup [here](https://github.com/cosmos/interchain-security/blob/e2e-setup-cleanup/x/ccv/provider/keeper/proposal.go#L61) and [here](https://github.com/cosmos/interchain-security/blob/e2e-setup-cleanup/x/ccv/provider/keeper/proposal.go#L212), using correct convention of the client state's unbonding period being for the counterparty chain.

Aside: The nomenclature of how "providerClientState", "consumerClientState", etc. are used in this repo is pretty inconsistent. "consumerEndpointClientState", etc. is easier for me to understand as the client state used on the consumer's IBC endpoint

### DIFF
--- a/tests/e2e/channel_init.go
+++ b/tests/e2e/channel_init.go
@@ -21,8 +21,10 @@ func (suite *CCVTestSuite) TestConsumerGenesis() {
 
 	genesis := consumerKeeper.ExportGenesis(suite.consumerChain.GetContext())
 
-	suite.Require().Equal(suite.providerClient, genesis.ProviderClientState)
-	suite.Require().Equal(suite.providerConsState, genesis.ProviderConsensusState)
+	// Confirm that client and cons state are exported from consumer keeper properly
+	consumerEndpointClientState, consumerEndpointConsState := suite.GetConsumerEndpointClientAndConsState()
+	suite.Require().Equal(consumerEndpointClientState, genesis.ProviderClientState)
+	suite.Require().Equal(consumerEndpointConsState, genesis.ProviderConsensusState)
 
 	suite.Require().NotPanics(func() {
 		consumerKeeper.InitGenesis(suite.consumerChain.GetContext(), genesis)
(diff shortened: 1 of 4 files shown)


tests/e2e/channel_init.go
```diff
@@ -85,15 +87,6 @@ func (suite *CCVTestSuite) TestConsumerGenesis() {
 	})
 }
 
-// TestProviderClientMatches tests that the provider client managed by the consumer keeper matches the client keeper's client state
```
smarshall-spitzbart, line 88: This test was removed in favor of better assertions during test setup

tests/e2e/setup.go
```diff
@@ -94,40 +92,37 @@ func (suite *CCVTestSuite) SetupTest() {
...
-	suite.path.EndpointB.ClientID = consumerClient
-	// - set consumer endpoint's clientID
-	providerClient, found := consumerKeeper.GetProviderClientID(suite.consumerChain.GetContext())
-	suite.Require().True(found, "provider client not found")
-	suite.path.EndpointA.ClientID = providerClient
-	// - client config
-
-	trustingPeriodFraction := suite.providerApp.GetProviderKeeper().GetTrustingPeriodFraction(suite.providerCtx())
-	providerUnbondingPeriod := suite.providerApp.GetStakingKeeper().UnbondingTime(suite.providerCtx())
-	suite.path.EndpointB.ClientConfig.(*ibctesting.TendermintConfig).UnbondingPeriod = providerUnbondingPeriod
```
smarshall-spitzbart, line 126: These setters were removed because we don't use any functionality from the IBC testing package that require the endpoint client configs to be set.

tests/e2e/setup.go
```diff
@@ -28,16 +26,14 @@ import (
 // Any method implemented for this struct will be ran when suite.Run() is called.
 type CCVTestSuite struct {
 	suite.Suite
-	coordinator       *ibctesting.Coordinator
-	providerChain     *ibctesting.TestChain
-	consumerChain     *ibctesting.TestChain
-	providerApp       e2e.ProviderApp
-	consumerApp       e2e.ConsumerApp
-	providerClient    *ibctmtypes.ClientState
```
smarshall-spitzbart, line 36: removed these two confusingly named parameters in favor of more explicit inline assertions

mpoke (approved):
Nice work @smarshall-spitzbart.

smarshall-spitzbart:
@danwt the method with "test" now uses "validate", nice catch 👍, see https://github.com/cosmos/interchain-security/pull/442/commits/f9ed30bff747828a4f01d3427bf2b8e2add28e6b

danwt (approved):

PR #447, opened 2022-11-08
Fix faulty iterator callsite callback return booleans

danwt:
We had some errors in the way we were using iterators.

### DIFF
--- a/x/ccv/provider/keeper/relay.go
+++ b/x/ccv/provider/keeper/relay.go
@@ -172,7 +172,7 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
 				k.AppendPendingVSC(ctx, chainID, packetData)
 			}
 		}
-		return false // do not stop the iteration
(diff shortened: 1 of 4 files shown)


sainoe (approved):
LGTM. Thanks for the fix @danwt!

x/ccv/provider/keeper/relay.go
```diff
@@ -354,7 +354,7 @@ func (k Keeper) EndBlockCCR(ctx sdk.Context) {
 			return false
 		})
 		// continue to iterate through all consumers
-		return true
+		return false
```
mpoke, line 357: ditto: `IterateChannelToChain` change `stop` to `continue`

mpoke (approved):
Nice work @danwt.

danwt:
I merged to get the fix in as changing the `stop` words to `continue` is going to touch several places and go against the small diff spirit.

New issue

- https://github.com/cosmos/interchain-security/issues/449

Issue #435, opened 2022-11-07
Client expiring results in panic on provider


mpoke:
# Problem

This issue was found during the ICS mini-testnet.

Once the client to the consumer chain expires, every attempt to send a packet to the consumer will result in a panic on the provider. This is because `channelKeeper.SendPacket` is always called from `EndBlock`, which means that when it returns error, the provider panics. 

`SendPacket` when the client is expired: https://github.com/cosmos/ibc-go/blob/b1f494c643280c73f3696461abbb11a39303b190/modules/core/04-channel/keeper/packet.go#L73

Provider panics: https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/provider/keeper/relay.go#L166 and https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/provider/keeper/relay.go#L195

**Possible fix:** handle the `clienttypes.ErrClientNotActive` error in the CCV module. Probably a good idea to do so also on the consumer side. 

# Closing criteria

- [x] The issue described above is fixed. 
- [x] The packets that couldn't be sent due to a client being expired are collected and sent if the clients are being upgraded. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

MSalopek:
Closed by #448

MSalopek:
Some issues were rased on #448. Small refactoring is needed

PR #448, opened 2022-11-08
handle provider and consumer client expiration

mpoke:
Partial fix for https://github.com/cosmos/interchain-security/issues/435. The provider collects the IBC packets that cannot be sent to a consumer due to its client being expired. Once the client is upgraded, these packets are sent on the next validator change. 

Still need to implement the same functionality on the consumer side for `VSCMaturedPackets` and `SlashPackets`. To facilitate the reviewing process, those changes will be added to a different PR.


tests/e2e/common.go
```diff
@@ -351,24 +338,70 @@ func (suite *CCVTestSuite) commitSlashPacket(ctx sdk.Context, packetData ccv.Sla
 }
 
 // incrementTimeBy increments the overall time by jumpPeriod
+// while updating to not expire the clients
 func incrementTimeBy(s *CCVTestSuite, jumpPeriod time.Duration) {
```
MSalopek, line 342: Can this be just `incrementTime` so the other function can be `incrementTimeWithoutClientUpdate`?

x/ccv/provider/keeper/keeper.go
```diff
@@ -698,10 +698,10 @@ func (k Keeper) GetPendingVSCs(ctx sdk.Context, chainID string) (packets []ccv.V
 
 // AppendPendingVSC adds the given ValidatorSetChange packet to the list
 // of pending ValidatorSetChange packets stored under chain ID
-func (k Keeper) AppendPendingVSC(ctx sdk.Context, chainID string, packet ccv.ValidatorSetChangePacketData) {
+func (k Keeper) AppendPendingVSC(ctx sdk.Context, chainID string, newPackets []ccv.ValidatorSetChangePacketData) {
```
MSalopek, line 701: Maybe rename to `AppendPendingVSCs` sice `newPackets` is a slice of VSC packets and keep it consistent with other methods (`GetPendingVSCs`, `DeletePendingVSCs`)

x/ccv/provider/keeper/proposal.go
```diff
@@ -153,7 +153,7 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, lockUbd, clos
 
 	k.DeleteInitChainHeight(ctx, chainID)
 	k.ConsumeSlashAcks(ctx, chainID)
```
MSalopek, line 155: Nit (you can ignore it because it's out of scope): Maybe rename `ConsumeSlashAcks` to `DeleteSlashAcks` to stay consistent.

x/ccv/provider/keeper/relay.go
```diff
@@ -153,7 +153,7 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
 			// check whether there is an established CCV channel to this consumer chain
 			if channelID, found := k.GetChainToChannel(ctx, chainID); found {
 				// send this validator set change packet data to the consumer chain
-				err := utils.SendIBCPacket(
+				expiredClient, err := utils.SendIBCPacket(
```
MSalopek, line 156: Maybe we can refactor the whole `sendValidatorUpdates` keeper method and use the refactored approach in every other places we use `utils.SendIBCPacket`.

`SendIBCPacket` has more validation/preparation logic than it does send logic.

Something like this could help simplify error handling and client expiration tracking (sorry for the code dump):

```golang
// x/ccv/utils/utils.go
// expireClient is a good idea, but it can be handled without touching the actual send method and the logic can be simplified.
// 
// Main idea is to split SendIBCPacket into 2 parts -> "prep work" that checks channels, capabilities and sequences & the actual send
// - any error in "prep" should result in panic (e.g missing capability should panic, missing sequence should panic)
// - the send step no longer needs to be its own function, it can be done with channelKeeper.SendPacket(ctx, capability, packetData)

// SendIBCPacket sends an IBC packet with packetData
// over the source channelID and portID
func SendIBCPacket(
	ctx sdk.Context,
	channelKeeper ccv.ChannelKeeper,
	channelCap *types.Capability
	packetData *channeltypes.Packet
) error {
	// send packet over IBC channel
	return channelKeeper.SendPacket(ctx, channelCap, packetData)
}

// PrepareIBCPacket checks that all conditions for sending a packet are fullfilled and returns a prepared packet.
func PrepareIBCPacket(
	ctx sdk.Context,
	scopedKeeper ccv.ScopedKeeper,
	channelKeeper ccv.ChannelKeeper,
	channelID string,
	portID string,
	packetData []byte,
	timeoutPeriod time.Duration) (*channeltypes.Packet, error) {
	channel, ok := channelKeeper.GetChannel(ctx, portID, channelID)
	if !ok {
		return nil, sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "channel not found for channel ID: %s", channelID)
	}
	channelCap, ok := scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(portID, channelID))
	if !ok {
		return nil, sdkerrors.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
	}

	// get the next sequence
	sequence, found := channelKeeper.GetNextSequenceSend(ctx, portID, channelID)
	if !found {
		return nil, sdkerrors.Wrapf(
			channeltypes.ErrSequenceSendNotFound,
			"source port: %s, source channel: %s", portID, channelID,
		)
	}
	packet := channeltypes.NewPacket(
		packetData, sequence,
		portID, channelID,
		channel.Counterparty.PortId, channel.Counterparty.ChannelId,
		clienttypes.Height{}, uint64(ctx.BlockTime().Add(timeoutPeriod).UnixNano()),
	)

	return &packet, nil
}

// x/ccv/provider/keeper/relay.go

// SendValidatorUpdates sends latest validator updates to every registered consumer chain
func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
	k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID, clientID string) (stop bool) {
		// check whether there is an established CCV channel to this consumer chain
		channelID, found := k.GetChainToChannel(ctx, chainID)
		if found {
			// Send pending VSC packets to consumer chain
			k.SendPendingVSCPackets(ctx, chainID, channelID)
		}


		valUpdateID := k.GetValidatorSetUpdateId(ctx) 	        // get current ValidatorSetUpdateId
		valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)  // get the validator updates from the staking module
		// check whether there are changes in the validator set;
		// note that this also entails unbonding operations
		// w/o changes in the voting power of the validators in the validator set
		unbondingOps, _ := k.GetUnbondingOpsFromIndex(ctx, chainID, valUpdateID)
		if len(valUpdates) > 0 || len(unbondingOps) > 0 {
			packetData := ccv.NewValidatorSetChangePacketData(valUpdates, valUpdateID, k.ConsumeSlashAcks(ctx, chainID))

			if !found {
				// store the packet data to be sent once the CCV channel is established
				k.AppendPendingVSC(ctx, chainID, []ccv.ValidatorSetChangePacketData{packetData})
			}
                         
                         // we only care about the packet if there's no error
			packet, err := PrepareIBCPacket(
				ctx,
				k.scopedKeeper,
				k.channelKeeper,
				channelID,          // source channel id
				ccv.ProviderPortID, // source port id
				packetData.GetBytes(),
				k.GetParams(ctx).CcvTimeoutPeriod,
			)

			// REVIEWER NOTE:
			// the returned errs are
			// - channeltypes.ErrChannelNotFound
			// - channeltypes.ErrChannelCapabilityNotFound,
			// - channeltypes.ErrSequenceSendNotFound
			// I'm not sure all of them should panic
			if err != nil {
				// something went wrong when sending the packet
				panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
			}

			// this can also be utils.SentIBCPacket(...)
			err = channelKeeper.SendPacket(ctx, channelCap, packetData)
			if err != nil && clienttypes.ErrClientNotActive.Is(err) {
				// IBC client expired:
				// store the packet data to be sent once the client is upgraded
				k.AppendPendingVSC(ctx, chainID, []ccv.ValidatorSetChangePacketData{packetData})
			} else if err != nil {
				// something went wrong when sending the packet
				panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
			} else {
				// successful send:
				// set the VSC send timestamp for this packet
				k.SetVscSendTimestamp(ctx, chainID, packetData.ValsetUpdateId, ctx.BlockTime())
			}	
		}
		return true // do not stop the iteration
	})
	k.IncrementValidatorSetUpdateId(ctx)
}

```

x/ccv/provider/keeper/relay.go
```diff
@@ -179,10 +186,13 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
...
-	pendingPackets := k.ConsumePendingVSCs(ctx, chainID)
-	for _, data := range pendingPackets {
+	pendingPackets, found := k.GetPendingVSCs(ctx, chainID)
+	if !found {
+		return
+	}
+	for i, data := range pendingPackets {
 		// send packet over IBC
-		err := utils.SendIBCPacket(
+		expiredClient, err := utils.SendIBCPacket(
```
MSalopek, line 195: See other comment about handling IBC packet sends

x/ccv/provider/keeper/relay.go
```diff
@@ -191,14 +201,26 @@ func (k Keeper) SendPendingVSCPackets(ctx sdk.Context, chainID, channelID string
 			data.GetBytes(),
 			k.GetParams(ctx).CcvTimeoutPeriod,
 		)
+		if expiredClient {
```
MSalopek, line 204: See other comment about handling IBC packet sends

MSalopek (commented):
I left some comments on how to make this code a bit more manageable. Let me know if you think that's a good approach.

x/ccv/provider/keeper/proposal.go
mpoke: Let's not add more changes than necessary. We can have these type of changes as a refactoring PR.

x/ccv/utils/utils.go
```diff
@@ -41,6 +42,50 @@ func GetChangePubKeyAddress(change abci.ValidatorUpdate) (addr []byte) {
...
+		portID, channelID,
+		channel.Counterparty.PortId, channel.Counterparty.ChannelId,
+		clienttypes.Height{}, uint64(ctx.BlockTime().Add(timeoutPeriod).UnixNano()),
+	)
+	return &packet, channelCap, nil
+}
+
 // SendIBCPacket sends an IBC packet with packetData
 // over the source channelID and portID
 func SendIBCPacket(
```
mpoke, line 91: Note that this function is removed in https://github.com/cosmos/interchain-security/pull/452

x/ccv/provider/keeper/relay.go
```diff
@@ -150,39 +153,58 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
...
+			}
+
+			// send packet over IBC channel
+			err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
+			if clienttypes.ErrClientNotActive.Is(err) {
+				// IBC client expired:
+				// store the packet data to be sent once the client is upgraded
+				k.AppendPendingVSCs(ctx, chainID, []ccv.ValidatorSetChangePacketData{packetData})
+			} else if err == nil {
+				// successful send:
```
MSalopek, line 185: Nit: missed this on first pass

```golang
if err == nil {
        // set the VSC send timestamp for this packet
        k.SetVscSendTimestamp(ctx, chainID, packetData.ValsetUpdateId, ctx.BlockTime())
} else if err != nil && clienttypes.ErrClientNotActive.Is(err) {
        // IBC client expired:
        // store the packet data to be sent once the client is upgraded
        k.AppendPendingVSCs(ctx, chainID, []ccv.ValidatorSetChangePacketData{packetData})
} else {
        // something went wrong when sending the packet
        panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
}
```

x/ccv/provider/keeper/relay.go
```diff
@@ -192,13 +214,32 @@ func (k Keeper) SendPendingVSCPackets(ctx sdk.Context, chainID, channelID string
 			k.GetParams(ctx).CcvTimeoutPeriod,
 		)
 		if err != nil {
+			// something went wrong when preparing the packet
+			panic(fmt.Errorf("packet could not be prepared for IBC send: %w", err))
+		}
+
+		// send packet over IBC channel
+		err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
```
MSalopek, line 222: Nit again - feel free to resolve.
```golang
err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
if err != nil && clienttypes.ErrClientNotActive.Is(err) {
	// IBC client expired during iteration
	// this should never happen
	if i != 0 {
		panic(fmt.Errorf("client expired while sending pending packets: %w", err))
	}
	// leave the packet data stored to be sent once the client is upgraded
	return
} else if err != nil {
	// something went wrong when sending the packet
	panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
}
```

MSalopek (approved):
LGTM.
Approved with some nitpicky comments you can safely ignore.

I'd wait for shawn to confirm e2e test, he has more context than me.

smarshall-spitzbart (approved):
LGTM, just small comments

app/consumer-democracy/app.go
```diff
@@ -785,6 +785,11 @@ func (app *App) GetE2eEvidenceKeeper() e2e.E2eEvidenceKeeper {
 	return app.EvidenceKeeper
 }
 
+// GetUpgradeKeeper implements the ConsumerApp interface.
+func (app *App) GetUpgradeKeeper() upgradekeeper.Keeper {
```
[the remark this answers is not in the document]
mpoke: actually, GetUpgradeKeeper is no longer needed

x/ccv/provider/keeper/keeper.go
sainoe: It's also a good place to use a  variadic function:
```
// AppendPendingVSC takes an arbitrary number of `ValidatorSetChangePacketData`
func (k Keeper) AppendPendingVSC(ctx sdk.Context, chainID string, newPackets ...ccv.ValidatorSetChangePacketData) {}

k.AppendPendingVSCs(ctx, chainID, cs.PendingValsetChanges...)

k.AppendPendingVSCs(ctx, chainID, packetData)
```

sainoe (approved):
LGTM. Great work @mpoke

x/ccv/provider/keeper/relay.go
```diff
@@ -150,55 +153,93 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
...
-				if err != nil {
-					panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
-				}
+			if !found {
+				// CCV channel not established:
+				// store the packet data to be sent once the CCV channel is established
+				k.AppendPendingVSCs(ctx, chainID, packetData)
+				return true // go to next consumer chain
+			}
+
```
danwt, line 162: I think an easier to understand pattern for SendValidatorUpdates would be something similar to

```
func GetPendingVSCPackets(..) {
	packets := []Packet{}
	// read from store
	for bytes in decoded store data { // if nothing is in the store then nothing is iterated
		packets = append(packets, Unmarshal(bytes))
	}
	return packets
}

func ConsumePendingVSCPackets(..) {
	packets := GetPendingVSCPackets()
	DeletePendingVscPackets()
	return packets
}

func SendValidatorUpdates(..) {
        ...
	packets := ConsumePendingVSCPackets()
	if len(valUpdates) != 0 || len(unbondingOps) != 0 {
		//...
		packets = append(packets, newPackets)
	}
	TrySendPackets(packets)
}

```

Dittoing my other comment I think much of the packet code could be simpler and have less cases, but this would require refactoring of existing code, perhaps outside of the scope of the PR.

danwt (commented):
Happy to merge this if someone were to make an immediate follow up PR to refactor, otherwise I think it could do with some refactoring beforehand.

jtremback (changes requested):
I'm requesting the same refactor here as for the other PR https://github.com/cosmos/interchain-security/pull/452

x/ccv/utils/utils.go
```diff
@@ -75,5 +121,11 @@ func SendIBCPacket(
...
 		clienttypes.Height{}, uint64(ctx.BlockTime().Add(timeoutPeriod).UnixNano()),
 	)
-	return channelKeeper.SendPacket(ctx, channelCap, packet)
+	// send packet over IBC channel
+	err = channelKeeper.SendPacket(ctx, channelCap, packet)
+	if clienttypes.ErrClientNotActive.Is(err) {
+		// IBC client expired
+		expiredClient = true
+	}
+	return expiredClient, err
```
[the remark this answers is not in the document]
mpoke: The entire function will be removed. See https://github.com/cosmos/interchain-security/pull/448/files/82770fb648df22640cda7bab7d735b55b5357fc1#r1017904501

x/ccv/provider/keeper/relay.go
```diff
@@ -150,55 +153,93 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
...
+		// send packet over IBC channel
+		err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
+		if err != nil {
+			if clienttypes.ErrClientNotActive.Is(err) {
+				// IBC client expired:
+				// store the packet data to be sent once the client is upgraded
+				if i != 0 {
+					// this should never happen
+					panic(fmt.Errorf("client expired while sending pending packets: %w", err))
+				}
```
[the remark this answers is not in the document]
mpoke: I agree. We can get rid of this.

mpoke:
> I'm requesting the same refactor here as for the other PR #452

@jtremback See https://github.com/cosmos/interchain-security/pull/452#issuecomment-1311447354

MSalopek:
Applying above comments and making refactors where convenient.

Larger refactor (returning `found` vs using the slice) will be done in a separete PR.

x/ccv/provider/keeper/genesis.go
```diff
@@ -112,12 +110,9 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 				)
 				return true
 			})
-		} else {
```
smarshall-spitzbart, line 115: Why were pending valset changes only stored  if the channelID was not found, previously?

x/ccv/provider/keeper/keeper.go
```diff
@@ -683,11 +683,13 @@ func (k Keeper) DeleteInitChainHeight(ctx sdk.Context, chainID string) {
 }
 
 // GetPendingVSCs returns the list of pending ValidatorSetChange packets stored under chain ID
-func (k Keeper) GetPendingVSCs(ctx sdk.Context, chainID string) (packets []ccv.ValidatorSetChangePacketData, found bool) {
+func (k Keeper) GetPendingVSCs(ctx sdk.Context, chainID string) []ccv.ValidatorSetChangePacketData {
+	var packets []ccv.ValidatorSetChangePacketData
```
smarshall-spitzbart, line 687: This is equiv to `packets := []ccv.ValidatorSetChangePacketData{}` right? Just confirming the slice gets initialized to empty with var syntax

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
-func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
-	// get current ValidatorSetUpdateId
-	valUpdateID := k.GetValidatorSetUpdateId(ctx)
-	// get the validator updates from the staking module
-	valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)
+// sendValidatorUpdates iterates over chains and sends VSCs to
+// consumer chains with established CCV channels
+// if ccv channel is not established for consumer chain
+// the updates will remain queued until the channel is established
+func (k Keeper) SendValidatorUpdates(ctx sdk.Context) {
```
smarshall-spitzbart, line 143: Easy for me to understand 👍

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
 		)
 		if err != nil {
+			// something went wrong when preparing the packet
+			panic(fmt.Errorf("packet could not be prepared for IBC send: %w", err))
+		}
+
+		err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
+
+		if err != nil && clienttypes.ErrClientNotActive.Is(err) {
+			// IBC client is expired!
```
smarshall-spitzbart, line 175: just to confirm, this sending of the packet will be retried next block? ie. nothing with SendValidatorUpdates changes

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
 		}
 		// set the VSC send timestamp for this packet;
 		// note that the VSC send timestamp are set when the packets
 		// are actually sent over IBC
 		k.SetVscSendTimestamp(ctx, chainID, data.ValsetUpdateId, ctx.BlockTime())
 	}
+	k.DeletePendingVSCs(ctx, chainID)
+}
+
+// queueVSCPackets queues latest validator updates for every registered consumer chain
```
smarshall-spitzbart, line 191:
```suggestion
// QueueValidatorUpdates queues latest validator updates for every registered consumer chain
```

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+func (k Keeper) QueueValidatorUpdates(ctx sdk.Context) {
+	valUpdateID := k.GetValidatorSetUpdateId(ctx) // curent valset update ID
+	valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)
+
+	k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID, clientID string) (stop bool) {
+		// check whether there are changes in the validator set;
```
smarshall-spitzbart, line 197: This comment should be above `	valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)`

x/ccv/provider/keeper/relay_test.go
```diff
@@ -0,0 +1,83 @@
...
+			},
+			expectNextValsetUpdateId: 1,
+			expectedQueueSize:        1,
+		},
+	}
+
+	chainID := "consumer"
+
+	for _, tc := range testCases {
+		keeperParams := testkeeper.NewInMemKeeperParams(t)
```
smarshall-spitzbart, line 53: This area can be simplified unless you're doing custom param setup
```suggestion
		pk, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
```

x/ccv/utils/utils.go
```diff
@@ -51,20 +98,21 @@ func SendIBCPacket(
 	portID string,
 	packetData []byte,
 	timeoutPeriod time.Duration,
-) error {
+) (expiredClient bool, err error) {
+	expiredClient = false
 	channel, ok := channelKeeper.GetChannel(ctx, portID, channelID)
 	if !ok {
-		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "channel not found for channel ID: %s", channelID)
+		return expiredClient, sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "channel not found for channel ID: %s", channelID)
```
smarshall-spitzbart, line 105: might as well return true/false inline. This makes it seem that the client is expired as a quick reader. But really `expiredClient` is false

x/ccv/utils/utils.go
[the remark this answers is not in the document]
smarshall-spitzbart: still a little confused about this stuff... Exact same check is done in relay.go

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+		}
+
+		err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
+
+		if err != nil && clienttypes.ErrClientNotActive.Is(err) {
+			// IBC client is expired!
+			// leave the packet data stored to be sent once the client is upgraded
+			// the client cannot expire during iteration (in the middle of a block)
+			return
+		} else if err != nil {
```
smarshall-spitzbart, line 179: lets remove the else if and make this a nested if instead

x/ccv/utils/utils.go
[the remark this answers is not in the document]
MSalopek: The function is marked as deprecated. It is removed in #452 and all callsites are refactored.

x/ccv/provider/keeper/genesis.go
MSalopek: Because we were sending them immediately (in every EndBlock).

We would only queue valset changes if the client was not active. Now we always queue regardless of client

x/ccv/provider/keeper/keeper.go
MSalopek: Behaviour is equivalent.

Technically it's a nil slice but all operations on it are permitted and will not panic unless you try to access an index.

It is safe to iterate over and append to it

x/ccv/provider/keeper/relay.go
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Done!

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+	k.SendValidatorUpdates(ctx)
 }
 
-// SendValidatorUpdates sends latest validator updates to every registered consumer chain
-func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
-	// get current ValidatorSetUpdateId
-	valUpdateID := k.GetValidatorSetUpdateId(ctx)
-	// get the validator updates from the staking module
-	valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)
+// sendValidatorUpdates iterates over chains and sends VSCs to
```
sainoe, line 139:
```suggestion
// SendValidatorUpdates iterates over chains and sends VSCs to
```

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+// consumer chains with established CCV channels
+// if ccv channel is not established for consumer chain
+// the updates will remain queued until the channel is established
+func (k Keeper) SendValidatorUpdates(ctx sdk.Context) {
 	k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID, clientID string) (stop bool) {
-		// check whether there is an established CCV channel to this consumer chain
+		// check if CCV channel is establish and send
```
sainoe, line 145:
```suggestion
		// check if CCV channel is establishes and send
```

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,86 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
 	})
-	k.IncrementValidatorSetUpdateId(ctx)
 }
 
-// Sends all pending ValidatorSetChangePackets to the specified chain
-func (k Keeper) SendPendingVSCPackets(ctx sdk.Context, chainID, channelID string) {
-	pendingPackets := k.ConsumePendingVSCs(ctx, chainID)
+// SendVSCPacketsToChain sends all queued ValidatorSetChangePackets to the specified chain
+func (k Keeper) SendVSCPacketsToChain(ctx sdk.Context, chainID, channelID string) {
+	pendingPackets := k.GetPendingVSCs(ctx, chainID)
```
sainoe, line 155: Can we rename `pendingPackets` to `pendingVSCs

sainoe:
Thanks for the refactoring @MSalopek !!

MSalopek:
> Thanks for the refactoring @MSalopek !! Just few nitpick comments

Comments addressed, all suggested changes applied.

x/ccv/consumer/keeper/validators.go
```diff
@@ -78,13 +78,14 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
...
-// Slash sends a slashing request to the provider chain
+// Slash queues a slashing request for the the provider chain
+// All queued slashing requests will be cleared in EndBlock
 func (k Keeper) Slash(ctx sdk.Context, addr sdk.ConsAddress, infractionHeight, power int64, _ sdk.Dec, infraction stakingtypes.InfractionType) {
 	if infraction == stakingtypes.InfractionEmpty {
 		return
 	}
 
-	k.SendSlashPacket(
+	k.QueueSlashPacket(
```
MSalopek, line 88: Instead of sending slash packets in BeginBlock, we only queue them. All packets are sent in EndBlock.

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,87 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+	valUpdates := k.stakingKeeper.GetValidatorUpdates(ctx)
+
+	k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID, clientID string) (stop bool) {
+		// note that this also entails unbonding operations
+		// w/o changes in the voting power of the validators in the validator set
+		unbondingOps, _ := k.GetUnbondingOpsFromIndex(ctx, chainID, valUpdateID)
+		if len(valUpdates) != 0 || len(unbondingOps) != 0 {
+			// construct validator set change packet data
+			packet := ccv.NewValidatorSetChangePacketData(valUpdates, valUpdateID, k.ConsumeSlashAcks(ctx, chainID))
+			k.AppendPendingVSCs(ctx, chainID, packet)
```
jtremback, line 205: Might want to think about changing the name to `AppendPendingPackets` to match the consumer logic and reinforce the generality of this mechanism.

x/ccv/utils/utils.go
```diff
@@ -41,30 +42,37 @@ func GetChangePubKeyAddress(change abci.ValidatorUpdate) (addr []byte) {
 	return pk.Address()
 }
 
-// SendIBCPacket sends an IBC packet with packetData
-// over the source channelID and portID
-func SendIBCPacket(
+// PrepareIBCPacketSend prepares the send an IBC packet with packetData
+// over the source channelID and portID. If successful, it returns the
+// packet and the channel capability.
+func PrepareIBCPacketSend(
```
jtremback, line 48: Let's put combine and prepare back into one step

jtremback (changes requested):
LGTM except:

- Put preparing and sending IBC packets back into one step
- Use same name for pending packet queue operations on provider and consumer
- Need answer about why pending packets are not being saved in consumer genesis

jtremback:
LGTM except:

- Put preparing and sending IBC packets back into one step
- Use same name for pending packet queue operations on provider and consumer
- Need answer about why pending packets are not being saved in consumer genesis

MSalopek:
> LGTM except:
> 
> * Put preparing and sending IBC packets back into one step
> * Use same name for pending packet queue operations on provider and consumer
> * Need answer about why pending packets are not being saved in consumer genesis

3. consumer genesis refactoring is handled in #382. All genesis handling was removed by Marius.
I would opt to not touch it in this PR and handle it in #382.

x/ccv/provider/keeper/relay.go
```diff
@@ -126,80 +127,87 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+		}
+		return true // do not stop the iteration
```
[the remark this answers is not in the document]
MSalopek: Yes, but is uses a counter intuitive pattern and does not in fact stop iteration when `true` is returned.

```golang
func (k Keeper) IterateConsumerChains(ctx sdk.Context, cb func(ctx sdk.Context, chainID, clientID string) (stop bool)) {
	iterator := sdk.KVStorePrefixIterator(store, []byte{types.ChainToClientBytePrefix})
	defer iterator.Close()

		// remove 1 byte prefix from key to retrieve chainID
		chainID := string(iterator.Key()[1:])
		clientID := string(iterator.Value())

		if !cb(ctx, chainID, clientID) {
			return
		}
	}
}
```

Will be refactored in #459

x/ccv/provider/keeper/relay.go
MSalopek: Done.

x/ccv/utils/utils.go
MSalopek: Done.

jtremback (approved):

jtremback:
@MSalopek Looks good to me with your changes. If you've done everything you want to, can you hit merge on this and start bringing in the iterator refactor?

x/ccv/provider/keeper/relay.go
```diff
@@ -190,16 +162,45 @@ func (k Keeper) SendPendingVSCPackets(ctx sdk.Context, chainID, channelID string
...
+		unbondingOps, _ := k.GetUnbondingOpsFromIndex(ctx, chainID, valUpdateID)
+		if len(valUpdates) != 0 || len(unbondingOps) != 0 {
+			// construct validator set change packet data
+			packet := ccv.NewValidatorSetChangePacketData(valUpdates, valUpdateID, k.ConsumeSlashAcks(ctx, chainID))
+			k.AppendPendingPackets(ctx, chainID, packet)
+		}
+		return true // do not stop the iteration
+	})
+
+	k.IncrementValidatorSetUpdateId(ctx)
```
mpoke, line 203: Take this line out from `QueueVSCPackets`. Makes more sense to be directly in `EndBlockVSU`.

x/ccv/consumer/keeper/relay.go
```diff
@@ -104,43 +94,26 @@ func (k Keeper) SendVSCMaturedPackets(ctx sdk.Context) error {
...
+			vscPacket := ccv.NewVSCMaturedPacketData(vscId)
+
+			// append VSCMatured packet to pending packets
+			// sending packets is attempted each EndBlock
+			// unsent packets remain in the queue until sent
+			k.AppendPendingPacket(ctx, types.ConsumerPacket{
+				Type: types.VscMaturedPacket,
+				Data: vscPacket.GetBytes(),
+			})
 			k.DeletePacketMaturityTime(ctx, vscId)
```
mpoke, line 107: We shouldn't delete inside the iterator.

x/ccv/consumer/keeper/relay.go
```diff
@@ -104,43 +94,26 @@ func (k Keeper) SendVSCMaturedPackets(ctx sdk.Context) error {
...
-			ctx.EventManager().EmitEvent(
```
mpoke, line 124: Why is this event deleted?

mpoke (commented):
@MSalopek thanks for the refactoring. Here are some comments.

x/ccv/provider/keeper/relay.go
MSalopek: It's convienient to keep it there since it makes unittesting the function easier.

The unittest checks that the updateId was incremented. It's a lot more involved to test the `EndBlockVSU` to see if the updateId got incremented after queueing.

It is also makes the logic easier to follow (just IMHO). You can see immediately that the counter is incremented inside `QueueVSCPackets` and don't have to check other functions called in `EndBlock`.

I'll refactor if you feel strongly about it.

mpoke: Okay. Let's leave it then.

x/ccv/consumer/keeper/relay.go
MSalopek: There's no sending going on. The packets are only queued for sending.

It would make more sense to add logging here and emit an event when the packet is sent.
Since we store all packets in the same queue, we'd need to unmarshal the data, check that it contains a maturedVSC packet and use it to build the event. The event seemed to have low value so it was dropped.

I'll reintroduce it in `SendPackets`

mpoke: Yes, that's the event that we care about. The actual packet being sent it's an IBC event anyways.

PR #451, opened 2022-11-08
Key Assignment

danwt:
Addresses https://github.com/cosmos/interchain-security/issues/26 

Adds a new feature KeyAssignment

Please see

- [API TLA+](https://github.com/cosmos/interchain-security/blob/danwt/key-assignment-v4/docs/key-assignment/api.tla)
- [Design MD](https://github.com/cosmos/interchain-security/blob/danwt/key-assignment-v4/docs/key-assignment/design.md)

This PR still needs some polish (docs) and testing (integration) but it's at a good milestone to open this. 

Please note that, in my view, the PR is tested to a high degree. There is an extensive suite in [key_assignment_test.go](https://github.com/cosmos/interchain-security/blob/5b172bb19425f9a59192381a957125a7fb5dce22/x/ccv/provider/keeper/key_assignment_test.go) and I also extensively diff tested the work in [a working branch](https://github.com/cosmos/interchain-security/compare/80b751c1d3afb5d2f8684868548faf8d3995ae3f..90bc365c2bd4730579e2eea768cf2a310ee4a0da#diff-b25e1f0e5637c373a56be66dc062507bbf96c673bbe86de716d129eb0ab08331). Additionally, there are tests for tx and query and other tests have been updated where needed. I have left the diff tests out as I feel it is not crucial, and should go in a separate PR (it should be refactored ect, and is a large amount of changes).

In short I'm happy with the milestone in this PR and from this point I think things can be added and improved later.


mpoke:
@jtremback Notes from discussion with @danwt:
* MVP: High-level overview
    * Send assignKey TX to provider
    * Result in mapping providerKey <-> consumerKey (stored on provider)
    * Mapping used for provider to consumer communication
        * InitGenesis (adding InitValset to consumer CCV genesis state)
        * VSCPackets (validator updates)
    * Consumer receives only consumerKey (i.e., doesn’t know the providerKey)
    * Validator changing key from consumer perspective:
        * Validator with previous key unbonds 
        * Validator with new key bond
        * Consequence: VSCPackets will be larger when the keys are changed
    * Slashing works the same on the consumer (i.e., send SlashPacket with consAddr)
        * Consequence: Provider CCV module needs history of consumerKeys
            * Consequence: Require pruning mechanism  
    * Pruning mechanism
        * assignKey TX (new consumerKey) <-> vscID (not necessarily the vscID of when the TX was received)
        * When receiving VSCMaturedPacket with this vscID prune old consumerKey 
    * Consequences: 
        * All logic on provider
        * Checking (through queries) Validator Set Replication property could be hard. This query is not really needed though.
* Alternative (keep mapping on both)
    * No history on provider, but needed on consumer (which means pruning)
    * Sending more data (both in genesis.state and VSCPackets)
    * **No clear benefit!**

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,602 @@
...
+
+	sdkcryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+
+	abci "github.com/tendermint/tendermint/abci/types"
+
+	ccv "github.com/cosmos/interchain-security/x/ccv/types"
+	tmprotocrypto "github.com/tendermint/tendermint/proto/tendermint/crypto"
+)
+
+// In this branch I am icing things for a few days.
```
jtremback, line 18: ?

danwt, line 18:
```suggestion
```

proto/interchain_security/ccv/provider/v1/keyassignment.proto
```diff
@@ -0,0 +1,39 @@
...
+option go_package = "github.com/cosmos/interchain-security/x/ccv/provider/types";
+
+import "gogoproto/gogo.proto";
+import "tendermint/crypto/keys.proto";
+
+message LastUpdateMemo {
+  tendermint.crypto.PublicKey consumer_key = 1;
+  tendermint.crypto.PublicKey provider_key = 2;
+  uint64 vscid = 4;
+  int64 power = 5;
```
jtremback, line 14: I forgot- why is the power on here?

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+
+func TMCryptoPublicKeyToConsAddr(k tmprotocrypto.PublicKey) sdk.ConsAddress {
+	sdkK, err := sdkcryptocodec.FromTmProtoPublicKey(k)
+	if err != nil {
+		panic("could not get public key from tm proto public key")
+	}
+	return sdk.GetConsAddress(sdkK)
+}
+
+type Store interface {
```
smarshall-spitzbart, line 32: Would it be much effort to refactor these abstractions into methods defined for the provider keeper? My understanding of the `/keeper` directory is that it should contain methods defined on the keeper struct, typically state transitions that affect the store. 

I don't necessarily see the value in introducing another layer of abstraction, but maybe I'm missing something

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+		return power, true
+	}
+	return -1, false
+}
+
+// KeyToLastUpdateMemo is an implementation of a map from a public key to a
+// memo storing the last update.
+// It is used because public keys are not comparable in Go, so they cannot be
+// used as builtin map keys.
+type KeyToLastUpdateMemo struct {
```
smarshall-spitzbart, line 194: These types all share an inner and canonical map, along with some shared methods. Could we use go generics to reduce the amount of code that is introduced?

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+	s.inner[pk] = struct{}{}
+}
+
+// Add is a typical set has/contains operation.
+func (s *KeySet) Has(pk tmprotocrypto.PublicKey) bool {
+	_, found := s.canonical[DeterministicStringify(pk)]
+	return found
+}
+
+// getProviderKeysForUpdate returns the all the provider keys of validators for which
```
smarshall-spitzbart, line 261: This is where I left off for the first pass of a review, lets go over some of this file via zoom call

smarshall-spitzbart (commented):
Great stuff! I got to the meat of the PR, but would love to review the remaining code over a call. Maybe we can add comments to pieces of code as we go along? Definitely a good amount of comments here, but it never hurts to add more from someone who's pretty unfamiliar with the feature

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -68,3 +80,16 @@ message Chain {
   string chain_id = 1;
   string client_id = 2;
 }
+
+message QueryConsumerChainValidatorKeyAssignmentRequest {
```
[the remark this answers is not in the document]
danwt: Yes, but not in this MVP!

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -39,6 +41,16 @@ service Query {
     option (google.api.http).get =
         "/interchain_security/ccv/provider/consumer_chain_stop_proposals";
   }
+
+  // QueryConsumerChainValidatorKeyAssignment queries the consensus key
+  // assigned by a validator for a consumer chain.
+  rpc QueryConsumerChainValidatorKeyAssignment(
+      QueryConsumerChainValidatorKeyAssignmentRequest)
+      returns (QueryConsumerChainValidatorKeyAssignmentResponse) {
+    // TODO: make sure path is correct.
```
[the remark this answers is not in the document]
danwt: Thanks for the reminder I'll check

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -0,0 +1,29 @@
...
+import "google/protobuf/any.proto";
+
+// Msg defines the Msg service.
+service Msg {
+  rpc AssignConsensusPublicKeyToConsumerChain(
+      MsgAssignConsensusPublicKeyToConsumerChain)
+      returns (MsgAssignConsensusPublicKeyToConsumerChainResponse);
+}
+
+message MsgAssignConsensusPublicKeyToConsumerChain {
```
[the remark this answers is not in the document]
danwt: In a sense, yes!
The assignment dimensions are `validator X chain` (each validator can assign for each chain).

I'm not convinced that your suggestion is more clear but I acknowledge that the original name isn't that good either. Maybe we can think of something else.

tests/e2e/setup.go
```diff
@@ -171,6 +172,15 @@ func (suite *CCVTestSuite) SetupTest() {
 	suite.transferPath.EndpointB.ChannelConfig.PortID = transfertypes.PortID
 	suite.transferPath.EndpointA.ChannelConfig.Version = transfertypes.Version
 	suite.transferPath.EndpointB.ChannelConfig.Version = transfertypes.Version
+
+	for _, validator := range suite.providerChain.Vals.Validators {
+		// Assign a mapping from the validator consensus key to itself to be able to lookup
```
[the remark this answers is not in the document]
danwt: Thanks you're right this block isn't needed at all.

> during the provider's InitGenesis?

It's when a consumer is added

testutil/keeper/expectations.go
```diff
@@ -26,22 +26,17 @@ import (
...
-				"ChainId", expectedChainID).Field(
-				"LatestHeight", expectedLatestHeight,
-			),
-			gomock.Any(),
-		).Return("clientID", nil).Times(1),
-	}
-
-	expectations = append(expectations, GetMocksForMakeConsumerGenesis(ctx, mocks, time.Hour)...)
-	return expectations
+	expectations := GetMocksForMakeConsumerGenesis(ctx, mocks, time.Hour)
```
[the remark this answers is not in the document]
danwt: Because the order _is_ changed in the code. MakeConsumerGenesis is now called before CreateConsumerClient.

x/ccv/provider/keeper/key_assignment.go
danwt: The approach taken is to try to encapsulate the business logic of a feature, and the (access to) the state it needs into an isolated unit which can be more easily understood, reasoned about and maintained and tested, without needing additional context

In summary the approach is, in psuedocode

```

// Usage example
k.KeyAssignment(ctx, "foochain").BlaBla()

func (k ProviderKeeper) KeyAssignment(ctx, chainID) KeyAssignment {
    // Return a KeyAssignment instance which encapsulates store access and business logic
}

func (ka KeyAssignment) BlaBla() {
    // business logic
    store.DoStuff()
    // business logic
    store.DoOtherStuff()
}
```

I like this for at least these several reasons

1. It's no longer necessary to pass ctx and chainID everywhere. This reduces the amount of code. Also it reduces the chance of error by substituting chainID for another string type or another string type for a chainID. 
2. By having a KeyAssignment class without full access to the provider keeper it is guaranteed that all of its methods will not call out somewhere else in the codebase through the provider keeper and its friends. This makes it much easier to reason about this feature in isolation.
4. Less setup is needed to test the business logic of the feature. Although I took a meandering path to get the code to where it is, and ended up using a regular KV store in tests, in principle I could have used anything implementing the Store interface which would actually allow very fast in memory read/writes and way more efficient min memory testing. I actually was doing something similar to this very early on when getting the logic right.
5. When writing any code using KeyAssignment the available methods are suggested by the language server, and the suggestions arent cluttered by tons of other methods and fields as would be if using the ProviderKeeper for example.
6. It's possible to find every single place in the code where any key assignment logic is used by simply doing a 'find all references' for the KeyAssignment method on the provider


In my opinion, our codebase generally is implementing the god class antipattern https://en.wikipedia.org/wiki/God_object and I think this is a major contributing factor to our slow development, all the bugs we are finding, and our difficutly testing. I think it would be a good thing if we were using more OOP, encapsulation, single responsiblity principle ect.

I'm always willing to adapt to the needs of the team. If the team feels strongly that things should be done a particular way, I'd like to hear a compelling argument, and I'll be happy to change it.

x/ccv/provider/keeper/key_assignment.go
danwt: Yeah I'm happy to use generics. I didn't want to bust them out yet because

1. I've not used golang generics before (although I've used them plenty in other languages)
2. We only just upgraded to 1.18, on a technicality, the code is really written in 1.17 style. We have currently have 0 generics and I think most cosmos code is written without generics AFAIU
3. I'm lazy and this didn't seem like a big deal, it's only 100 lines. Trying to KISS

I wouldn't mind using generics tbh. If you feel strongly I can do it

x/ccv/provider/keeper/key_assignment.go
danwt: Sounds good, thanks for having a look!

proto/interchain_security/ccv/provider/v1/keyassignment.proto
danwt: Added a comment. It's necessary to compute which updates to send and which updates to prune.

docs/key-assignment/design.md
```diff
@@ -0,0 +1,149 @@
...
+
+Validators can improve their security by using different consensus keys for each chain. That way, different teams in an organization can operate a subset (can be size 1) of the total number of consumer chains associated to a provider chain. If one key leaks the other keys will not be at risk. It is possible to change the keys at any time by submitting a transaction.
+
+## User Overview
+
+The KeyAssignment feature is available via a provider chain API (transactions and queries). The provider chain validator operator submits an assignment transaction to the provider chain with a consumer chain ID and desired consensus key as parameters. The over-IBC protocol used by Interchain Security takes care of forwarding the assignment to the specified consumer chain. When the consumer chain receives the key, it will immediately start using it with Tendermint.
+
+It is possible to start validating a consumer chain with the same key as used for the provider. This is the default behavior. It is also possible to specify another key to use when joining the validator set. Moreover it is possible to change the used key at any time, any multiple times, with some minor restrictions.
```
smarshall-spitzbart, line 11: > It is also possible to specify another key to use when joining the validator set.

This is a tad unclear, we're talking about consumer genesis correct?

proto/interchain_security/ccv/provider/v1/query.proto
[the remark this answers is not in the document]
smarshall-spitzbart: sounds reasonable, maybe just an issue

docs/key-assignment/design.md
```diff
@@ -0,0 +1,149 @@
+# KeyAssignment v4
```
smarshall-spitzbart, line 1: Could we add a section in here that explains the two types of pruning we discussed? Would also be useful to define semantics besides "pruning" and "deletion"

Side note could explain that unbonded validator keys are not pruned unless they reassign

lmk if this already exists in the doc, I'm skimming during a zoom :)

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+type Store interface {
+	SetProviderConsAddrToConsumerPublicKey(ProviderConsAddr, ConsumerPublicKey)
+	GetProviderConsAddrToConsumerPublicKey(ProviderConsAddr) (ConsumerPublicKey, bool)
+	DelProviderConsAddrToConsumerPublicKey(ProviderConsAddr)
+	IterateProviderConsAddrToConsumerPublicKey(func(ProviderConsAddr, ConsumerPublicKey) bool)
+	SetConsumerPublicKeyToProviderPublicKey(ConsumerPublicKey, ProviderPublicKey)
+	GetConsumerPublicKeyToProviderPublicKey(ConsumerPublicKey) (ProviderPublicKey, bool)
+	DelConsumerPublicKeyToProviderPublicKey(ConsumerPublicKey)
+	IterateConsumerConsAddrToLastUpdateMemo(func(ConsumerConsAddr, providertypes.LastUpdateMemo) bool)
+	SetConsumerConsAddrToLastUpdateMemo(ConsumerConsAddr, providertypes.LastUpdateMemo)
```
smarshall-spitzbart, line 41: imo might be useful to add VSC into here instead of just "update", or put it in comments

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+// getProviderKeysForUpdate returns the all the provider keys of validators for which
+// an update must be sent. An update must be sent to the consumer if, either:
+//  1. The consumer has the validator in its active set AND the assigned key for the
+//     validator has changed since the last update.
+//  2. The voting power of the validator has changed, and the validator is in the active
+//     set.
+func (ka *KeyAssignment) getProviderKeysForUpdate(stakingUpdates KeyToPower) KeySet {
+
+	ret := MakeKeySet()
+
```
smarshall-spitzbart, line 270: simple way to put it "find validators in which we need to update their key on a consumer through a vsc update"

mpoke (commented):
Partial review. See my comments below. 

Got stuck at `key_assignment.go`. As general feedback, I'm not sure I understand the reason for this level of abstraction. I find it very difficult to go over the code.

testutil/keeper/expectations.go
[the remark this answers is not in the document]
smarshall-spitzbart: Nice, I assume the order of those two calls doesn't matter, correct?

x/ccv/provider/keeper/key_assignment.go
smarshall-spitzbart: Interesting points and well thought out! I agree with a lot of what you've said here. 

I typically value design consistency to the same extent that I value OOP fundamentals. So it's a trade off for this particular feature. Your decisions all seem reasonable to me.

If we were to redesign cosmos sdk, there could be a generic submodule type or interface in which features like this could implement that interface. One module keeper could manage many submodule types. That way we reap the benefits of everything you've mentioned here, but still have consistency of design between submodules.

Obvi we're committed to the patterns of cosmos sdk for most of this repo, but nice thoughts for the future!

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+	tmprotocrypto "github.com/tendermint/tendermint/proto/tendermint/crypto"
+)
+
+type VSCID = uint64
+type ProviderPublicKey = tmprotocrypto.PublicKey
+type ConsumerPublicKey = tmprotocrypto.PublicKey
+type ProviderConsAddr = sdk.ConsAddress
+type ConsumerConsAddr = sdk.ConsAddress
+
+func TMCryptoPublicKeyToConsAddr(k tmprotocrypto.PublicKey) sdk.ConsAddress {
```
smarshall-spitzbart, line 24: This probably belongs in the key util file right?

x/ccv/provider/keeper/key_assignment.go
smarshall-spitzbart: Also removing "memo" as we talked about before, I'm sure that'd touch a lot of places in code tho

x/ccv/provider/keeper/key_assignment.go
danwt: @mpoke @jtremback @smarshall-spitzbart Thanks for the feedback guys on this slightly out-of-the-ordinary storage abstraction. I understand that consistency is valued a lot - I will change the code accordingly.

I do recommend that going into the next projects that we try to do things better as I don't think I'm wrong in identifying a major problem with the status quo here.

Thanks

x/ccv/provider/keeper/hooks.go
```diff
@@ -60,7 +60,15 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) {
 // Define unimplemented methods to satisfy the StakingHooks contract
 func (h Hooks) AfterValidatorCreated(ctx sdk.Context, valAddr sdk.ValAddress) {
 }
-func (h Hooks) AfterValidatorRemoved(ctx sdk.Context, _ sdk.ConsAddress, valAddr sdk.ValAddress) {
+func (h Hooks) AfterValidatorRemoved(ctx sdk.Context, ca sdk.ConsAddress, valAddr sdk.ValAddress) {
+	// Delete any key assignments that are associated with this
+	// validator across all consumer chains.
+	h.k.IterateConsumerChains(ctx, func(ctx sdk.Context, chainID, clientID string) (stop bool) {
```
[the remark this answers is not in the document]
danwt: True, I have left that out of the MVP because the proposal timing stuff in general is still using the temporary solution of allowing assignments at any time.

x/ccv/provider/keeper/genesis.go
```diff
@@ -118,6 +129,29 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 			}
 		}
 
+		keyAssignment := func() *types.KeyAssignment {
```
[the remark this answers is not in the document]
danwt: Readability, performance hit is irrelevant. I've now compressed into not even using a temporary `keyAssignment`

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,739 @@
...
+		return false
+	})
+	for _, cca := range toDel {
+		ka.Store.DelConsumerConsAddrToLastUpdateMemo(cca)
+	}
+}
+
+// DeterministicStringify returns a deterministic string representation of the
+// key. This is useful to identify like keys with different representations.
+func DeterministicStringify(k tmprotocrypto.PublicKey) string {
```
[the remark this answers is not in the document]
danwt: Marshall serializes to bytes such that ANY protobuf compatible language can deserialize the bytes to a native representation if they are aware of the protobuf definition. So I don't think Marshalled data contains anything to do with Go.

You can't use tmprotocrypto.PublicKey as a golang map key because it is a struct with an interface inside

https://github.com/cosmos/interchain-security/blob/4444034d1dfa72328fd7ac2a8bf38915933b5a67/x/ccv/provider/keeper/key_assignment_test.go#L521

danwt:
Superseded by [marius/key-assignment](https://github.com/cosmos/interchain-security/tree/marius/key-assignment)

Issue #368, opened 2022-10-04
Document why pending consumer initiated slash requests are sent in reverse order to provider


danwt:
# Problem

Pending slash requests are sent in reverse order

https://github.com/cosmos/interchain-security/blob/d995b56976d4a650bbc690b9001a8614dc067240/x/ccv/consumer/keeper/relay.go#L167-L181

This is confusing because the reason is not documented.

# Closing criteria

Document the code so that the reader isn't confused.

mpoke:
I agree, this is not clear. It is documented in the [spec](https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-ccf-sndpeslash1) (see the Note), i.e.,
> Iterating over pending SlashRequests in reverse order ensures that validators that are down for multiple blocks during channel initialization will be slashed for the latest downtime evidence.


Basically we send a slash request for the most recent infraction height.

mpoke:
No longer the case after https://github.com/cosmos/interchain-security/pull/448

PR #452, opened 2022-11-09
Consumer handles expired client to provider

mpoke:
Partial fix for https://github.com/cosmos/interchain-security/issues/435 (see https://github.com/cosmos/interchain-security/pull/448 for the other part). The consumer collects the IBC packets that cannot be sent to the provider due to the client being expired. Once the client is upgraded, these packets are sent in the next block.


mpoke:
@sainoe This PR removes the SlashRequests from the genesis state. Instead, we need to add the pending data packets, but it's a bit more involved. See https://github.com/cosmos/interchain-security/pull/264#discussion_r994333498 for more details.

x/ccv/consumer/keeper/relay.go
```diff
@@ -153,55 +183,67 @@ func (k Keeper) SendSlashPacket(ctx sdk.Context, validator abci.Validator, valse
...
-			// set validator outstanding downtime flag to true
-			if downtime {
-				k.SetOutstandingDowntime(ctx, sdk.ConsAddress(slashReq.Packet.Validator.Address))
+		// send packet over IBC channel
+		err = k.channelKeeper.SendPacket(ctx, channelCap, packet)
+		if clienttypes.ErrClientNotActive.Is(err) {
+			// IBC client expired:
+			if i != 0 {
+				// this should never happen
+				panic(fmt.Errorf("client expired while sending pending packets: %w", err))
```
MSalopek, line 235: This is an edgecase also handled in #448.

I'm wondering if this should invoke a panic?

We could simply delete packets that were sent so we don't resend them when a new client is established?

mpoke: This should never happen. The client cannot expired in the middle of a block as the time doesn't change. Also, it cannot expire in the middle of a loop as the SDK is sequential. I guess we could remove this check to save space :)

tests/e2e/expired_client.go
```diff
@@ -76,20 +65,148 @@ func (s *CCVTestSuite) TestVSCPacketSendExpiredClient() {
...
+	s.providerChain.NextBlock()
+	// - relay 1 VSC packet from provider to consumer
+	relayAllCommittedPackets(s, s.providerChain, s.path, ccv.ProviderPortID, s.path.EndpointB.ChannelID, 1)
+	// - increment time so that the unbonding period ends on the provider
+	incrementTimeByUnbondingPeriod(s, Consumer)
+	// - relay 1 VSCMatured packet from consumer to provider
+	relayAllCommittedPackets(s, s.consumerChain, s.path, ccv.ConsumerPortID, s.path.EndpointA.ChannelID, 1)
+}
+
+// expireClient expires the client the `clientTo`
```
sainoe, line 167: typo

jtremback (changes requested):
The code structure here is very hard to follow IMO. As one example, the call to `SendPendingDataPackets` on line 89 of relay.go can return `false` for its `clientExpired` return value, even if the client is expired, if the queue is empty. This is then checked again lower down in the function, leading to a pretty circuitous logic path. There are other examples as well.

I believe that all of this could be made much simpler if the functions `SendVSCMaturedPackets` and `SendSlashPacket` would only ever append to the `PendingDataPacket` queue. Then, another endblock method would deal with sending everything in the `PendingDataPacket` queue and dealing with expired clients etc.

sainoe:
Maybe I missed sthg but sending the pending packet in `Endblock` won't work due to the channel order property. For instance, in a block `h1` the channel isn't established so all packets are appended to the queue. Then in the next block `h2` the channel is ready and a `SlashPacket` for downtime infraction is sent in `BeginBlock`. In this case the packets order isn't respected.

mpoke:
> Maybe I missed sthg but sending the pending packet in `Endblock` won't work due to the channel order property. For instance, in a block `h1` the channel isn't established so all packets are appended to the queue. Then in the next block `h2` the channel is ready and a `SlashPacket` for downtime infraction is sent in `BeginBlock`. In this case the packets order isn't respected.

The idea is that all packets to be send are added to the queue and then in `EndBlock` all the packets in the queue are sent. The order is maintain by the queue. In your example, the `SlashPacket` for downtime infraction is appended at the end of the queue in the `BeginBlock` of `h2`.

sainoe:
Okay I got it! Thanks for the correction.

danwt (commented):
I did read through and I like the semantic changes but I agree with other comments that it could do with some substantial refactoring. I'd accept a merge as long as it was followed up with refactoring immediately afterwards.

Nit: I think `DataPackets` as a term is a bit weird: all packets have data so it's not informative. Just `Packets` is fine

danwt:
By the way, will the changes here affect the fact that currently pending consumer slash requests are sent in reverse order

- https://github.com/cosmos/interchain-security/issues/368

MSalopek:
Merging this into `masa/435-client-expired` (#448) as advised by the rest of the team.

MSalopek:
Closing in favor of #448 which has all changes in a single PR.

Issue #449, opened 2022-11-08
Change our Iterators functions to take a `stop` callback to match the SDK iterator pattern


danwt:
# Problem

The pattern of what we expect from our `cb` callbacks passed to our iterators is internally inconsistent and usually doesn't match the SDK pattern.

# Closing criteria

Make all of our iterators consistent wrt. how the callback is used inside the iterator, the name of the callback, and the name of the callback named return parameter. We should really also have regression tests.

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project 

## Detail

For example we do

```
if !cb(ctx, chainID, clientID) {
			return
		}
```

_and_

```
if cb(ctx, chainID, clientID) {
			return
		}
```

https://github.com/cosmos/interchain-security/blob/d9a002ff8845a660e437d087c9bac1c6a0779589/x/ccv/provider/keeper/keeper.go#L141-L159

The SDK pattern can be seen here

```go
// IterateValidatorSigningInfos iterates over the stored ValidatorSigningInfo
func (k Keeper) IterateValidatorSigningInfos(ctx sdk.Context,
	handler func(address sdk.ConsAddress, info types.ValidatorSigningInfo) (stop bool)) {

	store := ctx.KVStore(k.storeKey)
	iter := sdk.KVStorePrefixIterator(store, types.ValidatorSigningInfoKeyPrefix)
	defer iter.Close()
	for ; iter.Valid(); iter.Next() {
		address := types.ValidatorSigningInfoAddress(iter.Key())
		var info types.ValidatorSigningInfo
		k.cdc.MustUnmarshal(iter.Value(), &info)
		if handler(address, info) {
			break
		}
	}
}
```

IMO we should name all the callbacks `stop`

Issue #446, opened 2022-11-08
Document iteration in `ConsumerAdditionPropsToExecute`


danwt:
# Problem

The function uses iterators with callbacks

https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/provider/keeper/proposal.go#L329-L334

and it's a bit unclear whether or not it is written as intended because of the `return true` and `return false`. I suspect it's correct because the proposals are ordered timestamp ascending. If that's true it would be good to have a comment there.

# Closing criteria

Add comments to the code to make the iteration pattern clearer.

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
See 

- https://github.com/cosmos/interchain-security/issues/513

for a possible related issue.

smarshall-spitzbart:
Closed in #459 by adding comments

Issue #454, opened 2022-11-09
Regression tests for faulty iterators


smarshall-spitzbart:
# Problem
https://github.com/cosmos/interchain-security/pull/447 and https://github.com/cosmos/interchain-security/pull/459 fixed multiple iterators being implemented incorrectly. We need to go back and make sure all the fixes have an appropriate UT to prove out the "correct" functionality (some functionality was changed, some was not). Its a bit concerning that some of the previous UTs did not catch the bugs.

# Closing criteria
Make sure there is a unit test for each fix AND that the unit test would have failed for previous code behavior (only where behavior was changed). Here's [one example](https://github.com/cosmos/interchain-security/pull/434#issuecomment-1304883178) of a UT that should exist

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
@MSalopek Please can you take a look at this and the code and decide if this should be closed or if there is still work to do?

See

- https://github.com/cosmos/interchain-security/commit/e0cbf9a8f17cb03d03604954ef8685d82d518ea5#diff-993a06db0f6114ea2c41155cce2ac1d530bedebba2ba051b3f0a0e5c0834a01f

Thanks

smarshall-spitzbart:
Thanks for checking into this. Note every iterator in our codebase should be tested against multiple consumers, multiple iterations, etc. via a unit or e2e test

jtremback:
Also, we have identified some other issues with the ordering in iterators. We should probably go over iterators in general and add a lot more tests, including tests that the order in the order we expect.

smarshall-spitzbart:
Here's the checklist of methods I've collected which use iterators in the [latest version](https://github.com/cosmos/interchain-security/pull/596/commits/9a04275e820ab38d8629b84937b76a38b9644de2) of the `iterators-to-arrays` branch, and therefore need to be regression tested to close this issue. Commits on top of the one referenced should add/subtract from this list if iterators are added/removed. 

These names were found by searching the repo for `KVStorePrefixIterator`, `KVStoreReversePrefixIterator`, `KVStorePrefixIteratorPaginated`, and `KVStoreReversePrefixIteratorPaginated`. If there's a flaw in the way I'm collecting these methods, lmk. Also, it may be useful to split up #596 by file and/or any PRs which add onto that branch. 

List:

### Consumer `keeper.go`
- [ ] GetElapsedPacketMaturityTimes
[...]

PR #459, opened 2022-11-09
Make iterators consistent with SDK pattern 

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/449, https://github.com/cosmos/interchain-security/issues/446

### DIFF
--- a/x/ccv/consumer/keeper/genesis_test.go
+++ b/x/ccv/consumer/keeper/genesis_test.go
@@ -386,7 +386,7 @@ func assertProviderClientID(t *testing.T, ctx sdk.Context, ck *consumerkeeper.Ke
 // assert that the given input match the height to valset update ID mappings in the store
 func assertHeightValsetUpdateIDs(t *testing.T, ctx sdk.Context, ck *consumerkeeper.Keeper, heighValsetUpdateIDs []consumertypes.HeightToValsetUpdateID) {
 	ctr := 0
-	ck.IterateHeightToValsetUpdateID(ctx, func(height uint64, vscID uint64) bool {
+	ck.IterateHeightToValsetUpdateID(ctx, func(height uint64, vscID uint64) (stop bool) {
 		require.Equal(t, heighValsetUpdateIDs[ctr].Height, height)
 		require.Equal(t, heighValsetUpdateIDs[ctr].ValsetUpdateId, vscID)
 		ctr++
--- a/x/ccv/consumer/keeper/genesis_test.go
+++ b/x/ccv/consumer/keeper/genesis_test.go
@@ -390,6 +390,6 @@ func assertHeightValsetUpdateIDs(t *testing.T, ctx sdk.Context, ck *consumerkeep
 		require.Equal(t, heighValsetUpdateIDs[ctr].Height, height)
 		require.Equal(t, heighValsetUpdateIDs[ctr].ValsetUpdateId, vscID)
 		ctr++
-		return true
+		return false
 	})
 }
--- a/x/ccv/provider/keeper/keeper_test.go
+++ b/x/ccv/provider/keeper/keeper_test.go
@@ -394,7 +394,7 @@ func TestVscSendTimestamp(t *testing.T) {
 	require.Equal(t, 0, i)
 }
 
-// TestIterateConsumerChains iterates over consumer chains
+// TestIterateConsumerChains tests IterateConsumerChains behaviour correctness
 func TestIterateConsumerChains(t *testing.T) {
 	pk, ctx, ctrl, _ := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
 	defer ctrl.Finish()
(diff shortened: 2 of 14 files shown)


x/ccv/provider/keeper/keeper_test.go
```diff
@@ -369,28 +369,28 @@ func TestVscSendTimestamp(t *testing.T) {
...
 		i++
-		return true
+		return false // do not stop
 	})
 	require.Equal(t, 2, i)
 
 	// delete VSC send timestamps
 	var ids []uint64
-	providerKeeper.IterateVscSendTimestamps(ctx, testCases[0].chainID, func(vscID uint64, _ time.Time) bool {
+	providerKeeper.IterateVscSendTimestamps(ctx, testCases[0].chainID, func(vscID uint64, _ time.Time) (stop bool) {
```
smarshall-spitzbart, line 382: Don't need to iterate here twice, but that's not important for this PR

x/ccv/provider/keeper/proposal.go
```diff
@@ -159,14 +159,14 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, lockUbd, clos
...
-		k.IterateOverUnbondingOpIndex(ctx, chainID, func(vscID uint64, ids []uint64) bool {
+		k.IterateOverUnbondingOpIndex(ctx, chainID, func(vscID uint64, ids []uint64) (stop bool) {
 			// iterate over the unbonding operations for the current VSC ID
 			var maturedIds []uint64
 			for _, id := range ids {
 				unbondingOp, found := k.GetUnbondingOp(ctx, id)
 				if !found {
 					err = fmt.Errorf("could not find UnbondingOp according to index - id: %d", id)
-					return false
+					return true // stop the iteration
```
smarshall-spitzbart, line 169: Also not relevant for this PR necessarily, but we might want to panic here? And/or consider if we should keep iterating if we don't panic

x/ccv/provider/keeper/relay.go
```diff
@@ -157,7 +157,7 @@ func (k Keeper) sendValidatorUpdates(ctx sdk.Context) {
 				// CCV channel not established:
 				// store the packet data to be sent once the CCV channel is established
 				k.AppendPendingVSCs(ctx, chainID, []ccv.ValidatorSetChangePacketData{packetData})
-				return true // go to next consumer chain
+				return false // go to next consumer chain
```
smarshall-spitzbart, line 160:
```suggestion
				return false // do not stop iteration, go to next consumer chain
```

smarshall-spitzbart (approved):
LGTM, thanks @mpoke! Just a couple small comments. Following those we can close https://github.com/cosmos/interchain-security/issues/446. 

https://github.com/cosmos/interchain-security/issues/454 will be useful to prove out correct behavior of iterators

x/ccv/provider/keeper/proposal.go
mpoke: We do return an error below, see https://github.com/cosmos/interchain-security/blob/f2e403a25f851678a0dd1b07fdaa7a8ca55abc38/x/ccv/provider/keeper/proposal.go#L195

x/ccv/provider/keeper/keeper_test.go
smarshall-spitzbart: The two calls to `IterateVscSendTimestamps` as they exist could be replaced with one iterator
```go
	var ids []uint64
	i = 0
	providerKeeper.IterateVscSendTimestamps(ctx, testCases[0].chainID, func(vscID uint64, ts time.Time) (stop bool) {
		ids = append(ids, vscID)
		require.Equal(t, vscID, testCases[i].vscID)
		require.Equal(t, ts, testCases[i].ts)
		i++
		return false // do not stop
	})
	require.Equal(t, 2, i)
	for _, vscID := range ids {
		providerKeeper.DeleteVscSendTimestamp(ctx, testCases[0].chainID, vscID)
	}
```

mpoke: Oh, yeah. Not important though.

danwt:
Whats the reason for this going into [marius/435-client-expired-consumer](https://github.com/cosmos/interchain-security/tree/marius/435-client-expired-consumer) and not main, aren't we trying to minimize diffs?

mpoke:
> Whats the reason for this going into [marius/435-client-expired-consumer](https://github.com/cosmos/interchain-security/tree/marius/435-client-expired-consumer) and not main, aren't we trying to minimize diffs?

Once #435 is merged, this PR will be based on main. I basically try to minimized conflicts.

danwt:
I checked and it doesn't seem that any iterators or callsites were missed or wrong 👍 Nice work

MSalopek:
Ready for another review cycle after rebasing onto main

mpoke (commented):
Thanks @MSalopek. LGTM. Cannot approve since I opened the PR.

smarshall-spitzbart (approved):
Looks great!

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -367,28 +368,174 @@ func TestVscSendTimestamp(t *testing.T) {
...
+	defer ctrl.Finish()
+
+	type chanToChain struct {
+		chainID   string
+		channelID string
+	}
+
+	cases := []chanToChain{
+		{
+			chainID:   "chain-1",
```
smarshall-spitzbart, line 447: 🙌

x/ccv/consumer/keeper/genesis_test.go
```diff
@@ -390,6 +390,6 @@ func assertHeightValsetUpdateIDs(t *testing.T, ctx sdk.Context, ck *consumerkeep
 		require.Equal(t, heighValsetUpdateIDs[ctr].Height, height)
```
sainoe, line 390: nit: can you also name the `bool` callback output as `stop` above (line 389)

sainoe (approved):
LGTM!

x/ccv/consumer/keeper/genesis_test.go
MSalopek: Done

Issue #458, opened 2022-11-09
Remove special case in consumer addition/removal proposal handlers


danwt:
# Problem

The handlers unnecessarily (IMO) fork the execution by having two paths

https://github.com/cosmos/interchain-security/blob/dc19c57f71fb1adb38a5b73db0bdd3a95d99f713/x/ccv/provider/keeper/proposal.go#L30-L35

https://github.com/cosmos/interchain-security/blob/dc19c57f71fb1adb38a5b73db0bdd3a95d99f713/x/ccv/provider/keeper/proposal.go#L108-L112

This doubles the number of code paths to be tested and reasoned about for little or no gain because all pending proposals are processed in BeginBlock.

# Closing criteria

Remove the 'immediate' branches

https://github.com/cosmos/interchain-security/blob/dc19c57f71fb1adb38a5b73db0bdd3a95d99f713/x/ccv/provider/keeper/proposal.go#L30-L33

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project 

# Detail

I can see some extremely minor benefit to ux, perhaps if a consumer has to be stopped in 'an emergency' or something.

# Priority

This is low priority because everything seems to be working, although there might be odd yet-to-be solved things going on with the proposals (see https://github.com/cosmos/interchain-security/issues/438)

danwt:
cc @MSalopek

danwt:
An example of this leading to a problem

- https://github.com/cosmos/interchain-security/issues/463

Issue #390, opened 2022-10-12
Outdated tags for `HandleConsumerAdditionProposal` and `HandleConsumerRemovalProposal`


mpoke:
# Problem

This [PR](https://github.com/cosmos/ibc/pull/855) on the spec side will make the comments (i.e., the tags) for both `HandleConsumerAdditionProposal` and `HandleConsumerRemovalProposal` outdated. See [here](https://github.com/cosmos/interchain-security/blob/6e0666275a219ebb729417557ef76f4c8fe6c10b/x/ccv/provider/keeper/proposal.go#L30) and [here](https://github.com/cosmos/interchain-security/blob/6e0666275a219ebb729417557ef76f4c8fe6c10b/x/ccv/provider/keeper/proposal.go#L101).

# Closing criteria

Update comments.

jtremback:
@mpoke did this get corrected?

PR #460, opened 2022-11-09
simplify consumer addition and removal propsal handling

MSalopek:
Immediate execution of CreateConsumerClient and StopConsumerClient is not necessary because all pending proposals are already handled in BeginBlock.

Closes: #458 
Closes: #390

### DIFF
--- a/x/ccv/provider/keeper/proposal.go
+++ b/x/ccv/provider/keeper/proposal.go
@@ -20,20 +20,9 @@ import (
 )
 
 // HandleConsumerAdditionProposal will receive the consumer chain's client state from the proposal.
-// If the spawn time has already passed, then set the consumer chain. Otherwise store the client
-// as a pending client, and set once spawn time has passed.
-//
-// Note: This method implements SpawnConsumerChainProposalHandler in spec.
-// See: https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-spccprop1
-// Spec tag: [CCV-PCF-SPCCPROP.1]
+// The proposal is stored into state and if spawn time has passed the client will be created set in the next BeginBlocker.
 func (k Keeper) HandleConsumerAdditionProposal(ctx sdk.Context, p *types.ConsumerAdditionProposal) error {
-	if !ctx.BlockTime().Before(p.SpawnTime) {
-		// lockUbdOnTimeout is set to be false, regardless of what the proposal says, until we can specify and test issues around this use case more thoroughly
-		return k.CreateConsumerClient(ctx, p.ChainId, p.InitialHeight, false)
-	}
-
-	err := k.SetPendingConsumerAdditionProp(ctx, p)
-	if err != nil {
+	if err := k.SetPendingConsumerAdditionProp(ctx, p); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


danwt:
Hey, looks like excellent work

Any chance we can incorporate 

- https://github.com/cosmos/interchain-security/issues/463

Or perhps #463 should go in a different pr

x/ccv/provider/keeper/proposal.go
```diff
@@ -99,16 +88,8 @@ func (k Keeper) CreateConsumerClient(ctx sdk.Context, chainID string,
 
 // HandleConsumerRemovalProposal stops a consumer chain and released the outstanding unbonding operations.
 // If the stop time hasn't already passed, it stores the proposal as a pending proposal.
-//
-// This method implements StopConsumerChainProposalHandler from spec.
-// See: https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-stccprop1
-// Spec tag: [CCV-PCF-STCCPROP.1]
+// If the stop time has passed the proposal will be executed in BeginBlocker.
```
danwt, line 91: We could consolidate comments on 90-91

```
The proposal is stored into state and will be executed in BeginBlocker once the spawn time has passed.
```

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -348,39 +334,23 @@ func TestHandleConsumerRemovalProposal(t *testing.T) {
...
-
 		err := providerKeeper.HandleConsumerRemovalProposal(ctx, tc.prop)
 		require.NoError(t, err)
 
-		if tc.expStop {
-			// Expect no pending proposal to exist
-			found := providerKeeper.GetPendingConsumerRemovalProp(ctx, tc.prop.ChainId, tc.prop.StopTime)
-			require.False(t, found)
-
-			testProviderStateIsCleaned(t, ctx, providerKeeper, tc.prop.ChainId, "channelID")
```
danwt, line 378: In replacing these lines we want to maintain the same amount of coverage. Here the proposal actually was processed so `testProviderStateIsCleaned` checks that the consumer was removed. An equivalent should be added. ie.

```
BeginBlocker()
testProviderStateIsCleaned
```

to check the same thing.

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -71,41 +70,29 @@ func TestHandleConsumerAdditionProposal(t *testing.T) {
...
-			)
-		}
-
 		tc.prop.LockUnbondingOnTimeout = false // Full functionality not implemented yet.
 
 		err := providerKeeper.HandleConsumerAdditionProposal(ctx, tc.prop)
 		require.NoError(t, err)
 
-		if tc.expCreatedClient {
-			testCreatedConsumerClient(t, ctx, providerKeeper, tc.prop.ChainId, "clientID")
```
danwt, line 99: Similiar to another comment, an equivalent check should probably be tested

```
BeginBlocker()
testCreatedConsumerClient
```

danwt (changes requested):
Hey, nice work!

I just noticed that the content of the tests have changed somewhat because the actual addition/removal parts are no longer tested. Thus, somewhere, it's needed to do BeginBlock() and process the pendings.

x/ccv/provider/keeper/proposal.go
```diff
@@ -20,20 +20,9 @@ import (
...
+// The proposal is stored into state and if spawn time has passed the client will be created set in the next BeginBlocker.
 func (k Keeper) HandleConsumerAdditionProposal(ctx sdk.Context, p *types.ConsumerAdditionProposal) error {
-	if !ctx.BlockTime().Before(p.SpawnTime) {
-		// lockUbdOnTimeout is set to be false, regardless of what the proposal says, until we can specify and test issues around this use case more thoroughly
-		return k.CreateConsumerClient(ctx, p.ChainId, p.InitialHeight, false)
-	}
-
-	err := k.SetPendingConsumerAdditionProp(ctx, p)
-	if err != nil {
+	if err := k.SetPendingConsumerAdditionProp(ctx, p); err != nil {
```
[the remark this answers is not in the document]
MSalopek: I'm not sure what you mean? The behaviour is the same with regards to error handling.

MSalopek:
Let's keep this on ice for a little bit. It is implemented per the spec. The issue remains open for further investigation.

Closing.

PR #461, opened 2022-11-09
Slashing related e2e test improvements

smarshall-spitzbart:
Improves a couple slashing related e2e tests by:
1. Removing duplicate test/setup code in favor of test cases and if-statements (TestSendSlashPacketDowntime and TestSendSlashPacketDoubleSign were 90% the same, now they are one test method)
2. Removing a hack which previously reconstructed a VSC packet from local variables, and manually sent/received that packet using IBC testing functions
3. Using `providerSlashingKeeper.SlashFractionDowntime` instead of hardcoding an expected slash fraction

The second point here is needed to properly tie https://github.com/cosmos/interchain-security/pull/462 into existing e2e tests, the hack was faultily making e2e tests fail.

### DIFF
--- a/tests/e2e/slashing.go
+++ b/tests/e2e/slashing.go
@@ -14,249 +14,174 @@ import (
 	clienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 
-	"github.com/cosmos/interchain-security/x/ccv/types"
-
 	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/crypto/ed25519"
 )
 
-// TestSendDowntimePacket tests consumer initiated slashing
-func (s *CCVTestSuite) TestSendSlashPacketDowntime() {
-	s.SetupCCVChannel()
-	s.SetupTransferChannel()
-	validatorsPerChain := len(s.consumerChain.Vals.Validators)
+const (
+	downtimeTestCase = iota
+	doubleSignTestCase
+)
 
-	providerStakingKeeper := s.providerApp.GetE2eStakingKeeper()
-	providerSlashingKeeper := s.providerApp.GetE2eSlashingKeeper()
-	consumerKeeper := s.consumerApp.GetConsumerKeeper()
-	providerKeeper := s.providerApp.GetProviderKeeper()
+func (s *CCVTestSuite) TestSendAndApplySlashPacket() {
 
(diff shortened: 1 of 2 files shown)


tests/e2e/slashing.go
```diff
@@ -14,249 +14,170 @@ import (
...
-		Address: tmVal.Address,
-		Power:   tmVal.VotingPower,
-	}
+		// check that the validator was removed from the provider validator set
+		s.Require().Len(s.providerChain.Vals.Validators, validatorsPerChain-1)
 
-	oldBlockTime := s.consumerCtx().BlockTime()
-	packetData := types.NewSlashPacketData(validator, valsetUpdateId, stakingtypes.DoubleSign)
+		// Relay the VSC packet to the consumer
+		relayAllCommittedPackets(s, s.providerChain, s.path, ccv.ProviderPortID, s.path.EndpointB.ChannelID, 1)
```
smarshall-spitzbart, line 125: This is the new way that we relay the VSC packet from provider to consumer. Note that this method is commonly used elsewhere

tests/e2e/slashing.go
```diff
@@ -14,249 +14,170 @@ import (
...
+		// check that the validator's tokens were slashed
+		var slashFraction sdk.Dec
+		if tc == downtimeTestCase {
+			slashFraction = providerSlashingKeeper.SlashFractionDowntime(s.providerCtx())
 
-	// update consumer client on the VSC packet sent from provider
-	err = s.path.EndpointA.UpdateClient()
-	s.Require().NoError(err)
-
-	// reconstruct VSC packet
```
smarshall-spitzbart, line 214: This was the old way we sent the VSC packet from provider to consumer

x/ccv/provider/keeper/keeper.go
```diff
@@ -830,6 +830,24 @@ func (k Keeper) SetVscSendTimestamp(
 	store.Set(types.VscSendingTimestampKey(chainID, vscID), timeBz)
 }
 
+// GetVscSendTimestamp returns a VSC send timestamp by chainID and vscID
+//
+// Note: This method is used only for testing.
```
danwt, line 835: 👍

danwt (approved):
Really nice work, the new test is very clear 👍

sainoe (approved):
LGTM! Just a small comment.

tests/e2e/slashing.go
smarshall-spitzbart: I've also updated the comment for `TestRelayAndApplySlashPacket` to make this more clear

PR #467, opened 2022-11-10
testutils refactors and cleanup

smarshall-spitzbart:
This PR should not change test behavior in anyway. It should allow the implementation of multiple consumers in e2e tests to be cleaner. Changes included: 
- Remove the unused `testutil/sample` folder
- Consolidate `testutil/simibc` and `testutil/simapp` into a more descriptively named folder, `testuil/ibc_testing_utils`
- Clean up what was `testutil/simapp/simapp.go` and is now `testutil/ibc_testing_utils/setup.go` with semanitcs closer to the design of the IBC testing package. I've also made that file a bit more succinct
- Adjust e2e test instantiation to use the new IBC testing util functions


testutil/ibc_testing_utils/setup.go
```diff
@@ -0,0 +1,74 @@
...
+		chainID := ibctesting.GetChainID(i + 2)
+		testChain := ibctesting.NewTestChainWithValSet(t, coordinator,
+			appIniter, chainID, providerChain.Vals, providerChain.Signers)
+
+		coordinator.Chains[chainID] = testChain
+		consumers = append(consumers, testChain)
+	}
+
+	return consumers
+}
```
danwt, line 74: Nice, I like this split of creating the app and adding the consumers separately 👍

danwt (changes requested):
Nice job with the changes and adding multiple consumers!

I have one qualm which is that I don't see a compelling reason to put what was simibc and simapp into the same dir named ibc_testing_utils.

It's already clear that they are all testing utils because of the parent directory name. Also, the app stuff in setup.go is not necessary to use the other utils, and putting them in the same directory makes it seem like it is. Id rather keep it separate if that's ok.

Other than that I think the changes are great!

smarshall-spitzbart:
sorry I screwed up this branch haha, Dan I'll make your req changes in a new PR

PR #473, opened 2022-11-11
Use new cryptoId testutil across repo

smarshall-spitzbart:
Extends https://github.com/cosmos/interchain-security/pull/470 with usage of the new util. Also removes old/redundant helpers that are a less complete version of Dan's new util.

I'd love some sanity checks regarding whether I'm using the correct key types in the various tests


testutil/crypto/crypto.go
```diff
@@ -31,10 +32,16 @@ func NewCryptoIdentityFromBytesSeed(seed []byte) CryptoIdentity {
 	return CryptoIdentity{PV: privKey}
 }
 
-func NewCryptoIdentityFromIntSeed(i int) CryptoIdentity {
-	iUint64 := uint64(i)
+func NewCryptoIdentityFromRandSeed() CryptoIdentity {
+	b := make([]byte, 8)
+	_, _ = rand.Read(b)
+	seed := binary.BigEndian.Uint64(b)
```
danwt, line 38: Nice! I forgot to add this. I think you can do rand.Uint64() directly

https://pkg.go.dev/math/rand#Uint64

danwt (changes requested):
LGTM! Just a nit

testutil/crypto/crypto.go
```diff
@@ -31,10 +32,16 @@ func NewCryptoIdentityFromBytesSeed(seed []byte) CryptoIdentity {
...
+	return NewCryptoIdentityFromIntSeed(seed)
```
danwt, line 39:
```suggestion
	return NewCryptoIdentityFromIntSeed(rand.Uint64())
```

smarshall-spitzbart:
@danwt your idea was considered in an earlier commit of this branch. The `rand` package still needs to be seeded itself, otherwise it makes `NewCryptoIdentityFromIntSeed` pointless. See the gosec error message for commit `966d7292c3f44250cca4bd0b90f30b117913697f` 

```bash
[/github/workspace/testutil/crypto/crypto.go:38] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    37: 	rand.Seed(time.Now().UnixMicro())
  > 38: 	return NewCryptoIdentityFromIntSeed(rand.Int())
    39: }
```

The latest commits I have in this PR pass the gosec tests, and are functional for random seeding based on memory allocation and the crypto package. Heres a [stack overflow](https://stackoverflow.com/questions/12321133/how-to-properly-seed-random-number-generator) with different opinions on which way is best to seed in golang, but it didn't seem worth it to get too deep down that rabbit hole lol :), I just went with the way that passes tests

danwt:
@smarshall-spitzbart ah I see aha you are one step ahead.

I would suggest we assume any caller of ` NewCryptoIdentityFromRandSeed()` has already seeded rand properly how they want. What do you think?

smarshall-spitzbart:
> @smarshall-spitzbart ah I see aha you are one step ahead.
> 
> I would suggest we assume any caller of ` NewCryptoIdentityFromRandSeed()` has already seeded rand properly how they want. What do you think?

Yea if a caller wants to seed in some customized way (like from a trace, etc.) they can still call `NewCryptoIdentityFromRandSeed` directly. The additional helper function is useful for e2e and unit tests where we probably don't need to explicitly seed inline for each test, just reducing boilerplate in those scenarios

testutil/keeper/unit_test_helpers.go
```diff
@@ -177,16 +174,6 @@ func (params *InMemKeeperParams) RegisterSdkCryptoCodecInterfaces() {
 	params.Cdc = codec.NewProtoCodec(ir)
 }
 
-type PrivateKey struct {
```
smarshall-spitzbart, line 180: Removal of old helper here

tests/e2e/normal_operations.go
```diff
@@ -20,8 +20,8 @@ func (k CCVTestSuite) TestHistoricalInfo() {
 	// and then call track historical info in the next block
 	createVal := func(k CCVTestSuite) {
 		// add new validator to consumer states
-		pk := ed25519.GenPrivKey().PubKey()
-		cVal, err := consumertypes.NewCCValidator(pk.Address(), int64(1), pk)
+		cId := crypto.NewCryptoIdentityFromRandSeed()
+		cVal, err := consumertypes.NewCCValidator(cId.SDKValAddress(), int64(1), cId.SDKPubKey())
```
danwt, line 24: 1st arg I'm not sure about. I think it might need to be `cId.ABCIAddressBytes()`
3rd arg is fine

tests/e2e/slashing.go
```diff
@@ -546,22 +546,23 @@ func (suite *CCVTestSuite) TestValidatorDoubleSigning() {
...
+		})
 
 	// save next sequence before sending a slash packet
 	seq, ok := suite.consumerApp.GetIBCKeeper().ChannelKeeper.GetNextSequenceSend(ctx, ccv.ConsumerPortID, channelID)
 	suite.Require().True(ok)
 
 	// construct slash packet data and get the expcted commit hash
 	packetData := ccv.NewSlashPacketData(
-		abci.Validator{Address: consAddr.Bytes(), Power: power},
+		abci.Validator{Address: cryptoId.ABCIAddressBytes(), Power: power},
```
danwt, line 565: I thinkkkk it's ok 👍 . It's also confusing for me tbh.

testutil/crypto/crypto.go
```diff
@@ -31,10 +33,16 @@ func NewCryptoIdentityFromBytesSeed(seed []byte) CryptoIdentity {
...
+}
+
+func NewCryptoIdentityFromIntSeed(i uint64) CryptoIdentity {
 	seed := []byte("AAAAAAAAabcdefghijklmnopqrstuvwx") // 8+24 bytes
-	binary.LittleEndian.PutUint64(seed[:8], iUint64)
+	binary.LittleEndian.PutUint64(seed[:8], i)
 	return NewCryptoIdentityFromBytesSeed(seed)
 }
```
danwt, line 47: I would prefer a separate function `FromUint64Seed` if it's necessary. I used `int` as the arg to begin with because it's the easiest type to use when writing tests (in loops and such).

x/ccv/consumer/keeper/genesis_test.go
```diff
@@ -188,7 +185,7 @@ func TestExportGenesis(t *testing.T) {
 			name: "export a new chain",
 			malleate: func(ctx sdk.Context, ck consumerkeeper.Keeper, mocks testutil.MockedKeepers) {
 				// populate the states used by a new consumer chain
-				cVal, err := consumertypes.NewCCValidator(validator.Address.Bytes(), 1, pubKey)
+				cVal, err := consumertypes.NewCCValidator(validator.Address.Bytes(), 1, cId.SDKPubKey())
```
danwt, line 188: There are several instances of `validator.Address.Bytes()` I think they could be `cId.ABCIAddressBytes()`

danwt: Even if it's true it might not be worth changing them.

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -120,15 +116,14 @@ func TestCrossChainValidator(t *testing.T) {
 	consumerKeeper, ctx, ctrl, _ := testkeeper.GetConsumerKeeperAndCtx(t, keeperParams)
 	defer ctrl.Finish()
 
+	cId := crypto.NewCryptoIdentityFromRandSeed()
+
 	// should return false
-	_, found := consumerKeeper.GetCCValidator(ctx, ed25519.GenPrivKey().PubKey().Address())
+	_, found := consumerKeeper.GetCCValidator(ctx, cId.SDKValAddress())
```
danwt, line 122: I'm not sure if `cId.SDKValAddress()` is right here.

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -120,15 +116,14 @@ func TestCrossChainValidator(t *testing.T) {
...
 	require.False(t, found)
 
-	// Obtain derived private key
-	privKey := ed25519.GenPrivKey()
-
 	// Set cross chain validator
-	ccVal, err := types.NewCCValidator(privKey.PubKey().Address(), 1000, privKey.PubKey())
+	ccVal, err := types.NewCCValidator(cId.SDKValAddress(), 1000, cId.SDKPubKey())
```
danwt, line 126: I think the first arg maybe should be `cId.ABCIAddressBytes()`

x/ccv/consumer/keeper/validators_test.go
```diff
@@ -165,8 +166,8 @@ func GenerateValidators(t testing.TB) []*tmtypes.Validator {
 	numValidators := 4
 	validators := []*tmtypes.Validator{}
 	for i := 0; i < numValidators; i++ {
-		pubKey, err := testkeeper.GenPubKey()
-		require.NoError(t, err)
+		cryptoId := ccvcrypto.NewCryptoIdentityFromRandSeed()
+		pubKey := cryptoId.TMCryptoPubKey()
```
danwt, line 170: Could even make it a 1 liner 🤷

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -198,8 +198,9 @@ func TestHandleSlashPacketDoubleSigning(t *testing.T) {
 	keeperParams := testkeeper.NewInMemKeeperParams(t)
 	ctx := keeperParams.Ctx
 
+	cId := crypto.NewCryptoIdentityFromRandSeed()
 	slashPacket := ccv.NewSlashPacketData(
-		abci.Validator{Address: ed25519.GenPrivKey().PubKey().Address(),
+		abci.Validator{Address: cId.SDKValAddress(),
```
danwt, line 203: Maybe `ABCIAddressBytes()`

PR #482, opened 2022-11-15
Better documents HandleSlashPacket fatal cases

danwt:
I have added explicit comments around the critical error cases in HandleSlashPacket. I split the cases in two to draw attention to the problem. They should anyways be _logged_ separately when we add logging.


jtremback (changes requested):
The comment seems like a good addition, but I don't want to merge the code changes which did not change anything but just make the code longer and slightly more complicated.

PR #485, opened 2022-11-16
Upgrade SDK to v0.45.11

sainoe:
Upgrade SDK to v0.45.11 ([IS-fork branch](https://github.com/cosmos/cosmos-sdk/tree/interchain-security-rebase.0.45.11))

Minor changes to work with  the SDK team [optimizations on SDK 47 ](https://github.com/cosmos/cosmos-sdk/pull/13122/files)


x/ccv/provider/keeper/hooks.go
```diff
@@ -48,13 +48,14 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) {
 
 	// Set unbondingOp
 	if err := h.k.SetUnbondingOp(ctx, unbondingOp); err != nil {
-		panic(fmt.Errorf("unbonding op could not be persisted: %w", err))
+		return fmt.Errorf("unbonding op could not be persisted: %w", err)
```
jtremback, line 51: What happens on the other side when this error is returned? It's hard to find out what actual branch you're using from the go mod file.

jtremback (approved):
LGTM

x/ccv/provider/keeper/hooks.go
smarshall-spitzbart: I believe a logged error: https://github.com/cosmos/cosmos-sdk/blob/9c145c827001222df2e3e1101010874aeac20997/x/staking/keeper/delegation.go#L527

go.mod
```diff
@@ -12,19 +12,19 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/rakyll/statik v0.1.7
-	github.com/spf13/cast v1.4.1
-	github.com/spf13/cobra v1.4.0
-	github.com/stretchr/testify v1.7.1
+	github.com/spf13/cast v1.5.0
+	github.com/spf13/cobra v1.6.0
+	github.com/stretchr/testify v1.8.0
```
smarshall-spitzbart, line 17: This is the upgrade that is causing our CI to fail, is there a way to reference the older version of testify?

x/ccv/provider/keeper/hooks.go
smarshall-spitzbart: Previously we panicked for this error, and now we do not. Is this change in behavior ok and intentional? Does it even matter?

smarshall-spitzbart (changes requested):
Nice work! This PR seems safe, just want to confirm the changes in behavior are ok before merging

x/ccv/provider/keeper/hooks.go
sainoe: @jtremback  the SDK branch used is in the description

@smarshall-spitzbart `RunTx` normally recovers all the panic in tx executions https://docs.cosmos.network/v0.45/core/runtx_middleware.html so I think doesn't change the logic and it's more a matter of logging the error.

mpoke: I agree that this should return an error instead of panicking, especially since `AfterUnbondingInitiated` has error as a return field

mpoke (approved):
LGTM

go.mod
sainoe: We can downgrade them manually in the `go.mod` replace block. But what's failing exactly?

x/ccv/provider/keeper/hooks.go
smarshall-spitzbart: > I agree that this should return an error instead of panicking, especially since AfterUnbondingInitiated has error as a return field.

The error return field is introduced in this PR, I'm just pointing out that I think this PR will change behavior. Whether that matters, I'm not sure.

From the link @sainoe has given
> Depending on the panic type a different handler is used, for instance the default one prints an error log message.

I agree that we will log an error before and after this PR👍 

There is still a change in behavior AFAIK tho. If we panic, execution of the tx will halt and not be committed. Whereas with the new functionality, we log an error and continue tx execution. 

To make this PR not change any functionality, we'd need to panic on the sdk side of things

go.mod
smarshall-spitzbart: Here's [an issue](https://github.com/stretchr/testify/issues/1199) pointing to what was failing. Great point tho, easily fixable as https://github.com/cosmos/interchain-security/pull/485/commits/dcd0b09febbeb4e10ebe12d7addd29c8b7b3d362

smarshall-spitzbart (approved):
Approving and merging because even if a potential change in functionality from this PR matters, it could be fixed on the sdk end

x/ccv/provider/keeper/hooks.go
jtremback: Yea we need to leave the panics. They roll back the tx

Issue #397, opened 2022-10-14
Democracy consumer is not fully e2e tested


smarshall-spitzbart:
# Problem
The democracy consumer app does not run against the standard suite of e2e tests. Since this is an extension to the minimum functionality consumer, it should be ran against the same test cases as the minimum functionality consumer

# Closing criteria
Have all e2e tests run against normal consumer AND gov consumer. What about integration and diff?

jtremback:
Will this be easily solved through the test portability work you are doing rn?

smarshall-spitzbart:
This issue _should_ be easily solved by copying the pattern of `tests/e2e/instance_test.go` for both the democracy consumer and regular consumer. However when I run the democ consumer against the standard e2e tests I get this odd error coming from multiple tests
```
signature verification failed; please verify account number (0) and chain-id (testchain2): unauthorized
```

This might be related to https://github.com/cosmos/interchain-security/issues/151? (which still has a TODO comment in code)

@sainoe has the issue outlined in #151 produced similar errors in the past?

sainoe:
Yes it looks like a similar issue. The democracy module creates more accounts than a MVCC. It probably changes the account numbers. You should check if this test setup is still correct.

https://github.com/cosmos/interchain-security/blob/9dda4b606a7a71e7f6426feb5603600b66790a05/tests/e2e/setup_test.go#L112-L115

Btw where is located `tests/e2e/instance_test.go` ?

Issue #151, opened 2022-06-15
Fix sender accounts initialization in IBC testing


sainoe:
At the moment all the testing suites need to have sender accounts manually setup like so:
https://github.com/cosmos/interchain-security/blob/404529c8866c007bb02ba1309a249e369979486f/x/ccv/consumer/module_test.go#L90-L91

However it shouldn't be the case since all sender accounts are initialized in `ibc/testing/chain.go`. The bug can be identified when comparing the accounts number stored in the test chains against their value in the account module states.
```
cApp := suite.consumerChain.App.(*appConsumer.App)

acnt := suite.consumerChain.SenderAccount
gotAcnt := cApp.AccountKeeper.GetAccount(suite.consumerChain.GetContext(), acnt.GetAddress())

suite.Require().EqualValues(acnt.GetAccountNumber(), gotAcnt.GetAccountNumber())
```

jtremback:
What impact does this have?

PR #495, opened 2022-11-18
Fixes for issues #151 #397

stana-ethernal:
This PR contains fixes for issues #151 #397 
Auth module is  moved before ibc transfer module in SetOrderInitGenesis which eliminates the need to manually set account IDs (issue 151) and democracy consumer is tested against the same test cases as the minimum functionality consumer except that TestRewardsDistribution is skipped because democracy specific reward distribution is tested in ConsumerDemocracyTestSuite.


smarshall-spitzbart (approved):
Looks great! Thanks for knocking out those two issues

jtremback (approved):

Issue #487, opened 2022-11-16
Delete or use consumer addition proposal GenesisHash and BinaryHash fields


danwt:
# Problem

These fields seem to be unused

https://github.com/cosmos/interchain-security/blob/fb83bb7dbeef3532ef779adeb63b58b547189f21/x/ccv/provider/types/provider.pb.go#L48-L50

# Closing criteria

~~Remove the fields or use them~~
EDIT: Add comment explaining the purpose of these fields, see https://github.com/cosmos/interchain-security/issues/487#issuecomment-1321789297

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

sainoe:
These fields are meant to be used off-chain to validate consumer chain proposals against their associated binary and genesis file.

mpoke:
We should add a comment explaining that better.

PR #502, opened 2022-11-21
add consumer addition proposal documentation

MSalopek:
Closes: #487

### DIFF
--- a/proto/interchain_security/ccv/provider/v1/provider.proto
+++ b/proto/interchain_security/ccv/provider/v1/provider.proto
@@ -27,9 +27,13 @@ message ConsumerAdditionProposal {
     // the proposed initial height of new consumer chain.
     // For a completely new chain, this will be {0,1}. However, it may be different if this is a chain that is converting to a consumer chain.
     ibc.core.client.v1.Height initial_height = 4 [(gogoproto.nullable) = false];
-    // genesis hash with no staking information included.
+    // genesis hash contains initial chain genesis state without the cross-chain validating validator set
+    // it is used for off-chain confirmation of genesis.json validity by validators and other parties.
+    // i.e. hashing the genesis.json provided by the consumer chain project staff should produce this value.
     bytes genesis_hash = 5 ;
     // binary hash is the hash of the binary that should be used by validators on chain initialization.
+    // it is used for off-chain confirmation of binary validity by validators and other parties
+    // ie. the validators may build the consumer chain's binary from source and check if the resulting binary's hash matches this value.
     bytes binary_hash = 6 ;
     // spawn time is the time on the provider chain at which the consumer chain genesis is finalized and all validators
     // will be responsible for starting their consumer chain validator node.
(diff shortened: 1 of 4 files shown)


proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -27,9 +27,13 @@ message ConsumerAdditionProposal {
     // the proposed initial height of new consumer chain.
     // For a completely new chain, this will be {0,1}. However, it may be different if this is a chain that is converting to a consumer chain.
     ibc.core.client.v1.Height initial_height = 4 [(gogoproto.nullable) = false];
-    // genesis hash with no staking information included.
+    // genesis hash contains initial chain genesis state without the cross-chain validating validator set
+    // it is used for off-chain confirmation of genesis.json validity by validators and other parties.
+    // i.e. hashing the genesis.json provided by the consumer chain project staff should produce this value.
```
mpoke, line 32:
```suggestion
    // The hash of the genesis state needed for the consumer CCV module. 
    // It is used for off-chain confirmation of genesis.json validity by validators and other parties.
```
AFAIK it does contain also the initial valset, see https://github.com/cosmos/interchain-security/blob/87c21be74f121802d8be170d1275130466a9be59/x/ccv/provider/keeper/proposal.go#L277

mpoke (changes requested):

proto/interchain_security/ccv/provider/v1/provider.proto
MSalopek: True. I'm drawin my conclusions from Game of Chains testnet.

In the testnet the proposals were handled as follows:
1. the genesis.json hashed on proposal submission did not include any ccv state (no `initial_val_set` in genesis.json)
2. after spawn time, the genesis.json `initial_val_set` was populated and that was the genesis file used to start the chain.


Since the initial_val_set on proposal submission can be different that the actual valset at spawn time, does it make sense to have the `initial_val_set` set to anything but an empty array?

On spawn time the valset may be different, i.e. the validators on the provider may change - does that bring up any issues?

mpoke: Ah yes, that's true. The `genesis_hash` cannot contain any info about the CCV module. It's the hash of the genesis state without the CCV module data, as that part is computed by `MakeConsumerGenesis` when the proposal is handled by the provider chain.

mpoke: I've updated my suggestion above

MSalopek: Done

mpoke (approved):
Thanks @MSalopek

x/ccv/provider/keeper/proposal.go
```diff
@@ -219,7 +219,7 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, lockUbd, clos
 	return nil
 }
 
-// MakeConsumerGenesis constructs a consumer genesis state.
+// MakeConsumerGenesis constructs the consumer CCV module part of the genesis state.
```
smarshall-spitzbart, line 222: very clear 👍

smarshall-spitzbart (approved):

Issue #498, opened 2022-11-18
Upgrade IBC go to v3.4.0


smarshall-spitzbart:
# Problem

We're currently using a v3 beta version of IBC go in this repo. v3.4.0 is a more up to date version with changes to the testing package that will prevent e2e testing bugs. Note that the current version of ibc go we reference is causing #493 

# Closing criteria

Upgrade to v3.4 and implement any neccessary refactors therein. 

# Problem details

See #493 

https://github.com/cosmos/ibc-go/releases/tag/v3.4.0

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
Interesting! I'd definitely be happy to upgrade. I'm curious, what are the bugs that are fixed in 3.4?

Btw, just for reference, we have several of our own commits on our fork head, some of which might want to be rebased

https://github.com/informalsystems/ibc-go/commits/interchain-security-2

danwt:
cc @sainoe @glnro

smarshall-spitzbart:
Note, to fix the problems I was facing when this issue was made, I just added commits to our IBC 3.0 fork. I didn't get very far on upgrading to 3.4. Happy to help as needed tho!

sainoe:
Solved in https://github.com/cosmos/interchain-security/pull/559

PR #504, opened 2022-11-21
Improve how sent packets are stored in e2e tests 

### LINKED ISSUE #493: Multi-consumer e2e blocker
# Problem

I'm running into odd IBC related issues with e2e tests that setup the CCV channel for multiple in-mem consumers.

# Closing criteria

`TestRelayAndApplySlashPacket` passes with multiple consumers and multiple established CCV channels.

# Problem details

In developing #475, I refactored `CCVTestSuite.SetupTest()` and `CCVTestSuite.SetupCCVChannel()` to be compatible with a list of consumers. There should be no logical changes to individual consumer setup, just for-loops which expand the same setup to multiple consumers. 

The tests pass without error with these new setup refactors (and instantiating 5 consumer chains, apps, and paths). However, when the list contains more than one consumer, and there is more than one call to [SetupCCVChannel](https://github.com/cosmos/interchain-security/blob/e2e-multiple-consumers/tests/e2e/slashing.go#L43) for different consumer paths, the first relayed packet in `TestRelayAndApplySlashPacket` causes this error
```bash
failed to execute message; message index: 0: receive packet verification failed: packet source channel doesn't match the counterparty's channel (channel-1 ≠ channel-0): invalid packet
            	Test:       	TestCCVTestSuite/TestRelayAndApplySlashPacket
```
You can replicate this error by checking out `e2e-multiple-consumers` (which passes as committed) and adding this single line to `TestRelayAndApplySlashPacket`[here](https://github.com/cosmos/interchain-security/blob/e2e-multiple-consumers/tests/e2e/slashing.go#L43)

Line to add:
```go
s.SetupCCVChannel(s.consumerBundles["testchain3"].Path)
```
^ this sets up the second consumer path (chain ids start at "testchain2")

It seems there is some state that is being broken for the first consumer's `ibctesting.Path`, that is only caused by setting up the CCV channel for another consumer. This error could also be showing up if the generated packets from the provider CCV module are incorrectly referencing channel IDs for specific consumers. My hunch is that the problem lies in e2e setup and not the CCV module, but that hunch is unconfirmed. Any help is appreciated!

# TODOs

smarshall-spitzbart:
Closes #493 by referencing a new head of our [IBC fork](https://github.com/informalsystems/ibc-go/commits/interchain-security-2). See relevant IBC repo changes in https://github.com/informalsystems/ibc-go/pull/10

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -136,7 +136,7 @@ require (
 
 replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
-	github.com/cosmos/ibc-go/v3 => github.com/informalsystems/ibc-go/v3 v3.0.0-beta1.0.20221116144303-1e70792920f6
+	github.com/cosmos/ibc-go/v3 => github.com/informalsystems/ibc-go/v3 v3.0.0-beta1.0.20221121181828-c75c185d20b2
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/stretchr/testify => github.com/stretchr/testify v1.7.1
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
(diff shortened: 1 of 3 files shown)


mpoke (approved):
LGTM

MSalopek (approved):
LGTM!

I assume it was cleaner to exclude these changes from #498?

Issue #186, opened 2022-06-28
Fix invalid clientInfo and clientProposal keys


danwt:
The keys are wrong because in the call to Uint64 the `[1]`st entry is shorter than 64 bytes (shorter than 8 bytes).

These lines fail

https://github.com/cosmos/interchain-security/blob/335bd7c17fa5a9b33db25e44ddccfa899d675328/x/ccv/provider/keeper/proposal.go#L241-L245

https://github.com/cosmos/interchain-security/blob/335bd7c17fa5a9b33db25e44ddccfa899d675328/x/ccv/provider/keeper/proposal.go#L312-L316

due to impossible conversion

https://github.com/golang/go/blob/34f3ac5f165d50356d3a8940dc87b77e9b2b7fb9/src/encoding/binary/binary.go#L179

See

- https://github.com/cosmos/interchain-security/issues/165

danwt:
I was able to replicate using a fuzz generated test [here](https://gist.github.com/danwt/75657f06c17f36bdd5fc6c6319e1f963)

```go
func TestPendingClientKey(t *testing.T) {
	tests := []struct {
		name    string
		sec     int64
		nsec    int64
		chainId string
	}{
		{"malicious input", 83, 70, "foowiz"},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			ts := time.Unix(tt.sec, tt.nsec)
			suffixKey := PendingClientKey(ts, tt.chainId)
			splitKey := strings.Split(string(suffixKey), "/")
			binary.BigEndian.Uint64([]byte(splitKey[1]))
		})
	}
}
```

```
--- FAIL: TestPendingClientKey (0.00s)
    --- FAIL: TestPendingClientKey/malicious_input (0.00s)
panic: runtime error: index out of range [7] with length 5 [recovered]
	panic: runtime error: index out of range [7] with length 5

goroutine 23 [running]:
testing.tRunner.func1.2({0x1036d3460, 0x14000037110})
	/opt/homebrew/Cellar/go/1.18.3/libexec/src/testing/testing.go:1389 +0x1c8
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.18.3/libexec/src/testing/testing.go:1392 +0x384
panic({0x1036d3460, 0x14000037110})
	/opt/homebrew/Cellar/go/1.18.3/libexec/src/runtime/panic.go:838 +0x204
encoding/binary.bigEndian.Uint64(...)
[...]

PR #515, opened 2022-11-24
Key assignment

mpoke:
Closes https://github.com/cosmos/interchain-security/issues/26

Before merging:

- [x] Upgrade (state migration?) 
- [x] Fix current tests
- [x] provider genesis.go tests
- [x] Code to clean up key assignment state when a consumer chain is removed
- [x] E2e tests 
- [x] Integration with simple assign / query / test with invalid messages
  - [x] Happy path: Running chain, call key assign tx, check with both queries
  - [x] Assignment before a consumer chain is up
  - [x] Two validators assign same key
  - [x] Same validator assign same key
  - [x] Re-assign the keys of the entire valset at once (in one block). Can the client be updated afterwards?
- [ ] UTs for the other functions in assignment.go
  - [x] AssignConsumerKey
  - [x] ApplyKeyAssignment
  - [ ] GetProviderAddrFromConsumerAddr
  - [ ] PruneKeyAssignments
  - [ ] DeleteKeyAssignments
- [x] UTs for setters / getters in key_assignment.go 
  - [x] ValidatorConsumerPubKey
  - [x] ValidatorByConsumerAddr
  - [x] PendingKeyAssignment
  - [x] ConsumerAddrsToPrune


x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,619 @@
...
+		providerAddr := utils.TMCryptoPublicKeyToConsAddr(update.PubKey)
+		if consumerKey, found := k.GetValidatorConsumerPubKey(ctx, chainID, providerAddr); found {
+			newUpdates[i].PubKey = consumerKey
+			consumerAddr := utils.TMCryptoPublicKeyToConsAddr(consumerKey)
+			// set the mapping from this validator's new consensus address on the consumer
+			// to its consensus address on the provider
+			k.SetValidatorByConsumerAddr(ctx, chainID, consumerAddr, providerAddr)
+		}
+	}
+	return newUpdates
```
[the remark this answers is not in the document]
danwt: I think the difference is the call `k.SetValidatorByConsumerAddr(ctx, chainID, consumerAddr, providerAddr)`

I think there's a bug. Consider the following

A validator `V` assigns a consumer key `K` for a consumer chain which is not yet registered.
The initial validator set, when the consumer chain genesis is created does not contain `V`. 
The slash lookup assignment `K -> V` is never written.
A slash lookup returns `K->K` because `K->V` was never explicitly written.
`K` points to nothing so the validator can't be slashed.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,619 @@
...
+
+	// check whether the consumer chain is already registered,
+	// i.e., a client to the consumer was already created
+	if _, consumerRegistered := k.GetConsumerClientId(ctx, chainID); consumerRegistered {
+		// set the mapping from this validator's new consensus address on the consumer
+		// to its consensus address on the provider;
+		// otherwise, the mapping is added when the consumer is registered
+		// note: this state must be deleted through the pruning mechanism;
+		// see ConsumerValidatorsByVscID
+		k.SetValidatorByConsumerAddr(ctx, chainID, consumerAddr, providerAddr)
```
danwt, line 392: Bug: this is only called if the consumer is registered so it is possible for many validators to assign the same consumer key before the chain is registered.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,619 @@
...
+) (newUpdates []abci.ValidatorUpdate) {
+	newUpdates = updates
+	for i, update := range updates {
+		providerAddr := utils.TMCryptoPublicKeyToConsAddr(update.PubKey)
+		if consumerKey, found := k.GetValidatorConsumerPubKey(ctx, chainID, providerAddr); found {
+			newUpdates[i].PubKey = consumerKey
+			consumerAddr := utils.TMCryptoPublicKeyToConsAddr(consumerKey)
+			// set the mapping from this validator's new consensus address on the consumer
+			// to its consensus address on the provider
+			k.SetValidatorByConsumerAddr(ctx, chainID, consumerAddr, providerAddr)
```
danwt, line 456: Is this line the only reason to have a separate case for `ApplyKeyAssignmentToInitialValset`?

Given that there is already a pruning bug, can we design a solution doesn't require two methods?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -0,0 +1,73 @@
...
+	// TODO: In future, a mechanism will be added to limit assigning keys to chains
+	// which are approved or pending approval, only.
+	// Note that current attack potential is restricted because validators must sign
+	// the transaction, and the chainID size is limited.
+
+	providerAddr, err := sdk.ConsAddressFromBech32(msg.ProviderAddr)
+	if err != nil {
+		return nil, err
+	}
+
```
danwt, line 42: Why is this different from this version?

https://github.com/cosmos/interchain-security/blob/danwt/key-assignment-v4/x/ccv/provider/keeper/msg_server.go#L27-L39

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,619 @@
...
+
+	// Remove the pending key assignments
+	for _, addr := range addrToRemove {
+		k.DeletePendingKeyAssignment(ctx, chainID, addr)
+	}
+
+	// Remove the updates that need to be removed
+	for _, index := range updatesToRemove {
+		valUpdates = append(valUpdates[:index], valUpdates[index+1:]...)
+	}
```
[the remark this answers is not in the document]
mpoke: Nice catch.

x/ccv/provider/keeper/key_assignment.go
mpoke: `ApplyKeyAssignmentToInitialValset` is called from `MakeConsumerGenesis`, which may be called either from `BeginBlock` or `EndBlock`. The later means that the list of `PendingKeyAssignment`s is not empty, as the order of enblocks is `gov.EndBlock`, `staking.EndBlock`, `provider.EndBlock` (see app/provider/app.go). https://github.com/cosmos/interchain-security/issues/458 will probably solve this dependecy, but then by having a single `ApplyKeyAssignment` function, we will introduce another dependency on `EndBlock`s.

x/ccv/provider/keeper/msg_server.go
mpoke: Because the code is using `sdk.ConsAddress`, while `ValAddressFromBech32` returns a `sdk.ValAddress`, i.e.,
```golang
// ConsAddress defines a wrapper around bytes meant to present a consensus node.
// When marshaled to a string or JSON, it uses Bech32.
type ConsAddress []byte

// ValAddress defines a wrapper around bytes meant to present a validator's
// operator. When marshaled to a string or JSON, it uses Bech32.
type ValAddress []byte
```

x/ccv/provider/keeper/key_assignment.go
mpoke: Nice catch. See https://github.com/cosmos/interchain-security/pull/515/commits/86c597910b7ddd38a8ece96c7709813b3388591c

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,619 @@
...
+			// to enable multiple calls of AssignConsumerKey in the same block by the same validator
+			// the pending key assignment should not be overwritten
+			if _, found := k.GetPendingKeyAssignment(ctx, chainID, providerAddr); !found {
+				// store old key and power for modifying the valset update in EndBlock;
+				// note: this state is deleted at the end of the block
+				k.SetPendingKeyAssignment(
+					ctx,
+					chainID,
+					providerAddr,
+					abci.ValidatorUpdate{PubKey: oldConsumerKey, Power: oldPower},
```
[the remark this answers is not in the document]
mpoke: See https://github.com/cosmos/interchain-security/pull/515/commits/605d56264a1fbb4cec9c7b2c4a332d7dbcc8418f

x/ccv/provider/keeper/msg_server.go
sainoe: `msg.ProviderAddr` is  an operator address of type `ValAddress`
See
https://github.com/cosmos/interchain-security/blob/605d56264a1fbb4cec9c7b2c4a332d7dbcc8418f/x/ccv/provider/client/cli/tx.go#L83

So it shouldn't be converted to a `ConsAddress`

mpoke: Ah, didn't know that. I'll fix it.

mpoke: see https://github.com/cosmos/interchain-security/pull/515/commits/0c9f620017284fea541ac844167c8aaa2a69600a

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,618 @@
...
+
+// ApplyKeyAssignmentToValUpdates applies the key assignment to the validator updates
+// received from the staking module
+func (k Keeper) ApplyKeyAssignmentToValUpdates(
+	ctx sdk.Context,
+	chainID string,
+	valUpdates []abci.ValidatorUpdate,
+) (newUpdates []abci.ValidatorUpdate, err error) {
+	for _, valUpdate := range valUpdates {
+		providerAddr := utils.TMCryptoPublicKeyToConsAddr(valUpdate.PubKey)
```
MSalopek, line 480: If the provider did not assign a consumer key before the chain started and the chain has been running for some time, we also need to set the validator power to 0 for the provider's "normal" key.

Otherwise, the "normal" key stays in the validator set and the consumer key is added to the list.

From integration tests:
```bash
$: interchain-security-pd q tendermint-validator-set --node tcp://7.7.7.253:26658

block_height: "91"
total: "3"
validators:
// this is  the  "normal" key of the provider
- address: cosmosvalcons1qmq08eruchr5sf5s3rwz7djpr5a25f7xw4mceq
  proposer_priority: "-758"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: RrclQz9bIhkIy/gfL485g3PYMeiIku4qeo495787X10=
  voting_power: "511"
- address: cosmosvalcons1nx7n5uh0ztxsynn4sje6eyq2ud6rc6klc96w39
  proposer_priority: "379"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: mAN6RXYxSM4MNGSIriYiS7pHuwAcOHDQAy9/wnlSzOI=
  voting_power: "500"
- address: cosmosvalcons1ezyrq65s3gshhx5585w6mpusq3xsj3ayzf4uv6
  proposer_priority: "379"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: IHo4QEikWZfIKmM0X+N+BjKttz8HOzGs2npyjiba3Xk=
  voting_power: "500"


// check consumer valset after key assignment
$: interchain-security-cd q tendermint-validator-set --node tcp://7.7.8.253:26658

block_height: "65"
total: "4"
validators:
// NEW KEY -> pubkey assigned by provider
- address: cosmosvalcons16q4jna57d07rufql8qtzaqhwf3ud3hdf5yt3y2
  proposer_priority: "480"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: GJuUXISPjcWRIbEdzLTtVHzhnt9T98URH/gOA8KB7fA=
  voting_power: "511"
// OLD KEY -> the  "normal" key of the provider
- address: cosmosvalcons1qmq08eruchr5sf5s3rwz7djpr5a25f7xw4mceq
  proposer_priority: "-1500"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: RrclQz9bIhkIy/gfL485g3PYMeiIku4qeo495787X10=
  voting_power: "500"
- address: cosmosvalcons1nx7n5uh0ztxsynn4sje6eyq2ud6rc6klc96w39
  proposer_priority: "511"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: mAN6RXYxSM4MNGSIriYiS7pHuwAcOHDQAy9/wnlSzOI=
  voting_power: "500"
- address: cosmosvalcons1ezyrq65s3gshhx5585w6mpusq3xsj3ayzf4uv6
  proposer_priority: "511"
  pub_key:
    type: tendermint/PubKeyEd25519
    value: IHo4QEikWZfIKmM0X+N+BjKttz8HOzGs2npyjiba3Xk=
  voting_power: "500"
```

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -0,0 +1,494 @@
...
+	providerAddr   sdk.ConsAddress
+	consumerAddr   sdk.ConsAddress
+	consumerPubKey tmprotocrypto.PublicKey
+	valsetUpdate   abci.ValidatorUpdate
+	vscID          uint64
+}
+
+func TestValidatorConsumerPubKeyCRUD(t *testing.T) {
+	chainID := "consumer"
+	providerAddr := sdk.ConsAddress([]byte("providerAddr"))
```
danwt, line 25: You should use https://github.com/cosmos/interchain-security/blob/main/testutil/crypto/crypto.go to get test data

danwt: E.g. https://github.com/cosmos/interchain-security/blob/095a0760566dc795b4b840deb5cbe189891d10c0/testutil/crypto/crypto.go#L94-L96

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -0,0 +1,494 @@
...
+	require.Empty(t, consumerPubKey, "consumer pubkey was returned")
+	require.NotEqual(t, consumerPubKey, consumerKey)
+}
+
+func TestIterateValidatorConsumerPubKeys(t *testing.T) {
+
+	chainID := "consumer"
+	testAssignments := []testAssignment{
+		{
+			providerAddr:   sdk.ConsAddress([]byte("validator-1")),
```
danwt, line 50: It's better to use https://github.com/cosmos/interchain-security/blob/095a0760566dc795b4b840deb5cbe189891d10c0/testutil/crypto/crypto.go#L94-L96

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -0,0 +1,494 @@
...
+			consumerPubKey: cryptotestutil.NewCryptoIdentityFromIntSeed(1).TMProtoCryptoPublicKey(),
+		},
+		{
+			providerAddr:   sdk.ConsAddress([]byte("validator-2")),
```
danwt, line 54: ditto - and in several other places

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -0,0 +1,494 @@
...
+
+	keeper.IterateValidatorConsumerPubKeys(ctx, chainID, cbIterateOne)
+	require.Len(t, result, 1, "incorrect result len - should be 1, got %d", len(result))
+
+	require.Equal(t, testAssignments[0], result[0], "mismatched consumer key assignment in iterate one")
+
+}
+
+func TestIterateAllValidatorConsumerPubKeys(t *testing.T) {
+	providerAddr := sdk.ConsAddress([]byte("validator-1"))
```
danwt, line 103: ditto - use testutil

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -0,0 +1,494 @@
...
+	keeper.IterateAllValidatorConsumerPubKeys(ctx, cbIterateOne)
+	require.Len(t, result, len(testAssignments), "incorrect result len - should be 1, got %d", len(result))
+
+	require.Equal(t, testAssignments[0], result[0], "mismatched consumer key assignment in iterate one")
+}
+
+func TestValidatorByConsumerAddrCRUD(t *testing.T) {
+	chainID := "consumer"
+	providerAddr := sdk.ConsAddress([]byte("providerAddr"))
+	consumerAddr := sdk.ConsAddress([]byte("consumerAddr"))
```
danwt, line 167: ditto

x/ccv/provider/types/keys_test.go
```diff
@@ -124,43 +130,75 @@ func TestChainIdAndVscIdAndParse(t *testing.T) {
...
+		// Expected bytes = prefix + chainID length + chainID + consAddr bytes
+		expectedLen := 1 + 8 + len(test.chainID) + len(test.addr)
+		require.Equal(t, expectedLen, len(key))
+		parsedID, parsedConsAddr, err := types.ParseChainIdAndConsAddrKey(test.prefix, key)
+		require.Equal(t, test.chainID, parsedID)
+		require.Equal(t, test.addr, parsedConsAddr)
+		require.NoError(t, err)
+	}
+}
+
```
danwt, line 176: We could replace the custom <chainID, consAddr> pair implementation with a proto struct which we works, and we wouldn't have to test. Just a thought.

x/ccv/provider/keeper/key_assignment.go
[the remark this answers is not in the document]
danwt: That commit changes it to a better name indeed but I still think using abci.ValidatorUpdate is misleading. Just passing and receiving the two arguments is clearer IMO.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,618 @@
...
+		})
+
+		newConsumerKey, found := k.GetValidatorConsumerPubKey(ctx, chainID, pAddr)
+		if !found {
+			err = sdkerrors.Wrapf(types.ErrConsumerKeyNotFound, "consumer key not found for %s", pAddr)
+			return true
+		}
+		newUpdates = append(newUpdates, abci.ValidatorUpdate{
+			PubKey: newConsumerKey,
+			Power:  pendingKeyAssignment.Power,
```
danwt, line 539: This is using the old name ` pendingKeyAssignment`

x/ccv/provider/keeper/key_assignment.go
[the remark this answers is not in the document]
mpoke: Ah, you are talking about the function params, not the proto data structure use for marshaling, right?

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,618 @@
...
+			panic(err)
+		}
+		stop := cb(chainID, vscID, consumerAddrsToPrune.Addresses)
+		if stop {
+			break
+		}
+	}
+}
+
+// DeleteConsumerAddrsToPrune deletes the list of consumer addresses mapped to a given VSC ID
```
danwt, line 360: In this docstring, or somewhere else, the pruning invariant should be written. In other terms, when _exactly_ is it possible to get a provider addr from a given consumer addr using `GetValidatorByConsumerAddr` . The test can then be phrased similarly to how it was before

https://github.com/cosmos/interchain-security/blob/4743b4dbb7765647692574d3320b07806b51060c/x/ccv/provider/keeper/key_assignment_test.go#L329-L348

x/ccv/provider/keeper/key_assignment.go
danwt: I'm not sure I follow, `MakeConsumerGenesis` is only ever called after `GetConsumerClientId` returns false in `CreateConsumerClient` so when `MakeConsumerGenesis` is called there can never be any `Replacement`s (Pendings) when `MakeConsumerGenesis` is called.

Or not?

x/ccv/provider/keeper/key_assignment.go
danwt: What about

'
For all consumer addr CK in ValidatorByConsumerAddr:
  either (
  (
    there exists a provider addr PK in ValidatorConsumerPubKey st. ValidatorConsumerPubKey(PK) = CK
   ) or
  (
   there exists a vscid VID in ConsumerAddrsToPrune st. CK in ConsumerAddrsToPrune(VID)
  )
)
'

x/ccv/provider/keeper/key_assignment.go
mpoke: That's true. Didn't think of that. Okay, I'll get rid of `ApplyKeyAssignmentToInitialValset` then.

x/ccv/provider/keeper/key_assignment_test.go
MSalopek: In some places it was used, in others not.

In some places, `sdk.ConsAddress` was used so to make the tests easier to follow (mostly for brevity).

It was
`cryptotestutil.NewCryptoIdentityFromBytesSeed([]byte("validator-1")).SDKConsAddress()`

vs just

`sdk.ConsAddress([]byte("validator-1"))`

mpoke: @danwt What's the benefit from using crypto.go here?

x/ccv/provider/types/keys_test.go
mpoke: Can you use a proto struct as a key in the SDK store? If yes, this will be useful to a lot of code (not just here). We can open a refactoring issue.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,618 @@
...
+func (k Keeper) ApplyKeyAssignmentToInitialValset(
+	ctx sdk.Context,
+	chainID string,
+	updates []abci.ValidatorUpdate,
+) (newUpdates []abci.ValidatorUpdate) {
+	newUpdates = updates
+	for i, update := range updates {
+		providerAddr := utils.TMCryptoPublicKeyToConsAddr(update.PubKey)
+		if consumerKey, found := k.GetValidatorConsumerPubKey(ctx, chainID, providerAddr); found {
+			newUpdates[i].PubKey = consumerKey
```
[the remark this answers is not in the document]
mpoke: I removed the entire function.

x/ccv/provider/keeper/key_assignment.go
mpoke: Removed.

x/ccv/provider/keeper/key_assignment.go
mpoke: I added it to `AppendConsumerAddrsToPrune`

x/ccv/provider/keeper/key_assignment.go
mpoke: We already do that, see https://github.com/cosmos/interchain-security/blob/c396b3f704842b195a318f5e871fecd3e5b142da/x/ccv/provider/keeper/key_assignment.go#L422

This key is added to the `KeyAssignmentReplacement` here https://github.com/cosmos/interchain-security/blob/c396b3f704842b195a318f5e871fecd3e5b142da/x/ccv/provider/keeper/key_assignment.go#L432 and then used here https://github.com/cosmos/interchain-security/blob/c396b3f704842b195a318f5e871fecd3e5b142da/x/ccv/provider/keeper/key_assignment.go#L483

Can you output the VSCPacket sent?

x/ccv/provider/keeper/key_assignment_test.go
danwt: I just find it weird 

```
providerAddr:   sdk.ConsAddress([]byte("string")),
consumerPubKey: cryptotestutil.NewCryptoIdentityFromIntSeed(y).TMProtoCryptoPublicKey(),
```

instead of e.g.

```
providerAddr: cryptotestutil.NewCryptoIdentityFromIntSeed(x).SDKConsAddress(),
consumerPubKey: cryptotestutil.NewCryptoIdentityFromIntSeed(y).TMProtoCryptoPublicKey(),
```

But it's my opinion and not a big deal

x/ccv/provider/types/keys_test.go
danwt: Well of course you have to convert to bytes first. But the byte conversion is trivial and done for you without mistakes. We are just writing our own conversions everywhere which have caused bugs e.g

https://github.com/cosmos/interchain-security/issues/186

danwt: It would be cool if the sdk could do the .bytes() call, or whatever, for you.

x/ccv/provider/keeper/key_assignment.go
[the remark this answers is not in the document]
danwt: Yeah because that's the part you see when you read the business logic 👍

x/ccv/provider/client/handler.go
```diff
@@ -0,0 +1,24 @@
+package client
```
danwt, line 1: Why is this file and and the _test in provider/client? In the sdk they are in module directories. So it would be in provider/ here.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,621 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return consumerAddrsToPrune.Addresses
+}
+
+func (k Keeper) IterateConsumerAddrsToPrune(
+	ctx sdk.Context,
+	chainID string,
+	cb func(vscID uint64, consumerAddrsToPrune [][]byte) (stop bool),
```
danwt, line 326: `consumerAddrsToPrune` should be a concrete type or alias, not `[][]byte`

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,621 @@
...
+		stop := cb(vscID, consumerAddrsToPrune.Addresses)
+		if stop {
+			break
+		}
+	}
+}
+
+func (k Keeper) IterateAllConsumerAddrsToPrune(
+	ctx sdk.Context,
+	cb func(chainID string, vscID uint64, consumerAddrsToPrune [][]byte) (stop bool),
```
danwt, line 351: ditto: `consumerAddrsToPrune` should be a concrete type or alias, not `[][]byte`

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -0,0 +1,621 @@
...
+			panic(err)
+		}
+		stop := cb(vscID, consumerAddrsToPrune.Addresses)
+		if stop {
+			break
+		}
+	}
+}
+
+func (k Keeper) IterateAllConsumerAddrsToPrune(
```
danwt, line 349: This should either
Option A. Take a list of chains and use `IterateConsumerAddrsToPrune`
Option B. Document the reason why A is not done.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -71,3 +72,55 @@ func (k Keeper) QueryConsumerChainStops(goCtx context.Context, req *types.QueryC
...
+func (k Keeper) QueryValidatorConsumerAddr(goCtx context.Context, req *types.QueryValidatorConsumerAddrRequest) (*types.QueryValidatorConsumerAddrResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if _, found := k.GetConsumerClientId(ctx, req.ChainId); !found {
+		return nil, types.ErrUnknownConsumerChainId
+	}
```
danwt, line 85: As discussed on call, this check should be removed because keys should be assignable/queryable before chain addition.

x/ccv/provider/keeper/hooks.go
```diff
@@ -65,8 +67,33 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) error {
...
+		consumerKey tmprotocrypto.PublicKey,
+	) (stop bool) {
+		if providerAddr.Equals(valConsAddr) {
+			toDelete = append(toDelete, StoreKey{ChainID: chainID, ProviderAddr: providerAddr})
+		}
+		return false // do not stop
+	})
+	for _, key := range toDelete {
+		consumerKey, found := h.k.GetValidatorConsumerPubKey(ctx, key.ChainID, key.ProviderAddr)
+		if found {
```
danwt, line 89: Why is consumerKey queried if it was already known on line 80?

x/ccv/provider/keeper/key_assignment_test.go
danwt: @MSalopek if these nits are 'wont-fix' could you please resolve the comment boxes? Thank you!

x/ccv/provider/types/keys_test.go
mpoke: I'm resolving this conversation as it is outside the scope of this PR.

MSalopek:
> @MSalopek Hey, do you think you could please look at the failing gosec when you have time? It is about integration tests
> 
> ```shell
> [/github/workspace/tests/integration/state.go:285-292] - G204 (CWE-78): Subprocess launched with a potential tainted input or cmd arguments (Confidence: HIGH, Severity: MEDIUM)
>     284: 	}
>   > 285: 	bz, err := exec.Command("docker", "exec", tr.containerConfig.instanceName, tr.chainConfigs[chain].binaryName,
>   > 286: 
>   > 287: 		"query", "bank", "balances",
>   > 288: 		valDelAddress,
>   > 289: 
>   > 290: 		`--node`, tr.getQueryNode(chain),
>   > 291: 		`-o`, `json`,
>   > 292: 	).CombinedOutput()
>     293: 
> ```
> 
> Thank you!

That's a false positive. Patch was in `goc-december` branch and release. It will be included into this branch shortly

docs/architecture/adr-001-key-assignment.md
```diff
@@ -0,0 +1,267 @@
+# ADR 001: Key Assignment
```
danwt, line 1: What's the purpose of this ADR, or ADRs in our repo in general? This is from a template, which is also new. Where did it come from? Different organisations have different ideas of ADRs.

The general breakdown into subheading ('On receiving a `VSCMatured`:', 'On stopping a consumer chain:' ect) is good, and the consequences and other written sections are good.

The part that is not clear to me is if the parts which are almost all code snippets are useful. I have questions. Should this be maintained? Who is the intended audience? Why would they prefer to read this instead of code? (Especially if we directed the energy away from this towards making the code more readable)?

I am still left with questions, which I believe an ADR should at least try to answer. What were the other alternative approaches to this solution that were considered and why were they not taken? What business or product goal does this solve? What security concerns arise from this feature? What testing concerns arise?

mpoke: > What's the purpose of this ADR, 

I needed a place where to add the main logic behind the key assignment design. Having a random file for each feature doesn't seem like a good practice. 

> or ADRs in our repo in general?

ADRs should describe features that are added to the repo. Unlike the spec, the ADRs should address the dependencies to Cosmos SDK and Tendermint.

>  Where did it come from? Different organisations have different ideas of ADRs.

https://github.com/cosmos/ibc-go/tree/main/docs/architecture

> Should this be maintained?

Yes. 

> Who is the intended audience?

Engineers.

> Why would they prefer to read this instead of code?

The implementation contains a lot of boiler plate code that is hard to parse for somebody not used to the SDK. We need a more concise description of a feature. 

> I am still left with questions, which I believe an ADR should at least try to answer. What were the other alternative approaches to this solution that were considered and why were they not taken?

Had we had ADRs before, every approach taken but not finalized would have been its own ADR. Or the same ADR with a commit history. Not sure that it makes sense to consumer resources to document design decisions not taken. 

> What business or product goal does this solve? What security concerns arise from this feature? What testing concerns arise?

Feel free to expand the ADR with all this info if you feel it's needed. Even better, open an issue beforehand.

danwt:
In my view the repo is at a point sufficient to merge into main/. Should we open the PR and request another round of reviews?

jtremback (approved):

MSalopek (approved):
Left some comments but I approve.

There are no logic issues at this point in time, any small issues that arise can be fixed in separate PRs.

x/ccv/consumer/keeper/genesis.go
```diff
@@ -197,30 +188,3 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) (genesis *consumertypes.GenesisSt
 
 	return
 }
-
-// verifyGenesisInitValset verifies the latest consensus state on provider client matches
-// the initial validator set of restarted chain thus
-func (k Keeper) verifyGenesisInitValset(ctx sdk.Context, genState *consumertypes.GenesisState) error {
-
-	consState, ok := k.clientKeeper.GetLatestClientConsensusState(ctx, genState.ProviderClientId)
```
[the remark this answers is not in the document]
mpoke: I think this is something that should be addressed by validation.

x/ccv/provider/keeper/hooks.go
```diff
@@ -62,11 +64,74 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) error {
 	return nil
 }
 
-// Define unimplemented methods to satisfy the StakingHooks contract
+func ValidatorConsensusKeyInUse(k *Keeper, ctx sdk.Context, valAddr sdk.ValAddress) bool {
+	// Find the validator being added in the staking module
+	// This is necessary because only the operator address is received
+	// as argument at it is not possible to directly query the validator using
```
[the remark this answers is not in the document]
danwt: Damn my bad! Left that comment but it should be deleted. Thanks for pointing it out.

PR #520, opened 2022-11-25
Danwt/key assignment unit tests

danwt:
Adds some unit tests and random testing. It's a bit unfinished, the slash query isn't properly checked yet and the code is unpolished. But it's something.

Tests:

- assignment specific cases
- random assignents interleaved with applying mapping to random updates
- pruning property
- validator set replication

Does not test:

- Reverse lookup for slashing


danwt:
Will be superseded

PR #542, opened 2022-12-02
Separate validation and handling for recv slash packets

### LINKED ISSUE #414: Be more precise about what kinds of consumer initated slash requests should trigger an error and channel close
# Problem

There is some divergence between the spec and the code, and the code seems a bit suspicous w.r.t when it raises an error in `HandleSlashPacket`.

# Closing criteria

Add clarity by documenting or changing the code and defining in the spec precise rules about what should raise an error.

# Detail

Spec

https://github.com/cosmos/ibc/blob/c919d8ec8bec086be00b68d6f0e9ca4b0a0b8c5e/spec/app/ics-028-cross-chain-validation/methods.md?plain=1#L1667-L1704

Code

https://github.com/cosmos/interchain-security/blob/3f9100acd62a0350432a1a87c7caf06efadfd614/x/ccv/provider/keeper/relay.go#L206-L279

The code returns `success,err=(bool,error)` and it's not really clear when exactly an error should be returned. Note that a non-`nil` error will close the channel.

I think we could make the function return `closeChannel,err=(bool,error)` and then an error should be created whenever there is any problem preventing slashing, and closeChannel should be true only when the channel should be closed.

# TODOs

smarshall-spitzbart:
# Description
This PR separates out validation and handling for slash packets received on the provider. This allows us to separate out the concerns of
1. At validation time, we return an IBC err ack immediately upon failure
2. At handling time, we stop handling (no-op), or panic the binary upon system critical failures like marshalling. No errors are returned

No functionality should be changed in this PR, but the refactors are needed to properly implement #462 

Note: VSCMatured packet handling is refactored in this PR as well (just adds an explicit handler method). Jehan and I decided it doesn't make sense to have a validate method for VSCMatured packets since there's not a case in which to return an IBC err ack

# Linked issues
Closes #414 

# Type of change

- [x] Non-breaking changes
- [x] Updates in store keepers or store keys
- [x] Testing

# How was the feature tested?

- [x] Unit tests
- [x] E2E tests

# Other information
Changes in functionality are reserved for #546 and https://github.com/cosmos/interchain-security/issues/547

# Checklist:

- [x] Other PRs depend on this feature: `#<issue>`
- [x] testutil/e2e/debug_test.go is up-to-date with additional or changed e2e test names


tests/e2e/slashing.go
```diff
@@ -255,29 +254,69 @@ func (suite *CCVTestSuite) TestHandleSlashPacketDoubleSigning() {
 	suite.Require().True(signingInfo.Tombstoned)
 }
 
-// TestHandleSlashPacketErrors tests errors for the HandleSlashPacket method in an e2e testing setting
-func (suite *CCVTestSuite) TestHandleSlashPacketErrors() {
+// TestOnRecvSlashPacketErrors tests errors for the OnRecvSlashPacket method in an e2e testing setting
+func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
```
smarshall-spitzbart, line 258: We could delete this test altogether in favor of the new unit tests that are more complete. This was my attempt at salvaging this test where possible

x/ccv/provider/keeper/relay.go
```diff
@@ -56,23 +71,16 @@ func (k Keeper) OnRecvVSCMaturedPacket(
 			// Delete unbonding op
 			k.DeleteUnbondingOp(ctx, unbondingOp.Id)
 		} else {
-			if err := k.SetUnbondingOp(ctx, unbondingOp); err != nil {
```
smarshall-spitzbart, line 59: No functionality change, panics were moved into setters

x/ccv/provider/keeper/relay.go
```diff
@@ -215,55 +223,63 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
...
-	var found bool
-	if data.ValsetUpdateId == 0 {
-		infractionHeight, found = k.GetInitChainHeight(ctx, chainID)
-	} else {
-		infractionHeight, found = k.GetValsetUpdateBlockHeight(ctx, data.ValsetUpdateId)
-	}
+// validateSlashPacket validates a recv slash packet before it is
+// handled or persisted in store. An error is returned if the packet is invalid,
+// and an error ack should be relayed to the sender.
+func (k Keeper) validateSlashPacket(ctx sdk.Context,
```
smarshall-spitzbart, line 241: This is where the new check will be added to address #546

x/ccv/provider/keeper/relay.go
```diff
@@ -286,11 +302,13 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 		slashFraction = k.slashingKeeper.SlashFractionDoubleSign(ctx)
 		jailTime = evidencetypes.DoubleSignJailEndTime
 		k.slashingKeeper.Tombstone(ctx, consAddr)
-	default:
```
smarshall-spitzbart, line 289: Default case removed since infraction type is checked in validation method

x/ccv/provider/keeper/relay.go
```diff
@@ -34,14 +34,30 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
+func (k Keeper) ValidateVSCMaturedPacket(ctx sdk.Context,
+	packet channeltypes.Packet, data ccv.VSCMaturedPacketData) error {
+
+	// check that a ccv channel is established via the dest channel of the recv packet
+	_ = k.getChainIdOrPanic(ctx, packet)
+
+	return nil
+}
+
+// TODO: Unit test this method.
```
smarshall-spitzbart, line 59: Leaving as a TODO because I'm not very familiar with expected behavior

x/ccv/provider/keeper/relay.go
```diff
@@ -34,14 +34,30 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
 ) exported.Acknowledgement {
-	// check that the channel is established
-	chainID, found := k.GetChannelToChain(ctx, packet.DestinationChannel)
-	if !found {
-		// VSCMatured packet was sent on a channel different than any of the established CCV channels;
-		// this should never happen
-		panic(fmt.Errorf("VSCMaturedPacket received on unknown channel %s", packet.DestinationChannel))
+
+	if err := k.ValidateVSCMaturedPacket(ctx, packet, data); err != nil {
+		return channeltypes.NewErrorAcknowledgement(err.Error())
```
jtremback, line 39: This never happens

x/ccv/provider/keeper/relay.go
```diff
@@ -273,6 +290,12 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 		slashFraction sdk.Dec
 	)
 
+	// Obtain infraction height or panic
+	infractionHeight, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId)
+	if !found {
+		panic("infraction height not found. But was found during slash packet validation")
```
jtremback, line 296: This will panic if someone forgets to clean up the slash packet queues

smarshall-spitzbart: now just a log 👍

x/ccv/provider/keeper/relay.go
```diff
@@ -213,57 +222,65 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
...
-	var found bool
-	if data.ValsetUpdateId == 0 {
-		infractionHeight, found = k.GetInitChainHeight(ctx, chainID)
-	} else {
-		infractionHeight, found = k.GetValsetUpdateBlockHeight(ctx, data.ValsetUpdateId)
-	}
+// validateSlashPacket validates a recv slash packet before it is
+// handled or persisted in store. An error is returned if the packet is invalid,
+// and an error ack should be relayed to the sender.
+func (k Keeper) ValidateSlashPacket(ctx sdk.Context,
```
smarshall-spitzbart, line 242: This is where the new check will be added to address https://github.com/cosmos/interchain-security/issues/546

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: True, I can remove the unused path if https://github.com/cosmos/interchain-security/pull/542#discussion_r1038339867 should not be validated

x/ccv/provider/keeper/relay.go
```diff
@@ -385,3 +404,26 @@ func (k Keeper) EndBlockCCR(ctx sdk.Context) {
...
+	if valsetUpdateID == 0 {
+		return k.GetInitChainHeight(ctx, chainID)
+	} else {
+		return k.GetValsetUpdateBlockHeight(ctx, valsetUpdateID)
+	}
+}
+
+// getChainIdOrPanic returns the chainID from a recv packet,
+// or panics if the packet was sent on a different channel than any of the established CCV channels.
+func (k Keeper) getChainIdOrPanic(ctx sdk.Context, packet channeltypes.Packet) string {
```
jtremback, line 421: I'd just get rid of this method by getting the chain id and explicitly panicing at the top of `OnRecvSlashPacket` and `OnRecvVSCMaturedPacket` and then passing it into where it is needed later in those functions.

jtremback (approved):
LGTM except for the refactor getting rid of the getChainIdOrPanic method. But @mpoke really needs to review it.

x/ccv/provider/keeper/relay.go
danwt: Beautiful!

x/ccv/provider/keeper/relay.go
```diff
@@ -213,57 +216,68 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
...
 	validator, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, consAddr)
 
 	// make sure the validator is not yet unbonded;
 	// stakingKeeper.Slash() panics otherwise
 	if !found || validator.IsUnbonded() {
-		// TODO add warning log message
-		// fmt.Sprintf("consumer chain %s trying to slash unbonded validator %s", chainID, consAddr.String())
-		return false, nil
+		// if validator is not found or is unbonded drop slash packet and log error
+		k.Logger(ctx).Error("validator not found or is unbonded", "validator", data.Validator.Address)
```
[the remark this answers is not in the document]
danwt: I'm wondering if we should maybe remove the consumer in this case? @jtremback @mpoke What do you think?

My reasoning is: the provider or the consumer is incorrect. If the provider is incorrect then we should kill the provider and all the consumers. If the consumer is incorrect we should kill it. In both cases the consumer is killed.

testutil/keeper/expectations.go
```diff
@@ -90,6 +91,63 @@ func GetMocksForStopConsumerChain(ctx sdk.Context, mocks *MockedKeepers) []*gomo
...
+	// Different calls are made depending on the type infraction.
+	if expectedPacketData.Infraction == stakingtypes.Downtime {
+		calls = append(calls,
+			mocks.MockSlashingKeeper.EXPECT().SlashFractionDowntime(ctx).Return(sdk.NewDec(1)).Times(1),
+			mocks.MockSlashingKeeper.EXPECT().DowntimeJailDuration(ctx).Return(time.Hour).Times(1),
+		)
+
+	} else if expectedPacketData.Infraction == stakingtypes.DoubleSign {
+		calls = append(calls,
+			mocks.MockSlashingKeeper.EXPECT().SlashFractionDoubleSign(ctx).Return(sdk.NewDec(1)).Times(1),
```
danwt, line 118: Nit: it might be nice to move the hardcoded NewDec(1)'s to the top of the function and caps name them.

x/ccv/provider/keeper/relay.go
```diff
@@ -34,14 +34,24 @@ func (k Keeper) OnRecvVSCMaturedPacket(
 	packet channeltypes.Packet,
 	data ccv.VSCMaturedPacketData,
 ) exported.Acknowledgement {
-	// check that the channel is established
+
+	// check that the channel is established, panic if not
```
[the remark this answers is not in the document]
mpoke: I disagree. The comment is for people to understand that when `ChannelToChain` is set, then the channel is established.

x/ccv/provider/keeper/relay.go
```diff
@@ -34,14 +34,24 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
 	chainID, found := k.GetChannelToChain(ctx, packet.DestinationChannel)
 	if !found {
 		// VSCMatured packet was sent on a channel different than any of the established CCV channels;
 		// this should never happen
 		panic(fmt.Errorf("VSCMaturedPacket received on unknown channel %s", packet.DestinationChannel))
 	}
 
+	// Note: no validation is needed (that'd return an IBC err ack) for recv VSCMatured packets,
+	// since the packet data only includes a valset update id that can take any uint64 value.
```
mpoke, line 47: Actually it cannot be `0`, but probably it's not worth checking this case. 

More relevant would be to check whether subsequent `VSCMaturedPacket`s have increasing `vscID`s. Otherwise, a malicious consumer could send a lot of `VSCMaturedPacket`s  with random `vscID`s (or `0`), which means that relayers will end up relaying garbage without even realizing it (no err ack). 

We could actually check if every `VSCMaturedPacket` has the next expected `vscID` (i.e., the next `vscID` in `UnbondingOpsFromIndex`).

x/ccv/provider/keeper/relay.go
```diff
@@ -34,14 +34,24 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
+
+	k.HandleVSCMaturedPacket(ctx, chainID, data)
+	return channeltypes.NewResultAcknowledgement([]byte{byte(1)})
+}
+
+// TODO: Unit test this method.
+func (k Keeper) HandleVSCMaturedPacket(ctx sdk.Context, chainID string, data ccv.VSCMaturedPacketData) {
```
mpoke, line 54: Add a comment describing in what conditions is this function panicking.

x/ccv/provider/keeper/relay.go
```diff
@@ -213,57 +216,68 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
 	k.SetValsetUpdateBlockHeight(ctx, valUpdateID, uint64(ctx.BlockHeight()+1))
 }
 
-// OnRecvSlashPacket slashes and jails the given validator in the packet data
+// OnRecvSlashPacket receives a slash packet, validates it, then handles it if valid.
```
mpoke, line 219: nit: the receive is done in the IBC module (hence the name, `OnRecv...`)
```suggestion
// OnRecvSlashPacket delivers a received slash packet: validates it and then handles it if valid.
```

x/ccv/provider/keeper/relay.go
```diff
@@ -213,57 +216,68 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
...
 func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, data ccv.SlashPacketData) exported.Acknowledgement {
-	// check that the channel is established
+
+	// check that the channel is established, panic if not
 	chainID, found := k.GetChannelToChain(ctx, packet.DestinationChannel)
 	if !found {
 		// SlashPacket packet was sent on a channel different than any of the established CCV channels;
 		// this should never happen
 		panic(fmt.Errorf("SlashPacket received on unknown channel %s", packet.DestinationChannel))
```
mpoke, line 227: I think we should return an err ack here.

x/ccv/provider/keeper/relay.go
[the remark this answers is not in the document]
mpoke: I suggest leaving the error for now. A consumer can always be remove via governance.

x/ccv/provider/keeper/relay.go
```diff
@@ -213,57 +216,68 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
...
+		// if validator is not found or is unbonded drop slash packet and log error
+		k.Logger(ctx).Error("validator not found or is unbonded", "validator", data.Validator.Address)
+		return
 	}
 
-	// tombstoned validators should not be slashed multiple times
+	// tombstoned validators should not be slashed multiple times.
 	if k.slashingKeeper.IsTombstoned(ctx, consAddr) {
-		return false, nil
+		// Drop packet if validator is tombstoned.
```
mpoke, line 279: Maybe log an info message here. See https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/x/slashing/keeper/infractions.go#L117 and https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/x/evidence/keeper/infraction.go#L79.

mpoke (changes requested):
Nice work @smarshall-spitzbart. See my comments below.

x/ccv/provider/keeper/relay.go
mpoke: > We could actually check if every VSCMaturedPacket has the next expected vscID (i.e., the next vscID in UnbondingOpsFromIndex).

This is actually incorrect since some `VSCPackets` can be sent without any unbonding operations.

smarshall-spitzbart: This point will be superseeded by #568, I've added a comment in that issue specifically about validating VSCMatured packets

x/ccv/provider/keeper/hooks.go
```diff
@@ -46,12 +46,11 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) error {
...
-	if err := h.k.SetUnbondingOp(ctx, unbondingOp); err != nil {
-		// If there was an error persisting the unbonding op, panic to end execution for
-		// the current tx and prevent committal of this invalid state.
-		panic(fmt.Errorf("unbonding op could not be persisted: %w", err))
-	}
+	// Set unbonding op.
+	// If there is an error persisting the unbonding op, the method will
+	// panic to end execution for the current tx and prevent committal
+	// of this invalid state.
+	h.k.SetUnbondingOp(ctx, unbondingOp)
```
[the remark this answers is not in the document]
smarshall-spitzbart: comment was requested in previous PR, agreed and removed

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: Maybe we can add a comment to #568, but I'd like to avoid functionality changes in this PR as much as possible

x/ccv/provider/keeper/relay.go
```diff
@@ -273,6 +287,13 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 		slashFraction sdk.Dec
 	)
 
+	// Obtain infraction height, or log error and drop packet
```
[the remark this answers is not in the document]
smarshall-spitzbart: Yea this code is self explanatory 👍, removed

mpoke (approved):
LGTM

Issue #514, opened 2022-11-24
Run happy path integration tests only on push/merge to main


MSalopek:
# Problem
Integration tests are run for all PRs, and on each push to an opened PR. This results in long pipeline run times which may not be desireable.

Since we have significant e2e and unittest coverage it would be beneficial to run integration tests once the PR is actually merged into the main branch. That way we can have long running (> 20 minutes) and more complex tests without any time constraints. This would turn integration tests into more of a simulation that we run on each merge to main to confirm expected behaviour in semi-real world scenarios.

We could set it up so that failure to execute integration tests results in some emails being sent out to maintainers and code owners, just so we are aware that something is wrong.

It is still expected that integration tests pass for each PR. This can be solved by running a subset of the integration tests for i.e. running only happy path steps.

It would be best to setup some workflows that can be run manually and have a `nightly` test run which will run after everyone finishes their work for the day. The `nightly` run should run all tests and pull from `main` branch

# Closing criteria
Integration tests are running on push/merge to main instead of each PR.

# TODOs
- [x] Labels have been added for issue
- [ ] Issue has been added to the ICS project

smarshall-spitzbart:
Along with what's already mentioned, the important thing is that integration tests are not commented out/skipped during the CI process. This is currently the case AFAIK

PR #543, opened 2022-12-02
Deduplicate code in IncrementValidatorSetUpdateId

danwt:
Came across this. Only took 2 mins to open PR

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -517,17 +517,8 @@ func (k Keeper) chanCloseInit(ctx sdk.Context, channelID string) error {
 }
 
 func (k Keeper) IncrementValidatorSetUpdateId(ctx sdk.Context) {
-	store := ctx.KVStore(k.storeKey)
-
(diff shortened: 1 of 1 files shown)


mpoke (approved):

jtremback:
Try re running the integration tests locally. If they pass there, we can force merge.

mpoke:
> Hmm it's strange that this fails integration tests. Do you have any idea what might be up @MSalopek ?

I think this is a known issue. IIUC, we are running to many nodes in parallel on shitty hardware and things are timing out.  This issue is related https://github.com/cosmos/interchain-security/issues/514. @MSalopek could probably give more details.

MSalopek:
> Hmm it's strange that this fails integration tests. Do you have any idea what might be up @MSalopek ?

It's resource related

Runing on basic github runner instances causes some tests to fail. Main issues are either a validator missing blocks or proposals not passing.

We run tests in parallel and these machines are not handling them too well. I had already remove some test and merged them into the happy path to make the tests pass.

MSalopek (approved):
LGTM!

Issue #468, opened 2022-11-11
Inconsistent handling of IBC timeouts on consumer


andrey-kuprianov:
_Surfaced from @informalsystems audit of [Interchain Security](https://github.com/cosmos/interchain-security/) at commit 6c0d718_

# Problem

[ccv/consumer/ibc_module.go: OnTimeoutPacket ()](https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/consumer/ibc_module.go#L263-L286) assumes that only a timeout for `SlashPacketData` can be received, while in reality CCV Consumer also sends `VSCMaturedPacketData` packets; so on receiving a timeout of this packet the function will return an error. On the other hand, for  `SlashPacketData`  the processing goes to [ccv/consumer/keeper/relay.go: OnTimeoutPacket()](https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/consumer/keeper/relay.go#L239-L241), which returns `nil`. This is inconsistent wrt. handling of timeouts of different packet types.

# Closing criteria

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

PR #550, opened 2022-12-02
make OnTimeoutPacket handling consistent in consumer

MSalopek:
# Description
`ccv/consumer/ibc_module.go: OnTimeoutPacket()` assumes that only a timeout for SlashPacketData can be received, while in reality CCV Consumer also sends VSCMaturedPacketData packets; so on receiving a timeout of this packet the function will return an error. On the other hand, for SlashPacketData the processing goes to `ccv/consumer/keeper/relay.go: OnTimeoutPacket()`, which returns nil. This is inconsistent wrt. handling of timeouts of different packet types.

# Linked issues

Closes: #468 

# Type of change

- [x] Non-breaking changes

# How was the feature tested?
- [x] E2E tests
- [x] Integration tests/simulation

# Other information
Part of ICS audit in late 2022.

# Checklist:


x/ccv/consumer/ibc_module.go
```diff
@@ -275,15 +275,23 @@ func (am AppModule) OnTimeoutPacket(
...
+		vscErr := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &vscMaturedData)
+		if vscErr != nil {
+			// packet is neither SlashPacketData nor VCSMaturedPacketData
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest,
+				"cannot unmarshal consumer packet data: %s",
+				"expected SlashPacketData or VCSMaturedPacketData")
+		}
 	}
 
+	// keeper.OnTimeoutPacket is a no-op
```
mpoke, line 292: Remove `keeper.OnTimeoutPacket` as it is a noop. 

Also, add the following comment:
```
  // the CCV channel state is changed to CLOSED 
  // by the IBC module as the channel is ORDERED
```

mpoke (changes requested):

x/ccv/consumer/ibc_module.go
MSalopek: Done

x/ccv/consumer/ibc_module.go
```diff
@@ -270,18 +270,25 @@ func (am AppModule) OnAcknowledgementPacket(
...
-	var data ccv.SlashPacketData
-	if err := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
-		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal consumer packet data: %s", err.Error())
-	}
+	var slashData ccv.SlashPacketData
+	var vscMaturedData ccv.VSCMaturedPacketData
 
-	if err := am.keeper.OnTimeoutPacket(ctx, packet, data); err != nil {
-		return err
+	err := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &slashData)
```
jtremback, line 283: This could still accidentally unmarshal one type as another in some cases if in the future we changed the field names or something. Probably not a problem right now, but pretty sketchy. We probably need to think of a concrete way to differentiate these packets types, other than the lucky fact that they might not be able to deserialize into each other for now. Is there already something somewhere else in the codebase? @mpoke

x/ccv/consumer/ibc_module.go
```diff
@@ -270,18 +270,25 @@ func (am AppModule) OnAcknowledgementPacket(
...
 	_ sdk.AccAddress,
 ) error {
-	var data ccv.SlashPacketData
-	if err := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
-		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal consumer packet data: %s", err.Error())
-	}
+	var slashData ccv.SlashPacketData
+	var vscMaturedData ccv.VSCMaturedPacketData
 
-	if err := am.keeper.OnTimeoutPacket(ctx, packet, data); err != nil {
```
jtremback, line 283: We don't stop the consumer chain any more on timeout? Seems like a pretty huge change in behavior. Is this what we're doing now, or a typo?

x/ccv/consumer/keeper/relay.go
```diff
@@ -239,7 +239,7 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
 	return nil
 }
 
-func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, data ccv.SlashPacketData) error {
+func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet) error {
```
jtremback, line 242: Why is this function even in here?

jtremback (changes requested):
Got a lot of questions about this one

x/ccv/consumer/ibc_module.go
MSalopek: The function was a no-op anyway, there was no way it could return an error.

```golang
func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet) error {
	return nil
}
```

It has been like that since at least this commit:
https://github.com/cosmos/interchain-security/commit/af0670240406a89923fb7734562b749d5bf0db30#diff-52b901441beea83792e0e8fd2ddc7c6bf5155c707642e9b5bae10327acf7a62dL205

x/ccv/consumer/keeper/relay.go
MSalopek: That's the same function that was removed from https://github.com/cosmos/interchain-security/pull/550/files/f6cec81f2b7022021698cf09c45d928dfd8f85a0#r1040203683.

It's a no-op. it can be removed.

x/ccv/consumer/ibc_module.go
MSalopek: This function seemed like a sanity check and nothing more. It unmarshals to see if the type is correct and outputs an event if it is, otherwise it errors.

mpoke: I agree that we don't need to unmarshal here.

 Regarding the more general question about differentiating between different packet types, we do the same on the provider side, i.e., https://github.com/cosmos/interchain-security/blob/f6cec81f2b7022021698cf09c45d928dfd8f85a0/x/ccv/provider/ibc_module.go#L170. The easiest way would be to add a packet type in `ccv.proto`.

x/ccv/consumer/ibc_module.go
mpoke: Yes, this is consistent with the spec: https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-ccf-toslash1 and https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-ccf-tomat1

x/ccv/consumer/keeper/relay.go
mpoke: IMO, we should remove it.

jtremback (approved):
Looks good to LGTM to me

mpoke (approved):
LGTM

PR #556, opened 2022-12-05
Fixes #503 prevents two key assignment key overlap security issues

### LINKED ISSUE #503: KeyAssignment - must prevent validator creation using existing consumer key
# Problem

Consider a provider validator set consisting of validator V0 with provider key A and assigned consumer key B. We must prevent a new validator V1 being created with provider key B.

# Closing criteria

Prevent the scenario by placing appropriate checks in e.g. staking module.

# Problem details

The described scenario will lead to a fault, assuming that validator V1 does not assign a consumer key C with C!=B, because if V1 does not expliclity assign a key then a default value B will be chosen. This will lead to two validators having the assigned consumer key B. A clear violation.

Note that this problem appears even if requiring signatures for all keys, as one real world entity can create as many validators as they like.

# TODOs

danwt:
Fixes

- https://github.com/cosmos/interchain-security/issues/503

### DIFF
--- a/x/ccv/provider/keeper/hooks.go
+++ b/x/ccv/provider/keeper/hooks.go
@@ -70,15 +70,10 @@ func validatorConsensusKeyInUse(ctx sdk.Context, k *Keeper, valAddr sdk.ValAddre
 	// as argument at it is not possible to directly query the validator using
 	// the operator address. Nor is it possible to convert an operator addr
 	// to a consensus addr.
-	var val stakingtypes.ValidatorI
-	k.stakingKeeper.IterateValidators(ctx, func(_ int64, validator stakingtypes.ValidatorI) bool {
-		if validator.GetOperator().Equals(valAddr) {
-			val = validator
-			return true
-
-		}
-		return false
-	})
+	val, found := k.stakingKeeper.GetValidator(ctx, valAddr)
+	if !found {
+		panic("did not find newly created validator in staking module")
+	}
 
 	// Get the consensus address of the validator being added
 	cons, err := val.GetConsAddr()
--- a/x/ccv/provider/keeper/hooks.go
+++ b/x/ccv/provider/keeper/hooks.go
@@ -64,8 +64,49 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) error {
 	return nil
 }
 
+func validatorConsensusKeyInUseOnConsumerChain(ctx sdk.Context, k *Keeper, valAddr sdk.ValAddress) bool {
+	// Find the validator being added in the staking module
+	// This is necessary because only the operator address is received
+	// as argument at it is not possible to directly query the validator using
+	// the operator address. Nor is it possible to convert an operator addr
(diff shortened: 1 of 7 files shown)


x/ccv/provider/keeper/key_assignment.go
```diff
@@ -363,17 +358,32 @@ func (k Keeper) AssignConsumerKey(
...
 	consumerAddr := utils.TMCryptoPublicKeyToConsAddr(consumerKey)
 
 	providerAddr, err := validator.GetConsAddr()
+
 	if err != nil {
 		return err
 	}
 
+	if existingVal, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, consumerAddr); found {
+		if existingVal.OperatorAddress != validator.OperatorAddress {
```
mpoke, line 371: Why is this extra check needed?

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -363,17 +358,32 @@ func (k Keeper) AssignConsumerKey(
...
+			// consumer key is already the key belonging to another existing
+			// and different provider
+			// prevent a validator from assigning a key which is the *provider*
+			// key of another validator
```
mpoke, line 375: Comment unclear.

x/ccv/provider/keeper/hooks.go
```diff
@@ -64,8 +64,43 @@ func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, ID uint64) error {
 	return nil
 }
 
-// Define unimplemented methods to satisfy the StakingHooks contract
+func ValidatorConsensusKeyInUse(k *Keeper, ctx sdk.Context, valAddr sdk.ValAddress) bool {
```
mpoke, line 67: Why the need of an extra function instead of adding the logic directly in the `AfterValidatorCreated` hook? For UTs, we can use the hook directly?

mpoke (commented):
Nice work @danwt. See my comments below.

x/ccv/provider/keeper/hooks.go
danwt: What would be the benefit of inlining into AfterValidatorCreated? I think that would make it harder to understand the functionality and it would be less self documenting.

The status quo in writing software is *not* to inline everything AFAIU.

x/ccv/provider/keeper/key_assignment.go
danwt: I've changed the comment

```
		// If there is a validator with using the consumer key to validate on the provider
		// we prevent assigning the consumer key, unless the validator is assigning validator.
		// This ensures that a validator joining the active set who has not explicitly assigned
		// a consumer key, will be able to use their provider key as consumer key (as per default)
```

x/ccv/provider/keeper/hooks.go
mpoke: Then move it to `key_assignment.go` and the test to `key_assignment_test.go`.

PR #557, opened 2022-12-05
Throttle garbage collection

smarshall-spitzbart:
Implements the throttling related cleanup that is needed when a consumer is stopped. Queued slash packet and vsc matured packet data is thrown out


x/ccv/provider/keeper/proposal.go
```diff
@@ -218,6 +218,15 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, lockUbd, clos
 		k.DeleteUnbondingOpIndex(ctx, chainID, id)
 	}
 
+	// Remove any existing throttling related queue entries from the
+	// global queue for this consumer.
+	k.DeletePendingSlashPacketEntriesForConsumer(ctx, chainID)
+
+	// Remove all pending slash packets and vsc matured packets queued for this consumer.
+	// Note: queued VSC matured packets can be safely removed from the per-chain queue,
```
smarshall-spitzbart, line 226: Could someone confirm for me that it is indeed safe to throw out queued VSC matured packets when stopping a consumer? This seems safe since the unbonding ops are already released above

jtremback: Seems good to me. @mpoke?

jtremback (approved):
LGTM

x/ccv/provider/keeper/proposal.go
mpoke: Removing the queued VSCMatured packets is fine. I don't like the idea that we remove queued slash packets. These packets were already relayed on the provider and the only reason were not handled is the throttling mechanism. IMO, a consumer state should not be removed until the queue of received packets is empty.

mpoke (changes requested):
See my comment below.

danwt:
If the decision is that both maturities and slash packets are able to be removed, I would suggest, instead of removing them from the queue, to just ignore them when they are pulled from the queue.

If the decision is that one of the packet times should not be removable, I would suggest applying that decision to both packet types (addressing Marius' 'IMO, a consumer state should not be removed until the queue of received packets is empty.')

smarshall-spitzbart:
@mpoke issue as mentioned over call https://github.com/cosmos/interchain-security/issues/566

smarshall-spitzbart:
> If the decision is that both maturities and slash packets are able to be removed, I would suggest, instead of removing them from the queue, to just ignore them when they are pulled from the queue.

Although this may be the better solution, and less of a big-hammer, it'd be inconsistent with how all other consumer-specific state, including pending/queued state, is handled in `StopConsumerChain`

As far as whether slash/vsc matured packets are able to be safely removed/dropped, this topic will be tracked in https://github.com/cosmos/interchain-security/issues/566

mpoke (approved):

Issue #562, opened 2022-12-06
Add slash throttle queries


MSalopek:
# Problem

Add queries to check contents of slash queues.

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

PR #563, opened 2022-12-06
add slash throttling related queries

MSalopek:
# Description
This PR adds 2 new query methods on the provider

`interchain-security-pd q provider pending-consumer-packets <chain_id>`
* lists all pending data packets (both slash and VSC matured) for `chain_id`
* Eg.
```
{
    "chain_id": "consu",
    "packets": [],
    "size": "0"
}
```

`interchain-security-pd q provider pending-slash-packets`
* lists all pending slash packets with some data about the slash meter
* Eg.
```
{
    "slash_meter": "50",
    "slash_meter_allowance": "75",
    "last_replenish": "2022-12-06T14:47:14.249756250Z",
    "packets": []
}
```

# Linked issues

Closes: #562 

# Type of change

- [x] Non-breaking changes
- [x] New feature (adding functionality)
- [x] Proto files changes

# Checklist:

- [x] `make proto-gen` was run (for changes in `.proto` files)


jtremback:
Looks OK, will have to review it further- Have you ever been able to test this out? What does the serialization look like with fuller queues? Can you add an integration test?

x/ccv/provider/types/keys.go
```diff
@@ -275,38 +275,41 @@ func ParsePendingPacketDataKey(key []byte) (string, uint64, error) {
...
 		sdk.Uint64ToBigEndian(uint64(timeBzL)),
 		timeBz,
-		HashBytes(packetEntry.ValAddr),
+		sdk.Uint64ToBigEndian(packetEntry.IbcSeqNum),
 		[]byte(packetEntry.ConsumerChainID),
 	)
 }
 
 // ParsePendingSlashPacketEntryKey returns the received time and chainID for a pending slash packet entry key
-func ParsePendingSlashPacketEntryKey(bz []byte) (time.Time, string) {
```
jtremback, line 295: Can you explain more about why this needed to be changed?

MSalopek: There's 2 queues:
1. global queue holding SlashPacketEntry
2. per-chain queue holding both VSCMaturedPacketData and SlashPacketData

When querying the global queue, the data in it has no value to validators. SlashPacketEntry is just this:
```golang

// A persisted queue entry indicating that a slash packet data instance needs to be handled
type SlashPacketEntry struct {
	RecvTime time.Time

	ConsumerChainID string

	IbcSeqNum uint64

	ValAddr []byte
}
```
It was brought to my attention that this data structure is not informative enough for the normal user since it is used for internal processing. Instead, it is better to provide the actual slash packet data (as transferred over IBC) since it contains useful information relevant to the user.

```golang
type SlashPacketData struct {
	Validator types.Validator
	ValsetUpdateId uint64
	Infraction types1.InfractionType
}
```

`SlashPacketData` is not available in the global queue -> we need to get it from the per-chain queue.

It was not possible to get it from the per-chain queue without adding the newer changes from `circuit-breaker` branch. The changes in question expand the per-chain queue key and add `IbcSeq` to `SlashPacketEntry`. With all data readily available we can fetch the exact slash packet that's waiting to be processed (validator, infraction, valsetUpdateId) and enrich the query output

TL;DR:
In the `goc-december` version of throttling we did not save the IBC sequence number inside the `SlashPacketEntry` in the global queue, so we could not connect the `SlashPacketEntry` to the per-chain `SlashPacketData`

MSalopek:
Closing in favour of: #600.

This branch is too far out of date to be worth the effort.

Issue #544, opened 2022-12-02
Outdated `SlashPackets` are not dropped


mpoke:
# Problem

~A corrupt consumer chain could send `SlashPackets` with `data.vscId` after sending a `VSCMaturedPacket` for the same `vscId`. These packets should be dropped.~ 

After discussions with @MSalopek, we realized that it's correct behavior to send `SlashPackets` with `data.vscId` after sending a `VSCMaturedPacket` for the same `vscId`. This is because the `data.vscId` in a `SlashPacket` is the `vscId` of the `VSCPacket` that last updated the validator set of the block when the infraction was committed. The misbehaving validator must be punished regardless of the infraction height. `data.vscId` is mapped on a height on the  provider that is used to determined what undelegations and redelegations should be slashed (see `Slash` in the staking module). 

Thus, this issue is about reducing as much as possible the scenarios in which a corrupt consumer chain ends up slashing a validator without following the rules. A more complete solution will be added as part of future versions, i.e., the provider can verify the evidence for all the `SlashPackets`.  

# Closing criteria

~Drop `SlashPackets` for which a corresponding `VSCMaturedPacket` was already received.~ 

- [ ] Drop `SlashPackets` for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself. 
- [x] ~Drop `SlashPackets` for downtime infraction were received from the same consumer for infraction heights smaller than the latest downtime infraction height received from that consumer.~
- [ ] Run`SlashPacketData.ValidateBasic` on receiving `SlashPackets`

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

smarshall-spitzbart:
Nice! This logic can be added to `validateSlashPacket` after #542 is merged. I won't address this issue in 542 however, as I want to make sure that PR does not introduce any functionality changes.

mpoke:
> This logic can be added to validateSlashPacket after https://github.com/cosmos/interchain-security/pull/542 is merged.

If we add it to `validateSlashPacket`, we need to check `VSCMaturedPackets` in the queue. Alternatively, we can drop the packets when we handle them, but then we don't sent back err ack.

smarshall-spitzbart:
[...]

PR #565, opened 2022-12-06
Throttle bug fixes + req refactors

### LINKED ISSUE #538: Throttle uses consumer validator address to lookup validator power but it should use provider address
# Problem

The slash throttle uses the consumer validator address to lookup the voting power on the provider in order to subtract from the slash meter, so it will not find any validator if a consumer key is assigned. It will get always retrieve `0` power.

# Closing criteria

Ensure the validator power is queried using the provider validator address

# Problem details

Slash is queued

https://github.com/cosmos/interchain-security/blob/goc-december/x/ccv/provider/keeper/relay.go#L243-L256

Validator power is looked up

https://github.com/cosmos/interchain-security/blob/circuit-breaker/x/ccv/provider/keeper/throttle.go#L18-L37

# TODOs

smarshall-spitzbart:
# Description

This PR solves two bugs found from diff tests on #462, thank you for finding these @danwt! It also closes an issue raised about the integration of key assignment and throttling.

Bug 1: https://github.com/cosmos/interchain-security/pull/462#discussion_r1041337359
Bug 2: https://github.com/cosmos/interchain-security/pull/462#issuecomment-1339845142
Issue fixed: #538 

# Linked issues

Closes #538 

# Type of change

- [x] Bugfixes

# How was the feature tested?

- E2E tests
- Integration

# Other information

This will need to be cherry-picked into `goc-december` and `circuit-breaker`

### DIFF
--- a/tests/integration/config.go
+++ b/tests/integration/config.go
@@ -300,7 +300,8 @@ func MultiConsumerTestRun() TestRun {
 					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
-					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"",
 			},
 			chainID("consu"): {
 				chainId:        chainID("consu"),
--- a/tests/integration/config.go
+++ b/tests/integration/config.go
@@ -182,7 +182,8 @@ func KeyAssignmentTestRun() TestRun {
 					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
-					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"", // This disables slash packet throttling
 			},
 			chainID("consu"): {
 				chainId:        chainID("consu"),
@@ -221,7 +222,8 @@ func DefaultTestRun() TestRun {
 					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
-					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"", // This disables slash packet throttling
 			},
 			chainID("consu"): {
 				chainId:        chainID("consu"),
@@ -260,7 +262,8 @@ func DemocracyTestRun() TestRun {
 					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
-					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"", // This disables slash packet throttling
 			},
 			chainID("democ"): {
 				chainId:        chainID("democ"),
@@ -301,7 +304,7 @@ func MultiConsumerTestRun() TestRun {
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
 					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
-					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"",
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"", // This disables slash packet throttling
 			},
 			chainID("consu"): {
 				chainId:        chainID("consu"),
--- a/x/ccv/provider/keeper/relay.go
+++ b/x/ccv/provider/keeper/relay.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
@@ -249,12 +250,17 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
 		panic(fmt.Errorf("SlashPacket received on unknown channel %s", packet.DestinationChannel))
 	}
 
+	// The slash packet validator address may be known only on the consumer chain,
+	// in this case, it must be mapped back to the consensus address on the provider chain
+	consumerConsAddr := sdktypes.ConsAddress(data.Validator.Address)
+	providerConsAddr := k.GetProviderAddrFromConsumerAddr(ctx, chainID, consumerConsAddr)
+
 	// Queue a pending slash packet entry to the parent queue, which will be seen by the throttling logic
 	k.QueuePendingSlashPacketEntry(ctx, providertypes.NewSlashPacketEntry(
-		ctx.BlockTime(), // recv time
-		chainID,         // consumer chain id that sent the packet
-		packet.Sequence, // IBC sequence number of the packet
-		data.Validator.Address))
+		ctx.BlockTime(),   // recv time
+		chainID,           // consumer chain id that sent the packet
+		packet.Sequence,   // IBC sequence number of the packet
+		providerConsAddr)) // Provider consensus address of val to be slashed
 
 	// Queue slash packet data in the same (consumer chain specific) queue as vsc matured packet data,
 	// to enforce order of handling between the two packet types.
--- a/x/ccv/provider/keeper/relay.go
+++ b/x/ccv/provider/keeper/relay.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
@@ -252,7 +251,7 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
 
 	// The slash packet validator address may be known only on the consumer chain,
 	// in this case, it must be mapped back to the consensus address on the provider chain
-	consumerConsAddr := sdktypes.ConsAddress(data.Validator.Address)
+	consumerConsAddr := sdk.ConsAddress(data.Validator.Address)
 	providerConsAddr := k.GetProviderAddrFromConsumerAddr(ctx, chainID, consumerConsAddr)
 
 	// Queue a pending slash packet entry to the parent queue, which will be seen by the throttling logic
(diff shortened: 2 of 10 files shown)


jtremback (approved):

danwt (commented):
Heyoo
It's very important to define precise invariant(s) and property(s) of your ideal system before you start implementing a solution to get there. There are problems cropping up because the invariants defined so far are _not_ precise.

The invariant is here 

https://github.com/cosmos/interchain-security/blob/b28e176ffcb609d877baeb71395fdb8078e2083a/docs/throttle.md?plain=1#L81-L101

but it's not sensible in the current form

> The time it takes to jail `X`% of the provider validator set from consumer initiated slash requests ...

because the validator set is always changing every block. So the sentence doesn't really make sense: which validator set is being talked about?

I bring this up because development without invariants leads to many problems. The core issue is that you cannot judge a solution/implementation without invariants. Moreover, you cannot describe what the code is supposed to be doing, except in a hand wavy way.

I bring this up here because, in this PR, changes are being made which affect the slash guarantees in subtle ways, and it cannot be said if they're correct or not, because the goal (the invariant) isn't defined.

To be concrete, with a problem _here_. One of the issues being addressed is the one from this comment 

https://github.com/cosmos/interchain-security/pull/462#issuecomment-1339845142

which says that setting `replenishFraction=1` is not sufficient to avoid packet delays. But this code does not address that AFAIU.

Note that _inactive_ validators are not the same as _unbonded_ valiators. All unbonded validators are inactive, but not all inactive validators are unbonded. Some inactive validators are _unbonding_, _and they must be slashable_. Slash packets targeting active and inactive (but not unbonded) validators can all arrive at the same time and they must all be handled as per whatever invariant.

I think it's very important to precisely write down the throttle invariants, and then work from there. We can work together on doing that!

MSalopek (approved):
Approving because needed in goc-december testnet release.

Integration tests will be added in some other PR.

tests/integration/config.go
```diff
@@ -300,7 +300,8 @@ func MultiConsumerTestRun() TestRun {
 					".app_state.slashing.params.signed_blocks_window = \"2\" | " +
 					".app_state.slashing.params.min_signed_per_window = \"0.500000000000000000\" | " +
 					".app_state.slashing.params.downtime_jail_duration = \"2s\" | " +
-					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\"",
+					".app_state.slashing.params.slash_fraction_downtime = \"0.010000000000000000\" | " +
+					".app_state.provider.params.slash_meter_replenish_fraction = \"1.0\"",
```
MSalopek, line 304: Hey, can you include this in all test runs?

`MultiConsumerTestRun`
`DemocracyTestRun`
`DefaultTestRun`

x/ccv/types/ccv.go
```diff
@@ -76,3 +76,10 @@ func (vdt SlashPacketData) GetBytes() []byte {
 	valDowntimeBytes := ModuleCdc.MustMarshalJSON(&vdt)
 	return valDowntimeBytes
 }
+
+// Unmarshals json bytes to SlashPacketData or panics if the bytes are invalid.
+func MustUnmarshalJsonBzToSlashPacketData(bz []byte) SlashPacketData {
```
mpoke, line 81: It seems that this function is only needed for testing in only one place. Could you please remove it?

x/ccv/provider/keeper/throttle.go
```diff
@@ -29,9 +29,25 @@ func (k Keeper) HandlePendingSlashPackets(ctx sdktypes.Context) {
 	// Iterate through ordered (by received time) slash packet entries from any consumer chain
 	k.IteratePendingSlashPacketEntries(ctx, func(entry providertypes.SlashPacketEntry) (stop bool) {
 
+		// Obtain validator from consensus address.
+		// The slash packet validator address may be known only on the consumer chain
+		// in this case, it must be mapped back to the consensus address on the provider chain
+		consumerConsAddr := sdktypes.ConsAddress(entry.ValAddr)
```
mpoke, line 35: `entry.ValAddr` could be set to the validator operator address in `OnRecvSlashPacket`. That's where I'd also remap consumer addresses to provider addresses before queueing `PendingSlashPacketData`.

x/ccv/provider/keeper/throttle.go
```diff
@@ -29,9 +29,25 @@ func (k Keeper) HandlePendingSlashPackets(ctx sdktypes.Context) {
...
+		providerConsAddr := k.GetProviderAddrFromConsumerAddr(ctx, entry.ConsumerChainID, consumerConsAddr)
+
+		// Note: if validator is not found or unbonded, this will be handled appropriately in HandleSlashPacket
+		val, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerConsAddr)
+
 		// Obtain the validator power relevant to the slash packet that's about to be handled
 		// (this power will be removed via jailing or tombstoning)
-		valPower := k.stakingKeeper.GetLastValidatorPower(ctx, entry.ValAddr)
+		var valPower int64
+		if val.IsJailed() || !found {
```
mpoke, line 44: If `found==false` and `val==nil`, wouldn't this code panic?

mpoke (changes requested):

tests/integration/config.go
smarshall-spitzbart: Solid suggestion, done

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: `val` is a `stakingtypes.Validator` concrete type, so it can never be nil. If `found` is `false`, then a zero value [is returned](https://github.com/cosmos/cosmos-sdk/blob/9c145c827001222df2e3e1101010874aeac20997/x/staking/keeper/validator.go#L41) for the validator type. 


If val was a pointer to a `stakingtypes.Validator` then this code could panic. However that'd require a breaking change to the staking keeper API

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: SGTM, changes done in https://github.com/cosmos/interchain-security/pull/565/commits/f2a5522a4dfcd7025129ab6b264af7adb6f4fcb1. Note this makes the PR include refactors as well. So naming is changed accordingly

smarshall-spitzbart:
Everything passes locally, including `golancgci-lint run`, so ignore the failed GH actions

smarshall-spitzbart:
Responding to @danwt 

> There are problems cropping up because the invariants defined so far are not precise.

The problems that have came up so far seem like normal bugs that're a part of the development process. E2e/unit tests are limited, this is why I wasn't comfortable merging the slash packet throttling feature into main until we either had diff tests passing, or throttle specific integration tests passing (both of which are capable of finding this [critical bug](https://github.com/cosmos/interchain-security/pull/462#discussion_r1041337359)). Finding bugs now is a good thing imo and I welcome such efforts!

> the goal (the invariant) isn't defined.

Imo this is not the case. There is a lack of finalized spec rn, but that does not mean we're throwing invariants out the window. `Provider Slashing Warranty` is the only invariant that slash packet throttling should adjust in the current CCV spec. If there are more existing CCV invariants that this throttling implementation violates, then we can have a discussion about that, and make changes to spec or implementation as needed.

> which validator set is being talked about?

You have a valid concern about the invariant I've defined in `throttle.md`. I will improve that document to define that I'm referring to the validator set that exists at the start of any such attack. If it interests you to collab on making the language more granular, than by all means I'd welcome the help, CC @mpoke who will write the spec.

This PR does address https://github.com/cosmos/interchain-security/pull/462#issuecomment-1339845142. See `TestSlashSameValidator`, which would not pass in `circuit-breaker` without this PR. Setting `SlashMeterReplenishFraction = 1` should now allow all slash packets to be handled immediately, even if they're recv in the same block. If I'm understanding the bug incorrectly and it's not fixed by this PR, then lets discuss.

> Note that inactive validators are not the same as unbonded valiators.

None of the functionality you've written in this paragraph should be violated by slash packet throttling. If it is violated, that's a bug we need to fix.

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: Although, swapping the order as I've done [here](https://github.com/cosmos/interchain-security/pull/565/commits/74441c2d56a13a570a975526a2b1d155bf342cbb) will allow us not to rely on the value of `val` when `found = false`

x/ccv/provider/keeper/relay.go
```diff
@@ -249,12 +249,17 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
...
-		chainID,         // consumer chain id that sent the packet
-		packet.Sequence, // IBC sequence number of the packet
-		data.Validator.Address))
+		ctx.BlockTime(),   // recv time
+		chainID,           // consumer chain id that sent the packet
+		packet.Sequence,   // IBC sequence number of the packet
+		providerConsAddr)) // Provider consensus address of val to be slashed
 
 	// Queue slash packet data in the same (consumer chain specific) queue as vsc matured packet data,
 	// to enforce order of handling between the two packet types.
```
mpoke, line 265: Why not update the packet data to include the `providerConsAddr` and avoid calling `GetProviderAddrFromConsumerAddr` later on?

x/ccv/provider/keeper/relay.go
```diff
@@ -249,12 +249,17 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
...
 	// Queue a pending slash packet entry to the parent queue, which will be seen by the throttling logic
 	k.QueuePendingSlashPacketEntry(ctx, providertypes.NewSlashPacketEntry(
-		ctx.BlockTime(), // recv time
-		chainID,         // consumer chain id that sent the packet
-		packet.Sequence, // IBC sequence number of the packet
-		data.Validator.Address))
+		ctx.BlockTime(),   // recv time
+		chainID,           // consumer chain id that sent the packet
+		packet.Sequence,   // IBC sequence number of the packet
+		providerConsAddr)) // Provider consensus address of val to be slashed
```
mpoke, line 262: Why not using the validator's operator address instead of the consensus address? Is this used anywhere else besides in `HandlePendingSlashPackets`.

mpoke (approved):
LGTM. Just some suggestions on simplifying the code. Also, why are the GH actions failing?

x/ccv/provider/keeper/throttle.go
```diff
@@ -29,9 +29,20 @@ func (k Keeper) HandlePendingSlashPackets(ctx sdktypes.Context) {
...
 		// (this power will be removed via jailing or tombstoning)
-		valPower := k.stakingKeeper.GetLastValidatorPower(ctx, entry.ValAddr)
+		var valPower int64
+		if !found || val.IsJailed() {
+			// If validator is not found, or found but jailed, it's power is 0. This path is explicitly defined since the
+			// staking keeper's LastValidatorPower values are not updated till the staking keeper's endblocker.
+			valPower = 0
+		} else {
+			valPower = k.stakingKeeper.GetLastValidatorPower(ctx, val.GetOperator())
+		}
```
danwt, line 45: I might be missing something but doesn't this change allow validator to be slashed all the way to 0 in a single block?

If a consumer sends many slash packets for the same validator?

_______________________________________________________________________________________________


Besides that, I don't think this addressing the problem in this comment

https://github.com/cosmos/interchain-security/pull/462#issuecomment-1339845142

To state the problem more clearly:

We want it to be possible to set the slashFraction or other params so that the throttle is 'invisible'. Ie. the behavior of the system is exactly as in the current [spec](https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md#consumer-initiated-slashing),
This means, we do not want any slash packets to be held up/delayed (still talking about when fraction=1, replenishPeriod = 1 second ect).

Consider the following

Imagine there are n validators total on the provider, with k<n active. Imagine the actives are v1,v2..vk and the inactives are vk+1,..vn

A consumer sends slash packets for every single provider validator v1..,vn and they are all delivered to the provider in the same block.

It is possible, that all of the slashes for the active validators v1..vk are handled first. This will set the meter to 0 so the slashes for vk+1,.. vn will be blocked.

smarshall-spitzbart: With this design we can slash up to but not including 100% of the "initial val set" instantaneously when the replenish fraction is 1. If we want to include 100%, its a one line inequality change. I like 100% not being allowed, but I'm not super attached to that idea

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: > Why not update the packet data

By this, I believe you're referring to the `SlashPacketData` type. This is the protobuf generated type that is sent over the wire from consumer to provider. Adding a new field to this type (that's only used on the provider) seems unnecessary to me. However, if you feel this is important, I can refactor the base branch once the merge conflicts are solved. I'm hesitant to do much more refactoring in this PR due to how disconnected from main it is at this time :)

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: The change we discussed today is included in https://github.com/cosmos/interchain-security/pull/565/commits/777a3f6f5a771aabf73d1a591c8de7ae413f4be1, thank you for the efforts and feedback. Even with disagreements at times, the dialogue ended up useful and productive! It will be cool to get this feature properly specced and thoroughly tested before prod

smarshall-spitzbart:
> Also, why are the GH actions failing?

I think this has to do with some oddities of how GH handles CI for a branch off a branch that's disconnected from main 😆, I just ran `make test` locally on the most recent commit and it all checks out

smarshall-spitzbart:
@mpoke @danwt, I've merged, but have done my best to address your PR comments here. Please lmk in the base PR if there's anything that feels unaddressed, or incorrect. 

Refactors will be a lot easier to reason about once `circuit-breaker` is up to date with main, and particularly #542

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: and before I forget, 

> I might be missing something but doesn't this change allow validator to be slashed all the way to 0 in a single block?

Yes this is possible with or without slash packet throttling. Does https://github.com/cosmos/interchain-security/issues/544 address your concern? If not, might be useful to add onto that issue or create a new one

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: Ah gotcha, that's a reasonable change, done in https://github.com/cosmos/interchain-security/commit/1f920b4eb76189ca9901f0bb6578eba689ddd3e2

Issue #534, opened 2022-11-30
difftest core: Amend diff model with throttle changes once spec has been updated


danwt:
# Problem

The slash throttle PR changes the way things work when there are a lot of incoming slash requests. The diff test model should be changed to account for this, after the upstream canonical spec has been changed.

# Closing criteria

After the ICS spec has been updated to account for throttling, the diff model should be updated to match.

# Problem details

See

- https://github.com/cosmos/interchain-security/issues/533

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
cc @smarshall-spitzbart

smarshall-spitzbart:
@mpoke @MSalopek I see this is marked as high priority. Which issue should I put effort into first: this one, or https://github.com/cosmos/interchain-security/issues/615?

mpoke:
@smarshall-spitzbart This one should have priority as the other is just documentation.

smarshall-spitzbart:
It might be beneficial to introduce all throttle related changes to the ICS spec before tackling this issue

mpoke:
@smarshall-spitzbart is this issue fixed?

smarshall-spitzbart:
> @smarshall-spitzbart is this issue fixed?

This issue hasn't been completed. Maybe it's being worked on by @sainoe or @MSalopek? Lmk if I should make this my priority

p-offtermatt:
Closing since difftest model was deprecated in favor of MBT, may reopen if we want this in the MBT model

Issue #533, opened 2022-11-30
difftest core: Amend diff model with throttle changes (normal operation)


danwt:
# Problem

The slash throttle PR changes the way things work under normal operation (and abnormal operation). The diff test model should be changed to account for this.

# Closing criteria

Update the diff test core model so that the tests pass using the throttle when the Replenish Period and Allowance params are large.

# Problem details

The throttle should work almost as normal when the slash allowance is very large. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
cc @smarshall-spitzbart

smarshall-spitzbart:
Looks good, this issue is about using diff tests as a regression test, and will need to be fixed before #462 is merged. Whereas #534 is future work to test the throttling logic (not regression testing)

danwt:
See https://github.com/cosmos/interchain-security/pull/571

PR #580, opened 2022-12-09
#533 throttle normal operation diff testing PR rebase 2

danwt:
Adjusts the diff tests core model to tests scenarios where the throttle `fraction=1.0` and the throttle is replenished every block.

```
	prams.SlashMeterReplenishFraction = "1.0"
	prams.SlashMeterReplenishPeriod = time.Second * 1
```

This should allow all slashes to occur as per the spec as if the throttle didn't exist. This is a good baseline to get into the testnet and main until the full spec can be flesh out and more throttle scenarios can be tested (#534)

Issue

- https://github.com/cosmos/interchain-security/issues/533


smarshall-spitzbart:
rebased onto current version of `circuit-breaker`

Issue #537, opened 2022-11-30
At least one instance of incorrect iteration in the codebase- there may be more


jtremback:
# Problem

Some iterator(s) in the codebase iterate over keys which include a length prefix. This means that the length prefix is ordering the iteration, not the actual value in the key. We need to find out if this could cause any problems. For example, there are indexes with keys that have time in them, but the time is prefixed by a length. This may be ok, or it may cause incorrect ordering, depending on serialization details.

For example, in this function:

```
func (k Keeper) IteratePendingConsumerAdditionProps(ctx sdk.Context, cb func(spawnTime time.Time, prop types.ConsumerAdditionProposal) bool) {
	iterator := k.PendingConsumerAdditionPropIterator(ctx)
	defer iterator.Close()

	for ; iterator.Valid(); iterator.Next() {
		key := iterator.Key()
		spawnTime, _, err := types.ParsePendingCAPKey(key)
		if err != nil {
			panic(fmt.Errorf("failed to parse pending client key: %w", err))
		}

		var prop types.ConsumerAdditionProposal
		k.cdc.MustUnmarshal(iterator.Value(), &prop)

		if !cb(spawnTime, prop) {
			return
		}
	}
}
```
We think we are iterating in order of time, but we are really iterating in order of the serialized length of the timestamps.

# Closing criteria

Someone needs to study the iterators and figure out how to remove uncertainty. 

For example- don't use length prefixes on any key that needs to be iterated over in order. 

Or- don't use length prefixes at all. They are only used to enable parsing values back out of a key (otherwise you can just hash values to a constant length in a key). Values in keys could be stored alongside the values in the value instead of needing to parse the key.

Another fix- don't break out of iterations, making the order not matter

smarshall-spitzbart:
#592 is related, but tracked as separate issue with more focus. This issue will be closed by #583

PR #583, opened 2022-12-09
Check and fix iterators

mpoke:
# Description

- All SDK iterators behave as expected, i.e., the iteration order is correct.
- For some iterators, the option to stop the loop was removed. 
- All keys  under the same prefix are unique.
- Replace `PacketMaturityTime` with `VSCPacketQueue` (similar to the `UBDQueue` from the staking module). 

# Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/537

# Type of change

- [x] Breaking change (fix, feature or refactoring that changes existing functionality)
- [x] Requires state migrations
- [x] Proto files changes
- [x] Updates in store keepers or store keys
- [x] Changes in genesis (import/export)
- [x] Testing

# How was the feature tested?

- [x] Unit tests

# Checklist:

- [x] `make proto-gen` was run (for changes in `.proto` files)
- [x] Added iterators follow SDK pattern (`IterateX(ctx sdk.Context, cb func(arg1, arg2) (stop bool))`)


smarshall-spitzbart:
Opening a PR at Midnight, talk about quality assurance 🙌

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,50 +206,95 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
 
-		stop := cb(seq, timeNs)
-		if stop {
-			break
-		}
+	return validatorSetChangeIDs.Ids
+}
+
+// SetVSCPacketQueueTimeSlice sets a specific VSC packet queue timeslice.
+func (k Keeper) SetVSCPacketQueueTimeSlice(ctx sdk.Context, timestamp time.Time, vscIDs ...uint64) {
```
MSalopek, line 227: I fail to see why there's 2 functions regarding store set operation? (SetVSCPacketQueueTimeSlice and InsertVSCPacketQueue)

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,50 +206,95 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
+	store := ctx.KVStore(k.storeKey)
+	bz := k.cdc.MustMarshal(&types.ValidatorSetChangeIDs{Ids: vscIDs})
+	store.Set(types.VSCPacketQueueKey(timestamp), bz)
+}
+
+// InsertVSCPacketQueue inserts the ID of a VSC packet to the appropriate timeslice
+// in the VSCPacketQueue.
+func (k Keeper) InsertVSCPacketQueue(ctx sdk.Context, vscID uint64, maturityTime time.Time) {
```
MSalopek, line 235: Continuing on comment above, if we need both functions, can this be refactored to:
```golang
func (k Keeper) InsertVSCPacketQueue(ctx sdk.Context, vscID uint64, maturityTime time.Time) {
	timeSlice := append(k.GetVSCPacketQueueTimeSlice(ctx, maturityTime), vscID)
	k.SetVSCPacketQueueTimeSlice(ctx, maturityTime, timeSlice...)
}
```

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,50 +206,95 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
-	return binary.BigEndian.Uint64(bz)
+
+	return vscIDs
 }
 
-// DeletePacketMaturityTimes deletes the packet maturity time for given received VSC packet ids
-func (k Keeper) DeletePacketMaturityTimes(ctx sdk.Context, vscIds ...uint64) {
+// GetAllVSCPacketMaturityTimes returns the maturity times of all received VSC packets.
+//
+// Note: The order of iteration is irrelevant as it returns all maturity times.
```
MSalopek, line 276: Is there an expectation about the order of iteration expressed somewhere else in the code?

If not, it is confusing why a comment about order of iteration was necessary. The function returns a slice of all times, the slice can be sorted  by `MaturingVSCPacket.MaturityTime` it the caller needs specific sort ordering

x/ccv/consumer/keeper/keeper.go
```diff
@@ -384,7 +429,11 @@ func (k Keeper) DeleteCCValidator(ctx sdk.Context, addr []byte) {
 	store.Delete(types.CrossChainValidatorKey(addr))
 }
 
-// GetAllCCValidator returns all cross-chain validators
+// GetAllCCValidator returns all cross-chain validators.
+//
+// Note that the cross-chain validators are stored under keys with the following format:
+// CrossChainValidatorBytePrefix | address
+// Thus, the iteration is in ascending order of addresses.
```
MSalopek, line 436: Why is order of iteration important if you return a slice and not an iterator? I'm not sure the implications from iteration should be applied to a slice - it can be sorted in any appropriate way by the caller.

Are we adding a comment about the presumed order of iteration on all functions from here onward?

x/ccv/consumer/keeper/keeper.go
```diff
@@ -341,19 +384,21 @@ func (k Keeper) DeleteOutstandingDowntime(ctx sdk.Context, consAddress string) {
...
 }
 
-// IterateOutstandingDowntime iterates over the validator addresses of outstanding downtime flags
-func (k Keeper) IterateOutstandingDowntime(ctx sdk.Context, cb func(address string) (stop bool)) {
+// IterateOutstandingDowntime iterates over the outstanding downtime flags.
+//
+// Note that the outstanding downtime flags are stored under keys with the following format:
+// OutstandingDowntimeBytePrefix | consAddress
+// Thus, the iteration is in ascending order of consAddresses.
+func (k Keeper) IterateOutstandingDowntime(ctx sdk.Context, cb func(address string)) {
```
MSalopek, line 392: `IterateAllOutstandingDowntime` since the iterator is always completely exhusted.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -299,22 +342,22 @@ func (k Keeper) DeleteHeightValsetUpdateID(ctx sdk.Context, height uint64) {
...
 }
 
-// IterateHeightToValsetUpdateID iterates over the block height to valset update ID mapping in store
-func (k Keeper) IterateHeightToValsetUpdateID(ctx sdk.Context, cb func(height, vscID uint64) (stop bool)) {
+// IterateHeightToValsetUpdateID iterates over the block height to valset update ID mapping in store.
+//
+// Note that the block height to valset update ID mapping is stored under keys with the following format:
+// HeightValsetUpdateIDBytePrefix | height
+// Thus, the iteration is in ascending order of heights.
+func (k Keeper) IterateHeightToValsetUpdateID(ctx sdk.Context, cb func(height, vscID uint64)) {
```
MSalopek, line 350: `IterateAllHeightToValsetUpdateID` since the iterator is exhausted every time.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,50 +206,95 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
-	store.Set(types.PacketMaturityTimeKey(vscId), timeBytes)
+	return store.Iterator([]byte{types.VSCPacketQueueBytePrefix},
+		sdk.InclusiveEndBytes(types.VSCPacketQueueKey(endTime)))
 }
 
-// GetPacketMaturityTime gets the maturity time for a given received VSC packet id
-func (k Keeper) GetPacketMaturityTime(ctx sdk.Context, vscId uint64) uint64 {
+// DequeueAllMatureVSCPacketQueue returns a concatenated list of all IDs of matured VSC packets
+// and deletes the timeslices from the VSC packet queue.
+func (k Keeper) DequeueAllMatureVSCPacketQueue(ctx sdk.Context) (vscIDs []uint64) {
```
MSalopek, line 254: This deletes iterator keys during iteration.

From the rest of our codebase it seems that we used to prefer explicit `DeleteX` calls inside functions, ie:
```golang
func HandleItems() {
  items := GetItems()
  doAction(items)
  DeleteItems(items...)
}

```

Are we attempting to move from the previous pattern?

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -67,8 +72,13 @@ func (k Keeper) QueryConsumerChainStops(goCtx context.Context, req *types.QueryC
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
+	// get all pending consumer removal proposals
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	props := k.GetAllConsumerRemovalProps(ctx)
+	props := types.ConsumerRemovalProposals{}
```
MSalopek, line 77: Why not have a function `GetPendingConsumerRemovalProps`?

x/ccv/consumer/keeper/relay.go
```diff
@@ -86,46 +85,30 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
...
-				),
-			)
-
-			maturedVscIds = append(maturedVscIds, vscId)
-		} else {
-			break
-		}
-		maturityIterator.Next()
-	}
+	vscIDs := k.DequeueAllMatureVSCPacketQueue(ctx)
```
MSalopek, line 88: I would personally prefer that we don't delete inside `DequeueAllMatureVSCPacketQueue` and delete in `QueueVSCMaturedPackets`.

The implementation reads in a weird way. This function seems as if it is simply providing the caller with a list of vscIDs.

x/ccv/provider/keeper/keeper.go
```diff
@@ -535,52 +575,50 @@ func (k Keeper) GetValidatorSetUpdateId(ctx sdk.Context) (validatorSetUpdateId u
...
 }
 
-// IterateSlashAcks iterates through the slash acks set in the store
-func (k Keeper) IterateValsetUpdateBlockHeight(ctx sdk.Context, cb func(valsetUpdateId, height uint64) (stop bool)) {
+// IterateValsetUpdateBlockHeight iterates through the mapping from vscIDs to block heights.
+//
+// Note that the mapping from vscIDs to block heights is stored under keys with the following format:
+// ValsetUpdateBlockHeightBytePrefix | vscID
+// Thus, the iteration is in ascending order of vscIDs.
+func (k Keeper) IterateValsetUpdateBlockHeight(ctx sdk.Context, cb func(vscID, height uint64)) {
```
MSalopek, line 605: `IterateAllValsetUpdateBlockHeights`

x/ccv/provider/keeper/keeper.go
```diff
@@ -624,8 +662,13 @@ func (k Keeper) ConsumeSlashAcks(ctx sdk.Context, chainID string) (acks []string
...
 // IterateSlashAcks iterates through the slash acks set in the store.
-// Note: this method is only used in testing
-func (k Keeper) IterateSlashAcks(ctx sdk.Context, cb func(chainID string, acks []string) (stop bool)) {
+//
+// Note that the slash acks are stored under keys with the following format:
+// SlashAcksBytePrefix | chainID
+// Thus, the iteration is in ascending order of chainIDs.
+//
+// Note: This method is only used in testing
+func (k Keeper) IterateSlashAcks(ctx sdk.Context, cb func(chainID string, acks []string)) {
```
MSalopek, line 671: `IterateAllSlashAcks`

x/ccv/provider/keeper/keeper.go
```diff
@@ -763,20 +776,20 @@ func (k Keeper) DeleteInitTimeoutTimestamp(ctx sdk.Context, chainID string) {
 }
 
 // IterateInitTimeoutTimestamp iterates through the init timeout timestamps in the store.
-// Note: as the keys have the `bytePrefix | chainID` format, the iteration is NOT done in timestamp order.
-func (k Keeper) IterateInitTimeoutTimestamp(ctx sdk.Context, cb func(chainID string, ts uint64) (stop bool)) {
+//
+// Note that the init timeout timestamps are stored under keys with the following format:
+// InitTimeoutTimestampBytePrefix | chainID
+// Thus, the iteration is in ascending order of chainIDs (NOT in timestamp order).
+func (k Keeper) IterateInitTimeoutTimestamp(ctx sdk.Context, cb func(chainID string, ts uint64)) {
```
MSalopek, line 783: `IterateAllIniTimeoutTimestamp`

MSalopek (changes requested):
This is valuable work and an excellent cleanup! Thank you!

However, if it is possible could we consolidate how we use iterators, how we comment their docstrings and how we name them?

If an iterator is always exhausted before returning maybe we should denote it as `IterateAll`, if iteration can be exited early maybe we denote it as `Iterate`.

We discussed multiple times about avoiding deletion inside an iterator.

I would be happy to jump in and take some of the workload.

x/ccv/consumer/keeper/keeper.go
mpoke: I removed `SetVSCPacketQueueTimeSlice`

x/ccv/consumer/keeper/keeper.go
mpoke: It's important as some code could assume a certain order of the elements in the returned slice. None of the functions return iterators. They iterate over a set of keys and apply a callback function to each entry. The callback here is implicit, i.e., collect all entries in a slice. 

> Are we adding a comment about the presumed order of iteration on all functions from here onward?

No. Just on those ones that have an iterator inside.

x/ccv/consumer/keeper/keeper.go
mpoke: It's just the code from the SDK. But I agree that we should keep to our pattern.

x/ccv/provider/keeper/grpc_query.go
mpoke: What would be the benefit?

MSalopek: Consistency and easier testing.

This is done inside a query  function - those have lots of moving parts and can access many parts of the store. Previously we opted to cover all functions inside queries by unit tests so we don't have to mock everything out.

MSalopek: But actually, nevermind :D

I see that that's not consistently applied.

This is a personal nitpick.
I tend to prefer a function with a discriptive name over multiple imperative steps that are preceeded by a comment explaining what the steps do.

MSalopek (approved):

x/ccv/provider/types/keys.go
```diff
@@ -143,30 +142,36 @@ func InitTimeoutTimestampKey(chainID string) []byte {
...
+// PendingCRPKey returns the key under which pending consumer removal proposals are stored.
+// The key has the following format: PendingCRPBytePrefix | timestamp | chainID
 func PendingCRPKey(timestamp time.Time, chainID string) []byte {
-	return TsAndChainIdKey(PendingCRPBytePrefix, timestamp, chainID)
-}
-
-// ParsePendingCRPKey returns the time and chain ID for a pending consumer removal proposal key or an error if unparseable
-func ParsePendingCRPKey(bz []byte) (time.Time, string, error) {
-	return ParseTsAndChainIdKey(PendingCRPBytePrefix, bz)
+	timeBz := sdk.FormatTimeBytes(timestamp)
```
danwt, line 162: good 👍

x/ccv/provider/keeper/proposal.go
```diff
@@ -466,73 +447,52 @@ func (k Keeper) BeginBlockCCR(ctx sdk.Context) {
...
+		return true
 	})
 
 	return propsToExecute
 }
 
+// IteratePendingConsumerRemovalProps iterates through the pending consumer removal proposals.
+//
+// Note that the pending consumer removal proposals are stored under keys with the following format:
+// PendingCRPBytePrefix | stopTime | chainID
```
danwt, line 475: opinionated nit: I would generally not put this kind of implementation detail in the function docstring. It's not bad to have it inside though.

x/ccv/provider/keeper/proposal.go
```diff
@@ -466,73 +447,52 @@ func (k Keeper) BeginBlockCCR(ctx sdk.Context) {
...
+// Thus, the iteration is in ascending order of stopTimes.
```
danwt, line 476:
```suggestion
// IteratePendingConsumerRemovalProps iterates through the pending consumer removal proposals in
// stopTime ascending order.
```

x/ccv/provider/keeper/proposal.go
```diff
@@ -466,73 +447,52 @@ func (k Keeper) BeginBlockCCR(ctx sdk.Context) {
...
 		}
-		// No more proposals to check, since they're stored/ordered by timestamp.
-		return true // stop
+		// stop iteration; proposals are ordered by stop time,
+		// so no additional pending props are ready to act upon
+		return true
 	})
 
 	return propsToExecute
 }
```
danwt, line 470: I suppose we only really need to store the time and the chainID and we only really need to return the list of chainIDs. But it's good not to change it like that in this PR 👍

x/ccv/provider/keeper/proposal.go
```diff
@@ -466,73 +447,52 @@ func (k Keeper) BeginBlockCCR(ctx sdk.Context) {
 	k.DeletePendingConsumerRemovalProps(ctx, propsToExecute...)
 }
 
-// ConsumerRemovalPropsToExecute iterates over the pending consumer removal proposals
-// and returns an ordered list of consumer removal proposals to be executed,
-// ie. consumer chains to be stopped and removed from the provider chain.
+// GetConsumerRemovalPropsToExecute returns an ordered list of pending consumer
+// removal proposals that are ready to be executed,
```
danwt, line 451: Nit:

> ordered list

Ordered by what? All lists are technically ordered.

x/ccv/provider/keeper/proposal.go
```diff
@@ -342,91 +349,71 @@ func (k Keeper) BeginBlockInit(ctx sdk.Context) {
...
-		}
-
 		var prop types.ConsumerAdditionProposal
 		k.cdc.MustUnmarshal(iterator.Value(), &prop)
 
-		stop := cb(spawnTime, prop)
+		stop := cb(prop)
 		if stop {
 			break
 		}
```
danwt, line 391: You could instead the name the callback `stop` and have

```
if stop(prop) {
  return
}
```

x/ccv/consumer/keeper/relay.go
danwt: I would name `DequeueAllMatureVSCPacketQueue` something like `ConsumeMatureVSCPacketQueue` because that makes it more clear that it deletes stuff.

danwt (approved):
Great work! 

Only thing I'd change is the iterator docstrings which tell you the bytes ect. This is all implementation detail. It's better to just concisely write which field is used to sort, that's all.

x/ccv/provider/keeper/proposal.go
mpoke: For now, I would like to have clarity about what's happening inside the iterator. Once we have guidelines of how to use the SDK store, we can drop them. Maybe a cleaner way is to use `cosmos-sdk/store/prefix`, see https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/x/capability/keeper/keeper.go#L455

x/ccv/provider/keeper/proposal.go
mpoke: See https://github.com/cosmos/interchain-security/pull/583#discussion_r1047209424

x/ccv/consumer/keeper/relay.go
mpoke: I prefer to keep it consistent with the `DequeueAllMatureUBDQueue` from the SDK.

x/ccv/provider/keeper/proposal.go
mpoke: Is there an advantage to do so besides saving 3 lines of code. The `cb` name is more accurate since it's a callback. Given that we had problems with this kind of callbacks, I prefer this approach.

x/ccv/provider/keeper/proposal.go
smarshall-spitzbart: Imo comments of these type are useful

x/ccv/consumer/keeper/genesis.go
```diff
@@ -118,35 +118,6 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) (genesis *consumertypes.GenesisSt
 			panic("provider client does not exist")
 		}
 
-		maturingPackets := []consumertypes.MaturingVSCPacket{}
```
smarshall-spitzbart, line 121: This logic is replaced by the `GetAllVSCPacketMaturityTimes` and `GetOutstandingDowntimes` calls below, correct?

mpoke: Yes

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -845,69 +853,88 @@ func TestBeginBlockCCR(t *testing.T) {
...
+		{ChainId: "2", StopTime: now.Add(2 * time.Hour)},
 	}
 
 	keeperParams := testkeeper.NewInMemKeeperParams(t)
 	providerKeeper, ctx, ctrl, _ := testkeeper.GetProviderKeeperAndCtx(t, keeperParams)
 	defer ctrl.Finish()
 
 	for _, prop := range props {
-		providerKeeper.SetPendingConsumerRemovalProp(ctx, prop.ChainId, prop.StopTime)
+		require.NotPanics(t, func() {
```
smarshall-spitzbart, line 921: Curious why this `NotPanics` explicit call is needed? Tests fail upon panic right

x/ccv/provider/types/keys.go
```diff
@@ -143,30 +142,36 @@ func InitTimeoutTimestampKey(chainID string) []byte {
...
-// ParsePendingCAPKey returns the time and chain ID for a pending consumer addition proposal key
-// or an error if unparsable
-func ParsePendingCAPKey(bz []byte) (time.Time, string, error) {
-	return ParseTsAndChainIdKey(PendingCAPBytePrefix, bz)
+	timeBz := sdk.FormatTimeBytes(timestamp)
+	return AppendMany(
+		// Append the prefix
+		[]byte{PendingCAPBytePrefix},
+		// Append the time bytes
+		timeBz,
```
smarshall-spitzbart, line 153: These bytes are not of set length, correct? If that's the case, it should be at the end of the key, or hashed to a known byte length. Also, does `TsAndChainIdKey` still exist?

smarshall-spitzbart: you could put the hashed chainID before the time bytes

smarshall-spitzbart (commented):
Great work! In addition to my comments, I'm a bit concerned with a large iterator refactor and/or change in behavior without confirmation that a regression test exists for every iterator. 

Ie. https://github.com/cosmos/interchain-security/issues/454 should be closed by this PR, or fixed before this PR is merged

x/ccv/consumer/keeper/genesis.go
```diff
@@ -59,7 +59,7 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
 			k.SetProviderChannel(ctx, state.ProviderChannelId)
 			// set all unbonding sequences
 			for _, mp := range state.MaturingPackets {
-				k.SetPacketMaturityTime(ctx, mp.VscId, mp.MaturityTime)
+				k.InsertVSCPacketQueue(ctx, mp.VscId, mp.MaturityTime)
```
jtremback, line 62: IMO this rename makes the code harder to understand

jtremback:
@mpoke I don't think we should merge this as is. IMO there are several different pull requests in this one:

- Addition of comments about key formats and iteration order. These are great, and if they were the only thing in the PR, it could be merged immediately.
- Some tiny refactors where you used a utility method to serialize uints instead of having the code inline. This is great, but makes the PR bigger. Maybe should be in its own.
- Conversion of some iterators to IterateAll. I'm hesitant to merge this right now, not because it isn't a good idea, but because we probably want to do a refactor where we replace iterators with getters that return arrays. Changing to IterateAll before then accomplishes little, since it will all be changed again anyway. But if we don't do that refactor, then the change to IterateAll is good.
- The new `VSCPacketQueueTimeSlice` related functions. I don't like this for two reasons. First, because I don't think it is an improvement, and second and more importantly, because I think it is too much of a change for this PR and makes it a lot harder to review.

jtremback (changes requested):
@mpoke I don't think we should merge this as is. IMO there are several different pull requests in this one:

- Addition of comments about key formats and iteration order. These are great, and if they were the only thing in the PR, it could be merged immediately.
- Some tiny refactors where you used a utility method to serialize uints instead of having the code inline. This is great, but makes the PR bigger. Maybe should be in its own.
- Conversion of some iterators to IterateAll. I'm hesitant to merge this right now, not because it isn't a good idea, but because we probably want to do a refactor where we replace iterators with getters that return arrays. Changing to IterateAll before then accomplishes little, since it will all be changed again anyway. But if we don't do that refactor, then the change to IterateAll is good.
- The new `VSCPacketQueueTimeSlice` related functions. I don't like this for two reasons. First, because I don't think it is an improvement, and second and more importantly, because I think it is too much of a change for this PR and makes it a lot harder to review.

mpoke:
> I don't think we should merge this as is. IMO there are several different pull requests in this one:

So let's split the PR. 

> The new VSCPacketQueueTimeSlice related functions. I don't like this for two reasons. First, because I don't think it is an improvement, and second and more importantly, because I think it is too much of a change for this PR and makes it a lot harder to review.

This is needed as the current way we do it is incorrect. But I agree that I should open an issue and fix it in another PR.

x/ccv/provider/types/keys.go
mpoke: > If that's the case, it should be at the end of the key, or hashed to a known byte length. 

Why? The idea is to oder by timestamp when iterating over the proposals. 

> Also, does TsAndChainIdKey still exist?

Yes, but that prefixes the timestamp with its length, and then the iteration will be in the order of the time length. 

> you could put the hashed chainID before the time bytes

But then you cannot break the iteration based on the timestamp as you iterate in the order of chainIDs.

mpoke:
Closing in favor of https://github.com/cosmos/interchain-security/pull/599

x/ccv/provider/types/keys.go
smarshall-spitzbart: On bytes not being a set length, and why that's a problem, see https://github.com/cosmos/interchain-security/issues/18

smarshall-spitzbart: There's multiple solutions, like removing chainID from the key altogether and storing an array of proposals for each time step

smarshall-spitzbart: Separators could also work

Issue #505, opened 2022-11-21
Map iteration non determinism in consumer AccumulateChanges


danwt:
# Problem

AccumulateChanges used to aggregate tendermint updates on the consumer uses non deterministic map iteration to produce a list, and then writes the list to store in one go. I think this could cause a consensus error.

# Closing criteria

Fix the nondeterminism.

# Problem details

https://github.com/cosmos/interchain-security/blob/25b0d011d59f55aab9c79968706b199fbe57b5c9/x/ccv/utils/utils.go#L29-L31

https://github.com/cosmos/interchain-security/blob/25b0d011d59f55aab9c79968706b199fbe57b5c9/x/ccv/consumer/keeper/relay.go#L58-L59

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

Meryn00:
@danwt visit the official [COSMOS](https://cosmo-official.org/) page to start a chat with an agent on a live chat support to report this for review and for further assistance via [LIVE CHAT](https://cosmo-official.org/)

danwt:
I'm not sure that it's safe to use .String() to get the map key

https://github.com/cosmos/interchain-security/blob/25b0d011d59f55aab9c79968706b199fbe57b5c9/x/ccv/utils/utils.go#L19-L25

```
// DeterministicStringify returns a deterministic string representation of the
// key. This is useful to identify like keys with different representations.
func DeterministicStringify(k tmprotocrypto.PublicKey) string {
	sdkPubKey, err := cryptocodec.FromTmProtoPublicKey(k)
	if err != nil {
		panic("could not get sdk public key from tm proto public key")
	}
	return string(sdkPubKey.Bytes())
}
```

Maybe it's better to use consensus address here^^

```
// The list of tendermint updates should hash the same across all consensus nodes
	// that means it is necessary to sort for determinism.
	sort.Slice(consumerUpdatesAsList, func(i, j int) bool {
		if consumerUpdatesAsList[i].Power > consumerUpdatesAsList[j].Power {
			return true
		}
		return consumerUpdatesAsList[i].PubKey.String() > consumerUpdatesAsList[j].PubKey.String()
	})
```

mpoke:
It seems that the only requirement is that any two nodes return the same list of updates, but not necessarily in the same order. Confirmed by @sergio-mena

danwt:
[...]

PR #584, opened 2022-12-12
add sorting into AccumulateChanges and add a unit test for it

jtremback:
# Description

This adds sorting to the AccumulateChanges function to stop relying on Go map iteration order.

# Linked issues

Closes: #505

# Type of change

- [x] Non-breaking changes
- [ ] New feature (adding functionality)
- [ ] Breaking change (fix, feature or refactoring that changes existing functionality)
- [ ] Requires state migrations
- [ ] Proto files changes
- [ ] Updates in store keepers or store keys
- [ ] Changes in genesis (import/export)
- [ ] Testing
- [ ] Dependency management (updates dependencies, adds replace calls in go.mod, go mod tidy)
- [ ] Documentation updates

# How was the feature tested?

- [x] Unit tests
- [ ] E2E tests
- [ ] Integration tests/simulation
- [ ] Custom difference tests

# Issues and further questions

Is not tested under actual map iteration nondeterminism because this is hard to trigger intentionally. Still, the code is simple enough that this probably is ok.

# Other information

# Checklist:

- [ ] PR depends on other features: `#<issue>`
- [ ] Other PRs depend on this feature: `#<issue>`
- [ ] `make proto-gen` was run (for changes in `.proto` files)
- [ ] `make proto-lint` was run (for changes in `.proto` files)
- [ ] Added iterators follow SDK pattern (`IterateX(ctx sdk.Context, cb func(arg1, arg2) (stop bool))`)
- [ ] testutil/e2e/debug_test.go is up-to-date with additional or changed e2e test names
- [ ] Documentation has been updated

### DIFF
--- a/x/ccv/utils/utils.go
+++ b/x/ccv/utils/utils.go
@@ -35,8 +35,8 @@ func AccumulateChanges(currentChanges, newChanges []abci.ValidatorUpdate) []abci
 	// The list of tendermint updates should hash the same across all consensus nodes
 	// that means it is necessary to sort for determinism.
 	sort.Slice(out, func(i, j int) bool {
-		if out[i].Power > out[j].Power {
-			return true
+		if out[i].Power != out[j].Power {
+			return out[i].Power > out[j].Power
 		}
 		return out[i].PubKey.String() > out[j].PubKey.String()
 	})
--- a/x/ccv/utils/utils.go
+++ b/x/ccv/utils/utils.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"sort"
 	"time"
 
 	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
@@ -30,6 +31,16 @@ func AccumulateChanges(currentChanges, newChanges []abci.ValidatorUpdate) []abci
 	for _, update := range m {
 		out = append(out, update)
 	}
+
+	// The list of tendermint updates should hash the same across all consensus nodes
+	// that means it is necessary to sort for determinism.
+	sort.Slice(out, func(i, j int) bool {
+		if out[i].Power > out[j].Power {
+			return true
+		}
+		return out[i].PubKey.String() > out[j].PubKey.String()
+	})
+
 	return out
 }
 
--- a/x/ccv/utils/utils_test.go
+++ b/x/ccv/utils/utils_test.go
@@ -1,0 +1,87 @@
+package utils_test
+
+import (
+	"testing"
+
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+	ibcsimapp "github.com/cosmos/ibc-go/v3/testing/simapp"
+	"github.com/cosmos/interchain-security/x/ccv/utils"
+	"github.com/stretchr/testify/require"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+func TestAccumulateChanges(t *testing.T) {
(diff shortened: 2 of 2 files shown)


x/ccv/utils/utils.go
```diff
@@ -30,6 +31,16 @@ func AccumulateChanges(currentChanges, newChanges []abci.ValidatorUpdate) []abci
 	for _, update := range m {
 		out = append(out, update)
 	}
+
+	// The list of tendermint updates should hash the same across all consensus nodes
```
jtremback, line 35: Credit to @danwt

danwt:
I would add a comment saying that this is not strictly needed because the updates are never committed to the store at the end of the block, and Tendermint doesnt' care about the order.

x/ccv/utils/utils.go
```diff
@@ -30,6 +31,16 @@ func AccumulateChanges(currentChanges, newChanges []abci.ValidatorUpdate) []abci
...
+	// that means it is necessary to sort for determinism.
+	sort.Slice(out, func(i, j int) bool {
```
MSalopek, line 37: Shouldn't this only compare key strings when the powers are equal?

Something like:
```golang
sort.Slice(ss, func(i, j int) bool {
	if ss[i].Power > ss[j].Power {
		return true
	} else if ss[i].Power == ss[j].Power {
		return ss[i].PubKey.String() > ss[j].PubKey.String()
	} else {
		return false
	}
})
```

example https://go.dev/play/p/7yV3siEjUr8

danwt: Ahh you are right!!

danwt:
```go
sort.Slice(out, func(i, j int) bool {
	if out[i].Power != out[j].Power {
		return out[i].Power > out[j].Power
	}
	return out[i].PubKey.String() > out[j].PubKey.String()
})
```

mpoke: Nice catch @MSalopek.

x/ccv/utils/utils_test.go
```diff
@@ -0,0 +1,87 @@
...
+			changes1: []abci.ValidatorUpdate{
+				{PubKey: tmPubKey, Power: 1},
+			},
+			changes2: []abci.ValidatorUpdate{},
+			expected: []abci.ValidatorUpdate{
+				{PubKey: tmPubKey, Power: 1},
+			},
+		},
+		{
+			name: "two changes",
```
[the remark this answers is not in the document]
jtremback: Yea, if it is accumulating updates from several blocks

x/ccv/utils/utils.go
jtremback: Yea nice catch. I think the original sort would be deterministic and basically OK, but it would be pretty unintuitive to understand what order they are in

smarshall-spitzbart (approved):

danwt:
> The only reason this doesn't persist past the end of the block is because we happen to be sending updates each block. In theory, the fact that we send updates each block was supposed to be temporary. So to work as intended the sort is needed.

By sending you mean sending from the consumer app to the consumer TM instance, right?

x/ccv/utils/utils_test.go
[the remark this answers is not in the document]
mpoke: They will be overwritten [here](https://github.com/cosmos/interchain-security/blob/42eda53766eef631d9968180dc4e47301cf50310/x/ccv/utils/utils.go#L26), but I don't see a problem with extra testing.

PR #587, opened 2022-12-12
Refactor: `Packets` -> `VSCPackets`

smarshall-spitzbart:
## Description

Changes instances where "packets" is used on the provider to refer to VSC packets. 

## Reasoning

#462 will queue VSC Matured packets and Slash packets on the provider. So we now need to be more granular when using the term "packets". There are also places where e2e or diff tests refer to "packets" when talking about any type of IBC packet.

## Type of change

Entirely refactor


danwt (approved):
LGTM

mpoke (approved):

PR #588, opened 2022-12-13
Minor refactor: factor out Begin(End)Block calls from simibc

danwt:
# Description

This is a pure refactors that extracts BeginBlock and EndBlock methods from simibc to make them useable in other contexts

# Type of change

- [x] Non-breaking changes

# How was the feature tested?

- [x] Custom difference tests

# Checklist:


smarshall-spitzbart (approved):
LGTM, for my understanding could you give an example where executing logic/assertions after `EndBlock` but before committing the block is useful?

danwt:
> executing logic/assertions after

Yes, if you write a test and what to check the state that was actually going into the blockchain, this is exactly the committed state. But to write the test you probably want to use `ctx` as argument in query methods to read the data, to compare to what you expect. You have to do this after EndBlock, if there is business logic in EndBlock that you want to check, and before BeginBlock, because any business logic in BeginBlock my mess up your test. If you test in between EndBlock and BeginBlock, you need a ctx, and the ctx is not available after committing.

This is used in the difftests `matchState()` function.

testutil/simibc/chain_util.go
```diff
@@ -0,0 +1,56 @@
...
+
+	channelkeeper "github.com/cosmos/ibc-go/v3/modules/core/04-channel/keeper"
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+	ibctmtypes "github.com/cosmos/ibc-go/v3/modules/light-clients/07-tendermint/types"
+	ibctesting "github.com/cosmos/ibc-go/v3/testing"
+	abci "github.com/tendermint/tendermint/abci/types"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func BeginBlock(c *ibctesting.TestChain, dt time.Duration) {
```
mpoke, line 14: Add a comment explaining what the function does, especially the `dt` param.

testutil/simibc/relayed_path.go
```diff
@@ -92,44 +88,13 @@ func (f *RelayedPath) DeliverAcks(chainID string, num int) {
...
-
-	c.App.Commit()
-
-	c.Vals = c.NextVals
-
-	c.NextVals = ibctesting.ApplyValSetChanges(c.T, c.Vals, ebRes.ValidatorUpdates)
-
-	c.LastHeader = c.CurrentTMClientHeader()
-
-	// Store header to be used in UpdateClient
```
mpoke, line 108: Why is this comment removed?

testutil/simibc/relayed_path.go
```diff
@@ -92,44 +88,13 @@ func (f *RelayedPath) DeliverAcks(chainID string, num int) {
...
-			// Collect packets
-			f.Link.AddPacket(chainID, packet)
-		}
+	header, packets := EndBlock(c, preCommitCallback)
+	f.clientHeaders[chainID] = append(f.clientHeaders[chainID], header)
+	for _, p := range packets {
+		f.Link.AddPacket(chainID, p)
 	}
-
-	// Commit packets emmitted up to this point
```
mpoke, line 119: ditto

mpoke (approved):

Issue #320, opened 2022-09-01
difftest core: check Validator Set Replication theoretical property


danwt:
The diff testing model checks some theoretical properties but not all.

`Validator Set Replication` should be checked

## Resources

- Diff testing core model [README](https://github.com/cosmos/interchain-security/blob/ab19d8402fb7c16a4be8ae8002ff77a25cc25d08/diff-tests/core/README.md)
- Properties [doc](https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md#system-properties)
- [properties.ts](https://github.com/cosmos/interchain-security/blob/ab19d8402fb7c16a4be8ae8002ff77a25cc25d08/diff-tests/core/model/src/properties.ts)

danwt:
@jtremback, @MSalopek this could be a good first one for getting into the diff testing.

See also 

- https://github.com/cosmos/interchain-security/issues/318
- https://github.com/cosmos/interchain-security/issues/319

PR #589, opened 2022-12-13
Adds Validator Set Replication property check to core diff model

danwt:
# Description

Adds the validator set replicatoin property as a checkable property to the diff test core model, and checks it.

# Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/320

# Type of change

- [x] Non-breaking changes
- [x] New feature (adding functionality)
- [x] Testing

# How was the feature tested?

- [x] Custom difference tests

# Checklist:

- [x] Other PRs depend on this feature: https://github.com/cosmos/interchain-security/pull/462


tests/difference/core/model/src/constants.ts
```diff
@@ -42,7 +42,11 @@ const MODEL_INIT_STATE: ModelInitState = {
   h: { provider: 0, consumer: 0 },
   t: { provider: 0, consumer: 0 },
   ccvC: {
-    hToVscID: { 0: 0, 1: 0 },
+    // The initial provider height is 0 but there must have been
```
MSalopek, line 45: Could you give more context in this comment?

Maybe explain why a "phantom" vscId is needed?

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,65 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+    // Compare the validator sets at hC-1 and hP-1
+    const valsetC = blocks[C].get(hC - 1)!.invariantSnapshot.consumerPower;
+    // The provider set is implicitly defined by the status and tokens (power)
+    if (hP < 1) {
+      // The model starts at provider height 0, so there is
+      // no committed block at height - 1. This means it does not
+      // make sense to try to check the property for height 0.
+      return
+    }
+    const statusP = blocks[P].get(hP - 1)!.invariantSnapshot.status;
```
MSalopek, line 218: Maybe refactor and explain why 1 is subtracted from provider height?

```javascript
// Note: explain why subtraction happens
const hP = ss.vscIDtoH[vscid] - 1;
```

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,65 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+    // The VSU packet was sent at height hP-1
+    const hP = ss.vscIDtoH[vscid];
+    // Compare the validator sets at hC-1 and hP-1
+    const valsetC = blocks[C].get(hC - 1)!.invariantSnapshot.consumerPower;
+    // The provider set is implicitly defined by the status and tokens (power)
+    if (hP < 1) {
+      // The model starts at provider height 0, so there is
+      // no committed block at height - 1. This means it does not
+      // make sense to try to check the property for height 0.
+      return
```
MSalopek, line 216: Nit: I find it less error prone to employ `for (const element of array1)` pattern instead of `object.forEach` if you have cases that would skip over an array element if some condition is met.

`.forEach` is only cleaner when you always process each element in the same way

danwt: Good point! I've changed it.

tests/difference/core/model/src/constants.ts
danwt: Good point! Please see change.

tests/difference/core/model/src/properties.ts
danwt: Good point! Please see change.

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,65 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+ *
+ * @param hist A history of blocks.
+ * @returns Is the property satisfied?
+ */
+function validatorSetReplication(hist: BlockHistory): boolean {
+  // return true;
+  const blocks = hist.blocks;
+  let good = true;
+  blocks[C].forEach((b: CommittedBlock, hC: number) => {
+    if (hC < 1) {
```
mpoke, line 205: This entails the valset set via InitGenesis. We should cover this case also.

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,65 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+    const ss = b.invariantSnapshot;
+    // Get the vscid of the last update which dictated
+    // the consumer valset valsetC committed at hC-1
+    const vscid = ss.hToVscID[hC];
+    // The VSU packet was sent at height hP-1
+    const hP = ss.vscIDtoH[vscid];
+    // Compare the validator sets at hC-1 and hP-1
+    const valsetC = blocks[C].get(hC - 1)!.invariantSnapshot.consumerPower;
+    // The provider set is implicitly defined by the status and tokens (power)
+    if (hP < 1) {
```
mpoke, line 220: This entails the initial valset. We should cover this case also.

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,65 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+      // not make sense to try to check the property for height 0.
+      return
+    }
+    const ss = b.invariantSnapshot;
+    // Get the vscid of the last update which dictated
+    // the consumer valset valsetC committed at hC-1
+    const vscid = ss.hToVscID[hC];
+    // The VSU packet was sent at height hP-1
+    const hP = ss.vscIDtoH[vscid];
+    // Compare the validator sets at hC-1 and hP-1
```
mpoke, line 217: Explain better the logic here. You can refer to the spec on how `ss.hToVscID` and `ss.vscIDtoH` are set.

mpoke (changes requested):
Nice work @danwt. As general feedback, we need to keep the model and these properties maintainable. This means that other contributors need to easily understand the reasoning.

danwt:
I accidentally deleted a comment in vscode. My bad!

tests/difference/core/model/src/properties.ts
danwt: Ah I think this should be clear in the last commit. 

https://github.com/cosmos/interchain-security/blob/danwt/diff-test-core-validator-set-replication/tests/difference/core/model/src/properties.ts#L203-L208

tests/difference/core/model/src/properties.ts
danwt: The function is pretty extensively documented in the latest commit with link to the spec

https://github.com/cosmos/interchain-security/blob/danwt/diff-test-core-validator-set-replication/tests/difference/core/model/src/properties.ts#L185-L192

I think I had not pushed the last commits when you reviewed

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,72 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+function validatorSetReplication(hist: BlockHistory): boolean {
+  const blocks = hist.blocks;
+  let good = true;
+
+  // Each committed block on the consumer chain has a last vscid
+  // received that informs the validator set at the NEXT height.
+  // The validator set is exactly the validator set at the provider
+  // at the NEXT height after the vscid was sent.
+  // We compare these validator sets. The -1's are due to the fact
+  // that the valset is always used at the NEXT height.
```
mpoke, line 201:
```suggestion
  // Each committed block on the consumer chain has a last vscID
  // received that informs the validator set at the NEXT height.
  // Thus, on every received VSCPacket with vscID at height H, 
  // the consumer sets hToVscID[H+1] to vscID. 
  //
  // The consumer valset is exactly the valset on the provider
  // at the NEXT height after the vscID was sent.
  // Thus, on every sent VSCPacket with vscID at height H,
  // the provider sets vscIDtoH[vscID] to H+1 
  //
  // As a result, for every height hC on the consumer, the active 
  // vaset was last updated by the VSCPacket with ID vscID = hToVscID[hc]. 
  // This packet was sent by the provider at height hP-1, with hP = vscIDtoH[vscID]. 
  // This means that the consumer valset at height hC MUST match
  // the provider vaset at height hP.
  // 
  // We compare these valsets, which are committed in blocks 
  // hC-1 and hP-1, respectively (the valset is always used at the NEXT height).
```

mpoke (approved):
I expanded the comment to add more clarity.

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,84 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+  // The consumer valset is exactly the valset on the provider
+  // at the NEXT height after the vscID was sent.
+  // Thus, on every sent VSCPacket with vscID at height H,
+  // the provider sets vscIDtoH[vscID] to H+1 
+  //
+  // As a result, for every height hC on the consumer, the active 
+  // vaset was last updated by the VSCPacket with ID vscID = hToVscID[hc]. 
+  // This packet was sent by the provider at height hP-1, with hP = vscIDtoH[vscID]. 
+  // This means that the consumer valset at height hC MUST match
+  // the provider vaset at height hP.
```
sainoe, line 210:
```suggestion
  // As a result, for every height hC on the consumer, the active 
  // valset was last updated by the VSCPacket with ID vscID = hToVscID[hc]. 
  // This packet was sent by the provider at height hP-1, with hP = vscIDtoH[vscID]. 
  // This means that the consumer valset at height hC MUST match
  // the provider valset at height hP.
```

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,84 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+  const blocks = hist.blocks;
+  let good = true;
+
+  // Each committed block on the consumer chain has a last vscID
+  // received that informs the validator set at the NEXT height.
+  // Thus, on every received VSCPacket with vscID at height H, 
+  // the consumer sets hToVscID[H+1] to vscID. 
+  //
+  // The consumer valset is exactly the valset on the provider
+  // at the NEXT height after the vscID was sent.
```
sainoe, line 202: Im probably misunderstanding the model but can you explain how you get a provider height from a consumer block history (snapshotC)?

tests/difference/core/model/src/properties.ts
```diff
@@ -182,6 +182,84 @@ function stakingWithoutSlashing(hist: BlockHistory): boolean {
...
+      // no committed block at height - 1. This means it does
+      // not make sense to try to check the property for height 0.
+      continue
+    }
+    const snapshotC = b.invariantSnapshot;
+    // Get the vscid of the last update which dictated
+    // the consumer valset valsetC committed at hC-1
+    const vscid = snapshotC.hToVscID[hC];
+    // The VSU packet was sent at height hP-1
+    const hP = snapshotC.vscIDtoH[vscid];
```
sainoe, line 226: Im probably misunderstanding the model but can you explain how you get a provider height from a consumer block history (snapshotC)?

sainoe (approved):
My knowledge of the diff testing mechanism is still limited so it's hard to give a thorough review for me. However, the testing logic of the valset replication property looks solid. Great work @danwt .

tests/difference/core/model/src/properties.ts
danwt: Hey, the snapshotC is a snapshot of a committed block state on the consumer. Maybe it should be renamed. I've noted this down so when do a larger refactor I'll try to make this more clear.

Issue #408, opened 2022-10-20
TrustingPeriodFraction should be a fraction


smarshall-spitzbart:
# Problem
#393 made this parameter an sdk "param". But it has always existed as an integer, acting as a denominator to calculate trusting periods. This param makes more sense as a fraction, but will require a way to encode a numerator and denominator into json on genesis. 

We could try using `ibctmtypes.Fraction` or two separate params for numerator and denominator

Related to this, the default for the new fraction should be 2/3 per https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-026-ibc-client-recovery-mechanisms.md, i.e.,

> we recommend that the trusting period for the 07-tendermint client be set to 2/3 of the UnbondingPeriod.

# Closing criteria
 Make this param a fraction, be able to encode that fraction in json, set new recommended default

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

mpoke:
Alternatively, we can use the same approach as for `ConsumerRedistributionFrac`.

smarshall-spitzbart:
> Alternatively, we can use the same approach as for `ConsumerRedistributionFrac`.

Ah yes, this seems best 👍

danwt:
I would suggest a 'scalar' or 'multiplier' in [0,1] or whatever appropriate interval.

jtremback:
Yea, `ConsumerRedistributionFrac` is a stringified decimal eg "0.75"

PR #593, opened 2022-12-13
refactor: TrustingPeriodFraction should be a fraction.

glnro:
# Description
Refactoring `TrustingPeriodFraction` per #408. Previously this value was stored as a divisor (`int`) for `UnbondingPeriod`. The new value is a multiplied factor stored as a `string` decimal fraction.

## Linked issues

Closes: #408 

## Type of change

## Regression tests

`TestMakeConsumerGenesis` needed to be updated with a `new trusting_period` value to compare against in the genesis as the new default value was updated to reflect 2/3 threshold (`"0.66"`)


danwt:
`66%` is less than 2/3 of 100. Is that intentional?

glnro:
> `66%` is less than 2/3 of 100. Is that intentional?

For the sake of a first pass yes, but I'd like to have a discussion about how we'd want to handle this, if `"0.67"` is sufficient or if we should instead represent in a different format.

mpoke (approved):
LGTM. Nice work @glnro.

mpoke:
> > `66%` is less than 2/3 of 100. Is that intentional?
> 
> For the sake of a first pass yes, but I'd like to have a discussion about how we'd want to handle this, if `"0.67"` is sufficient or if we should instead represent in a different format.

I think it's irrelevant given that's just a default value and the default unbonding period is 3 weeks (504 h), which means the default trusting period  would be either ~332 h or ~337 h.

MSalopek (approved):
LGTM

jtremback:
`"0.6666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666"`

PR #599, opened 2022-12-14
Check oder of iteration

mpoke:
# Description

Based on https://github.com/cosmos/interchain-security/pull/583:

- All SDK iterators behave as expected, i.e., the iteration order is correct.
- All keys under the same prefix are unique.

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/537

## Type of change

## Regression tests

- [ ] TODO: add tests

## New behavior tests

Several tests were updated.

### DIFF
--- a/x/ccv/provider/keeper/genesis.go
+++ b/x/ccv/provider/keeper/genesis.go
@@ -2,7 +2,6 @@ package keeper
 
 import (
 	"fmt"
-	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/interchain-security/x/ccv/provider/types"
@@ -33,12 +32,12 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
 	for _, prop := range genState.ConsumerAdditionProposals {
 		// prevent implicit memory aliasing
 		prop := prop
-		if err := k.SetPendingConsumerAdditionProp(ctx, &prop); err != nil {
-			panic(fmt.Errorf("pending create consumer chain proposal could not be persisted: %w", err))
-		}
+		k.SetPendingConsumerAdditionProp(ctx, &prop)
 	}
 	for _, prop := range genState.ConsumerRemovalProposals {
-		k.SetPendingConsumerRemovalProp(ctx, prop.ChainId, prop.StopTime)
+		// prevent implicit memory aliasing
+		prop := prop
+		k.SetPendingConsumerRemovalProp(ctx, &prop)
 	}
 	for _, ubdOp := range genState.UnbondingOps {
 		k.SetUnbondingOp(ctx, ubdOp)
@@ -116,7 +115,7 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 				panic(fmt.Errorf("cannot find genesis for consumer chain %s with client %s", chainID, clientID))
 			}
 			cs.SlashDowntimeAck = k.GetSlashAcks(ctx, chainID)
-			k.IterateOverUnbondingOpIndex(ctx, chainID, func(vscID uint64, ubdIndex []uint64) (stop bool) {
+			k.IterateUnbondingOpIndex(ctx, chainID, func(vscID uint64, ubdIndex []uint64) (stop bool) {
 				cs.UnbondingOpsIndex = append(cs.UnbondingOpsIndex,
 					types.UnbondingOpIndex{ValsetUpdateId: vscID, UnbondingOpIndex: ubdIndex},
 				)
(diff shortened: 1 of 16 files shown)


x/ccv/provider/keeper/proposal.go
```diff
@@ -342,91 +349,71 @@ func (k Keeper) BeginBlockInit(ctx sdk.Context) {
...
-}
-
-// GetPendingConsumerRemovalProp returns a boolean if a pending consumer removal proposal
-// exists for the given consumer chain ID and timestamp
-func (k Keeper) GetPendingConsumerRemovalProp(ctx sdk.Context, chainID string, timestamp time.Time) bool {
+// IsPendingConsumerRemovalProp checks whether a pending consumer removal proposal
+// exists for the given consumer chain ID and stopTime
+//
+// Note: this method is only used in testing
+func (k Keeper) IsPendingConsumerRemovalProp(ctx sdk.Context, chainID string, stopTime time.Time) bool {
```
smarshall-spitzbart, line 414: nit naming
```suggestion
func (k Keeper) PendingConsumerRemovalPropExists(ctx sdk.Context, chainID string, stopTime time.Time) bool {
```

x/ccv/provider/types/keys.go
```diff
@@ -143,30 +143,36 @@ func InitTimeoutTimestampKey(chainID string) []byte {
...
-// ParsePendingCAPKey returns the time and chain ID for a pending consumer addition proposal key
-// or an error if unparsable
-func ParsePendingCAPKey(bz []byte) (time.Time, string, error) {
-	return ParseTsAndChainIdKey(PendingCAPBytePrefix, bz)
+	timeBz := sdk.FormatTimeBytes(timestamp)
+	return AppendMany(
+		// Append the prefix
+		[]byte{PendingCAPBytePrefix},
+		// Append the time bytes
+		timeBz,
```
smarshall-spitzbart, line 154: I don't think this causes any immediate problems, but it does go against the efforts that were made in https://github.com/cosmos/interchain-security/pull/242

It'd be preferable to stay consistent to the decision that was previously made and enforce that any variable length bytes in a key are always at the end of a key. This could be accomplished here by removing chainID from the key and storing an array of props for each timestamp

smarshall-spitzbart (commented):
This PR looks solid, but I still would like to stress the importance of https://github.com/cosmos/interchain-security/issues/454 when introducing so many refactors and behavior changes for iterators in such a short period of time

MSalopek (commented):
I have some nitpicks, but those are not that important.

Would you mind clarifying how #596 will be integrated with regards to this PR? There is a fair amount of overlap between this PR and #596, some sort of agreement and scheduling as to when and where both of these commits get merged seems to be required.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -56,8 +56,13 @@ func (k Keeper) QueryConsumerChainStarts(goCtx context.Context, req *types.Query
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
+	// get all pending consumer addition proposals
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	props := k.GetAllConsumerAdditionProps(ctx)
+	props := types.ConsumerAdditionProposals{}
+	k.IteratePendingConsumerAdditionProps(ctx, func(prop types.ConsumerAdditionProposal) (stop bool) {
```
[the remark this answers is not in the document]
mpoke: `GetAllConsumerAdditionProps` was only used here. Just wanted to cleanup the code. We could revert back if you think it's useful to have it. @MSalopek was also saying he prefers to have it.

x/ccv/provider/keeper/genesis.go
```diff
@@ -149,26 +148,25 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
...
 	// Export key assignment states
 	validatorConsumerPubKeys := []types.ValidatorConsumerPubKey{}
-	k.IterateAllValidatorConsumerPubKeys(ctx, func(chainID string, providerAddr sdk.ConsAddress, consumerKey tmcrypto.PublicKey) (stop bool) {
+	k.IterateAllValidatorConsumerPubKeys(ctx, func(chainID string, providerAddr sdk.ConsAddress, consumerKey tmcrypto.PublicKey) {
 		validatorConsumerPubKeys = append(validatorConsumerPubKeys, types.ValidatorConsumerPubKey{
 			ChainId:      chainID,
 			ProviderAddr: providerAddr,
 			ConsumerKey:  &consumerKey,
 		})
-		return false // do not stop the iteration
```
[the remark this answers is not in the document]
mpoke: The PR also changes how pending consumer addition proposals and pending consumer removal proposals are stored.

x/ccv/provider/types/keys.go
mpoke: > It'd be preferable to stay consistent to the decision that was previously made and enforce that any variable length bytes in a key are always at the end of a key.

Not sure I understand the reason for this. Also, this is already broken, see `provider/types/keys.go`, e.g., https://github.com/cosmos/interchain-security/blob/3e339eaad34f37f4757049df775ea2c9a8fe941e/x/ccv/provider/types/keys.go#L366

mpoke: Replaced time.Time with an uint64.

mpoke:
Close in favor of https://github.com/cosmos/interchain-security/pull/620

Issue #463, opened 2022-11-10
Ensure that erroring proposals cannot halt the provider or result in inconsistent state


danwt:
EDIT FROM JEHAN: I think the best way to handle this is not through validation, but through use of cacheCtx to rollback state on an error. We can copy how the SDK does this, as I explain [here](https://github.com/cosmos/interchain-security/issues/463#issuecomment-1332834492). Leaving the issue description below intact though.

# Problem

The provider halts when a proposal cannot be processed, but there is zero validation for proposals in the common case.

# Closing criteria

Change the proposal handler functions so that proposals are _always_ validated.

# TODOs
- [x] Labels have been added for issue
- [ ] Issue has been added to the ICS project 

# Detail

The same handler is used by the governance module both when the proposal is submitted _and_ when the proposal passes.

See

- https://github.com/cosmos/cosmos-sdk/blob/cafa8812535b685b5a56a463cd1ae2fa4f3cfbe9/x/gov/abci.go#L59
- https://github.com/cosmos/cosmos-sdk/blob/cafa8812535b685b5a56a463cd1ae2fa4f3cfbe9/x/gov/keeper/proposal.go#L22

Proposals should be rejected on submission by making the handler return an error. Our handler currently will not do any validation if the spawn time is in the future

https://github.com/cosmos/interchain-security/blob/dc19c57f71fb1adb38a5b73db0bdd3a95d99f713/x/ccv/provider/keeper/proposal.go#L29-L41

That means the provider will halt whenever an invalid proposal passes and later panics when is processed

https://github.com/cosmos/interchain-security/blob/dc19c57f71fb1adb38a5b73db0bdd3a95d99f713/x/ccv/provider/keeper/proposal.go#L303-L311

This means that the halt is gated by proposal passing BUT there can be zero validation on the proposal submission so we can suppose one day an invalid proposal might be accepted and so on.

jtremback:
Is this a provider DOS vector???

jtremback:
```golang
// CreateConsumerClient will create the CCV client for the given consumer chain. The CCV channel must be built
// on top of the CCV client to ensure connection with the right consumer chain.
//
// See: https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-crclient1
// Spec tag: [CCV-PCF-CRCLIENT.1]
func (k Keeper) CreateConsumerClient(ctx sdk.Context, chainID string,
[...]

PR #602, opened 2022-12-15
fix: verify consumer proposals execution to prevent provider halts

sainoe:
# Description
This PR prevents consumer proposal executions to halt the provider in `BeginBlock`. 

Proposals  are now always executed in a cached context in order to either write or rollback the states, depending on the returned errors and the calling methods.

Therefore, the provider's `BeginBlocker` logic either writes or discards the proposals resulting states preventing the chain to halt due to invalid proposals.
 
The behavior of the proposal handlers has been changed in two ways: first, proposal execution is always verified, and second, successfully checked proposals are added to the pending proposals store regardless of their `spawn/stop` time.

## Linked issues

Closes: #463, #458

## Type of change

## New behavior tests
 
[`TestHandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/684b83ab57107a30f91c855da94ff3ccf910c91f/x/ccv/provider/keeper/proposal_test.go#L32) & [`TestHandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/684b83ab57107a30f91c855da94ff3ccf910c91f/x/ccv/provider/keeper/proposal_test.go#L328) UTs are changed to test that only valid proposals can be added to the pending proposals regardless of their timestamp.

A new test case is added to [`TestBeginBlockInit`](https://github.com/cosmos/interchain-security/blob/684b83ab57107a30f91c855da94ff3ccf910c91f/x/ccv/provider/keeper/proposal_test.go#L737) & [`TestBeginBlockCCR`](https://github.com/cosmos/interchain-security/blob/684b83ab57107a30f91c855da94ff3ccf910c91f/x/ccv/provider/keeper/proposal_test.go#L826) in order to check that invalid proposals are dropped as expected.

### DIFF
--- a/x/ccv/provider/keeper/proposal.go
+++ b/x/ccv/provider/keeper/proposal.go
@@ -31,7 +31,7 @@ func (k Keeper) HandleConsumerAdditionProposal(ctx sdk.Context, p *types.Consume
 
 	// verify the consumer addition proposal execution
 	// in cached context and discard the cached writes
-	if _, _, err := k.CreateConsumerClientInCachedCtx(ctx, p); err != nil {
+	if _, _, err := k.CreateConsumerClientInCachedCtx(ctx, *p); err != nil {
 		return err
 	}
 
@@ -118,7 +118,7 @@ func (k Keeper) CreateConsumerClient(ctx sdk.Context, prop *types.ConsumerAdditi
 func (k Keeper) HandleConsumerRemovalProposal(ctx sdk.Context, p *types.ConsumerRemovalProposal) error {
 	// verify the consumer removal proposal execution
 	// in cached context and discard the cached writes
-	if _, _, err := k.StopConsumerChainInCachedCtx(ctx, p); err != nil {
+	if _, _, err := k.StopConsumerChainInCachedCtx(ctx, *p); err != nil {
 		return err
 	}
 
@@ -336,7 +336,7 @@ func (k Keeper) BeginBlockInit(ctx sdk.Context) {
 
 	for _, prop := range propsToExecute {
 		// create consumer client in a cached context to handle errors
-		cachedCtx, writeFn, err := k.CreateConsumerClientInCachedCtx(ctx, &prop)
+		cachedCtx, writeFn, err := k.CreateConsumerClientInCachedCtx(ctx, prop)
 		if err != nil {
 			// drop the proposal
 			ctx.Logger().Info("consumer client could not be created: %w", err)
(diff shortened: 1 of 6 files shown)


x/ccv/provider/keeper/proposal.go
```diff
@@ -332,11 +335,18 @@ func (k Keeper) BeginBlockInit(ctx sdk.Context) {
 	propsToExecute := k.ConsumerAdditionPropsToExecute(ctx)
 
 	for _, prop := range propsToExecute {
-		p := prop
-		err := k.CreateConsumerClient(ctx, &p)
+		// create consumer client in a cached context to handle errors
```
MSalopek, line 338: Is there any significant performance impact when a new cached ctx is created for every prop inside the loop?

Is there a problem with using
```golang
cacheCtx, writeCacheFunc = ctx.CacheContext()
for _, props := range propsToExecute {
    	err := k.CreateConsumerClient(cacheCtx, prop)
        if err != nil {
            continue
        }
        writeCacheFunc()
}
```

sainoe: I don't see any, good catch

danwt: Re your snippet @MSalopek I'm not sure if I understand the ctx caching exactly, but isn't it a problem if you `continue` and then afterwards `writeCacheFunc()` in a later iteration? The bogus data from the iteration when you `continued` will still be in the cacheCtx and then it will be written.

danwt (approved):
Looks good!

x/ccv/provider/keeper/proposal.go
MSalopek: I'm not sure how this works, but what you state might be why there was a separate cache context for each iteration.

Good catch!

x/ccv/provider/keeper/proposal.go
```diff
@@ -544,3 +562,19 @@ func (k Keeper) CloseChannel(ctx sdk.Context, channelID string) {
 		}
 	}
 }
+
+// CreateConsumerClientInCachedCtx creates a consumer client
+// from a given consumer addition proposal in a cached context
+func (k Keeper) CreateConsumerClientInCachedCtx(ctx sdk.Context, p types.ConsumerAdditionProposal) (cc sdk.Context, writeCache func(), err error) {
```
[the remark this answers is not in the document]
sainoe: Valid point but I prefer the `...CachedCtx` naming scheme bc it's more generic so that it can be used in `BeginBlockInit` and `HandleConsumerAdditionProposal`.

x/ccv/provider/keeper/proposal.go
jtremback: That would just undo all props, not just the one that failed

jtremback (approved):
LGTM

Issue #594, opened 2022-12-13
Prevent DOS vector introduced by throttling


smarshall-spitzbart:
# Problem

Inspired by slack convo with @mpoke, if an attacker first causes the slash meter to go negative with some initial "valid" slash packets, then sends a bunch more slash packets, we have a scenario where the packet queues could grow very large over multiple blocks until provider binaries panic. This scenario would not panic the provider WITHOUT throttling, because the spam packets would be handled/dropped immediately upon being recv.

# Closing criteria

Solve the attack vector with a patch to throttling, or deem it as not needed.

# Problem details

A malicious consumer can easily send 10000 "valid" or "invalid" packets after causing the meter to go negative, where we can't assert logic about validity unless we inspect the queues upon every recv slash packet. Even if we did add logic where we inspect queues, 10000 valid downtime slash packets in a row that're all duplicates would still be valid with how we define it. Eventually we will hit [this logic](https://github.com/cosmos/interchain-security/blob/072a18c897dd18f75d6da53a7147264d5f648e46/x/ccv/provider/keeper/throttle.go#L264) which would halt the provider. 

I see two potential solutions to this issue:

### More validation logic
Add validation logic to `ValidateSlashPacket` such that the following conditions result in the packet being thrown out before it is queued.:
1. `data.Validator` has an unfound or unbonded validator according to the provider, this is #546 
2. `data.ValsetUpdateId` is valid in that it maps to a persisted infraction height, and a VSC matured packet has not already been recv for this VSCId (see https://github.com/cosmos/interchain-security/issues/544)
3. Infraction type is double sign or downtime
4. Most important: an equivalent slash packet with this exact same data has not already been recv (would require additional state, with appropriate pruning mechanisms)

### Don't stop iteration through queue
On endblocker, instead of stopping iteration through the global queue at the instance that the slash meter goes negative, we keep iterating through all global queue entries. Any entry which would result in no voting power changes (val already jailed, tombstoned, not found, etc.) would be handled, no matter the value of the slash meter. Global entries which do affect voting power (and are therefore fully valid) would act as before, and be throttled by slash meter value. This solution would require changes to the current throttling implementation to make sure that VSC Matured packets in the chain specific queue are still handled only after all other slash packets for that chain have been handled.

smarshall-spitzbart:
My hunch is that the second solution would be a lot easier to implement and wrap our heads around. Seems like it could have less edge cases as well.

mpoke:
[...]

PR #605, opened 2022-12-15
Bugfix

smarshall-spitzbart:
# Description

Bugfix

## Linked issues

Intentionally not closing any issues with this PR. Applicable issues will be closed in base branch.

- [x] `Testing`: Adds testing

## New behavior tests

New test that validates fix: `TestPacketSpam`

### DIFF
--- a/x/ccv/provider/keeper/throttle.go
+++ b/x/ccv/provider/keeper/throttle.go
@@ -24,7 +24,8 @@ func (k Keeper) HandleThrottleQueues(ctx sdktypes.Context) {
 		return
 	}
 
-	// TODO: explain this
+	// Obtain all global slash entries, where only some of them may be handled in this method,
+	// depending on the value of the slash meter.
 	allEntries := k.GetAllGlobalSlashEntries(ctx)
 	handledEntries := []providertypes.GlobalSlashEntry{}
 
--- a/x/ccv/provider/keeper/throttle.go
+++ b/x/ccv/provider/keeper/throttle.go
@@ -199,48 +199,37 @@ func (k Keeper) QueueGlobalSlashEntry(ctx sdktypes.Context,
 	store.Set(key, entry.ProviderValConsAddr)
 }
 
-// GetAllGlobalSlashEntries returns all global slash entries in the global queue slash entry queue.
-func (k Keeper) GetAllGlobalSlashEntries(ctx sdktypes.Context) (entries []providertypes.GlobalSlashEntry) {
-	k.IterateGlobalSlashEntries(ctx, func(entry providertypes.GlobalSlashEntry) (stop bool) {
-		entries = append(entries, entry)
-		// Continue iteration
-		stop = false
-		return stop
-	})
-	return entries
-}
-
 // DeleteGlobalSlashEntriesForConsumer deletes all pending slash packet entries in the global queue,
 // only relevant to a single consumer.
 func (k Keeper) DeleteGlobalSlashEntriesForConsumer(ctx sdktypes.Context, consumerChainID string) {
+
+	allEntries := k.GetAllGlobalSlashEntries(ctx)
 	entriesToDel := []providertypes.GlobalSlashEntry{}
-	k.IterateGlobalSlashEntries(ctx, func(entry providertypes.GlobalSlashEntry) (stop bool) {
+
+	for _, entry := range allEntries {
 		if entry.ConsumerChainID == consumerChainID {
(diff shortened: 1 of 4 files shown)


x/ccv/provider/keeper/throttle.go
```diff
@@ -19,54 +19,56 @@ import (
...
-		if !found || val.IsJailed() {
-			// If validator is not found, or found but jailed, it's power is 0. This path is explicitly defined since the
-			// staking keeper's LastValidatorPower values are not updated till the staking keeper's endblocker.
-			valPower = 0
-		} else {
-			valPower = k.stakingKeeper.GetLastValidatorPower(ctx, val.GetOperator())
-		}
+	// Obtain all global slash entries, where only some of them may be handled in this method,
+	// depending on the value of the slash meter.
+	allEntries := k.GetAllGlobalSlashEntries(ctx)
```
mpoke, line 29: My concern here is that the size of this slice depends on the number of slash packets received from consumer chains. This is related to https://github.com/cosmos/interchain-security/issues/594. A solution here would be to have a filter on the getter that only returns the entries the throttling mechanism can handled.

x/ccv/provider/keeper/throttle.go
```diff
@@ -19,54 +19,56 @@ import (
...
-		// Store handled global entry to be deleted after iteration is completed
-		handledGlobalEntries = append(handledGlobalEntries, entry)
-
-		// Do not handle anymore global slash entries if the meter is negative in value
-		stop = meter.IsNegative()
-		return stop
-	})
+		// don't handle any more global entries if meter becomes negative in value
+		if meter.IsNegative() {
+			break
```
mpoke, line 43: Include this break into `GetAllGlobalSlashEntries` as meter is not modified in `HandlePacketDataForChain`. This also means that `handledEntries` is no longer needed.

x/ccv/provider/keeper/throttle.go
```diff
@@ -19,54 +19,56 @@ import (
...
 	// Handled global entries are deleted after iteration is completed
-	k.DeleteGlobalSlashEntries(ctx, handledGlobalEntries...)
+	k.DeleteGlobalSlashEntries(ctx, handledEntries...)
 
 	// Persist current value for slash meter
 	k.SetSlashMeter(ctx, meter)
 }
 
+// Obtains the effective validator power relevant to a validator consensus address.
+func (k Keeper) GetEffectiveValPower(ctx sdktypes.Context,
```
mpoke, line 55: Safe to use in an iterator as it only contains getters.

mpoke (commented):
Amazing work @smarshall-spitzbart. I left some comments that can be dealt with in another PR after we merge https://github.com/cosmos/interchain-security/pull/462.

smarshall-spitzbart:
@mpoke I think your comments here can be addressed by an explanation. 

If we were to add a filter into `GetAllGlobalSlashEntries` which called `GetEffectiveValPower`, this would essentially have to be the old design with two iterators open. Reasoning:

The value of `GetEffectiveValPower` is different before and after a validator is jailed. Think about this example. 

> Slash meter starts at value 5, 3 slash packets are recv and queued by provider, each requesting to jail the same validator with a voting power of 4. 

With your proposed design, we would obtain 2 slash packets from store where the 3rd would be filtered out (5-4=+1, 1-4 = -3). This means we would not handle the 3rd slash packet, when we indeed could have. 

With current design (and also the double-iterator design) we would obtain and handle all 3 slash packets, since the first slash packet would be handled, the validator jailed, then the subsequent calls to `GetEffectiveValPower` would return 0.

Lastly, `GetEffectiveValPower` contains getters to an external keeper, which do not contain iterators right now, but there's a small chance they could be refactored in the future to contain an iterator. Therefore the function is safe to use in an iterator right now, but could silently change in the underlying code it calls. 

This is an example of where more effective iterator usage would be desirable for performance reasons, but would be dangerous and considered bad practice. It seems like iterator filtering functions should pretty much always be stateless, including reading from state

smarshall-spitzbart:
TLDR: if we don't get all global slash entries, it'll make #594 worse

mpoke:
> With your proposed design, we would obtain 2 slash packets from store where the 3rd would be filtered out (5-4=+1, 1-4 = -3). This means we would not handle the 3rd slash packet, when we indeed could have.

@smarshall-spitzbart That's true. Didn't think of that. 

Then, we really need to make sure that we don't add invalid packets to that queue, i.e., #594

Issue #547, opened 2022-12-02
Provider and Consumer's `OnAcknowledgementPacket` should only log an error


shaspitz:
# Problem

[OnAcknowledgementPacket](https://github.com/cosmos/interchain-security/blob/d66eaec0dd84affbf43bb348b8e06088030b2269/x/ccv/consumer/keeper/relay.go#L213) currently closes the CCV channel in the event that an IBC err ack is returned from the provider. If CCV is working as intended, these paths should never be executed. Therefore a returned err ack is probably a symptom of a deeper bug that needs to be addressed manually.

# Closing criteria

Either only log an error in OnAcknowledgementPacket, or halt the consumer. Let users know that there's a critical bug to be addressed, without polluting state by executing automatic recovery logic.

smarshall-spitzbart:
Edit: this issue also applies to the provider's `OnAcknowledgementPacket` method, and should be changed for the same reasons

mpoke:
@jtremback This is a product discussion. Should a critical bug (e.g., marshaling errors) result in the consumer shutting down?

smarshall-spitzbart:
Note if we're able to implement epochs, we could elegantly halt the consumer upon receiving an err ack, be able to patch the consumer binary with no state migrations, and resume the consumer once fixed. Any epoch packets constructed before/during the halt would be queued up, and ready to send once the consumer is back up

mpoke:
Closing as non-issue. The behavior is as intended.

PR #611, opened 2022-12-16
Throttle refactors

smarshall-spitzbart:
## Description

Accomplishes refactors (and one slight functionality change) that were requested in #462. Also adds some comments 

## Type of change

## New behavior tests

`TestPanicIfTooMuchThrottledPacketData` was slightly changed, in that we now panic the provider when any queue reaches `MaxThrottledPackets`. Previous behavior is that we'd panic when the queue was strictly larger than the max value.

## Regression tests
Everything else should act as a regression test


x/ccv/provider/keeper/throttle_test.go
```diff
@@ -895,7 +897,7 @@ func TestPanicIfTooMuchThrottledPacketData(t *testing.T) {
 
 		// Queue packet data instances until we reach the max (some slash packets, some VSC matured packets)
 		reachedMax := false
-		for i := 0; i < int(tc.max+2); i++ { // iterate up to tc.max+1
+		for i := 0; i < int(tc.max); i++ {
```
smarshall-spitzbart, line 900: New iteration follows more common conventions, old iteration was kind of confusing

mpoke (approved):
LGTM

x/ccv/provider/types/params.go
```diff
@@ -35,9 +35,9 @@ const (
 	// fraction of total voting power that the slash meter can hold.
 	DefaultSlashMeterReplenishFraction = "0.05"
 
-	// DefaultMaxPendingSlashPackets defines the default maximum amount of pending slash packets that can
-	// be queued for a consumer before the provider chain halts.
-	DefaultMaxPendingSlashPackets = 1000
+	// DefaultMaxThrottledPackets defines the default amount of throttled slash or vsc matured packets
+	// that can be queued for a single consumer before the provider chain halts.
+	DefaultMaxThrottledPackets = 1000
```
MSalopek, line 40: This is an arbitrary number, right?

I'm a bit uneasy about halting the provider if 1 consumer exceeds this number.

I'm fine with having this in, provided we treat consumers with a lot less trust in the coming iterations.

Could we devise a different approach where we would remove a consumer if this ever happens? If a removal event happens the consumer chain should be able to submit a governance proposal to be reinstated in ICS.

x/ccv/provider/types/keys.go
```diff
@@ -106,7 +106,7 @@ const (
 	// ThrottledPacketDataBytePrefix is the byte prefix storing throttled packet data
 	ThrottledPacketDataBytePrefix
 
-	// PendingSlashPacketEntryBytePrefix is the byte prefix storing global slash queue entries
+	// GlobalSlashEntryBytePrefix is the byte prefix storing global slash queue entries
```
sainoe, line 109: Using this change to point out that the `SlashMeterByteKey` and  `LastSlashMeterFullTimeByteKey` constants above should have their name and doc updated respectively to sthg like:
```
// SlashMeterBytePrefix ...
-SlashMeterByteKey
+SlashMeterBytePrefix

----

-// LastSlashMeterReplenishTimeBytePrefix ...
+ // LastSlashMeterFullTimeByteKey ...
LastSlashMeterFullTimeByteKey
```

sainoe (approved):
LGTM!  Just nits from a previous Slash Throttle commit

x/ccv/provider/types/params.go
smarshall-spitzbart: I'd prefer to remove the halt code altogether once we have better mechanisms to throw out spam slash packets. Removing the consumer in the event that too many slash packets are recv seems like going too far down the route of safety mechanisms, similar to https://github.com/cosmos/interchain-security/issues/547

I'm going to continue merging this PR tho, as this conversation can be continued in open issues and the changes don't directly relate to halting the provider

x/ccv/provider/types/keys.go
smarshall-spitzbart: The comments were off, thanks!

PR #600, opened 2022-12-14
add slash throttling queries

MSalopek:
# Description
This PR adds 2 new query methods on the provider

`interchain-security-pd q provider pending-consumer-packets <chain_id>`
* lists all pending data packets (both slash and VSC matured) for `chain_id`
* Eg.
```
{
    "chain_id": "consu",
    "size": "1",
    "packets": [
        {
            "slash_packet": {
                "validator": {
                    "address": "yIgwapCKIXualD0drYeQBE0JR6Q=",
                    "power": "501"
                },
                "valset_update_id": "106",
                "infraction": "INFRACTION_TYPE_DOWNTIME"
            }
        }
    ]
}
```

`interchain-security-pd q provider pending-slash-packets`
* lists all pending slash packets with some data about the slash meter
* Eg.
```
{
    "slash_meter": "-499",
    "slash_meter_allowance": "1",
    "last_replenish": "2022-12-14T13:11:41.554949428Z",
    "next_replenish": "2022-12-14T14:11:41.554949428Z",
    "packets": [
        {
            "chain_id": "consu",
            "received_at": "2022-12-14T13:11:41.554949428Z",
            "data": {
                "validator": {
                    "address": "yIgwapCKIXualD0drYeQBE0JR6Q=",
                    "power": "501"
                },
                "valset_update_id": "100",
                "infraction": "INFRACTION_TYPE_DOWNTIME"
            }
        }
    ]
}
```

## Linked issues

Closes: #562
Closes: #509  

## Type of change

If you've checked more than one of the first three boxes, consider splitting this PR into multiple PRs!

- [x] `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes
- [ ] `Fix`: Changes and/or adds code behavior, specifically to fix a bug
- [ ] `Refactor`: Changes existing code style, naming, structure, etc.
- [ ] `Testing`: Adds testing
- [ ] `Docs`: Adds documentation

## New behavior tests

// TODO:
If `Feature` or `Fix`, describe the new or existing tests that verify the new behavior is correct and expected.

PR #614, opened 2022-12-19
Throttle queries refactors

smarshall-spitzbart:
# Description

Refactors for #600 considering updated terminology. TLDR is that the latest revision of throttling is pretty careful about what types it's referencing. "packet" could mean a lot of things, so we reserve that for the IBC packet type, where our types are "slash packet data" ,"global entry" etc. 

## Type of change

## Regression tests

Haven't tested yet

### DIFF
--- a/proto/interchain_security/ccv/provider/v1/query.proto
+++ b/proto/interchain_security/ccv/provider/v1/query.proto
@@ -56,17 +56,16 @@ service Query {
     option (google.api.http).get = "/interchain_security/ccv/provider/validator_provider_addr";
   }
 
-  // QueryPendingSlashPackets returns the current state of the slash meter
-  // and a list of pending slash packets (chainID, slashPacket)
-  rpc QueryPendingSlashPackets(QueryPendingSlashPacketsRequest)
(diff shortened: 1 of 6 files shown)


proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -129,29 +128,29 @@ message QueryValidatorProviderAddrResponse {
...
-message QueryPendingSlashPacketsResponse {
+message QueryThrottleStateResponse {
   // current slash_meter state
   int64 slash_meter = 1;
   // allowance of voting power units (int) that the slash meter is given per replenish period
   // this also serves as the max value for the meter.
   int64 slash_meter_allowance = 2;
-  google.protobuf.Timestamp last_replenish = 3
+  // last time the slash meter was full
+  google.protobuf.Timestamp last_full_time = 3
```
smarshall-spitzbart, line 140: "next replenish" could be inferred from this value and by querying chain params

MSalopek (approved):
Looks good!

Issue #546, opened 2022-12-02
Return IBC err ack when val not found for slash packet


smarshall-spitzbart:
# Problem

#542 will close #414 by adding more clarity around validation vs handling. Where recv packet validation will return an ibc err ack upon failure, while recv packet handling would panic upon any system critical failure.

#542 is intended be refactoring (no functionality changes), whereas this issue is concerned with changing provider functionality to return an IBC err ack on recv slash packet validation when the validator from the packet is not found or unbonded.

# Closing criteria

Change functionality of provider as described below, test new functionality properly.

# Problem details

Currently, when the validator from a recv slash packet is not found or unbonded, we simply log an error and drop the packet. There is no err ack that is returned to the consumer. We should return an err ack in this case, by adding a check to `validateSlashPacket()`

Note that changing this functionality seems trivial, but there are a good amount of tests that non-trivially rely on current provider functionality, `TestConsumerPacketSendExpiredClient` as example.

mpoke:
@smarshall-spitzbart what would be the benefit / impact of solving this issue?

smarshall-spitzbart:
Returning an ibc err ack would [offboard the consumer](https://github.com/cosmos/interchain-security/blob/cb6f856f1da105ad8f7e3f2803cd7a7dc49447fb/x/ccv/consumer/keeper/relay.go#L284).

However, with throttling fully implemented, the worst case scenario for spammed slash packets from a consumer is just wasted computation, and high fees paid by relayers. Ie. there's no security threat here. 

Also, offboarding consumers comes with it's own pros/cons.

I'll close this issue, feel free to re-open if you disagree

PR #619, opened 2022-12-21
add additional slash packet checks

MSalopek:
## Please do not merge
Difftests are not working due to additional slash packet validation (ValidateBasic).

# Description

Downtime slash packets will be dropped in `HandleSlashPacket` if a `SlashAck` already exists for a given chain and given validator. This prevents validators being spammed with downtime infractions without ever unjailing themselves.

## Linked issues

EDIT: Partial fix for #544 

## Type of change


tests/e2e/expired_client.go
```diff
@@ -124,19 +119,23 @@ func (s *CCVTestSuite) TestConsumerPacketSendExpiredClient() {
...
+	// should not return an errAck and not close the channel
 
-	// check that the packets were added to the list of pending data packets
-	consumerPackets = consumerKeeper.GetPendingPackets(s.consumerCtx())
-	s.Require().NotEmpty(consumerPackets)
-	s.Require().Equal(4, len(consumerPackets.GetList()), "unexpected number of pending data packets")
+	// try to send slash packet for downtime infraction
+	// addr := ed25519.GenPrivKey().PubKey().Address()
+	// val := abci.Validator{Address: addr}
+	// consumerKeeper.QueueSlashPacket(s.consumerCtx(), val, 2, stakingtypes.Downtime)
```
MSalopek, line 129: Sending these packets would cause an error in `provider.OnRecvSlashPacket` since the slash packet would be considered outdated.

x/ccv/provider/keeper/relay.go
```diff
@@ -287,15 +294,19 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
 func (k Keeper) ValidateSlashPacket(ctx sdk.Context, chainID string,
 	packet channeltypes.Packet, data ccv.SlashPacketData) error {
 
-	_, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId)
 	// return error if we cannot find infraction height matching the validator update id
-	if !found {
-		return fmt.Errorf("cannot find infraction height matching "+
-			"the validator update id %d for chain %s", data.ValsetUpdateId, chainID)
+	if _, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId); !found {
+		return providertypes.ErrSlashPacketInfractionHeightInvalid
```
MSalopek, line 299: Error message can be improved or reverted to the previous approach. I'm unopinionated about how they look, it was easier (imho) to have errors defined in types so they can be checked more easily.

tests/e2e/expired_client.go
smarshall-spitzbart: I was also running into problems with the tests in this file when first attempting to solve #546. 

Imo we should change the test behavior instead of commenting out

tests/e2e/common.go
```diff
@@ -396,7 +396,7 @@ func (suite *CCVTestSuite) commitSlashPacket(ctx sdk.Context, packetData ccv.Sla
 
 // constructSlashPacketFromConsumer constructs a slash packet to be sent from consumer to provider,
 func (s *CCVTestSuite) constructSlashPacketFromConsumer(bundle icstestingutils.ConsumerBundle,
-	tmVal tmtypes.Validator, infractionType stakingtypes.InfractionType, ibcSeqNum uint64) channeltypes.Packet {
+	tmVal tmtypes.Validator, infractionType stakingtypes.InfractionType, ibcSeqNum uint64) (channeltypes.Packet, ccv.SlashPacketData) {
```
smarshall-spitzbart, line 399: Nice addition 👍, I need this change for my own PR as well

x/ccv/provider/keeper/relay.go
```diff
@@ -285,15 +292,19 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
...
 	}
 
 	if data.Infraction != stakingtypes.DoubleSign && data.Infraction != stakingtypes.Downtime {
-		return fmt.Errorf("invalid infraction type: %s", data.Infraction)
+		return providertypes.ErrSlashPacketInfractionTypeInvalid
+	}
+
+	// VSCMaturedPacket had already been received and processed for this ValsetUpdateId
+	// return error because slash packet is out of date
+	if _, found := k.GetVscSendTimestamp(ctx, chainID, data.ValsetUpdateId); !found {
```
mpoke, line 306: Remove this check as it's not correct.

x/ccv/provider/keeper/relay.go
```diff
@@ -285,15 +292,19 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
...
 	packet channeltypes.Packet, data ccv.SlashPacketData) error {
 
-	_, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId)
 	// return error if we cannot find infraction height matching the validator update id
-	if !found {
-		return fmt.Errorf("cannot find infraction height matching "+
-			"the validator update id %d for chain %s", data.ValsetUpdateId, chainID)
+	if _, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId); !found {
+		return providertypes.ErrSlashPacketInfractionHeightInvalid
 	}
```
mpoke, line 299: Drop SlashPackets for downtime infraction were received from the same consumer for infraction heights smaller than the latest downtime infraction height received from that consumer, e.g.,
```
if stakingtypes.Downtime
  if data.ValsetUpdateId < latest received data.ValsetUpdateId for same chain downtime
    return error
```
Not sure how important is this check and whether it is worth the trouble.

x/ccv/provider/keeper/relay.go
```diff
@@ -285,15 +292,19 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
 func (k Keeper) ValidateSlashPacket(ctx sdk.Context, chainID string,
```
mpoke, line 292: I'm adding this here as I cannot add it to `HandleSlashPacket`.

When handling a slash packet, drop it if `providerConsAddr` is in `SlashAck(chainID)`

tests/e2e/expired_client.go
mpoke: Uncomment the tests and use the following reasoning to populate the `valsetUpdateID` of `consumerKeeper.QueueSlashPacket`:
```golang
// a downtime infraction identified at the current height
// has the infraction height set as follows (cf. SDK slashing module)
infractionHeight := s.consumerCtx().BlockHeight() - sdk.ValidatorUpdateDelay - 1
```

```golang
// for a double-sing infraction at consumer height H
// has the infraction height set as follows (cf. SDK evidence module)
infractionHeight := H - sdk.ValidatorUpdateDelay
```

For both downtime and double-signing, send SlashPackets using the following `valsetUpdateID`
```golang
infractionVSCid := consumerKeeper.GetHeightValsetUpdateID(s.consumerCtx(), uint64(infractionHeight))
```

mpoke (changes requested):

tests/e2e/slashing.go
```diff
@@ -441,12 +446,13 @@ func (suite *CCVTestSuite) TestHandleSlashPacketDistribution() {
 	// the test cases verify that only the unbonding tokens get slashed for the VSC ids
 	// mapping to the block heights before and during the undelegation otherwise not.
 	testCases := []struct {
+		name     string
 		expSlash bool
 		vscID    uint64
 	}{
-		{expSlash: true, vscID: vscIDs[0]},
-		{expSlash: true, vscID: vscIDs[1]},
```
MSalopek, line 448: @smarshall-spitzbart I think this needs to be redone on my part. I think the point of the test got lost, right?

In this version, the second slash cannot be applied since the SlashAck for given validator already exists. The slash acks are consumed (deleted) in endblock.

x/ccv/provider/keeper/relay.go
MSalopek: Done.

x/ccv/provider/keeper/relay.go
MSalopek: Done.

x/ccv/provider/keeper/relay.go
MSalopek: done

tests/e2e/slashing.go
smarshall-spitzbart: Hmm it's not completely clear what the original intention of the test was, considering the comment: 

> The test cases verify that only the unbonding tokens get slashed for the VSC ids mapping to the block heights before and during the undelegation otherwise not.

I think Marius may be more familiar with this functionality. @mpoke do you know what this test is validating?

smarshall-spitzbart (changes requested):
Regarding slash acks as a mechanism to achieve

> Drop SlashPackets for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself.

(From https://github.com/cosmos/interchain-security/issues/544) 

Slash acks are appended for a particular chain only once that slash [has been handled](https://github.com/cosmos/interchain-security/blob/57b47af1edc71301e5456c3323b97bb066e09191/x/ccv/provider/keeper/relay.go#L354). Therefore I don't think slash acks work as a mechanism to achieve the closing criteria defined in the issue.

A malicious consumer could still send a bunch of downtime slash packets that are recv all in block N, those packets would all be queued during block N. During the end blocker of block N, one of the queued downtime slash packets _could be_ handled, which would append the slash ack. Note that if the slash meter is negative in block N, no slash ack would be appended in block N. Even if a slash ack was appended in block N, this would only prevent spam until [that slash ack is consumed](https://github.com/cosmos/interchain-security/blob/57b47af1edc71301e5456c3323b97bb066e09191/x/ccv/provider/keeper/relay.go#L217), and the spam packets would only be dequeued if the slash ack exists at the time of handling. 

To prevent spam, we need to be adding logic to `ValidateSlashPacket`, which runs before the slash packet is added to the queue.

x/ccv/provider/keeper/keeper.go
```diff
@@ -887,3 +897,21 @@ func (k Keeper) GetFirstVscSendTimestamp(ctx sdk.Context, chainID string) (vscSe
 
 	return types.VscSendTimestamp{}, false
 }
+
+// SetLastDowntimeValsetUpdateId sets the latest valsetUpdateId when consumer initiated downtime slashing occured on chainID.
+func (k Keeper) SetLastDowntimeValsetUpdateId(ctx sdk.Context, chainID string, valsetUpdateID uint64) {
```
mpoke, line 902: Add UTs.

x/ccv/provider/keeper/keeper.go
```diff
@@ -662,6 +662,16 @@ func (k Keeper) AppendSlashAck(ctx sdk.Context, chainID, ack string) {
 	k.SetSlashAcks(ctx, chainID, acks)
 }
 
+// SlashAckExists checks if a SlashAck for chainID exists for given valConsAddr.
+func (k Keeper) SlashAckExists(ctx sdk.Context, chainID, valConsAddr string) bool {
```
mpoke, line 666: Add UT.

x/ccv/provider/keeper/relay.go
```diff
@@ -376,6 +394,11 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 	}
 	k.slashingKeeper.JailUntil(ctx, providerConsAddr, jailTime)
 
+	// LastDowntimeValsetUpdateId increases monotonically
+	// all slash requests that are less than LastDowntimeValsetUpdateId will be dropped
+	// during slash request validation in OnRecvSlashPacket
+	k.SetLastDowntimeValsetUpdateId(ctx, chainID, data.ValsetUpdateId)
```
mpoke, line 400: This should only be set for downtime infractions. Also, we should set it on receiving a SlashPacket instead of on handling it.

x/ccv/provider/keeper/relay.go
```diff
@@ -340,13 +349,22 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 
 	infractionHeight, found := k.getMappedInfractionHeight(ctx, chainID, data.ValsetUpdateId)
 	if !found {
-		k.Logger(ctx).Error("infraction height not found. But was found during slash packet validation")
+		k.Logger(ctx).Error(
+			"infraction height not found buut was found during slash packet validation. Dropping slash packet")
```
mpoke, line 353: nit: typo

x/ccv/provider/keeper/relay.go
```diff
@@ -250,7 +250,14 @@ func (k Keeper) OnRecvSlashPacket(ctx sdk.Context, packet channeltypes.Packet, d
 	}
 
 	if err := k.ValidateSlashPacket(ctx, chainID, packet, data); err != nil {
-		return channeltypes.NewErrorAcknowledgement(err.Error())
+		return channeltypes.NewErrorAcknowledgement(
```
mpoke, line 253: Can you also log an error here.

x/ccv/provider/keeper/keeper.go
```diff
@@ -887,3 +897,21 @@ func (k Keeper) GetFirstVscSendTimestamp(ctx sdk.Context, chainID string) (vscSe
 
 	return types.VscSendTimestamp{}, false
 }
+
+// SetLastDowntimeValsetUpdateId sets the latest valsetUpdateId when consumer initiated downtime slashing occured on chainID.
```
mpoke, line 901: Delete this state in `StopConsumer`. Also add logic for export/import.

mpoke (changes requested):

x/ccv/provider/keeper/keeper.go
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Not sure about the flow where you set it on receive.

Seems more logical to me to set it once the packet is actually processed, not when it was first seen.

Can you elaborate?

mpoke: https://github.com/cosmos/interchain-security/pull/619#pullrequestreview-1228226404

mpoke: If `LastDowntimeValsetUpdateId` is set on handle, then you could receive multiple packets for the same downtime infractions and all of them will be queued.

x/ccv/provider/keeper/keeper.go
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Done

x/ccv/provider/types/keys.go
```diff
@@ -122,6 +122,10 @@ const (
 	// ConsumerAddrsToPruneBytePrefix is the byte prefix that will store the mapping from VSC ids
 	// to consumer validators addresses needed for pruning
 	ConsumerAddrsToPruneBytePrefix
+
+	// LastDowntimeValsetUpdateIdBytePrefix is the byte prefix that will store the mapping from chainID
+	// to valsetUpdateId needed to track that the last valsetUpdateId that had consumer initiated downtime slashing applied
+	LastDowntimeValsetUpdateIdBytePrefix
```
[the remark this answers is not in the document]
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Done

x/ccv/provider/keeper/keeper.go
MSalopek: Done

tests/e2e/slashing.go
MSalopek: Setting this to draft since difftests are not covering new functionality. Current difftest do not account for the fact that `LastDowntimeValsetUpdateId` must increase monotonically and sometimes creates traces where the valsetUpdateId in the slash packet is less than `LastDowntimeValsetUpdateId`.

Updating difftests will take will take an unknown amount of time for me to update.

MSalopek:
> Regarding slash acks as a mechanism to achieve
> 
> > Drop SlashPackets for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself.
> 
> (From #544)
> 
> Slash acks are appended for a particular chain only once that slash [has been handled](https://github.com/cosmos/interchain-security/blob/57b47af1edc71301e5456c3323b97bb066e09191/x/ccv/provider/keeper/relay.go#L354). Therefore I don't think slash acks work as a mechanism to achieve the closing criteria defined in the issue.
> 
> A malicious consumer could still send a bunch of downtime slash packets that are recv all in block N, those packets would all be queued during block N. During the end blocker of block N, one of the queued downtime slash packets _could be_ handled, which would append the slash ack. Note that if the slash meter is negative in block N, no slash ack would be appended in block N. Even if a slash ack was appended in block N, this would only prevent spam until [that slash ack is consumed](https://github.com/cosmos/interchain-security/blob/57b47af1edc71301e5456c3323b97bb066e09191/x/ccv/provider/keeper/relay.go#L217), and the spam packets would only be dequeued if the slash ack exists at the time of handling.
> 
> To prevent spam, we need to be adding logic to `ValidateSlashPacket`, which runs before the slash packet is added to the queue.

After adding the requested changes I'm setting this to draft since difftests are not covering the new functionality.

Current difftest do not account for the fact that `LastDowntimeValsetUpdateId` must increase monotonically and thus creates traces where the valsetUpdateId in the slash packet is less than `LastDowntimeValsetUpdateId`.

Updating difftests will take will take an unknown amount of time for me to handle.

MSalopek:
Please don't merge - difftests are not working due to additional slash packet validation (ValidateBasic).

mpoke (approved):

smarshall-spitzbart:
If this PR will still close #544, it'll need to address the closing criteria of 

> RunSlashPacketData.ValidateBasic on receiving SlashPackets

Otherwise we can make that criteria it's own issue. 

@MSalopek @mpoke I still do not understand how slash acks accomplish the following behavior in this PR

> Drop SlashPackets for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself.

AFAIK slash acks have nothing to do with unjailing, and slash acks are deleted once included in a VSC packet. It seems [this comment](https://github.com/cosmos/interchain-security/pull/619#pullrequestreview-1228226404) still applies, and the current changes prevent spam only in a very small subset of cases, not the cases defined in the issue

MSalopek:
> If this PR will still close #544, it'll need to address the closing criteria of
> 
> > RunSlashPacketData.ValidateBasic on receiving SlashPackets
> 
> Otherwise we can make that criteria it's own issue.
> 
> @MSalopek @mpoke I still do not understand how slash acks accomplish the following behavior in this PR
> 
> > Drop SlashPackets for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself.
> 
> AFAIK slash acks have nothing to do with unjailing, and slash acks are deleted once included in a VSC packet. It seems [this comment](https://github.com/cosmos/interchain-security/pull/619#pullrequestreview-1228226404) still applies, and the current changes prevent spam only in a very small subset of cases, not the cases defined in the issue

Hi, thanks for the insight.

`ValidateBasic` was removed since it broke difftests in a way that is not super trivial to fix since it adds new requirements that need to be reflected in the difftest setup/trace generation.

This PR insures the validator has unjailed itself in case of multiple subsequent downtime requests from the same chain.

`ValidateBasic` addition can be tracked in the existing issue or in a new one if you prefer. I would prefer the current issue remain open, so no context is lost.

mpoke:
> AFAIK slash acks have nothing to do with unjailing, and slash acks are deleted once included in a VSC packet.

@smarshall-spitzbart True. Which means that if two `SlashPackets` are received in the same block from the same chain requesting to slash the same validator for downtime, then the second packet is dropped. If the two packets are received in separate blocks, then the second packet will go through and the validator will be slashed. Thus, you are right Shawn. This doesn't fix the issue. @MSalopek

MSalopek:
@mpoke @smarshall-spitzbart Thanks for your insights.

> Drop SlashPackets for downtime infraction for the same validator that were received from the same consumer without the validator having the chance to Unjail itself.

After discussing with @smarshall-spitzbart it seems that this requirement cannot be fulfilled with basic checks that are currently proposed in this PR. We would need to keep additional state to achieve that, and that would need to be thoroughly tested.

The basic checks may still hold some value, but I am in favor of closing this PR as it stands now.

mpoke:
Closing and opening new issues: https://github.com/cosmos/interchain-security/issues/635 and https://github.com/cosmos/interchain-security/issues/634

Issue #598, opened 2022-12-14
Iteration through `PacketMaturityTime`s assumes maturity time order


mpoke:
# Problem

`PacketMaturityTime`s are stored on the consumer under a key with the format `prefix | vscID`. As a result, the iteration is in the order of `vscID`s and not `maturityTimes` as [expected](https://github.com/cosmos/interchain-security/blob/3e339eaad34f37f4757049df775ea2c9a8fe941e/x/ccv/consumer/keeper/relay.go#L123). Since the `unbondingPeriod` on the consumer can be updated, ordering by `vscID`s is not the same as ordering by `maturityTimes`.

# Closing criteria

Ensure that the iteration over `PacketMaturityTime`s is in order of `maturityTimes`. 

# Problem details

A suggestion would be to use the same approach as in the staking module with the `MatureUBDQueue`, see https://github.com/cosmos/cosmos-sdk/blob/b46d83850270b82496e4cd61421d9ebc8bac87f2/x/staking/keeper/delegation.go#L376. This solution was implemented in https://github.com/cosmos/interchain-security/pull/583. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

smarshall-spitzbart:
In the PR that fixes this issue, could we use `time.Time` instead of uint64 for the business logic. There should be easy conversion between the two types

mpoke:
Yes, use uint64 in the key and time.Time in the value

PR #622, opened 2022-12-21
Fix: Iteration through PacketMaturityTimes assumes maturity time order

mpoke:
# Description

`PacketMaturityTimes` are stored using the following index:
```
PacketMaturityTimeBytePrefix | maturityTime.UnixNano | vscID -> MaturingVSCPacket{VscId, MaturityTime}
```
As a result, iterating over packet maturities is done in order of maturity times. 

- [x] Based on https://github.com/cosmos/interchain-security/pull/596

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/598

## Type of change

- [x] `Testing`: Adds testing

## New behavior tests

`TestPacketMaturityTime` was updated to check the expected order.


proto/interchain_security/ccv/consumer/v1/genesis.proto
```diff
@@ -41,13 +41,6 @@ message GenesisState {
   [ (gogoproto.nullable) = false ];
 }
 
-// MaturingVSCPacket defines the genesis information for the
```
mpoke, line 44: Moved to `consumer.proto` and change the `maturity_time` type.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,76 +207,79 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
+		MaturityTime: maturityTime,
+	}
+	bz, err := maturingVSCPacket.Marshal()
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal MaturingVSCPacket: %w", err))
+	}
+	store.Set(types.PacketMaturityTimeKey(vscId, maturityTime), bz)
 }
 
-// GetPacketMaturityTime gets the maturity time for a given received VSC packet id
```
mpoke, line 263: `GetPacketMaturityTime` was only used in testing. I replace it with `PacketMaturityExists` as you need both `vscID` and `maturityTIme` to access it. If you have that, you don't really need to access it. 

@MSalopek @jtremback this means it would be hard to query the maturity of a VSCPacket (as we don't know the maturity time).

x/ccv/provider/types/keys.go
```diff
@@ -341,20 +343,12 @@ func ParseGlobalSlashEntryKey(bz []byte) (
 	return recvTime, chainID, ibcSeqNum
 }
 
-// AppendMany appends a variable number of byte slices together
```
mpoke, line 344: Moved to `utils/utils.go` as it is needed by the consumer code also.

proto/interchain_security/ccv/consumer/v1/consumer.proto
```diff
@@ -91,3 +92,9 @@ message ConsumerPackets {
   repeated ConsumerPacket list = 1
   [ (gogoproto.nullable) = false ];
 }
+
+// MaturingVSCPacket contains the maturing time of a received VSCPacket
```
smarshall-spitzbart, line 96: Prob worth noting that if this time is being stored on the consumer at all, the maturity time is in the future. Otherwise a VSCMatured packet is sent from consumer -> provider.

Is my understanding correct?

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,76 +207,79 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
+// PacketMaturityExists checks whether the packet maturity time for a given vscId and maturityTime exists
+func (k Keeper) PacketMaturityTimeExists(ctx sdk.Context, vscId uint64, maturityTime time.Time) bool {
 	store := ctx.KVStore(k.storeKey)
-	bz := store.Get(types.PacketMaturityTimeKey(vscId))
-	if bz == nil {
-		return 0
-	}
-	return binary.BigEndian.Uint64(bz)
+	bz := store.Get(types.PacketMaturityTimeKey(vscId, maturityTime))
+	return bz != nil
```
smarshall-spitzbart, line 276: Would an empty byte array still be considered valid here?

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -90,60 +90,54 @@ func TestPacketMaturityTime(t *testing.T) {
...
+	// sorting by MaturityTime
 	sort.Slice(expectedGetAllOrder, func(i, j int) bool {
-		return expectedGetAllOrder[i].VscId < expectedGetAllOrder[j].VscId
+		return expectedGetAllOrder[i].MaturityTime.Before(expectedGetAllOrder[j].MaturityTime)
 	})
 	expectedGetElapsedOrder := []types.MaturingVSCPacket{}
 	for _, packet := range packets {
-		// only packets with MaturityTime <= nsNow
-		if packet.MaturityTime <= nsNow {
+		// only packets with MaturityTime <= now
```
smarshall-spitzbart, line 118:
```suggestion
		// only packets with MaturityTime before or equal to now
```

x/ccv/consumer/keeper/keeper_test.go
```diff
@@ -90,60 +90,54 @@ func TestPacketMaturityTime(t *testing.T) {
...
 	sort.Slice(expectedGetElapsedOrder, func(i, j int) bool {
-		return expectedGetElapsedOrder[i].VscId < expectedGetElapsedOrder[j].VscId
+		return expectedGetAllOrder[i].MaturityTime.Before(expectedGetAllOrder[j].MaturityTime)
 	})
 
 	for _, packet := range packets {
 		ck.SetPacketMaturityTime(ctx, packet.VscId, packet.MaturityTime)
 	}
 
-	for _, packet := range packets {
```
smarshall-spitzbart, line 133: Should we keep this check with the new "exists" method?

x/ccv/provider/types/keys.go
```diff
@@ -7,6 +7,8 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	utils "github.com/cosmos/interchain-security/x/ccv/utils"
```
smarshall-spitzbart, line 11: Nit: we have a lot of things that we refer to as utils
```suggestion
	ccvutils "github.com/cosmos/interchain-security/x/ccv/utils"
```

smarshall-spitzbart (approved):
LGTM, woo last iterator refactor PR

proto/interchain_security/ccv/consumer/v1/consumer.proto
mpoke: Almost. This is set either in `InitGenesis` on a consumer restart in which case the maturity time could be in the past, which would result in `VSCMaturedPackets` sent in the first `EndBlock`. 

Most likely though, the maturity time is set in `OnRecvVSCPacket`, i.e., 
```golang
maturityTime := ctx.BlockTime().Add(k.GetUnbondingPeriod(ctx))
k.SetPacketMaturityTime(ctx, newChanges.ValsetUpdateId, maturityTime)
```

x/ccv/consumer/keeper/keeper.go
mpoke: As long as the state is not corrupted, `bz` should not be an empty byte array. Anyway, this method is only used in testing.

x/ccv/consumer/keeper/keeper_test.go
mpoke: The exists method is only used in testing. Do we need UTs for those?

mpoke: Added it anyway.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -206,76 +207,81 @@ func (k Keeper) DeletePendingChanges(ctx sdk.Context) {
...
 
 	for ; iterator.Valid(); iterator.Next() {
-		vscId := binary.BigEndian.Uint64(iterator.Key()[1:])
-		maturityTime := binary.BigEndian.Uint64(iterator.Value())
+		var maturingVSCPacket consumertypes.MaturingVSCPacket
+		if err := maturingVSCPacket.Unmarshal(iterator.Value()); err != nil {
+			panic(fmt.Errorf("failed to unmarshal MaturingVSCPacket: %w", err))
+		}
 
 		// If the maturity time is after the current time, then stop the iteration;
```
MSalopek, line 224: Maybe change the comment to reflect the surrounding code (use "before" in both the comment and the function.

```
// If the current block time is before maturity time then stop the iteration;
// this is possible since the iteration over PacketMaturityTimes is in order
// of maturity times
```

x/ccv/consumer/keeper/relay.go
```diff
@@ -85,24 +85,22 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
...
-			panic(fmt.Errorf("maturity time %d is greater than current time %d", maturityTime.MaturityTime, currentTime))
+		if ctx.BlockTime().Before(maturityTime.MaturityTime) {
+			panic(fmt.Errorf("maturity time %s is greater than current time %s", maturityTime.MaturityTime, ctx.BlockTime()))
 		}
 		// construct validator set change packet data
 		vscPacket := ccv.NewVSCMaturedPacketData(maturityTime.VscId)
 
-		// append VSCMatured packet to pending packets
-		// sending packets is attempted each EndBlock
+		// append VSCMatured packet to pending packets;
```
MSalopek, line 95: Do semicolons improve readability or were they added by accident?

```golang
// Append VSCMatured packet to pending packets queue.
// Sending packets is attempted each EndBlock and all
// unsent packets will remain in the queue until sent.
```

x/ccv/consumer/keeper/relay.go
```diff
@@ -85,24 +85,22 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
 // operations that resulted in validator updates included in that VSC have matured on
 // the consumer chain.
 func (k Keeper) QueueVSCMaturedPackets(ctx sdk.Context) {
-	currentTime := uint64(ctx.BlockTime().UnixNano())
-
 	for _, maturityTime := range k.GetElapsedPacketMaturityTimes(ctx) {
-		if currentTime < maturityTime.MaturityTime {
-			panic(fmt.Errorf("maturity time %d is greater than current time %d", maturityTime.MaturityTime, currentTime))
+		if ctx.BlockTime().Before(maturityTime.MaturityTime) {
+			panic(fmt.Errorf("maturity time %s is greater than current time %s", maturityTime.MaturityTime, ctx.BlockTime()))
```
MSalopek, line 90: Nit:
```golang
maturity time %s is after than current time %s
```

MSalopek (approved):
Looks good!
We'll figure something out if we need to query maturity times for VSCs. We might consider adding another mapping (VSCId->maturityTime) when we introduce queries. 

Some nits were commented, you can safely ignore those

Issue #573, opened 2022-12-08
Streamline conditionals in slash packet throttling code


jtremback:
# Problem

In the code we currently handle VSCMatured packets immediately if there is nothing in the slash packet throttle queue. Instead, it would be good to queue every VSCMatured packet and always handle them in endblock to reduce the number of code paths. To do this, we then need to have a separate loop in the endblocker which clears out VSCMatured packets that are in the queue and do not have any slash packets in front of them.

# Problem details

Please describe the problem in all detail.

Here is the conditional that we would like to eliminate.

```
// If no packets are in the per chain queue, immediately handle the vsc matured packet data
	if k.GetPendingPacketDataSize(ctx, chainID) == 0 {
		k.HandleVSCMaturedPacket(ctx, chainID, data)
	} else {
		// Otherwise queue the packet data as pending (behind one or more pending slash packet data instances)
		k.QueuePendingVSCMaturedPacketData(ctx, chainID, packet.Sequence, data)
	}
```
Should become

```
k.QueuePendingVSCMaturedPacketData(ctx, chainID, packet.Sequence, data)
```

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

PR #624, opened 2022-12-22
Always queue vsc matured packet data

smarshall-spitzbart:
# Description

VSC matured packet data is now _always_ queued when recv. When VSC matured packet data is not preceded by slash packet data in the chain-specific throttle queue, it'll be handled in a new end blocker method, `HandleLeadingVSCMaturedPackets`. 

This PR is effectively a refactor, but I'm marking it as a feature since it does slightly change behavior (vsc matured packet data not preceded by slash packet data will now be handled on endblock. Whereas with previous behavior, the data would have been handled immediately after being recv).

## Linked issues

Closes https://github.com/cosmos/interchain-security/issues/573

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

`TestOnRecvVSCMaturedPacket`, existing throttle related e2e tests.

## New behavior tests

`TestGetLeadingVSCMaturedData`, `TestHandleLeadingVSCMaturedPackets`, `TestLeadingVSCMaturedAreDequeued`.


x/ccv/provider/keeper/relay.go
```diff
@@ -17,16 +17,6 @@ import (
 	utils "github.com/cosmos/interchain-security/x/ccv/utils"
 )
 
-func removeStringFromSlice(slice []string, x string) (newSlice []string, numRemoved int) {
```
smarshall-spitzbart, line 20: Moved to bottom of file

mpoke (approved):
LGTM

x/ccv/provider/keeper/throttle.go
```diff
@@ -321,6 +322,43 @@ func (k Keeper) QueueThrottledPacketData(
...
+	defer iterator.Close()
+
+	// Iterate over the throttled packet data queue,
+	// and return vsc matured packet data instances until we encounter a slash packet data instance.
+	vscMaturedData = []ccvtypes.VSCMaturedPacketData{}
+	ibcSeqNums = []uint64{}
+	for ; iterator.Valid(); iterator.Next() {
+
+		bz := iterator.Value()
+		if bz[0] == slashPacketData {
```
MSalopek, line 343: Nit:
Maybe use a switch in this case since all processing happens only for vscs, slash packet causes a break and an unknown type panics.

```golang
bz := iterator.Value()
switch bz[0] {
case vscMaturedPacketData:
	var data ccvtypes.VSCMaturedPacketData
	err := data.Unmarshal(bz[1:])
	if err != nil {
		panic(fmt.Sprintf("failed to unmarshal vsc matured packet data: %v", err))
	}
	
	vscMaturedData = append(vscMaturedData, data)
	_, ibcSeqNum := providertypes.MustParseThrottledPacketDataKey(iterator.Key())
	ibcSeqNums = append(ibcSeqNums, ibcSeqNum)
case slashPacketData:
	break
default:
	panic(fmt.Sprintf("unexpected packet data type: %d", bz[0]))
}
```

MSalopek (approved):
LGTM!

Added some nits.

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: This does look more readable, however the `break` you've defined in this code would break out of the switch case, and not the for-loop. Therefore we'd either need a break label, or the current code which uses if/else. 

A recent PR of mine had a break label, but it was decided from reviews that the pattern is not desirable

Issue #629, opened 2022-12-23
Rewrite some of legacy integration tests


shaspitz:
**EDIT**: e2e tests were renamed to integration tests 

# Problem

A lot of our current integration tests were migrated over in PRs similar to https://github.com/cosmos/interchain-security/pull/297, which tried to clean up testing conventions without changing functionality. The original tests were written in various non-standardized ways, in various files, and may be testing outdated functionality. Some of these testing pitfalls are still present in `main`, and we need to fix this where applicable.

# Closing criteria

All tests (legacy integration tests especially), are relevant to the current implementation of the ccv protocol, and have clear intention of what functionality is actually being tested. 

# Problem details

Here is an [example](https://github.com/cosmos/interchain-security/blob/0c62722c4c8ac138ff795a9b09f9125da3ed2bd7/tests/e2e/slashing.go#L380) of an integration test in which the functionality being tested is outdated? @mpoke created a more up-to-date test that fails. See https://github.com/cosmos/interchain-security/commit/5928146642857ff4bfc73db115c00ededcf61928

smarshall-spitzbart:
Regarding lessons for the future:

Standardizing test conventions, intention, organization, etc. should be the first thing we prioritize for a new repo! Test standardization should not be an afterthought that is prioritized half way through implementing a protocol.

PR #633, opened 2023-01-03
replace TestHandleSlashPacketDistribution w/ TestSlashUndelegation

mpoke:
# Description

`TestHandleSlashPacketDistribution` is replaced with `TestSlashUndelegation`, which tests 
> the slashing of an undelegation balance in various scenarios

## Linked issues

Related to https://github.com/cosmos/interchain-security/issues/629

## Type of change

- [x] `Testing`: Adds testing


tests/e2e/slashing.go
```diff
@@ -374,106 +375,396 @@ func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
...
-	// setup the test with a delegation, a no-op and an undelegation
-	setupOperations := []struct {
-		fn func(suite *CCVTestSuite) error
+	testCases := []struct {
+		name                  string
+		slash                 func(consAddr sdk.ConsAddress)
+		checkDelegatorBalance func(delAddr sdk.AccAddress, initBalance sdk.Int)
 	}{
+		// infraction - delegate - undelegate - slash - mature consumer - mature provider
+		// TODO: this behavior is unexpected
```
mpoke, line 404: This behavior is unexpected according to my understanding of slashing in Cosmos. Need to discuss with TM team.

tests/e2e/slashing.go
```diff
@@ -374,106 +375,396 @@ func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
...
+		hToVSCids = consumerKeeper.GetAllHeightToValsetUpdateIDs(suite.consumerCtx())
+		found = false
+		for _, hToVSCid := range hToVSCids {
+			if hToVSCid.ValsetUpdateId == vscID {
+				undelegateConsumerHeight = hToVSCid.Height
+				found = true
+				break
+			}
+		}
+		suite.Require().True(found, "cannot find height mapped to vscID; test: "+tc.name)
```
sainoe, line 740: Maybe we can move this into a helper function that takes a VSC id as input and either returns a height or fails.

tests/e2e/slashing.go
```diff
@@ -374,106 +375,396 @@ func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
...
 
-		isUbdSlashed := ubdBalance.GT(ubd.Entries[0].Balance)
-		suite.Require().True(tc.expSlash == isUbdSlashed)
+		tc.checkDelegatorBalance(delAddr, initBalance)
 
-		// update balance
-		ubdBalance = ubd.Entries[0].Balance
+		if i+1 < len(testCases) {
+			// reset suite to reset provider client
+			suite.SetupTest()
```
sainoe, line 764: This can be moved to the very beginning of the `for` loop.

sainoe:
Great work @mpoke. Just a few suggestions to shorten the test length.

sainoe (approved):

tests/e2e/slashing.go
mpoke: Yes, but then it will be called also in the first iteration and that's not needed. I don't think it save too much code.

tests/e2e/slashing.go
```diff
@@ -374,106 +375,371 @@ func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
...
 
-	// execute the setup operations, distributed uniformly in three blocks.
-	// For each of them, save their current VSC Id value which map correspond respectively
-	// to the block heights lesser, equal and greater than the undelegation creation height.
-	vscIDs := make([]uint64, 0, 3)
-	for _, so := range setupOperations {
-		err := so.fn(suite)
-		suite.Require().NoError(err)
+	for i, tc := range testCases {
+		providerKeeper := suite.providerApp.GetProviderKeeper()
```
smarshall-spitzbart, line 638: To fully reset the state, we need to call SetupTest on e2e tests that have multiple cases. Otherwise I believe we're using the same store between cases
```suggestion
                suite.SetupTest()
		providerKeeper := suite.providerApp.GetProviderKeeper()
```

tests/e2e/slashing.go
```diff
@@ -374,106 +375,371 @@ func (suite *CCVTestSuite) TestOnRecvSlashPacketErrors() {
...
-
-	// new delegator shares used
-	testShares := sdk.Dec{}
-
-	// setup the test with a delegation, a no-op and an undelegation
-	setupOperations := []struct {
-		fn func(suite *CCVTestSuite) error
+	testCases := []struct {
+		name                  string
+		slash                 func(consAddr sdk.ConsAddress)
```
smarshall-spitzbart, line 400: It doesn't seem like we need to define `slash` and `checkDelegatorBalance` callbacks for each test case, since it's only the function parameters that are different between cases. Something similar to this is more understandable and prob less code:
```suggestion
		incrementBy time.Duration
		infractionHeight int64
		expectedEndBalance sdk.Int
		// etc. - shows what func parameters are different between test cases, no need to vary func calls
```

Using callbacks to define test cases should be used sparingly imo, as it makes things less readable unless the callbacks are brief

tests/e2e/slashing.go
smarshall-spitzbart: Is this already tracked as an issue?

smarshall-spitzbart (approved):
LGTM once the comments are addressed. Great stuff, this test is definitely more intentioned than the previous one

tests/e2e/slashing.go
mpoke: Already doing it at the end of the for loop.

tests/e2e/slashing.go
mpoke: Thanks for the fix :)

tests/e2e/slashing.go
mpoke: https://github.com/cosmos/interchain-security/issues/643

mpoke:
`make test` passes locally

Issue #621, opened 2022-12-21
Security feedback: Consider employing structured error handling


andrey-kuprianov:
_Surfaced from @informalsystems security feedback for [Interchain Security](https://github.com/cosmos/interchain-security/), at commit 0f57a24_

# Problem

The current error handling in the codebase represents a mixture of scattered panics (132 in the non-testing code at the time of writing) with the standard Go error handling. Having so many panics in the codebase is dangerous, and may lead to e.g. halting the chains. We recommend to refactor the code and to employ structured error handling, with a few well-justified panics at the top level.

# Closing criteria

The code is refactored such that:

- [ ] There are only a few well-justified panics at the top level of the codebase.
- [ ] The code reachable from `BeginBlockers` or `EndBlockers` is guaranteed not to panic.

# Problem details

In the present Interchain Security (ICS) codebase, there is a mixed collection of error handling patterns:
 - standard Golang error handling, via returning an instance of the `error` interface;
 - **an abundant number of panics: 132 (in the non-testing code) at the time of writing**.

To illustrate, we will take as an example the function [IterateVscSendTimestamps()](https://github.com/cosmos/interchain-security/blob/0f57a2470cc7f0a11d90affce0c5080512083118/x/ccv/provider/keeper/keeper.go#L826-L851): the function calls two functions, `types.ParseVscSendingTimestampKey` and `sdk.ParseTimeBytes`, each returning an instance of `error`, and panics in both cases in case an error is indeed returned:

```go
func (k Keeper) IterateVscSendTimestamps(
	ctx sdk.Context,
	chainID string,
	cb func(vscID uint64, ts time.Time) (stop bool),
) {
	store := ctx.KVStore(k.storeKey)
	iterator := sdk.KVStorePrefixIterator(store, types.ChainIdWithLenKey(types.VscSendTimestampBytePrefix, chainID))
	defer iterator.Close()

	for ; iterator.Valid(); iterator.Next() {
		key := iterator.Key()
		_, vscID, err := types.ParseVscSendingTimestampKey(key)
		if err != nil {
			panic(fmt.Errorf("failed to parse VscSendTimestampKey: %w", err))
		}
		ts, err := sdk.ParseTimeBytes(iterator.Value())
		if err != nil {
			panic(fmt.Errorf("failed to parse timestamp value: %w", err))
		}

		stop := cb(vscID, ts)
		if stop {
			return
		}
	}
}
```

Considering the call sites of the above function, we see the following two cases:

[...]

Issue #634, opened 2023-01-03
`ValidateBasic` is not called when receiving packets from consumers


mpoke:
# Problem

When receiving packets from a consumer chain, the provider doesn't validate them (by calling `ValidateBasic`, for [VSCMaturedPacketData](https://github.com/cosmos/interchain-security/blob/0c62722c4c8ac138ff795a9b09f9125da3ed2bd7/x/ccv/types/ccv.go#L42) and [SlashPacketData](https://github.com/cosmos/interchain-security/blob/0c62722c4c8ac138ff795a9b09f9125da3ed2bd7/x/ccv/types/ccv.go#L62)), i.e., https://github.com/cosmos/interchain-security/blob/0c62722c4c8ac138ff795a9b09f9125da3ed2bd7/x/ccv/provider/ibc_module.go#L184 

Related to https://github.com/cosmos/interchain-security/issues/544

# Closing criteria

`ValidateBasic` is called in `OnRecvPacket`.

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

sainoe:
I think that https://github.com/cosmos/interchain-security/issues/609 should be solved beforehand since it entails renaming the consumer packet type in `ccv.proto`.

mpoke:
Fixed by https://github.com/cosmos/interchain-security/pull/1460

PR #640, opened 2023-01-05
Throttle related panic cleanup

smarshall-spitzbart:
# Description

For every panic in `throttle.go` or closely related files, this PR either: 
1. Clarifies with comments why the panic is still included
2. Removes the panic in favor of an error log to prevent undesirable binary behavior

## Linked issues

Works toward https://github.com/cosmos/interchain-security/issues/621

## Type of change

## New behavior tests

n/a, fix was small and only related to queries

### DIFF
--- a/x/ccv/provider/keeper/throttle.go
+++ b/x/ccv/provider/keeper/throttle.go
@@ -165,7 +165,7 @@ func (k Keeper) ReplenishSlashMeter(ctx sdktypes.Context) {
 func (k Keeper) GetSlashMeterAllowance(ctx sdktypes.Context) sdktypes.Int {
 
 	strFrac := k.GetSlashMeterReplenishFraction(ctx)
-	// The below func should not panic, since the (string representation) of the slash meter replenish fraction
+	// MustNewDecFromStr should not panic, since the (string representation) of the slash meter replenish fraction
 	// is validated in ValidateGenesis and anytime the param is mutated.
 	decFrac := sdktypes.MustNewDecFromStr(strFrac)
 
@@ -227,7 +227,7 @@ func (k Keeper) GetAllGlobalSlashEntries(ctx sdktypes.Context) []providertypes.G
 	entries := []providertypes.GlobalSlashEntry{}
 
 	for ; iterator.Valid(); iterator.Next() {
-		// The below func should not panic, since we should be iterating over keys that're
+		// MustParseGlobalSlashEntryKey should not panic, since we should be iterating over keys that're
 		// assumed to be correctly serialized in QueueGlobalSlashEntry.
 		recvTime, chainID, ibcSeqNum := providertypes.MustParseGlobalSlashEntryKey(iterator.Key())
 		valAddr := iterator.Value()
@@ -335,7 +335,7 @@ func (k Keeper) QueueThrottledPacketData(
 		}
 		bz = append([]byte{vscMaturedPacketData}, bz...)
 	default:
-		// Default case would indicate a developer error, this method should only be called
+		// Indicates a developer error, this method should only be called
 		// by tests, QueueThrottledSlashPacketData, or QueueThrottledVSCMaturedPacketData.
 		panic(fmt.Sprintf("unexpected packet data type: %T", data))
 	}
--- a/x/ccv/provider/keeper/throttle.go
+++ b/x/ccv/provider/keeper/throttle.go
@@ -559,6 +559,9 @@ func (k Keeper) GetSlashMeter(ctx sdktypes.Context) sdktypes.Int {
 //
 // Note: the value of this int should always be in the range of tendermint's [-MaxTotalVotingPower, MaxTotalVotingPower]
 func (k Keeper) SetSlashMeter(ctx sdktypes.Context, value sdktypes.Int) {
+
+	// TODO: remove these invariant panics once https://github.com/cosmos/interchain-security/issues/534 is solved.
+
 	// The following panics are included since they are invariants for slash meter value.
 	//
 	// Explanation: slash meter replenish fraction is validated to be in range of [0, 1],
--- a/x/ccv/provider/keeper/throttle.go
+++ b/x/ccv/provider/keeper/throttle.go
@@ -165,10 +165,9 @@ func (k Keeper) ReplenishSlashMeter(ctx sdktypes.Context) {
 func (k Keeper) GetSlashMeterAllowance(ctx sdktypes.Context) sdktypes.Int {
 
 	strFrac := k.GetSlashMeterReplenishFraction(ctx)
-	decFrac, err := sdktypes.NewDecFromStr(strFrac)
-	if err != nil {
-		panic(fmt.Sprintf("failed to parse slash meter replenish fraction: %s", err))
-	}
+	// The below func should not panic, since the (string representation) of the slash meter replenish fraction
+	// is validated in ValidateGenesis and anytime the param is mutated.
+	decFrac := sdktypes.MustNewDecFromStr(strFrac)
 
 	// Compute allowance in units of tendermint voting power (integer),
 	// noting that total power changes over time
@@ -229,7 +228,9 @@ func (k Keeper) GetAllGlobalSlashEntries(ctx sdktypes.Context) []providertypes.G
 	entries := []providertypes.GlobalSlashEntry{}
 
 	for ; iterator.Valid(); iterator.Next() {
-		recvTime, chainID, ibcSeqNum := providertypes.ParseGlobalSlashEntryKey(iterator.Key())
+		// The below func should not panic, since we should be iterating over keys that're
+		// assumed to be correctly serialized in QueueGlobalSlashEntry.
+		recvTime, chainID, ibcSeqNum := providertypes.MustParseGlobalSlashEntryKey(iterator.Key())
 		valAddr := iterator.Value()
 		entry := providertypes.NewGlobalSlashEntry(recvTime, chainID, ibcSeqNum, valAddr)
 		entries = append(entries, entry)
@@ -270,7 +271,7 @@ func (k Keeper) SetThrottledPacketDataSize(ctx sdktypes.Context, consumerChainID
 
 	// Sanity check to ensure that the chain-specific throttled packet data queue does not grow too
 	// large for a single consumer chain. This check ensures that binaries would panic deterministically
-	// if the queue does grow too large.
+	// if the queue does grow too large. MaxThrottledPackets should be set accordingly (quite large).
 	if size >= uint64(k.GetMaxThrottledPackets(ctx)) {
 		panic(fmt.Sprintf("throttled packet data queue for chain %s is too large: %d", consumerChainID, size))
 	}
@@ -322,16 +323,22 @@ func (k Keeper) QueueThrottledPacketData(
 	case ccvtypes.SlashPacketData:
 		bz, err = data.Marshal()
 		if err != nil {
+			// Slash packet data is assumed to be validated in OnRecvPacket and/or OnRecvSlashPacket,
+			// hence an error here would indicate something is very wrong.
 			panic(fmt.Sprintf("failed to marshal slash packet data: %v", err))
 		}
 		bz = append([]byte{slashPacketData}, bz...)
 	case ccvtypes.VSCMaturedPacketData:
 		bz, err = data.Marshal()
 		if err != nil {
+			// VSC matured packet data is assumed to be validated in OnRecvPacket and/or OnRecvVSCMaturedPacket,
+			// hence an error here would indicate something is very wrong.
 			panic(fmt.Sprintf("failed to marshal vsc matured packet data: %v", err))
 		}
 		bz = append([]byte{vscMaturedPacketData}, bz...)
 	default:
+		// Default case would indicate a developer error, this method should only be called
+		// by tests, QueueThrottledSlashPacketData, or QueueThrottledVSCMaturedPacketData.
 		panic(fmt.Sprintf("unexpected packet data type: %T", data))
 	}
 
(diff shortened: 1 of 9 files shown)


x/ccv/provider/keeper/throttle.go
```diff
@@ -431,8 +450,8 @@ func (k Keeper) GetSlashAndTrailingData(ctx sdktypes.Context, consumerChainID st
 
 // GetAllThrottledPacketData returns all throttled packet data for a specific consumer chain.
 //
-// Note: This method is only used by tests, hence why it returns redundant data as different types.
-// This method is not unit tested, as it is a test util.
+// Note: This method is only used by tests and queries, hence why it returns redundant data as different types.
+// Since this method executes on query, no panics are explicitly included.
 func (k Keeper) GetAllThrottledPacketData(ctx sdktypes.Context, consumerChainID string) (
```
smarshall-spitzbart, line 455: This method contains the only functionality changes in the PR

x/ccv/provider/keeper/throttle.go
```diff
@@ -322,16 +323,22 @@ func (k Keeper) QueueThrottledPacketData(
...
 		if err != nil {
+			// Slash packet data is assumed to be validated in OnRecvPacket and/or OnRecvSlashPacket,
+			// hence an error here would indicate something is very wrong.
 			panic(fmt.Sprintf("failed to marshal slash packet data: %v", err))
 		}
 		bz = append([]byte{slashPacketData}, bz...)
 	case ccvtypes.VSCMaturedPacketData:
 		bz, err = data.Marshal()
 		if err != nil {
+			// VSC matured packet data is assumed to be validated in OnRecvPacket and/or OnRecvVSCMaturedPacket,
```
smarshall-spitzbart, line 334: This assumption is wrong until #634 is solved btw, maybe that issue is worth prioritizing

x/ccv/provider/keeper/throttle.go
```diff
@@ -322,16 +323,22 @@ func (k Keeper) QueueThrottledPacketData(
 	case ccvtypes.SlashPacketData:
 		bz, err = data.Marshal()
 		if err != nil {
+			// Slash packet data is assumed to be validated in OnRecvPacket and/or OnRecvSlashPacket,
```
mpoke, line 326: `QueueThrottledPacketData` is called only when receiving a packet (in DeliverTX). Thus, it's cleaner to return an error and send back an err ack if something goes wrong.

x/ccv/provider/keeper/throttle.go
```diff
@@ -431,8 +450,8 @@ func (k Keeper) GetSlashAndTrailingData(ctx sdktypes.Context, consumerChainID st
 
 // GetAllThrottledPacketData returns all throttled packet data for a specific consumer chain.
 //
-// Note: This method is only used by tests, hence why it returns redundant data as different types.
-// This method is not unit tested, as it is a test util.
+// Note: This method is only used by tests and queries, hence why it returns redundant data as different types.
+// Since this method executes on query, no panics are explicitly included.
```
mpoke, line 454: Should we return the error during the query? It seems like a cleaner approach than to just log error messages.

mpoke: What actually happens if we panic in a query?

x/ccv/provider/keeper/throttle.go
```diff
@@ -515,29 +541,40 @@ func (k Keeper) GetSlashMeter(ctx sdktypes.Context) sdktypes.Int {
...
 	}
 	return value
 }
 
 // SetSlashMeter sets the slash meter to the given signed int value
 //
-// Note: the value of this int should always be in the range of tendermint's [-MaxVotingPower, MaxVotingPower]
+// Note: the value of this int should always be in the range of tendermint's [-MaxTotalVotingPower, MaxTotalVotingPower]
 func (k Keeper) SetSlashMeter(ctx sdktypes.Context, value sdktypes.Int) {
+	// The following panics are included since they are invariants for slash meter value.
```
mpoke, line 562: Could we just set the meter to `MaxTotalVotingPower` in this case?

x/ccv/provider/keeper/throttle.go
```diff
@@ -515,29 +541,40 @@ func (k Keeper) GetSlashMeter(ctx sdktypes.Context) sdktypes.Int {
...
+	//
+	// Explanation: slash meter replenish fraction is validated to be in range of [0, 1],
+	// and MaxMeterValue = MaxAllowance = MaxReplenishFrac * MaxTotalVotingPower = 1 * MaxTotalVotingPower.
 	if value.GT(sdktypes.NewInt(tmtypes.MaxTotalVotingPower)) {
 		panic("slash meter value cannot be greater than tendermint's MaxTotalVotingPower")
 	}
+	// Further, HandleThrottleQueues should never subtract more than MaxTotalVotingPower from the meter,
```
mpoke, line 569: ditto

mpoke (changes requested):
Nice work @smarshall-spitzbart. See my comments below. Also, the following throttling methods are called from EndBlock:
```golang
k.CheckForSlashMeterReplenishment(ctx)
k.HandleLeadingVSCMaturedPackets(ctx)
k.HandleThrottleQueues(ctx)
```
Could you add for each one a docstring describing the scenarios when they panic?

x/ccv/provider/keeper/throttle.go
```diff
@@ -229,7 +228,9 @@ func (k Keeper) GetAllGlobalSlashEntries(ctx sdktypes.Context) []providertypes.G
 	entries := []providertypes.GlobalSlashEntry{}
 
 	for ; iterator.Valid(); iterator.Next() {
-		recvTime, chainID, ibcSeqNum := providertypes.ParseGlobalSlashEntryKey(iterator.Key())
+		// The below func should not panic, since we should be iterating over keys that're
+		// assumed to be correctly serialized in QueueGlobalSlashEntry.
+		recvTime, chainID, ibcSeqNum := providertypes.MustParseGlobalSlashEntryKey(iterator.Key())
```
danwt, line 233:
```suggestion
		// MustParseGlobalSlashEntryKeyshould not panic, since we should be iterating over keys that're
		// assumed to be correctly serialized in QueueGlobalSlashEntry.
		recvTime, chainID, ibcSeqNum := providertypes.MustParseGlobalSlashEntryKey(iterator.Key())
```

x/ccv/provider/keeper/throttle.go
```diff
@@ -165,10 +165,9 @@ func (k Keeper) ReplenishSlashMeter(ctx sdktypes.Context) {
 func (k Keeper) GetSlashMeterAllowance(ctx sdktypes.Context) sdktypes.Int {
 
 	strFrac := k.GetSlashMeterReplenishFraction(ctx)
-	decFrac, err := sdktypes.NewDecFromStr(strFrac)
-	if err != nil {
-		panic(fmt.Sprintf("failed to parse slash meter replenish fraction: %s", err))
-	}
+	// The below func should not panic, since the (string representation) of the slash meter replenish fraction
+	// is validated in ValidateGenesis and anytime the param is mutated.
+	decFrac := sdktypes.MustNewDecFromStr(strFrac)
```
danwt, line 170:
```suggestion
	// MustNewDecFromStr should not panic, since the (string representation) of the slash meter replenish fraction
	// is validated in ValidateGenesis and anytime the param is mutated.
	decFrac := sdktypes.MustNewDecFromStr(strFrac)
```

x/ccv/provider/keeper/throttle.go
```diff
@@ -270,7 +271,7 @@ func (k Keeper) SetThrottledPacketDataSize(ctx sdktypes.Context, consumerChainID
 
 	// Sanity check to ensure that the chain-specific throttled packet data queue does not grow too
 	// large for a single consumer chain. This check ensures that binaries would panic deterministically
-	// if the queue does grow too large.
+	// if the queue does grow too large. MaxThrottledPackets should be set accordingly (quite large).
 	if size >= uint64(k.GetMaxThrottledPackets(ctx)) {
 		panic(fmt.Sprintf("throttled packet data queue for chain %s is too large: %d", consumerChainID, size))
```
danwt, line 276: Maybe we should use `Must` prefix for all functions which directly have a `panic`. So `SetThrottledPacketDataSize` becomes `Must..`

x/ccv/provider/keeper/throttle.go
```diff
@@ -322,16 +323,22 @@ func (k Keeper) QueueThrottledPacketData(
...
 		bz, err = data.Marshal()
 		if err != nil {
+			// VSC matured packet data is assumed to be validated in OnRecvPacket and/or OnRecvVSCMaturedPacket,
+			// hence an error here would indicate something is very wrong.
 			panic(fmt.Sprintf("failed to marshal vsc matured packet data: %v", err))
 		}
 		bz = append([]byte{vscMaturedPacketData}, bz...)
 	default:
+		// Default case would indicate a developer error, this method should only be called
+		// by tests, QueueThrottledSlashPacketData, or QueueThrottledVSCMaturedPacketData.
```
danwt, line 341:
```suggestion
		// Indicates a developer error, this method should only be called
		// by tests, QueueThrottledSlashPacketData, or QueueThrottledVSCMaturedPacketData.
```

x/ccv/provider/keeper/throttle.go
```diff
@@ -453,21 +472,28 @@ func (k Keeper) GetAllThrottledPacketData(ctx sdktypes.Context, consumerChainID
...
 			rawOrderedData = append(rawOrderedData, d)
 		default:
-			panic(fmt.Sprintf("invalid packet data type: %v", bz[0]))
+			k.Logger(ctx).Error(fmt.Sprintf("invalid packet data type: %v", bz[0]))
+			continue
+		}
+		_, ibcSeqNum, err := providertypes.ParseThrottledPacketDataKey(iterator.Key())
+		if err != nil {
+			k.Logger(ctx).Error(fmt.Sprintf("failed to parse throttled packet data key: %v", err))
+			continue
```
danwt, line 495: I'm not sure I understand the justification for this. Why not just panic?
If you can't unmarshal there's a serious problem so you should assume unexpected behavior in general, and panic. 
Failures tend to correlate so if there's one unmarshal failure there might be many more, which won't be handled, and things could deteriorate.

danwt (commented):
Suggestion: prefix every func which calls panic directly with Must
Tiny nit: some of the comments are a bit verbose ('Explanation:', 'Further', 'In this case'). They are very clear though so that's good

Overall good job!

mpoke:
> prefix every func which calls panic directly with Must

@danwt Prefixing getters with Must may imply that the key must exist. I think it's okay to go ahead with the pattern that getters and setters panic if marshaling / unmarshaling fails. We just need to make sure that the data for setters is valid.

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: Agreed 👍, my previous understanding was that "this should never happen" types of marshaling errors should always panic, as it's assumed calling methods handle validation. Just confirming we're okay moving away from that pattern?

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: If the slash meter is attempted to be set out of the range of `[-MaxTotalVotingPower, MaxTotalVotingPower]`, either something is incorrect/unexpected with the throttling logic, or someone has found a vulnerability. Allowing binaries to safely carry along in this case is pretty dangerous. 

Regardless, if the throttling logic is correct (and esp our understanding of its properties), then the slash meter should stay in the expected range. We could potentially remove this invariant logic altogether if we're confident in throttling correctness. 

Personally I'm not 100% confident in its correctness until we spec it out and such :)

smarshall-spitzbart: For now, leaving the panics and adding a todo to remove them once https://github.com/cosmos/interchain-security/issues/534 is solved

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: I've addressed this comment in https://github.com/cosmos/interchain-security/commit/8f5bd9462f269d262fddd14c9f31e807257b1c82. It's notable that returning errors like this does introduce more boilerplate. We should eventually settle on a standardized pattern for external data that needs to be validated

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: See comment for this method

> // Since this method executes on query, no panics are explicitly included.

`GetAllThrottledPacketData` is only used by queries and tests, it's best not to expose panics to queries imo. For example, maybe there's a reason why this method would fail, while `GetSlashAndTrailingData` would not fail. The latter method should be held to a higher standard because it runs on endblock

For any code that runs on txs and end/begin blockers, I completely agree with your sentiments, granted that external data is properly validated somehow

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: I see where you're coming from here, and Marius also makes a good point below about implications of "must" for getters. Personally I could go either way. Maybe create an issue and we can address this at a broader level?

x/ccv/provider/keeper/throttle.go
mpoke: @MSalopek any views on this?

mpoke (approved):
Nice work @smarshall-spitzbart

x/ccv/provider/keeper/throttle.go
MSalopek: Panics in queries could be dangerous, since it will bring down the node that tried to respond to the query.

However, simply logging an error will not communicate to the caller that there were issues during query execution.

What do you think about concatenating a bunch of errors and returning them? Invalid entries will never be returned so that may be a bit confusing (queue count is different than the number of returned entries).

Just an idea:
```golang
func (k Keeper) GetAllThrottledPacketData(ctx sdktypes.Context, consumerChainID string) (
	slashData []ccvtypes.SlashPacketData, vscMaturedData []ccvtypes.VSCMaturedPacketData, errors []error,
	rawOrderedData []interface{}, ibcSeqNums []uint64) {
        // ...
       if err != nil {
           errors = append(errors.New("err handling packet with sequence %d: err: %s, seq, err)
       }
```

MSalopek:
Overall, LGTM!
I am in favor of adding a `Must` prefix to functions that perform logic other than Get/Set since the pattern of panicking inside Set/Get is widespread in cosmos-sdk.

Maybe we could aggregate errors that happened during queue iteration and return them as part of the query result in another PR?

Besides that, queries should use pagination to avoid dumping large number of records from the throttle queues. Not sure, but that could also be a very easy DoS attack vector. I doubt it would affect validator nodes (provided the validators use sentry nodes) but the RPC/proxy/sentry nodes could be affected.

MSalopek (approved):

danwt (approved):
Great work!

MSalopek:
> Re "must" naming conventions, see #650. We can then address this for the entire repo
> 
> > Maybe we could aggregate errors that happened during queue iteration and return them as part of the query result in another PR?
> 
> Seems like a good idea 👍, thanks for the input @MSalopek. Could you make an issue for this, as you may have more context as to whether this error accumulation pattern could be applied to other queries as well

Issue opened: #651 

Thanks!

PR #642, opened 2023-01-06
Key assignment related panic cleanup

smarshall-spitzbart:
# Description

WIP, optionally we can audit the key related panics in `x/ccv/utils/utils.go` in this PR as well.

**_EDIT:_** 

- [x] `utils.TMCryptoPublicKeyToConsAddr` should return an error instead. 

## Linked issues

Works toward https://github.com/cosmos/interchain-security/issues/621

## Type of change

If you've checked more than one of the first three boxes, consider splitting this PR into multiple PRs!

- [ ] `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes
- [ ] `Fix`: Changes and/or adds code behavior, specifically to fix a bug
- [ ] `Refactor`: Changes existing code style, naming, structure, etc.
- [ ] `Testing`: Adds testing
- [ ] `Docs`: Adds documentation

## Regression tests

If `Refactor`, describe the new or existing tests that verify no behavior was changed or added where refactors were introduced.

## New behavior tests

If `Feature` or `Fix`, describe the new or existing tests that verify the new behavior is correct and expected.

PR #644, opened 2023-01-06
Improved assign consumer key msg validation

danwt:
This is a tiny fix to make sure the AssignConsumerKey message is validated properly. The pattern is similar what they use in the staking module for the CreateValidatorMsg.

https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/types/msg.go#L89

See comment

https://github.com/cosmos/interchain-security/pull/642#discussion_r1063385536

### DIFF
--- a/x/ccv/provider/types/msg.go
+++ b/x/ccv/provider/types/msg.go
@@ -6,7 +6,6 @@ import (
 	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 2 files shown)


x/ccv/provider/types/msg.go
```diff
@@ -76,8 +75,9 @@ func (msg MsgAssignConsumerKey) ValidateBasic() error {
 	if 128 < len(msg.ChainId) {
 		return ErrBlankConsumerChainID
 	}
-	if msg.ProviderAddr == "" {
```
mpoke, line 79: Do not remove this check?

danwt: https://github.com/cosmos/cosmos-sdk/blob/f771f20da49afc0fa3e0e75da431733382ddf3dd/types/address.go#L325-L328

smarshall-spitzbart (approved):

Issue #572, opened 2022-12-08
Improve tests to make sure we are using different keys for Operator Address and Cons Address


danwt:
# Problem

Our unit and e2e tests frequently use the same bytes for validator operators and consensus addres. That means we dont' properly test paths where they should be treated differently.

# Closing criteria

Improve the tests and cryptotestutils to use different bytes for each type.

# Problem details

The best way to achieve this is to make both addresses stem from different private keys.

See 

- https://github.com/cosmos/interchain-security/pull/462#discussion_r1041366884

for originating context

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

sainoe:
Solved in #672

Issue #486, opened 2022-11-16
Check if any of the cryptographicIdentity util methods are equivalent


danwt:
# Problem

When writing any code involving keys or addresses it is hard to know the exact meanings of each type required in a method signature. The cryptographic utilities implement helpers to get those types. Some of those types _may refer to the same thing_. 

# Closing criteria

Check if any of the return values from the crypto identity methods are equivalent, and if they are, document them as such

https://github.com/cosmos/interchain-security/blob/main/testutil/crypto/crypto.go

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

smarshall-spitzbart:
My hunch is that some of the utils do return equivalents. At least, I was able to use multiple different types of addresses when making a first pass at #473. The new util file is still a huuuuge improvement from before 🙏

smarshall-spitzbart:
Note that https://github.com/cosmos/interchain-security/pull/473 was closed because it got too behind. A future PR could use that PR as inspiration but use the utils correctly, maybe fix this issue in the same PR

danwt:
Note, this issue is related to https://github.com/cosmos/interchain-security/issues/572

mpoke:
@danwt does https://github.com/cosmos/interchain-security/pull/672 address this issue?

danwt:
Will be superseded by pending solution to https://github.com/cosmos/interchain-security/issues/572

PR #672, opened 2023-01-18
Improve crypto testutil

sainoe:
# Description

Improve the [crypto testutil](https://github.com/cosmos/interchain-security/blob/main/testutil/crypto/crypto.go) by differentiating the validator consensus and operator address type bytes. This expends the tests to cover scenarios where these addresses need to be treated differently.

## Linked issues

Closes: 
- #572
- #486 (TBD)

## Type of change

- [x] `Testing`: Adds testing


danwt (approved):
Great!

testutil/crypto/crypto.go
```diff
@@ -38,55 +46,42 @@ func NewCryptoIdentityFromIntSeed(i int) *CryptoIdentity {
...
-func (v *CryptoIdentity) SDKPubKey() sdkcryptotypes.PubKey {
-	tmcryptoPubKey := v.TMCryptoPubKey()
-	ret, err := sdkcryptocodec.FromTmPubKeyInterface(tmcryptoPubKey)
-	if err != nil {
-		panic(err)
-	}
-	return ret
+func (v *CryptoIdentity) ConsensusSDKPubKey() sdkcryptotypes.PubKey {
+	return v.consensus.PubKey()
 }
```
smarshall-spitzbart, line 79: For consistency we could also define `OperatorSDKPubKey` to be used in `SDKValOpAddress`
```suggestion
}

func (v *CryptoIdentity) OperatorSDKPubKey() sdkcryptotypes.PubKey {
	return v.operator.PubKey()
}
```

x/ccv/provider/handler_test.go
```diff
@@ -103,7 +103,7 @@ func TestAssignConsensusKeyForConsumerChain(t *testing.T) {
 			tc.setup(ctx, k, mocks)
 
 			msg, err := types.NewMsgAssignConsumerKey(tc.chainID,
-				testValProvider.SDKValAddress(), testValConsumer.SDKPubKey(),
+				testValProvider.SDKValOpAddress(), testValConsumer.ConsensusSDKPubKey(),
```
smarshall-spitzbart, line 106: Are we sure `NewMsgAssignConsumerKey` should accept a provider op address and a consumer cons pub key? Seems like it'd accept two cons or two operator addresses. Although I'm not that familiar myself 

CC @mpoke

x/ccv/provider/keeper/hooks_test.go
```diff
@@ -53,14 +53,14 @@ func TestValidatorConsensusKeyInUse(t *testing.T) {
 
 		gomock.InOrder(
 			mocks.MockStakingKeeper.EXPECT().GetValidator(ctx,
-				newValidator.SDKStakingOperator(),
+				newValidator.SDKStakingValidator().GetOperator(),
```
smarshall-spitzbart, line 56: Does this make the line more understandable?
```suggestion
				newValidator.OperatorSDKPubKey(),
```

x/ccv/provider/keeper/hooks_test.go
```diff
@@ -53,14 +53,14 @@ func TestValidatorConsensusKeyInUse(t *testing.T) {
...
 			).Return(newValidator.SDKStakingValidator(), true),
 		)
 
 		tt.setup(ctx, k)
 
 		t.Run(tt.name, func(t *testing.T) {
-			if actual := providerkeeper.ValidatorConsensusKeyInUse(&k, ctx, newValidator.SDKStakingOperator()); actual != tt.expect {
+			if actual := providerkeeper.ValidatorConsensusKeyInUse(&k, ctx, newValidator.SDKStakingValidator().GetOperator()); actual != tt.expect {
```
smarshall-spitzbart, line 63: Is it intentional to pass in an operator address to a method that seems like it accepts a consensus key? 

If this is intentional, we should add a comment or change the naming of that method

smarshall-spitzbart (commented):
Great work! Just a couple small things

x/ccv/provider/handler_test.go
sainoe: I think that `NewMsgAssignConsumerKey` mimics `NewMsgCreateValidator` in order to work with a similar CLI.

>  Seems like it'd accept two cons or two operator addresses

Actually only the former would work since consumer chains don't store operator addresses.

x/ccv/provider/keeper/hooks_test.go
sainoe: `GetValidator` takes a validator op address, not a pubkey. But this line can definitely be simplified, thanks. 
```suggestion
				newValidator.SDKValOpAddress()
,
```

x/ccv/provider/keeper/hooks_test.go
sainoe: It's intentional. We could rename this function to sthg like `IsConsensusAddressOfValidatorInUse`. Wdyt?

smarshall-spitzbart: Your idea works, also something like `IsValAddrInUse` could work, since the addrs we're iterating over shouldn't matter to the caller. Changing the parameter name from `valAddr` to `valOpAddr` [here](https://github.com/cosmos/interchain-security/blob/a5b78e375e089773d65f75636e9c36dc0b3a1b9b/x/ccv/provider/keeper/hooks.go#L70) is useful too

Aside: It seems like the `sdk.ValAddr` type should always imply operator address? I didn't know this previously. Cons addresses could still be assigned to that type unfortunately

x/ccv/provider/handler_test.go
smarshall-spitzbart: Got it, I think my confusion is just stemming from sdk's address/key semantics. This API is good

smarshall-spitzbart (approved):
LGTM!

Issue #437, opened 2022-11-07
Add test for consumer initiated slash via evidence module for no-longer-active validator


danwt:
# Problem

I think the case

```
validator is active on consumer
validator is no longer active on consumer
double sign evidence is submitted to consumer for the validator
```

is not really tested and I'm slightly suspicous of it. Either way the consumer code has

https://github.com/cosmos/interchain-security/blob/6c0d718d8c59339d112feb5850589258a504756e/x/ccv/consumer/keeper/validators.go#L76-L79

which is called in evidence/infraction.go, so the flow here should be commented/explained somewhere.

# Closing criteria

Test the scenario and document the relevant code flows by adding comments

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

danwt:
Having seen 

https://github.com/cosmos/interchain-security/blob/danwt/%23437-document-evidence-flow/tests/e2e/slashing.go#L826

and read through the code again I don't think there's any reason to be suspicious. We can close this.

I opened a PR which just adds a comment explaining a suspicious looking method.

https://github.com/cosmos/interchain-security/pull/675

PR #675, opened 2023-01-18
Document ValidatorByConsAddr in consumer module

danwt:
# Description

When evidence is submitted to the consumer chain it will be handled by the evidence module. The evidence module will call `consumerKeeper.ValidatorByConsAddr(..)` so the returned value must satisfy this. 

https://github.com/cosmos/cosmos-sdk/blob/interchain-security-rebase.0.45.11/x/evidence/keeper/infraction.go#L54-L75

It's similar for downtime 

https://github.com/cosmos/cosmos-sdk/blob/interchain-security-rebase.0.45.11/x/slashing/keeper/infractions.go#L81-L82

This PR adds a comment which explains this.

## Linked issues

I think with this comment we can close the concern in 437

Closes https://github.com/cosmos/interchain-security/issues/437

## Type of change


smarshall-spitzbart (approved):
LGTM

sainoe (approved):

Issue #689, opened 2023-01-30
Disable consumer initiated slashing


mpoke:
# Problem

A malicious consumer could send SlashPackets without validators actually misbehaving. This would be solve by having some type of provable evidence sent to the provider chain. This will be implemented as part of the https://github.com/cosmos/interchain-security/milestone/4 milestone. 

In the meantime, the provider should only jail once (not slash) validators for downtime infractions. This requires the following changes:
- provider drops SlashPackets that are not for downtime
- provider doesn't call Slash ~~or JailUntil~~
- provider logs SlashPackets for double signing

# Closing criteria

Consumer initiated slashing is disabled.

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

PR #692, opened 2023-02-01
disable consumer initiated slashing + fix slash acks bug

sainoe:
# Description

Disable consumer-initiated slashing and also jailing for double-signing evidence.

Also includes #708 

## Linked issues

Closes: #689 and #693

## Type of change

- [x]  `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes

## Regression tests

- [x] Disable consumer-initiated slashing feature
- [x] Update e2e tests
- [x] Update integration tests
- [x] Update diff tests

## New behavior tests

All tests should be updated to follow the new expected behavior when the consumer-initiated slashing feature is disabled.


smarshall-spitzbart:
Imo `HandleSlashPacket` should be renamed to `HandleDowntimeSlashPacket`, `JailFromDowntimeSlashPacket`, or something that can now be more descriptive

x/ccv/provider/keeper/relay.go
```diff
@@ -404,60 +420,25 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
...
+	// append the validator address to the slash ack for its chain id
+	k.AppendSlashAck(ctx, chainID, providerConsAddr.String())
 
 	// jail validator
 	if !validator.IsJailed() {
 		k.stakingKeeper.Jail(ctx, providerConsAddr)
 		k.Logger(ctx).Info("validator jailed", "provider cons addr", providerConsAddr.String())
 	}
 
-	k.slashingKeeper.JailUntil(ctx, providerConsAddr, jailTime)
```
smarshall-spitzbart, line 452: @mpoke @sainoe I'm not sure we can remove this call to `JailUntil` altogether. Shouldn't the call just be moved into the if-statement where `Jail` is called? What happens if you jail a validator without setting `JailUntil`, and how does it interact with [this check](https://github.com/cosmos/cosmos-sdk/blob/9c145c827001222df2e3e1101010874aeac20997/x/slashing/keeper/unjail.go#L55)?

smarshall-spitzbart:
@mpoke do you think it's appropriate to remove `TestSlashUndelegation` altogether in this PR? I realize this test was recently added in https://github.com/cosmos/interchain-security/pull/633. We can keep some parts of `TestSlashUndelegation`, but it seems the only testing that'd be left is already covered by e2e tests in `unbonding.go`. Any opinions here?

x/ccv/provider/keeper/relay.go
sainoe: Don't think it causes a problem. AFAIU, we want validators to be able to unjail directly after beeing jailed. So by keeping the `JailUntil` untouched, i.e. to its zero value  1970-01-01 00:00:00 UTC,  [this check](https://github.com/cosmos/cosmos-sdk/blob/9c145c827001222df2e3e1101010874aeac20997/x/slashing/keeper/unjail.go#L55) won't go through and the unjailing will work.

smarshall-spitzbart: Addressed in https://github.com/cosmos/interchain-security/pull/692/commits/69086f8b6b05f2b654d5d42d52cee59d41e7ab0b

smarshall-spitzbart:
> @mpoke do you think it's appropriate to remove `TestSlashUndelegation` altogether in this PR? I realize this test was recently added in #633. We can keep some parts of `TestSlashUndelegation`, but it seems the only testing that'd be left is already covered by e2e tests in `unbonding.go`. Any opinions here?

This test was [removed](https://github.com/cosmos/interchain-security/pull/692/commits/7e38dbe9d5ec481a5e6d79e1f3de6f0ea245c156)

smarshall-spitzbart:
e2e tests are fixed ✅

sainoe:
Diff-tests updated ✅

tests/difference/core/model/src/model.ts
```diff
@@ -458,36 +465,25 @@ class CCVProvider {
...
 
-    this.m.staking.slash(data.val, infractionHeight);
+    // jail validator
     this.m.staking.jailUntil(data.val, this.m.t[P] + JAIL_SECONDS);
-    if (data.isDowntime) {
-      this.downtimeSlashAcks.push(data.val);
-    } else {
-      this.tombstoned[data.val] = true;
-    }
+    // update slashing acks
```
sainoe, line 484:
```suggestion
    // update slash acks
```

tests/difference/core/model/src/model.ts
```diff
@@ -458,36 +465,25 @@ class CCVProvider {
...
-      this.m.events.push(Event.RECEIVE_DOWNTIME_SLASH_REQUEST);
-    } else {
-      this.m.events.push(Event.RECEIVE_DOUBLE_SIGN_SLASH_REQUEST);
-    }
+    this.m.events.push(Event.RECEIVE_DOWNTIME_SLASH_REQUEST);
+
 
     if (this.tombstoned[data.val]) {
-      return;
+      return
```
sainoe, line 479:
```suggestion
      return;
```

tests/difference/core/model/src/model.ts
```diff
@@ -422,6 +422,13 @@ class CCVProvider {
   };
 
   onReceive = (data: PacketData) => {
+
+    // Drop slash packets for double-sign infraction
+    if ('isDowntime' in data && ! data.isDowntime) {
```
smarshall-spitzbart, line 427: Curious why these two checks involve downtime if the code is intended to drop double-sign packets?

tests/difference/core/model/src/model.ts
```diff
@@ -458,36 +465,25 @@ class CCVProvider {
...
       this.m.events.push(Event.RECEIVE_SLASH_REQUEST_UNBONDED);
       return;
     }
 
-    if (data.isDowntime) {
-      this.m.events.push(Event.RECEIVE_DOWNTIME_SLASH_REQUEST);
-    } else {
-      this.m.events.push(Event.RECEIVE_DOUBLE_SIGN_SLASH_REQUEST);
-    }
+    this.m.events.push(Event.RECEIVE_DOWNTIME_SLASH_REQUEST);
```
smarshall-spitzbart, line 475: Related to above would double-sign packets get to this control flow? It's a little unclear to me

tests/integration/actions.go
```diff
@@ -772,6 +772,7 @@ func (tr TestRun) relayPackets(
 	if err != nil {
 		log.Fatal(err, "\n", string(bz))
 	}
+	time.Sleep(10 * time.Second)
```
smarshall-spitzbart, line 775: @MSalopek comments here would be nice, but shouldn't block merging

tests/integration/steps_democracy.go
```diff
@@ -203,7 +203,7 @@ func stepsDemocracy(consumerName string) []Step {
 					ValPowers: &map[validatorID]uint{
 						validatorID("alice"): 511,
 						// 1% of bob's stake should be slashed as set in config.go
```
smarshall-spitzbart, line 205: Remove this comment on tokens getting slashed

tests/integration/steps_downtime.go
```diff
@@ -126,14 +133,16 @@ func stepsDowntime(consumerName string) []Step {
 					ValPowers: &map[validatorID]uint{
 						// Non faulty validators still maintain just above 2/3 power here
 						validatorID("alice"): 509,
-						validatorID("bob"):   495,
+						validatorID("bob"):   500,
+						// Carol's stake should be slashed
```
smarshall-spitzbart, line 137: Is this suggestion correct?
```suggestion
						// Carol's stake should be jailed
```

tests/difference/core/model/src/model.ts
sainoe: Yea, this is a bit confusing. PacketData is a union type in TS, meaning that it can either carry slash or VSC mature data. So the first check asserts that the `isDowntime` field exists and the second one asserts that the infraction isn't for downtime but for a double sign. It should definitely be improved by using an enum or adding a `isDoubleSign` field.

smarshall-spitzbart (approved):
Some comments but nothing that should block merging, LGTM

tests/difference/core/model/src/model.ts
sainoe: Nop, double-sign packets are dropped before in the [onReceive](https://github.com/cosmos/interchain-security/blob/0cb44b5942ac4a574ede045daf61a31bdf06b59d/tests/difference/core/model/src/model.ts#L424) function so they aren't added to the queue and processed here in `onReceiveSlash`.

tests/difference/core/model/src/model.ts
jtremback: To me, it looks like `data.isDowntime` will be true if it is IS a downtime packet, and thus the negation `!data.isDowntime` will be true if it IS NOT a downtime packet. So the code inside the conditional will be executed if the packet IS NOT a downtime packet, yet it seems to fire the downtime event

jtremback: I believe that you probably intended to fire a double sign event not a downtime event

jtremback:
Looks good, except for changes required in model.ts line 428

tests/difference/core/model/src/model.ts
sainoe: True good catch

tests/integration/actions.go
MSalopek: Atm this has a merge conflict.
We should retain the changes from main - those have some comments and actually wait for blocks to be produced instead of sleeping.

smarshall-spitzbart:
> Imo `HandleSlashPacket` should be renamed to `HandleDowntimeSlashPacket`, `JailFromDowntimeSlashPacket`, or something that can now be more descriptive

Naming refactors can be included in a future PR. See #706

tests/integration/actions.go
smarshall-spitzbart: Sounds good, merged main 👍

tests/integration/steps_democracy.go
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/pull/692/commits/238a69f4c62a62246591ab45234d5891c1a56a3b

smarshall-spitzbart:
https://github.com/cosmos/interchain-security/pull/708 should have been merged into main after this PR, all good tho, this PR will close two issues now. Description updated

tests/difference/core/model/src/model.ts
smarshall-spitzbart: LGTM now, resolving comment

Issue #693, opened 2023-02-01
SlashAck sent to consumer is wrong


mpoke:
# Problem

The provider appends SlashAcks when sending VSCPackets to a consumer. This acks consists of validator addresses that are used by the consumer to reset the downtime flags. Currently, these are provider addresses and should be consumer addresses, see https://github.com/cosmos/interchain-security/blob/b89c46ed51d18c97795e52edaf4811cf78086800/x/ccv/provider/keeper/relay.go#L427. 

# Closing criteria

Use the address from the original SlashPacket before it's converted into a provider address, i.e., https://github.com/cosmos/interchain-security/blob/b89c46ed51d18c97795e52edaf4811cf78086800/x/ccv/provider/keeper/relay.go#L318. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

smarshall-spitzbart:
Closed in #692 (which includes #708)

PR #694, opened 2023-02-01
fix Slash processing on provider and consumer

MSalopek:
# Description

While processing Slash packets on the provider the validator address to be slashed was translated from the address on the consumer chain to the address on the provider chain. When SlashAcks were written during `HandleSlashPacket` the address used in the SlaskAck was equal to the address on the provider - it should have been set to the address on the consumer.

This was not an issue before key assignment was introduced. Since key assignment allows validators to use different keys on consumer chains, SlashAcks needed to be refactored so the consumer chain would receive SlashAcks with validator addresses that exist on the consumer.

## Linked issues

Closes: #693 

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests
e2e and integration tests are passing.

## New behavior tests

Added new unit test that checks SlashAcks after processing Slash packet on the provider. Updated all related mock calls to reflect the new behaviour.

### DIFF
--- a/tests/e2e/slashing.go
+++ b/tests/e2e/slashing.go
@@ -63,15 +63,14 @@ func (s *CCVTestSuite) TestRelayAndApplySlashPacket() {
 		s.Require().Nil(err)
 		consAddr := sdk.GetConsAddress(pubkey)
 		// map consumer consensus address to provider consensus address
-		if providerAddr, found := providerKeeper.GetValidatorByConsumerAddr(
+		providerAddr, found := providerKeeper.GetValidatorByConsumerAddr(
 			s.providerCtx(),
 			s.consumerChain.ChainID,
 			consAddr,
-		); found {
-			consAddr = providerAddr
-		}
+		)
+		s.Require().True(found)
 
-		valData, found := providerStakingKeeper.GetValidatorByConsAddr(s.providerCtx(), consAddr)
+		valData, found := providerStakingKeeper.GetValidatorByConsAddr(s.providerCtx(), providerAddr)
 		s.Require().True(found)
 		valOldBalance := valData.Tokens
 
@@ -154,7 +153,7 @@ func (s *CCVTestSuite) TestRelayAndApplySlashPacket() {
 		}
 
 		// check that the validator is successfully jailed on provider
-		validatorJailed, ok := providerStakingKeeper.GetValidatorByConsAddr(s.providerCtx(), consAddr)
+		validatorJailed, ok := providerStakingKeeper.GetValidatorByConsAddr(s.providerCtx(), providerAddr)
 		s.Require().True(ok)
 		s.Require().True(validatorJailed.Jailed)
 		s.Require().Equal(validatorJailed.Status, stakingtypes.Unbonding)
@@ -173,7 +172,7 @@ func (s *CCVTestSuite) TestRelayAndApplySlashPacket() {
 		s.Require().Equal(resultingTokens, validatorJailed.GetTokens())
 
 		// check that the validator's unjailing time is updated on provider
-		valSignInfo, found := providerSlashingKeeper.GetValidatorSigningInfo(s.providerCtx(), consAddr)
+		valSignInfo, found := providerSlashingKeeper.GetValidatorSigningInfo(s.providerCtx(), providerAddr)
 		s.Require().True(found)
 		s.Require().True(valSignInfo.JailedUntil.After(s.providerCtx().BlockHeader().Time))
 
--- a/testutil/keeper/expectations.go
+++ b/testutil/keeper/expectations.go
@@ -92,7 +92,7 @@ func GetMocksForStopConsumerChain(ctx sdk.Context, mocks *MockedKeepers) []*gomo
 }
 
 func GetMocksForHandleSlashPacket(ctx sdk.Context, mocks MockedKeepers,
-	expectedPacketData ccv.SlashPacketData, valToReturn stakingtypes.Validator) []*gomock.Call {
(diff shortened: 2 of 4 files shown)


x/ccv/provider/keeper/relay.go
```diff
@@ -376,12 +372,12 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 		"infractionType", data.Infraction,
 	)
 
-	// Obtain provider chain consensus address from packet data
-	// (overwritten with proper provider chain cons address in OnRecvSlashPacket)
-	providerConsAddr := sdk.ConsAddress(data.Validator.Address)
+	consumerConsAddr := sdk.ConsAddress(data.Validator.Address)
```
sainoe, line 375: The log just above should print `providerConsAddr` no?

x/ccv/provider/keeper/relay_test.go
```diff
@@ -469,8 +473,92 @@ func TestHandleSlashPacket(t *testing.T) {
...
+		{
+			"full double sign packet handling, uses valid vsc id and non-jailed validator",
+			*ccv.NewSlashPacketData(abci.Validator{Address: consumerConsAddr.Bytes()}, validVscID, stakingtypes.DoubleSign),
+			func(ctx sdk.Context, mocks testkeeper.MockedKeepers,
+				expectedPacketData ccv.SlashPacketData,
+			) []*gomock.Call {
+				return testkeeper.GetMocksForHandleSlashPacket(
+					ctx, mocks, expectedPacketData, providerConsAddr, stakingtypes.Validator{Jailed: false})
+			},
+			0, // no slash acks for double - only tombstone
```
sainoe, line 561: nit: => // no slash acks for double "sign" - only tombstone

sainoe (approved):
Good work Matija! I just have a few suggestions.

x/ccv/provider/keeper/relay.go
jtremback: Agree

jtremback: Move it down to where you have providerConsAddr and then print both prvoiderConsAddr and ConsumerConsAddr in the log

jtremback (approved):

jtremback:
LGTM except for the comment about the log

x/ccv/provider/keeper/relay_test.go
MSalopek: Done

x/ccv/provider/keeper/relay.go
MSalopek: Done

testutil/keeper/expectations.go
```diff
@@ -92,7 +92,7 @@ func GetMocksForStopConsumerChain(ctx sdk.Context, mocks *MockedKeepers) []*gomo
 }
 
 func GetMocksForHandleSlashPacket(ctx sdk.Context, mocks MockedKeepers,
-	expectedPacketData ccv.SlashPacketData, valToReturn stakingtypes.Validator) []*gomock.Call {
+	expectedPacketData ccv.SlashPacketData, expectedProviderValConsAddr sdk.ConsAddress, expectedValidator stakingtypes.Validator) []*gomock.Call {
```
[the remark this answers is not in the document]
smarshall-spitzbart: Ah I see now, `expectedPacketData.Validator.Address` is the consumer address. I think that `expectedPacketData` isn't needed at all once #692 is merged

smarshall-spitzbart:
Imo this should be a PR into `689-disable-slashing`, and should be merged after #692. It's hard for me to keep track whether these diffs are correct with new (lack of) slashing behavior

smarshall-spitzbart:
Also, why does `AppendSlashAck` accept an ack as a string? Shouldn't it accept a cons address? or better yet if we implement #707, a `consumerConsAddr`

tests/e2e/slashing.go
```diff
@@ -63,15 +63,14 @@ func (s *CCVTestSuite) TestRelayAndApplySlashPacket() {
 		s.Require().Nil(err)
 		consAddr := sdk.GetConsAddress(pubkey)
 		// map consumer consensus address to provider consensus address
-		if providerAddr, found := providerKeeper.GetValidatorByConsumerAddr(
+		providerAddr, found := providerKeeper.GetValidatorByConsumerAddr(
 			s.providerCtx(),
 			s.consumerChain.ChainID,
 			consAddr,
-		); found {
-			consAddr = providerAddr
```
smarshall-spitzbart, line 71: It's notable that setting `consAddr = providerAddr` was concealing this bug previously, since `SetOutstandingDowntime` was being called on the consumer keeper with the provider address. We should learn from this and use type safety more often! See #707

x/ccv/provider/keeper/relay_test.go
```diff
@@ -469,8 +473,92 @@ func TestHandleSlashPacket(t *testing.T) {
...
+				expectedPacketData ccv.SlashPacketData,
+			) []*gomock.Call {
+				return testkeeper.GetMocksForHandleSlashPacket(
+					ctx, mocks, expectedPacketData, providerConsAddr, stakingtypes.Validator{Jailed: true})
+			},
+			0, // no slash acks for double sign - only tombstone
+		},
+		{
+			"full double sign packet handling, uses valid vsc id and non-jailed validator",
+			*ccv.NewSlashPacketData(abci.Validator{Address: consumerConsAddr.Bytes()}, validVscID, stakingtypes.DoubleSign),
```
smarshall-spitzbart, line 554: We don't need to test this method for double-sign infractions, it'll only be called with downtime slash packets. might be worth changing the name of the method in the future

smarshall-spitzbart:
Closing in favor of https://github.com/cosmos/interchain-security/pull/708. If 708 is incorrect for some reason, we could reopen this one

PR #687, opened 2023-01-27
fix: slash meter replenishment

smarshall-spitzbart:
# Description

Please include a summary of the changes and the related issue. If the issue was ambiguous try to clarify it in this section.

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/684

## Type of change

- [x] `Fix`: Changes and/or adds code behavior, specifically to fix a bug
- [x] `Refactor`: Changes existing code style, naming, structure, etc.

## Regression tests

If `Refactor`, describe the new or existing tests that verify no behavior was changed or added where refactors were introduced.

## New behavior tests

If `Feature` or `Fix`, describe the new or existing tests that verify the new behavior is correct and expected.

Issue #684, opened 2023-01-27
Throttling: negative value of the meter causes consecutive replenishments


ivan-gavran:
The [throttle.md](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/docs/throttle.md?plain=1#LL79C315-L79C315) file says:

> The slash meter can go negative in value, and will do so when handling a single slash packet that jails a validator with significant voting power. In such a scenario, the slash meter may take multiple replenishment periods to once again reach a positive value, meaning no other slash packets may be handled for multiple replenishment periods.

However, this is not what happens in the code:
If the meter has negative value, it will be replenished once. In the following block, it will be replenished again (because the meter was not full, so the `lastSlashMeterFullTime` variable will not get set ([here](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/provider/keeper/throttle.go#L127-L127)).
This enables the meter to get replenished every block until it finally becomes full and resets `lastSlashMeterFullTime`.

PR #696, opened 2023-02-01
Throttle integration test fixes

smarshall-spitzbart:
# Description

The diffs in #687 caused integration tests to fail, which is actually good and proves that the patch changed behavior. This PR fixes integration tests to be a bit more granular with the throttle behavior that's expected.

## Type of change

- [x] `Testing`: Adds testing

## New Behavior Testing

Full integration tests pass locally for me, `make test-integration`

### DIFF
--- a/tests/integration/actions.go
+++ b/tests/integration/actions.go
@@ -1227,7 +1227,7 @@ func (tr TestRun) waitForSlashThrottleDequeue(
 		}
 
 		if globalQueueSize == chainQueueSize && globalQueueSize == action.nextQueueSize {
-			return
+			break
 		}
 
 		if time.Now().After(timeout) {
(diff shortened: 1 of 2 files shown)


tests/integration/actions.go
```diff
@@ -1236,6 +1236,8 @@ func (tr TestRun) waitForSlashThrottleDequeue(
 
 		time.Sleep(500 * time.Millisecond)
 	}
+	// Sleep 20 seconds to pass a block, allowing the jailing to be incorporated into voting power
```
smarshall-spitzbart, line 1239: @MSalopek depending on block time when running in the cloud, this may need to be adjusted

MSalopek (approved):
LGTM!
Thanks for taking the time to look into this.

jtremback (approved):
LGTM

smarshall-spitzbart:
BTW I just tried running the same test with a longer replenish period, and longer timeout. The test still passes if the `slashThrottleDequeue` timeout is set correctly. Long term it might be worth making the configured replenish period something around ~5 minutes, as this is a more effective test that packets are actually throttled. 

ie. with a replenish period of 20 seconds and a block time of 20 seconds, throttling will act no different than allowing a packet to be handled once per block. This is why #684 was not caught by integration tests

Issue #631, opened 2023-01-03
Fix flaking integration tests


danwt:
# Problem

The current integration tests are flaky

# Closing criteria

Make them not flaky

# Problem details

See 

https://github.com/cosmos/interchain-security/actions/runs/3749048773/jobs/6367083292

for an example run.

Additional context (credit @MSalopek @mpoke @jtremback )

> Any time you push them past 2 containers that will happen

> It's a timing issue. There's no way around it besides running them in series

> Re-running the test usually solves it. Bear in mind that multichain tests are not running in the pipeline any longer due to the same issues you are experiencing now

> When I played with tests last week, one thing was apparent - block times are not very consistent, they can vary from 1-2s to anywhere around 20s. We have some actions that wait for a number of blocks to be produced - these often fail due to wait timeouts being reached (not enough blocks produced in specified window, usually 1 minute).

> We should have a chat with the Tendermint team and see how they do it. 

# TODOs
- [x] Labels have been added for issue
- [x] Issue has been added to the ICS project

MSalopek:
Initial idea Id like to test out is running only happy path tests on push to PR -> this reduces pipeline run times while giving us enough coverage during development iterations.

Full integration tests can be run in series after merge. There are [merge queues in GH](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) but they are somewhat experimental.

The desired behaviour can be used by creating custom github actions, either community contributed or our own. There should be plenty of examples.

mpoke:
Another option is to adopted the approach used by the TM team: run most of the integrations tests over the night on the main branch. If something fails, we investigate and if necessary, either push a fix or revert the problematic PR.

PR #701, opened 2023-02-02
wait after sending unjail tx in integration tests

MSalopek:
# Description

Adding a wait time after sending an unjail tx so it has a chance to be included in a block and the required packets to be commited.

Previously, when making state assertions after unjailing the validator's jailed state sould not change and their power would not be refreshed. 

## Linked issues

Closes: #631 

## Type of change

- [x] `Testing`: Adds testing

### DIFF
--- a/tests/integration/actions.go
+++ b/tests/integration/actions.go
@@ -1010,9 +1010,9 @@ func (tr TestRun) unjailValidator(action unjailValidatorAction, verbose bool) {
 		log.Fatal(err, "\n", string(bz))
 	}
 
-	// wait a bit so the tx get included in a block
+	// wait for 2 blocks to be sure that tx got included in a block
 	// and packets commited before proceeding
-	time.Sleep(20 * time.Second)
+	tr.waitBlocks(action.provider, 2, time.Minute)
 }
 
 type registerRepresentativeAction struct {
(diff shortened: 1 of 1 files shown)


tests/integration/actions.go
```diff
@@ -1008,6 +1009,10 @@ func (tr TestRun) unjailValidator(action unjailValidatorAction, verbose bool) {
 	if err != nil {
 		log.Fatal(err, "\n", string(bz))
 	}
+
+	// wait a bit so the tx get included in a block
+	// and packets commited before proceeding
+	time.Sleep(20 * time.Second)
```
smarshall-spitzbart, line 1015: We need to wait two blocks right? One for the tx to be included, and one for the val set to be committed. Might be worth making this 40 seconds considering the block time you've observed when these tests run in cloud

smarshall-spitzbart (approved):
Nice!

sainoe (approved):
LGTM

tests/integration/actions.go
MSalopek:
```suggestion
		tr.waitBlocks(action.provider, 2, time.Minute)
```
We can actually do this. Don't know why I didn't think of that right away :D

PR #632, opened 2023-01-03
Update IBC-Go to canonical v3.4.0

glnro:
# Description
Currently ICS uses informal's fork of IBC-Go. This PR brings ICS up to date with the official IBC-Go version as the changes made in informal's fork were strictly for testing purposes. Any modifications to TestApp and IBC SimApp were brought directly into ICS until it can upgrade to a version of IBC-Go (v5) where the necessary compatible changes have been back-ported.

## Linked issues
#585 

## Type of change

- [ ] `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes
- [ ] `Fix`: Changes and/or adds code behavior, specifically to fix a bug
- [X] `Refactor`: Changes existing code style, naming, structure, etc.
- [ ] `Testing`: Adds testing
- [ ] `Docs`: Adds documentation

## Regression tests
Changes made in TestApp and SimApp to to accommodate the updated staking keeper interface that consumer keeper must return have been directly brought into ics in their own modules. 

**NOTE:**  The changes made to informal's fork of ibc-go strictly for testing have been copied into a temporary directory `legacy_ibc_testing`. This is to be able to use a canonical version of ibc-go `v3.4.0` without having to drastically change the tests. The current e2e test framework relies on modifications to ibc-go that cannot be back-ported to `v3.4.0`. This temporary directory can be removed once ICS upgrades to a version that supports the testingmodifications, `v5` at the earliest.

PR #702, opened 2023-02-02
Fix lint error in main

smarshall-spitzbart:
# Description

`golangci-lint run` has been failing on main since #632 was merged. Not sure why this hasn't been caught by CI... Seems that command produces different results when running locally vs cloud

## Type of change

- [x] `Other`: Fixes lint errors 

## Regression tests

n/a

## New behavior tests

n/a


legacy_ibc_testing/testing/chain.go
```diff
@@ -494,6 +494,10 @@ func (chain *TestChain) CreateTMClientHeader(chainID string, blockHeight int64,
 	vsetHash := tmValSet.Hash()
 	nextValHash := nextVals.Hash()
 
+	if tmValSet == nil {
+		panic("tmValSet cannot be nil")
+	}
+
```
yaruwangway, line 500: can panic very early before 
```
	var (
		valSet      *tmproto.ValidatorSet
		trustedVals *tmproto.ValidatorSet
	)
	```

yaruwangway (approved):

legacy_ibc_testing/testing/chain.go
MSalopek: @yaruwangway I didn't get this? Can you elaborate?

MSalopek (approved):

legacy_ibc_testing/testing/chain.go
smarshall-spitzbart: Yea I'm happy to address anything actionable from this comment. For now I'll merge this PR as it's intended to be pretty trivial, just to make linters pass

PR #703, opened 2023-02-03
feat: Equivocation gov proposal

tbruyelle:
# Description

This change adds a new kind of gov proposal that will slash and tombstone validators for double-signing. Since slashing for double signing is about to be removed from ICS (https://forum.cosmos.network/t/slashing-updates-in-replicated-security/9571), this new proposal is here to cover the lack of punishment if a double-sign happens in a consumer chain.

The proposal handler is added in the `provider` module, and use the `evidence` module to handle the equivocations.

## Linked issues

Closes: no issue created.

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

## New behavior tests

Added new tests for the proposal handler.


x/ccv/provider/keeper/keeper.go
```diff
@@ -569,8 +572,8 @@ func (k Keeper) ConsumeMaturedUnbondingOps(ctx sdk.Context) []uint64 {
 
 // Retrieves the underlying client state corresponding to a connection ID.
 func (k Keeper) getUnderlyingClient(ctx sdk.Context, connectionID string) (
-	clientID string, tmClient *ibctmtypes.ClientState, err error) {
-
+	clientID string, tmClient *ibctmtypes.ClientState, err error,
```
jtremback, line 575: whitespace diff?

tbruyelle: It's because of gofumpt, I can revert those changes if you prefer.

PR #714, opened 2023-02-07
invariant explanation made more precise; some extra clarifications

ivan-gavran:
# Description

In this commit I tried to make some notions more precise, correct some typos, and add extra explanations where I thought they were necessary.

In particular, I replaced the assumption of the "final jailed provider smaller than F" with the assumption of "every provider smaller than F". The reason is that it is up to adversary to choose the last slashed provider and thus we need to have this stronger assumption.

Other changes are mostly about formatting and adding gluing explanations.


smarshall-spitzbart (approved):
All looks great, thanks!

Issue #707, opened 2023-02-03
Use type safety to prevent provider addr vs consumer addr bugs


smarshall-spitzbart:
# Problem

Issues like https://github.com/cosmos/interchain-security/issues/693 have proven that using provider and consumer addresses interchangeably as the same type can cause bugs. We should make a new protobuf key type wrapper that can delineate between a provider cons address and a consumer cons address.  

# Closing criteria

Protobuf type(s) are made and implemented across our repo to prevent developers from mistakenly using the wrong type of address. This issue could also be closed by simply using golang type aliases

smarshall-spitzbart:
See comments like https://github.com/cosmos/interchain-security/pull/694#issuecomment-1416507883 as an example. Methods like `SetSlashAcks` should accept `consumerConsAddr` instead of a string

PR #725, opened 2023-02-08
Key assignment type safety

smarshall-spitzbart:
# Description

Reduces foot guns in key assignment code which previously used `sdk.ConsAddress` for both provider and consumer addresses. This PR makes key assignment code leverage type safety better by introducing protobuf generated wrappers around consensus addresses, `ConsumerConsAddress` and `ProviderConsAddress`

Note that key assignment state is still persisted as hand crafted `sdk.ConsAddress` bytes, but future refactors could use these new pb types to store key assignment state in KV store value bytes using pb serialization methods.

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/707

Note this PR enables a good solution to #728, but I'll save that as future work.

## Type of change

## Regression tests

Existing key assignment tests should suffice, lmk if there's more targeted tests I should develop.


x/ccv/provider/keeper/key_assignment.go
```diff
@@ -566,6 +566,8 @@ func (k Keeper) MustApplyKeyAssignmentToValUpdates(
 
 // GetProviderAddrFromConsumerAddr returns the consensus address of a validator with
 // consAddr set as the consensus address on a consumer chain
+//
+// TODO: communicate better that this method may return a consumer addr
```
glnro, line 570: What would precipitate this function returning a consumer address?

MSalopek:
```golang
	// check if this address is known only to the consumer chain
	if providerConsAddr, found := k.GetValidatorByConsumerAddr(ctx, chainID, consAddr); found {
		return providerConsAddr
	}
	return consAddr
```

As it stands now, it is a bit messy.

Consumer address is equal to provider address in case the operator does not use key assignment (ie, they use the same validator key for provider and consumer)

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -196,9 +191,49 @@ message VscSendTimestamp {
   google.protobuf.Timestamp timestamp = 2 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
 }
 
+//
+// Key assignment section
+//
+
+message ConsumerConsAddress {
```
MSalopek, line 198: An edgecase crossed my mind:
What if the address format is different on provider and consumer?

eg. `cosmosvalconsjdfp0gks0dr9gs` vs `stridevalconsjdfp0gks0dr9gs`

Do we have/need any special handling in that case?

smarshall-spitzbart:
Totally my bad, this is very much a draft PR lol

x/ccv/provider/keeper/key_assignment.go
smarshall-spitzbart: @MSalopek has the correct understanding. That code behavior was not clear to me from reading code/comments

proto/interchain_security/ccv/provider/v1/provider.proto
smarshall-spitzbart: I don't believe this is an issue, but I'm not 100% sure. These are consensus addresses that are persisted as bytes. `ConsumerCoonsAddress` is just a type wrapper for a byte array

x/ccv/provider/keeper/genesis.go
```diff
@@ -73,16 +73,16 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
 
 	// Import key assignment state
 	for _, item := range genState.ValidatorConsumerPubkeys {
-		k.SetValidatorConsumerPubKey(ctx, item.ChainId, item.ProviderAddr, *item.ConsumerKey)
+		k.SetValidatorConsumerPubKey(ctx, item.ChainId, *item.ProviderAddr, *item.ConsumerKey)
```
smarshall-spitzbart, line 76: I believe these changes mean v1.1 will require state migrations from v1.0

x/ccv/provider/keeper/hooks_test.go
```diff
@@ -33,7 +33,10 @@ func TestValidatorConsensusKeyInUse(t *testing.T) {
 			setup: func(ctx sdk.Context, k providerkeeper.Keeper) {
```
smarshall-spitzbart, line 33: This test was hard to follow in some cases but I did not change which validator addrs are input to the methods

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -128,12 +130,16 @@ func (k Keeper) GetValidatorByConsumerAddr(
...
-	consumerAddr sdk.ConsAddress,
-	providerAddr sdk.ConsAddress,
+	consumerAddr types.ConsumerConsAddress,
+	providerAddr types.ProviderConsAddress,
 ) {
 	store := ctx.KVStore(k.storeKey)
 	// Cons address is a type alias for a byte string, no marshaling needed
-	bz := providerAddr
+	bz, err := providerAddr.Marshal()
+	if err != nil {
```
smarshall-spitzbart, line 139: Technically a change in behavior, however an invalid cons address getting persisted would eventually cause critical errors. Addresses should be validated by this point (correct me if I'm wrong)

See `TestValidatorByConsumerAddrCRUD`

x/ccv/provider/keeper/key_assignment_test.go
```diff
@@ -873,18 +878,18 @@ func TestSimulatedAssignmentsAndUpdateApplication(t *testing.T) {
 			// Build up the historicSlashQueries data structure
 			for i := range consumerValset.identities {
 				// For each active validator on the consumer chain
-				consC := consumerValset.identities[i].SDKValConsAddress()
+				consC := consumerValset.identities[i].ConsumerConsAddress()
 				if 0 < consumerValset.power[i] {
 					// Get the provider who assigned the key
 					consP := k.GetProviderAddrFromConsumerAddr(ctx, CHAINID, consC)
 
-					if _, found := historicSlashQueries[string(consC)]; !found {
```
smarshall-spitzbart, line 881: This test was hard to follow but I did my best to keep its intention the same

x/ccv/provider/keeper/throttle.go
```diff
@@ -195,7 +195,12 @@ func (k Keeper) GetSlashMeterAllowance(ctx sdktypes.Context) sdktypes.Int {
 func (k Keeper) QueueGlobalSlashEntry(ctx sdktypes.Context, entry providertypes.GlobalSlashEntry) {
 	store := ctx.KVStore(k.storeKey)
 	key := providertypes.GlobalSlashEntryKey(entry)
-	store.Set(key, entry.ProviderValConsAddr)
+	bz, err := entry.ProviderValConsAddr.Marshal()
+	if err != nil {
+		// This should never happen, since the provider val cons addr should be a valid sdk address
```
smarshall-spitzbart, line 200: We again assume the provider cons address has been properly validated by now. This is where it's obtained https://github.com/cosmos/interchain-security/blob/38a1540972d3cd1dda7285c5a602b8706840270f/x/ccv/provider/keeper/relay.go#L316

x/ccv/provider/types/keys.go
```diff
@@ -262,20 +262,20 @@ func ParseVscSendingTimestampKey(bz []byte) (string, uint64, error) {
 
 // ConsumerValidatorsKey returns the key under which the
 // validator assigned keys for every consumer chain are stored
-func ConsumerValidatorsKey(chainID string, addr sdk.ConsAddress) []byte {
-	return ChainIdAndConsAddrKey(ConsumerValidatorsBytePrefix, chainID, addr)
+func ConsumerValidatorsKey(chainID string, addr ProviderConsAddress) []byte {
```
smarshall-spitzbart, line 265: For these methods, provider vs consumer cons addr type was not previously specified. However I inspected code and found that all callers gave one of two types

smarshall-spitzbart:
Full `make test` passed locally for me on latest commit

tests/e2e/slashing.go
```diff
@@ -147,7 +148,7 @@ func (s *CCVTestSuite) TestRelayAndApplyDowntimePacket() {
 	// check that the outstanding slashing flag is reset on first consumer,
 	// since that consumer originally sent the slash packet.
 	// It's important to use the consumer's cons addr here.
-	pFlag := firstConsumerKeeper.OutstandingDowntime(s.consumerCtx(), consumerConsAddr)
+	pFlag := firstConsumerKeeper.OutstandingDowntime(s.consumerCtx(), consumerConsAddr.ToSdkConsAddr())
```
sainoe, line 151: Perhaps the `OutstandingDowntime` keeper's functions should also take a `ConsumerConsAddress` rather that `sdk.ConsAddress`.

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -196,9 +191,57 @@ message VscSendTimestamp {
...
+}
+
+// AddressList contains a list of consumer consensus addresses
+message ConsumerAddressList {
+  repeated ConsumerConsAddress addresses = 1;
+}
+  
 message KeyAssignmentReplacement {
-  bytes provider_addr = 1;
+  ProviderConsAddress provider_addr = 1;
```
sainoe, line 220: The CLI fills the `KeyAssignmentReplacement` msg using the validator operator address
see how we it's parsed [here](https://github.com/cosmos/interchain-security/blob/bf6cdae5f02ff31bb2982766bed9db4c664a0ec3/x/ccv/provider/keeper/msg_server.go#L38)

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -196,9 +191,57 @@ message VscSendTimestamp {
...
+}
+
+// A validator's consensus address on the provider chain
+message ProviderConsAddress {
+  // Do not generate stringer for this type, we'll use sdk.ConsAddress.String() instead
+  option (gogoproto.goproto_stringer) = false;
+  bytes address = 1;
+}
+
+// AddressList contains a list of consumer consensus addresses
```
sainoe, line 214: nit: //ConsumerAddressList

sainoe:
Great work @smarshall-spitzbart! LGTM. 
As a suggestion, we could use a `Bytes() ` method returning the raw address bytes of the `ConsumerConsAddress` and `ProviderConsAddress` in the serialization rather than using `ToSdkConsAddress()`.

sainoe (approved):

smarshall-spitzbart:
> Great work @smarshall-spitzbart! LGTM. As a suggestion, we could use a `Bytes() ` method returning the raw address bytes of the `ConsumerConsAddress` and `ProviderConsAddress` in the serialization rather than using `ToSdkConsAddress()`.

Thanks for the suggestion! I went with the current design just because there were some spots in code that previously accepted the `sdk.ConsAddress` type, and the `ToSdkConsAddress()` helper allows the changes to be more of a drop-in replacement. `sdk.ConsAddress` is still automatically casted to byte array where that type is accepted

MSalopek (approved):
Awesome job!

Full integration tests passed on the pipeline.

PR #734, opened 2023-02-14
Fix TestHandleEquivocationProposal

smarshall-spitzbart:
# Description

`TestHandleEquivocationProposal` was recently added, the test involved calling `GetConsensusAddress` for two different addresses in `HandleEquivocationProposal`, both which would return `nil`.

This PR fixes the test to use valid cons addresses, unblocking #707 

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

n/a

## New behavior tests

n/a

### DIFF
--- a/x/ccv/provider/keeper/proposal_test.go
+++ b/x/ccv/provider/keeper/proposal_test.go
@@ -1002,9 +1002,8 @@ func TestBeginBlockCCR(t *testing.T) {
 }
 
 func TestHandleEquivocationProposal(t *testing.T) {
-	keeperParams := testkeeper.NewInMemKeeperParams(t)
-	keeper, ctx, _, mocks := testkeeper.GetProviderKeeperAndCtx(t, keeperParams)
-	tcFail := []*evidencetypes.Equivocation{
+
+	equivocations := []*evidencetypes.Equivocation{
 		{
 			Time:             time.Now(),
 			Height:           1,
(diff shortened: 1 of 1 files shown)


x/ccv/provider/keeper/proposal_test.go
```diff
@@ -1045,8 +1045,14 @@ func TestHandleEquivocationProposal(t *testing.T) {
 	propPass := &providertypes.EquivocationProposal{
 		Equivocations: []*evidencetypes.Equivocation{tcPass[0], tcPass[1]},
 	}
-	keeper.SetSlashLog(ctx, tcPass[0].GetConsensusAddress())
```
smarshall-spitzbart, line 1048: Previously, `tcPass[0].GetConsensusAddress()` would return nil here. This caused odd behavior in `HandleEquivocationProposal`

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -1005,32 +1005,32 @@ func TestHandleEquivocationProposal(t *testing.T) {
...
 		},
-		&evidencetypes.Equivocation{
+		{
 			Time:             time.Now(),
 			Height:           1,
 			Power:            1,
-			ConsensusAddress: "addr4",
+			ConsensusAddress: "cosmosvalcons1ezyrq65s3gshhx5585w6mpusq3xsj3ayzf4uv6",
 		},
 	}
```
yaruwangway, line 1035: Hi, why `tcPass[1]` is the same test case as `tcFail[1]`

smarshall-spitzbart: I will refactor the test to use the same structure for each test case, good catch 👍

sainoe (approved):
LGTM

x/ccv/provider/keeper/proposal_test.go
smarshall-spitzbart: Done

MSalopek (approved):

PR #735, opened 2023-02-14
allow using gaia as provider in integration tests

MSalopek:
# Description

This PR updates integration test setup and allows using gaia as the provider chain.
Use any of these commands to use the **latest tagged version of gaia** in integration tests.
```bash
# run full integration tests in sequence (including multiconsumer) using latest tagged gaia
make test-gaia-integration

# OR

# run only happy path integration tests using latest tagged gaia
make test-gaia-integration-short

# OR

# run full integration tests in parallel (including multiconsumer) using latest tagged gaia
make test-gaia-integration-parallel
```

## Linked issues

Closes: No issue assigned.

## Type of change

If you've checked more than one of the first three boxes, consider splitting this PR into multiple PRs!

- [ ] `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes
- [ ] `Fix`: Changes and/or adds code behavior, specifically to fix a bug
- [ ] `Refactor`: Changes existing code style, naming, structure, etc.
- [x] `Testing`: Adds testing
- [ ] `Docs`: Adds documentation

## Regression tests
Using ICS provider app in integration tests is not changed.

PR #737, opened 2023-02-14
Gaia integration tests with local sdk

smarshall-spitzbart:
# Description

Expands on #735, with the ability to set the `--local-sdk-path` flag

The following command worked for me, where I've checked out [this branch](https://github.com/cosmos/cosmos-sdk/tree/v0.45.13-ics) to `path` locally. We can local debug prints as needed and inspect them in docker
```bash
go run ./tests/integration/... --happy-path-only --use-gaia --local-sdk-path "path"
```

## Linked issues

Closes: `#<issue>`

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

n/a

## New behavior tests

n/a

### DIFF
--- a/Dockerfile.gaia
+++ b/Dockerfile.gaia
@@ -13,9 +13,21 @@ ENV GOOS=linux
 ENV GOFLAGS="-buildvcs=false"
 
 # fetch gaia from tag and build it
-RUN if [  -z ${GAIA_TAG} ] ; then git clone https://github.com/cosmos/gaia.git && cd gaia && git checkout $(git tag | tail -1) && make build; else git clone https://github.com/cosmos/gaia.git && cd gaia && git checkout ${GAIA_TAG} && make build ; fi
-
-# if GAIA_TAG is not set, build the latest tagged version
+RUN git clone https://github.com/cosmos/gaia.git && \
+    cd gaia && \
+    # if GAIA_TAG is not set, build the latest tagged version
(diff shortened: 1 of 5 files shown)


glnro (approved):

Dockerfile.gaia
```diff
@@ -13,9 +13,21 @@ ENV GOOS=linux
 ENV GOFLAGS="-buildvcs=false"
 
 # fetch gaia from tag and build it
```
glnro, line 15: Would also be ideal if we could pass a local build of gaia

smarshall-spitzbart: Agreed, @MSalopek or I could prob knock this out pretty quickly by just pattern matching the `-use-local-sdk` flag does things

smarshall-spitzbart:
> Can this be set to draft? I figured out a more comprehensive way in #738. I need to clean it up since it was used for debugging purposes.
> 
> We will have this:
> 
> * pulling gaia from tag
> * using latest gaia release
> * using both local sdk and local gaia
>   @smarshall-spitzbart

Imo it's not neccessary to specify both a local gaia and sdk, because the local gaia could reference whatever sdk it wants from it's go.mod

I agree that we can consolidate things, lets chat on a zoom call and we can figure it out

smarshall-spitzbart:
Just confirmed expected behavior with the CLI command mentioned above, this PR should be good to go

sainoe (approved):

PR #740, opened 2023-02-15
New validation for keeper fields

smarshall-spitzbart:
# Description

https://github.com/cosmos/gaia/pull/2197 fixed an incorrect ordering of keeper initialization in gaia's `app.go`. Before Simon's fix, gaia's ccv provider keeper held a pointer reference to a zeroed (outdated) external keeper. In short, ccv keepers should never have references to nil pointers, or references to concrete types which are zero-valued for that type. This PR adds a test which validates what I've described.

This new test would have caught the mentioned evidence keeper bug, and will hopefully prevent new bugs or regressions of this same kind in the future. 

To confirm test behavior, I've created two gaia branches that import this PR's ICS branch, [see comparison](https://github.com/cosmos/gaia/compare/v9-should-fail...v9-should-pass). [One branch does fail w/o Simon's fix](https://github.com/cosmos/gaia/compare/release/v9.0.x...v9-should-fail), and the [other branch passes w/ the fix](https://github.com/cosmos/gaia/compare/release/v9.0.x...v9-should-pass) ✅ 

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

n/a

## New behavior tests

n/a

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -127,28 +127,6 @@ func (k Keeper) mustValidateFields() {
 	}
 }
 
-// ExposeAllFields exposes (only by value) all the un-exported fields of the ccv provider keeper,
-// purely for testing purposes. DO NOT Use this method outside of testing, as it breaks encapsulation.
-func (k Keeper) ExposeAllFields() (
-	codec.BinaryCodec,
-	sdk.StoreKey,
-	paramtypes.Subspace,
-	ccv.ScopedKeeper,
-	ccv.ChannelKeeper,
-	ccv.PortKeeper,
-	ccv.ConnectionKeeper,
-	ccv.AccountKeeper,
-	ccv.ClientKeeper,
-	ccv.StakingKeeper,
-	ccv.SlashingKeeper,
-	ccv.EvidenceKeeper,
-	string) {
-
-	return k.cdc, k.storeKey, k.paramSpace, k.scopedKeeper, k.channelKeeper,
-		k.portKeeper, k.connectionKeeper, k.accountKeeper, k.clientKeeper, k.stakingKeeper,
-		k.slashingKeeper, k.evidenceKeeper, k.feeCollectorName
-}
-
 // Logger returns a module-specific logger.
 func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", "x/"+host.ModuleName+"-"+types.ModuleName)
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -990,3 +990,25 @@ func (k Keeper) GetFirstVscSendTimestamp(ctx sdk.Context, chainID string) (vscSe
 
 	return types.VscSendTimestamp{}, false
 }
+
+// SetSlashLog updates validator's slash log for a consumer chain
+// If an entry exists for a given validator address, at least one
+// double signing slash packet was received by the provider from at least one consumer chain
+func (k Keeper) SetSlashLog(
+	ctx sdk.Context,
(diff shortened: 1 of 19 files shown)


sainoe (approved):
Great work @smarshall-spitzbart

x/ccv/provider/keeper/keeper.go
```diff
@@ -73,6 +73,28 @@ func NewKeeper(
 	}
 }
 
+// ExposeAllFields exposes (only by value) all the un-exported fields of the ccv provider keeper,
+// purely for testing purposes. DO NOT Use this method outside of testing, as it breaks encapsulation.
+func (k Keeper) ExposeAllFields() (
```
MSalopek, line 78: Although I see the point you are making, maybe consider making a validator func instead of exposing private fields. And run that validator func inside the provider keeper init procedure - the node should not start if the keepers are not available.

```golang

func (k Keeper) ValidateInit() {
	// make checks on the keeper
	panic("EvidenceKeeprs is not set")
}
```

MSalopek (changes requested):
Thank you for raising the issue and going through all this effort!

I understand the concern, but please consider making a validator func instead of exposing private fields.

The provided tests are already checking that the fields are not zero values.

smarshall-spitzbart:
> The provided tests are already checking that the fields are not zero values.

This is not true, as no e2e test caught the evidence keeper bug. All e2e tests have to be manually crafted and its easy to overlook their use, especially for external contributors.

That being said, I'm not opposed to the idea of making this a runtime check with panics instead of a test

MSalopek:
> > The provided tests are already checking that the fields are not zero values.
> 
> This is not true, as no e2e test caught the evidence keeper bug. All e2e tests have to be manually crafted and its easy to overlook their use, especially for external contributors.
> 
> That being said, I'm not opposed to the idea of making this a runtime check with panics instead of a test

Meant to say:
> the tests provided in this PR are only checking for zero values

Apologies for my bad english

MSalopek (approved):
@smarshall-spitzbart Thank you for updating the PR!

x/ccv/provider/keeper/keeper.go
```diff
@@ -71,6 +72,59 @@ func NewKeeper(
...
 		feeCollectorName: feeCollectorName,
 	}
+
+	k.mustValidateFields()
+	return k
+}
+
+// Validates that the provider keeper is initialized with non-zero and
+// non-nil values for all its fields. Otherwise this method will panic.
+func (k Keeper) mustValidateFields() {
```
glnro, line 82: Wondering in the future if it's worth refactoring to have `mustValidateFields` be generic for keepers, then pass in a set of validations instead per keeper....

```
validations := map[string]func(Keeper) bool{
	"number of fields in provider keeper is not 13": func(keeper Keeper) bool { return reflect.ValueOf(k).NumField() != 13 },
	"cdc is zero-valued or nil": func(k Keeper) bool { return reflect.ValueOf(k.cdc).IsZero() },
	"storeKey is zero-valued or nil": func(keeper Keeper) bool { return reflect.ValueOf(k.storeKey).IsZero()},
}

...

func (k Keeper) mustValidateFields(validations map[string]func(Keeper) bool) {
	for errMessage, failedValidation := range validations {
		if failedValidation(k) {
			panic(errMessage)
		}
	}
}
```

glnro (approved):

x/ccv/provider/keeper/keeper.go
smarshall-spitzbart: This is a solid idea! Although in our repo there will only ever be 2 keepers defined, so maybe not worth adding the additional code and complexity? That sort of functionality living within the sdk repo would make the most sense to me.

Lmk if you strongly disagree and I can refactor in a future PR

Issue #377, opened 2022-10-05
Tidy up docs directory


danwt:
# Problem

People don't read docs/ directories because they are usually unorganized and out of date. It is hard to work on docs for the same reasons. Ours is unorganized and partially out of date. We should improve it.

# Closing criteria

Remove anything that is outdated or abandonded. Make sure things that are up to date and used are in sensible places.

# Concretely

We have e.g. https://github.com/cosmos/interchain-security/tree/main/docs/unbonding-hooks which is abandonware? We have https://github.com/cosmos/interchain-security/tree/main/docs/figures which dont' have a clear use. Ect.

mpoke:
The figures are needed here https://github.com/cosmos/interchain-security/blob/main/docs/quality_assurance.md#interchain-security-workflow

PR #758, opened 2023-03-07
Tidy up docs directory

smarshall-spitzbart:
# Description

Deletes docs that're outdated. Future PRs will migrate uncategorized markdowns like [this](https://github.com/cosmos/interchain-security/blob/8161916d5c2b41584afc2de5679e585545d4169e/docs/info_for_validators.md) into docusaurus pages 

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/377

## Type of change

- [x] `Docs`: Adds/removes documentation

## Regression tests

n/a

## New behavior tests

n/a

## Versioning Implications
n/a

## Targeting


docs/interchain-security/proto-docs.md
```diff
@@ -1,518 +0,0 @@
-<!-- This file is auto-generated. Please do not modify it yourself. -->
```
mpoke, line 1: This is auto-generated. This is good to have.

docs/ccv-distribution/ccv-distribution.md
```diff
@@ -1,92 +0,0 @@
-# CCV Distribution
```
mpoke, line 1: This document is useful. Can we move it do a issue on how to improve the distribution. We may need this for other iterations of ICS.

mpoke (approved):
Nice. See my comments bellow.

docs/ccv-distribution/ccv-distribution.md
smarshall-spitzbart: Yea I can add this back, is it still representative of the code how it exists now?

smarshall-spitzbart: I think this file should still be removed as it's not documentation on how the current code in main works. The diagram is maintained and could be addressed in https://github.com/cosmos/interchain-security/issues/760

docs/interchain-security/proto-docs.md
smarshall-spitzbart: We could use something like https://github.com/pseudomuto/protoc-gen-doc. But I'm having trouble understanding what that documentation would be used for, since `.proto` files are already pretty self explanatory imo

MSalopek (approved):

PR #764, opened 2023-03-10
Add dependabot config

mpoke:
## Type of change

Adds a config file to update dependencies introduced by us directly (i.e., [version updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates)).

## Versioning Implications

## Targeting

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


smarshall-spitzbart:
This new file will allow GH to run a process weekly to generate PRs that fix security vulnerabilities, right?

smarshall-spitzbart (approved):
I've updated the PR description with your comment, LGTM

sainoe (approved):

Issue #785, opened 2023-03-16
`x/ccv/provider/types.SlashLogKey()` uses the wrong byte prefix


tbruyelle:
## Summary of Bug

The `x/ccv/provider/types.SlashLogKey()` function uses the `SlashAcksBytePrefix` byte prefix instead of `SlashLogBytePrefix`.

## Version

4c318b46c9cb13c67f6cce8b9d3e96ba62edc335
https://github.com/cosmos/interchain-security/blob/4c318b46c9cb13c67f6cce8b9d3e96ba62edc335/x/ccv/provider/types/keys.go#L447-L450

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

jtremback:
Great catch! This is definitely a mistake, but luckily it will not cause any problems in production. We'll put a fix in the next release.

Usually, `SlashAcksBytePrefix` is used to store records having to do with a particular chain id.

For this bug to have any effect at all, someone would need to pass a consumer creation proposal with the chain id being a serialized public key belonging to a validator.

Once this was done, the consumer chain could cause records to be written to the storage location by sending downtime packets. This would then make it possible for someone to create an equivocation slashing governance proposal for the validator with that key. This proposal would then also have to be approved for anything to happen.

PR #787, opened 2023-03-16
Improved key tests

smarshall-spitzbart:
# Description

Improves key tests s.t they would have caught bugs like this: https://github.com/cosmos/interchain-security/issues/785

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

Existing key tests that were not touched

## New behavior tests

New tests in keys_test.go

## Versioning Implications

## Targeting


smarshall-spitzbart:
@mpoke this PR will fail as expected until #786 is merged

mpoke (approved):
LGTM

sainoe (approved):
LGTM

Issue #747, opened 2023-02-28
Pause unbondings while equivocation slashing proposal is in voting period


jtremback:
# Problem

Currently, if an equivocation slashing proposal is created after more than one week has passed since the equivocation, it is possible that the validator in question could unbond and get away without being slashed, since the unbonding period is 3 weeks, and the voting period is 3 weeks. For this reason, it might be good to pause unbondings for validators named in an equivocation slashing proposal until the proposal's voting period is over.

NOTE: This will be kind of complicated and janky to implement, and it may be that we would be better off just getting automatic equivocation slashing to work correctly. Just writing it here for future reference.

# Solution

With current Cosmos Hub settings, it would be sufficient to simply pause all unbondings for a validator that were started before the proposal was created. This is because new unbondings created after the proposal could not complete before the proposal's voting period ended, because the voting period is shorter than the unbonding period. However, if for some reason the voting period was set to be longer than the unbonding period, this would no longer hold, and it would be necessary to pause unbondings started after the proposal was created.

There is no way to get the governance module to run some code when a proposal is created, so we would need to create an antehandler which recognizes a tx creating an equivocation slashing proposal, validates it, and pauses the unbondings. This is kind of janky, but would work.

The API necessary to pause unbondings was added to Cosmos-SDK as part of the changes that we made to enable replicated security. There are 3 types of unbondings: undelegations, redelegations, and validator self unbondings. To pause unbondings for a specific validator, we would first need to get a list of all undelegations, redelegations and self unbondings for that validator, and then pause them.

To get undelegations, call [GetValidatorDelegations](https://github.com/cosmos/cosmos-sdk/blob/d6c8598683ce1ac8efd39941109eae78b8211ddf/x/staking/keeper/delegation.go#LL57C17-L57C40). This should be an efficient function that uses an index and does not iterate any records that it does not need to, but we should verify that.

To get redelegations, call [GetRedelegationsFromSrcValidator](https://github.com/cosmos/cosmos-sdk/blob/d6c8598683ce1ac8efd39941109eae78b8211ddf/x/staking/keeper/delegation.go#LL442C17-L442C49).

I don't remember how to get validator self unbondings right now, but it should also be possible.

[...]

PR #791, opened 2023-03-17
feat: pause unbondings during equivocation proposal voting period

tbruyelle:
# Description

The change registers 2 gov module hooks in the provider module:

- `AfterProposalDeposit`: if the proposal is an equivocation proposal in voting period, then call `PutUnbondingOnHold` for each unbondings of each validators found in the proposal.

- `AfterProposalVotingPeriodEnded`: if the proposal is an equivocation proposal, then call `UnbondingCanComplete` for each unbondings of each validators found in the proposal.

A new key is also added, to store the equivocation proposal ID for which unbondings have been put on hold. This covers 2 specific cases:

- The gov module allows additional deposits even if the proposal is already in voting period. So when `AfterProposalDeposit` is invoked, we have to make sure the proposal is in voting period for the first time before puting the unbondings on hold. This is simply handled by checking if the proposal ID exists in the store at the beginning of the hook, and then storing it if not.

- If the provider chain is upgraded with this change and there's already an equivocation proposal in voting period, `AfterProposalVotingPeriodEnded` could be invoked without the initial `AfterProposalDeposit`, so some unbondings could be un-paused while they hadn't been paused (conflicts with `AfterUnbondingInitiated` hook). To prevent that, we check the proposal ID exists in the store, which means `AfterProposalDeposit` has been called prevouisly.

## Linked issues

Closes: #747

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

## New behavior tests

A new e2e test have been added in `tests/e2e/proposal.go`.

## Versioning Implications

## Targeting

### DIFF
--- a/app/provider/app.go
+++ b/app/provider/app.go
@@ -376,7 +376,7 @@ func New(
 		stakingtypes.NewMultiStakingHooks(
 			app.DistrKeeper.Hooks(),
 			app.SlashingKeeper.Hooks(),
-			app.ProviderKeeper.Hooks(),
+			app.ProviderKeeper.Hooks(app.GovKeeper),
 		),
 	)
 
@@ -426,7 +426,7 @@ func New(
 		AddRoute(providertypes.RouterKey, ibcprovider.NewProviderProposalHandler(app.ProviderKeeper)).
 		AddRoute(ibcclienttypes.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper))
 
-	app.GovKeeper = govkeeper.NewKeeper(
+	govKeeper := govkeeper.NewKeeper(
 		appCodec,
 		keys[govtypes.StoreKey],
 		app.GetSubspace(govtypes.ModuleName),
@@ -434,6 +434,11 @@ func New(
 		app.BankKeeper,
 		&stakingKeeper,
 		govRouter,
+	)
+	app.GovKeeper = *govKeeper.SetHooks(
+		govtypes.NewMultiGovHooks(
+			app.ProviderKeeper.Hooks(govKeeper),
+		),
 	)
 
 	app.TransferKeeper = ibctransferkeeper.NewKeeper(
@@ -761,6 +766,11 @@ func (app *App) GetProviderKeeper() ibcproviderkeeper.Keeper {
 // GetE2eStakingKeeper implements the ProviderApp interface.
 func (app *App) GetE2eStakingKeeper() e2e.E2eStakingKeeper {
 	return app.StakingKeeper
+}
+
+// GetE2eGovKeeper implements the ProviderApp interface.
+func (app *App) GetE2eGovKeeper() e2e.E2eGovKeeper {
+	return app.GovKeeper
 }
 
 // GetE2eBankKeeper implements the ProviderApp interface.
--- a/tests/e2e/proposal.go
+++ b/tests/e2e/proposal.go
@@ -1,0 +1,243 @@
+package e2e
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	providertypes "github.com/cosmos/interchain-security/x/ccv/provider/types"
+)
+
+func (s *CCVTestSuite) TestPauseUnbondingOnEquivocationProposal() {
+	tests := []struct {
+		name                string
+		setup               func() govtypes.Content
+		expectedWithoutProp unbondingsOnHold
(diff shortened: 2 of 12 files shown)


app/provider/app.go
```diff
@@ -376,7 +376,7 @@ func New(
 		stakingtypes.NewMultiStakingHooks(
 			app.DistrKeeper.Hooks(),
 			app.SlashingKeeper.Hooks(),
-			app.ProviderKeeper.Hooks(),
+			app.ProviderKeeper.Hooks(app.GovKeeper),
```
tbruyelle, line 379: As far as I can tell this is not very conventional, but on the other hand, adding the `govKeeper` to the `providerKeeper` is even more ugly, because the `govKeeper` needs the `providerKeeper` for the proposal routes. So we would have to write something like (pseudo code): 

```go
providerKeeper := providerKeeper.NewKeeper( ...  govKeeper ...) // govKeeper is nil for now
govKeeper := govKeeper.NewKeeper( ... providerKeeper ... )
providerKeeper.SetGovKeeper( govKeeper ) // finally set a non-nil govKeeper
```

jtremback:
Thanks! Will review this soon.

smarshall-spitzbart (commented):
Awesome work

tests/e2e/proposal.go
```diff
@@ -0,0 +1,243 @@
...
+
+			proposalID := submitProp(s, govContent)
+			checkStakingUnbondingOnHoldRefCount(s, valAddr, tt.expectedWithoutProp)
+			// assert that until voting period starts, there's no pause triggered
+			addDepositProp(s, proposalID, dustAmt)
+			checkStakingUnbondingOnHoldRefCount(s, valAddr, tt.expectedWithoutProp)
+			// activate prop
+			addDepositProp(s, proposalID, minDepositAmt)
+			checkStakingUnbondingOnHoldRefCount(s, valAddr, tt.expectedDuringProp)
+			// assert that an additionnal deposit done after the activation doesn't
```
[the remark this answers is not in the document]
tbruyelle: How can that happen ? I see no message in the gov module that could reset the voting period.

tests/e2e/proposal.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	addDepositProp := func(s *CCVTestSuite, proposalID uint64, depositAmt sdk.Coins) {
+		depositorAddr := s.providerChain.SenderAccount.GetAddress()
+		_, err := s.providerApp.GetE2eGovKeeper().AddDeposit(
+			s.providerCtx(), proposalID, depositorAddr, depositAmt,
+		)
+		s.Require().NoError(err)
+	}
+	waitFor := func(s *CCVTestSuite, period time.Duration) {
+		s.providerChain.CurrentHeader.Time = s.providerChain.CurrentHeader.Time.Add(period)
```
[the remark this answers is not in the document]
tbruyelle: Good point, we didn't noticed this function. [6df4b95](https://github.com/cosmos/interchain-security/pull/791/commits/6df4b95cb4aac2cc27e7275993ea1e89cf742b94)

app/provider/app.go
tbruyelle: I see one benefit of having an additional parameter in `Hooks()`: it breaks compilation after upgrade. Typically when gaia will upgrade its interchain-security dependency, the fix will be obvious, while on the other hand, having to call `SetGovKeeper` can be easily forgotten.
Is there any validation we could setup to ensure `SetGovKeeper` is actually called ? A little bit like `keeper.mustValidateField` but for setters.

faddat:
Do you guys figure that this one is complete? I think it might solve one of the test cases in the 47 work that we have been doing

mpoke:
Converting to draft as this feature requires an analysis on the potential risks. @tbruyelle @jtremback it would be useful to create an ADR to be able to discuss on it.

tbruyelle:
> Converting to draft as this feature requires an analysis on the potential risks. @tbruyelle @jtremback it would be useful to create an ADR to be able to discuss on it.

ADR ready for review https://github.com/cosmos/interchain-security/pull/964

x/ccv/provider/keeper/relay_test.go
```diff
@@ -567,7 +567,7 @@ func TestHandleVSCMaturedPacket(t *testing.T) {
 
 	// Start first unbonding without any consumers registered
 	var unbondingOpId uint64 = 1
-	err := pk.Hooks().AfterUnbondingInitiated(ctx, unbondingOpId)
+	err := pk.Hooks(mocks.MockGovKeeper).AfterUnbondingInitiated(ctx, unbondingOpId)
 	require.NoError(t, err)
 	// Check that no unbonding op was stored
 	_, found := pk.GetUnbondingOp(ctx, unbondingOpId)
```
witaliy12, line 573: cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn

Issue #781, opened 2023-03-15
Validate staking functionalities for sovereign -> consumer changeover


smarshall-spitzbart:
Slashing, jailing, etc. must work correctly after the upgrade and before the provider valset starts validating blocks. Confirm that consumer initiated slashing is functional even before CCV is established. We also need the "old" staking keeper around for at least the consumer unbonding period.

smarshall-spitzbart:
Notes from zoom call: All of this is about dealing with how to punish the old valset. We need the old staking keeper for at least the unbonding period

Start with Jail/unjail

smarshall-spitzbart:
@mpoke here's an update after brainstorming and drafting up some code:

### Concerns / notables

- How are delegation records migrated between staking keepers? 
- Consider that staking token before changeover is STRD, and after it’s ATOM
- previously sovereign staking module (that satisfies democracy module?) needs to have slashing capabilities to slash STRD

### Slash behavior

|  | After upgrade, before changeover complete | After changeover complete |
| --- | --- | --- |
| val in old Valset only | call sovStaking.Slash() to slash STRD | call sovStaking.Slash() for STRD. Would need to add some tests around this confirming STRD can still be burned. Sov staking should handle unbonding |
| val in new Valset only | no-op | call ccvStaking.Slash() to send slash req |
| In both sets | call sovStaking.Slash() | do one of two above boxes depending on infraction height |

### Jail Behavior - TLDR always No-op

|  | After upgrade, before changeover complete | After changeover complete |
| --- | --- | --- |
| Old Valset only | Call sovStaking.Jail(). essentially a no-op. new valset in effect by the time jailing happens  | Validator already has zero power. No way to punish (no-op) |
| New Valset only | no-op | ccvStaking.Jail() (no-op) |
| In both sets | no-op | ccvStaking.Jail() (no-op) |

### Unjail Behavior

Always a no-op as well. Unjailing from the new valset is done on provider. Unjailing from old valset is n/a after the changeover is complete. Unjailing from old valset after upgrade, before changeover is complete doesn't really do much since the new valset takes over in 2-3 blocks anyways

The remaining methods from `validators.go` can be handled on a case by case basis, but I don't think we'd break anything if we just leave them unimplemented. Further testing can confirm this. Lmk if you have any thoughts

smarshall-spitzbart:
[...]

Issue #756, opened 2023-03-06
EPIC: Standalone to consumer changeover


smarshall-spitzbart:
# Problem

https://github.com/cosmos/interchain-security/pull/697 gave us an awesome implementation of https://github.com/cosmos/ibc/pull/840. There are remaining tasks that we need to get into a release for Stride's standalone to consumer changeover

# Closing criteria

The below tasks are completed and merged into main, where all comments from https://github.com/cosmos/interchain-security/pull/697 are addressed. 

In accomplishing all these tasks, we should try our best to make the changes affect the consumer only.

# Task list

```[tasklist]
### Must have
- [x] Initial implementation is merged, follows ICS patterns -> https://github.com/cosmos/interchain-security/pull/757
- [ ] https://github.com/cosmos/interchain-security/issues/783
- [ ] https://github.com/cosmos/interchain-security/issues/781
- [ ] https://github.com/cosmos/interchain-security/issues/782
- [ ] https://github.com/cosmos/interchain-security/issues/885
- [ ] https://github.com/cosmos/interchain-security/issues/895
```

```[tasklist]
### Nice to have
- [ ] https://github.com/cosmos/interchain-security/issues/790
- [ ] https://github.com/cosmos/interchain-security/issues/780
- [ ] Various integration and UTs for things like endblockers, involving newly introduced code. Slashing edge cases, etc.
- [ ] https://github.com/cosmos/interchain-security/issues/894
- [ ] https://github.com/cosmos/interchain-security/issues/1069
```

PR #757, opened 2023-03-06
feat: standalone to consumer changeover part 1

smarshall-spitzbart:
# Description

PR based off branch in fork repo, where commits originally come from https://github.com/cosmos/interchain-security/pull/697

This PR serves as the beginnings to enabling a proper sovereign to consumer changeover process. Not all issues from the EPIC are solved here. See #756 for future work.

## Linked issues

#756 

## Type of change

- [x] `Feature`: Changes and/or adds code behavior, irrelevant to bug fixes
- [x] `Testing`: Adds testing

## Regression tests

A bunch of existing tests, there are more to be made for endblockers, etc. 

## New behavior tests

tests for new CRUD, keys tests, and `TestChangeoverToConsumer()` 

## Versioning Implications

- [x] This PR will affect [semantic versioning as defined for ICS](../CONTRIBUTING.md#semantic-versioning)

If the above box is checked, which version should be bumped?

- [ ] `MAJOR`: Consensus breaking changes to both the provider and consumers(s), including updates/breaking changes to IBC communication between provider and consumer(s)
- [x] `MINOR`: Consensus breaking changes which affect either only the provider or only the consumer(s)
- [ ] `PATCH`: Non consensus breaking changes

## Targeting

Please select one of the following:

- [x] This PR is only relevant to main

PR #794, opened 2023-03-20
Standalone to consumer changeover - staking functionalities

smarshall-spitzbart:
# Description

Implements the decisions about how the main staking methods should be handled for a chain going through a standalone to consumer changeover. 

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/781

Epic: #756 

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

Tests added in `validators_test.go`, integration test added in `slashing.go`

## New behavior tests

Behavior of normal consumer and normal democracy consumer shouldn't be affected

## Versioning Implications

## Targeting


smarshall-spitzbart:
TODO: explicit flag for `previousStandaloneChain`

bmorphism:
Is this borg? 👽

x/ccv/consumer/keeper/changeover.go
```diff
@@ -5,17 +5,30 @@ import (
...
+		panic("ChangeoverIsComplete should only be called on previously standalone consumers")
+	}
+	return ctx.BlockHeight() >= k.FirstConsumerHeight(ctx)
+}
+
+// The first height that the ccv valset will be in effect is 2 blocks after init genesis height
+// (aka height that the ccv module first returned updates to tendermint), because if init genesis is block N,
+// the new valset is committed in block N+1, and in effect for block N+2.
+func (k Keeper) FirstConsumerHeight(ctx sdk.Context) int64 {
+	return k.GetInitGenesisHeight(ctx) + 2
```
mpoke, line 20: Use `ValidatorUpdateDelay+1`, see https://github.com/cosmos/cosmos-sdk/blob/c0fe4f7da17b7ec17d9bea6fcb57b4644f044b7a/types/staking.go#L18 and https://github.com/cosmos/cosmos-sdk/blob/c0fe4f7da17b7ec17d9bea6fcb57b4644f044b7a/x/slashing/keeper/infractions.go#L85.

mpoke (approved):
LGTM. Great work @smarshall-spitzbart

app/consumer-democracy/app.go
```diff
@@ -427,7 +427,7 @@ func New(
 		authtypes.FeeCollectorName,
 	)
 
-	// Setting the staking keeper is only needed for standalone to consumer changeover chains
+	// Setting the standalone staking keeper is only needed for standalone to consumer changeover chains
```
glnro, line 430: Nomenclature is strange to me, why not just refer to it as Sovereign to Consumer? Or is there a precedent that was decided already for this?

tests/integration/slashing.go
```diff
@@ -665,3 +665,41 @@ func (suite *CCVTestSuite) TestQueueAndSendSlashPacket() {
...
+	// Check slash packet was queued
+	pendingPackets = consumerKeeper.GetPendingPackets(suite.consumerCtx())
+	suite.Require().Len(pendingPackets.List, 1)
+
+	// Pass 5 blocks, confirming the consumer doesn't panic
+	suite.consumerChain.NextBlock()
+	suite.consumerChain.NextBlock()
+	suite.consumerChain.NextBlock()
+	suite.consumerChain.NextBlock()
+	suite.consumerChain.NextBlock()
```
glnro, line 691: Why not just iterate here or write a function to pass time?

```go
func produceEmptyBlocks(suite *CCVTestSuite, blocks int){
  for blocks > 0 {
    suite.consumerChain.NextBlock()
    blocks--
  }
}
```

app/consumer-democracy/app.go
smarshall-spitzbart: Yea there is a precedent here: https://github.com/cosmos/interchain-security/pull/757#discussion_r1142231475. "Sovereign" could make it seem like consumers have no sovereignty

tests/integration/slashing.go
smarshall-spitzbart: "empty blocks" is kind of true since no txs are submitted. But when we do something like this it's usually a sanity check that beginblock/endblock do or do not trigger some code. Reasoning is usually a case by case basis

For loop makes sense tho https://github.com/cosmos/interchain-security/commit/cb9a1ec93fdbed4c8f09cd4152deac687e46dd45

x/ccv/consumer/keeper/changeover.go
```diff
@@ -5,17 +5,30 @@ import (
...
 
+// ChangeoverIsComplete returns whether the standalone to consumer changeover process is complete.
+func (k Keeper) ChangeoverIsComplete(ctx sdk.Context) bool {
+	if !k.IsPrevStandaloneChain() {
+		panic("ChangeoverIsComplete should only be called on previously standalone consumers")
+	}
+	return ctx.BlockHeight() >= k.FirstConsumerHeight(ctx)
+}
+
+// The first height that the ccv valset will be in effect is 2 blocks after init genesis height
```
sainoe, line 16:
```suggestion
// FirstConsumerHeight returns the first height that the ccv valset will be in effect is 2 blocks after init genesis height
```

x/ccv/consumer/keeper/validators.go
```diff
@@ -102,11 +104,20 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
...
-func (k Keeper) Slash(ctx sdk.Context, addr sdk.ConsAddress, infractionHeight, power int64, _ sdk.Dec, infraction stakingtypes.InfractionType) {
+func (k Keeper) Slash(ctx sdk.Context, addr sdk.ConsAddress, infractionHeight, power int64, slashFactor sdk.Dec, infraction stakingtypes.InfractionType) {
 	if infraction == stakingtypes.InfractionEmpty {
 		return
 	}
 
+	// If this is a previously standalone chain and infraction happened before the changeover was completed,
+	// slash only on the standalone staking keeper.
+	if k.IsPrevStandaloneChain() && infractionHeight < k.FirstConsumerHeight(ctx) {
+		k.standaloneStakingKeeper.Slash(ctx, addr, infractionHeight, power, slashFactor, infraction)
```
sainoe, line 115: nits: the `standaloneStakingKeeper`'s slash method doesn't use the infraction arg so it can be replaced by a the infraction type zero value.
```suggestion
		k.standaloneStakingKeeper.Slash(ctx, addr, infractionHeight, power, slashFactor, InfractionEmpty)
```

sainoe: nits: the `standaloneStakingKeeper`'s slash method doesn't use the infraction arg so it can be replaced by a the infraction type zero value.
```suggestion
		k.standaloneStakingKeeper.Slash(ctx, addr, infractionHeight, power, slashFactor, stakingtypes.InfractionEmpty)
```

sainoe, line 115: nits:  `standaloneStakingKeeper's` slash method doesn't use the infraction arg so the infraction type zero value can be passed.
```suggestion
		k.standaloneStakingKeeper.Slash(ctx, addr, infractionHeight, power, slashFactor, stakingtypes.InfractionEmpty)
```

sainoe (approved):
Great work @smarshall-spitzbart! Check my comments

Issue #799, opened 2023-03-23
StopConsumerChain not running in cachedContext


MSalopek:
## Summary of Bug

A bug was identified in ICS provider code where the `StopConsumerChain` is not actually called in a cachedContext. Instead, it is run in the global app context.
This means that the consumer chain can get removed unexpectedly and its CCV chanel was closed as soon as the proposal tx gets handled.
https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/proposal.go#L599

## Version

`v9.0.x`
____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

PR #802, opened 2023-03-24
fix StopConsumerChain not running in cachedContext

MSalopek:
# Description

This PR resolves a bug in `StopConsumerChain`.

A bug was identified in ICS provider code where the StopConsumerChain is not actually called in a cachedContext. Instead, it is run in the global app context.

This meant that the consumer chain can get removed unexpectedly and its CCV chanel was closed as soon as the proposal tx gets handled.

This bug is resolved i

## Linked issues
Closes: #799 

## Type of change

- [x] `Testing`: Adds testing

## New behavior tests

Expanded existing tests to confirm that the consumer is not removed prematurely and also that it is cleanly removed when a proposal passes.

## Versioning Implications

- [x] This PR will affect [v1.1.0](../CONTRIBUTING.md#semantic-versioning)

## Targeting

- [ ] This PR is only relevant to main
- [x] This PR is relevant to main, and should also be back-ported to `v1.1.0`

### DIFF
--- a/tests/integration/steps.go
+++ b/tests/integration/steps.go
@@ -20,9 +20,10 @@ var happyPathSteps = concatSteps(
 	stepsUnbond("consu"),
 	stepsRedelegate("consu"),
 	stepsDowntime("consu"),
-	stepsRejectEquivocationProposal("consu", 2),    // prop to tombstone bob is rejected
-	stepsDoubleSignOnProviderAndConsumer("consu"),  // carol double signs on provider, bob double signs on consumer
-	stepsSubmitEquivocationProposal("consu", 2),    // now prop to tombstone bob is submitted and accepted
+	stepsRejectEquivocationProposal("consu", 2),   // prop to tombstone bob is rejected
+	stepsDoubleSignOnProviderAndConsumer("consu"), // carol double signs on provider, bob double signs on consumer
+	stepsSubmitEquivocationProposal("consu", 2),   // now prop to tombstone bob is submitted and accepted
+	stepsStartHermes(),
 	stepsConsumerRemovalPropNotPassing("consu", 3), // submit removal prop but vote no on it - chain should stay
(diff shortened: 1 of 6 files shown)


x/ccv/provider/keeper/proposal_test.go
```diff
@@ -392,16 +392,21 @@ func TestHandleConsumerRemovalProposal(t *testing.T) {
...
+		// chainID of the consumer chain
+		// tests need to check that the CCV channel is not closed prematurely
+		chainId string
 	}
 
 	// Snapshot times asserted in tests
 	now := time.Now().UTC()
-	hourFromNow := now.Add(time.Hour).UTC()
+	hourAfterNow := now.Add(time.Hour).UTC()
+	hourBeforeNow := now.Add(-time.Hour).UTC()
```
yaruwangway, line 404: why not use `now.Sub()` ?

tests/integration/steps.go
```diff
@@ -23,7 +23,9 @@ var happyPathSteps = concatSteps(
 	stepsRejectEquivocationProposal("consu", 2),   // prop to tombstone bob is rejected
 	stepsDoubleSignOnProviderAndConsumer("consu"), // carol double signs on provider, bob double signs on consumer
 	stepsSubmitEquivocationProposal("consu", 2),   // now prop to tombstone bob is submitted and accepted
-	stepsStopChain("consu", 3),
+	stepsStartHermes(),
+	stepsConsumerRemovalPropNotPassing("consu", 3), // submit removal prop but vote no on it - chain should stay
```
glnro, line 27: Irrelevant for now but perhaps future we move away from hard coding the proposal number and pass in something that can auto increment.

jtremback: Nope, not how this is supposed to work, but we can talk about that later at the retreat with @p-offtermatt

PR #805, opened 2023-03-28
bump: sdk v0.45.15-ics

faddat:
# Description

This PR bumps:

* cosmos-sdk to v0.45.15
* comet-bft to v0.34.27

It also:

* suppresses copylocks for tests

## Linked issues

* https://github.com/cosmos/gaia/issues/2313

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

## New behavior tests

## Versioning Implications

## Targeting


faddat:
Discussed on tuesday and still here blocking consumers from launching with latest sdk

go.mod
```diff
@@ -135,10 +148,8 @@ require (
 )
 
 replace (
-	github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.2 //indirect
-	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.13-ics
+	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.15-ics
```
mpoke, line 151: This bump is not SM breaking. Why combine it with a bump to ibc-go 4.3.0 which is ST breaking? Please create separate PRs.

go.mod
```diff
@@ -3,54 +3,62 @@ module github.com/cosmos/interchain-security
 go 1.18
 
 require (
-	github.com/cosmos/cosmos-sdk v0.45.11
+	github.com/confio/ics23/go v0.9.0
```
mpoke, line 6: Please stop refactoring in the same PRs where changes are introduces as it makes it difficult to review.

mpoke (changes requested):
Thanks @faddat for your contribution. See my comments below.

faddat:
you're correct

faddat:
the need for the btcd replace is eliminated by using cometbft

go.mod
mpoke: Clearly `github.com/confio/ics23/go v0.9.0` was not added by this PR. It was just moved around the file which makes the review difficult. The contributions in a single PR should be minimal to improve the review process and avoid unwanted bugs.

faddat: @mpoke - SDK v0.45.15 does add that.

It is not a refactor. 

It is a part of changing the sdk version.

go.mod
faddat: I've removed the change to ibc-go v4.3.0

go.mod
faddat: go mod tidy organizes files in go

faddat:
We should use comet.

go.mod
faddat: Oh that's right I forgot I definitely got rid of the weird artificial group in the middle of the direct imports because if it's there there should be a comment explaining it but it's just sort of there so I got rid of it

faddat:
If we were on this repo, then merges would cease to be a diffculty

mpoke (approved):
LGTM. Thanks for the contribution @faddat

tests/integration/distribution.go
```diff
@@ -242,8 +242,8 @@ func (s *CCVTestSuite) TestEndBlockRD() {
 	}
 }
 
-// getEscrowBalance gets the current balances in the escrow account holding the transferred tokens to the provider
-func (s CCVTestSuite) getEscrowBalance() sdk.Coins {
+// getEscrowBalance gets the current balances in the escrow account holding the transfered tokens to the provider
```
smarshall-spitzbart, line 245: Are the changes here and onward intentional? If so could we add to the description of the PR why they are needed to bump sdk

faddat: that's intentional.  It set off the copy locks linter.  So I suppressed it because we are copying locks in the test.

I don't think there's any additional notation needed, but sure

mpoke (changes requested):
@faddat thanks for your contribution. Please resolve the conflicts and remove all other changes besides the ones in `go.mod` and `go.sum`. Bumping SDK to 0.45.15-ics shouldn't results in other changes.

faddat:
@mpoke that's true, but it does.  I can revert them, I expect the linter not to pass.

mpoke:
> @mpoke that's true, but it does. I can revert them, I expect the linter not to pass.
> 
> Overall the approach that we use at Notional, is to let the linter guide things like this.

Hi @faddat. It is weird that a change in go.mod break the linter. Regardless, could you please add back the changes that fix the linter. Thanks.

sainoe (approved):

mpoke (approved):
Thanks @faddat

PR #823, opened 2023-04-04
use ibc v4.3.0

faddat:
# Description

Separated from #805.  Equally important.

fix: do we want another cosmos hub quicksilver ica incident?

Since we do not want that, let's always ship with the latest available code.

PS: yes I grouped the direct imports.

PPS: no, that is not a refactor, if they are separated there should be a comment explaining grouping.

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

## New behavior tests

## Versioning Implications

## Targeting


go.mod
```diff
@@ -3,13 +3,18 @@ module github.com/cosmos/interchain-security
 go 1.18
 
 require (
-	github.com/cosmos/cosmos-sdk v0.45.11
+	github.com/confio/ics23/go v0.9.0
+	github.com/cosmos/cosmos-sdk v0.45.12
```
jtremback, line 7: Note for tomorrow- does this need to be 45.13?

go.mod
```diff
@@ -18,6 +23,7 @@ require (
 	github.com/tendermint/tm-db v0.6.7
 	github.com/tidwall/gjson v1.14.4
 	golang.org/x/crypto v0.5.0 // indirect
+	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
```
jtremback, line 26: What's this one?

jtremback (approved):
LGTM

go.mod
jtremback: We override it in the replace block with 45.13 anyway

smarshall-spitzbart (approved):

go.mod
faddat: Case it should say 13 but there's a reason why it was bundled together with the one that upgrades to 4515

faddat:
I fix

faddat:
thought this was approved?

Issue #782, opened 2023-03-15
Maintain transfer channel before/after sovereign -> consumer changeover


smarshall-spitzbart:
Transfer channel needs to be maintained through the migration. New transfer channel would result in undesirable legacy denoms

PR #832, opened 2023-04-07
Standalone to consumer changeover - recycle existing transfer channel

smarshall-spitzbart:
# Description

Allows previously standalone chains to use the already existing transfer channel from the standalone chain

## Linked issues

Closes: https://github.com/cosmos/interchain-security/issues/782

Epic: #756 

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

Existing handshake tests like `TestOnChanOpenAck`

## New behavior tests

CRUD unit tests, and new integration test in `changeover.go`

## Versioning Implications

## Targeting


MSalopek (approved):
Looks good!

e2e tests will be added at a later point

x/ccv/consumer/keeper/validators.go
```diff
@@ -111,7 +111,7 @@ func (k Keeper) Slash(ctx sdk.Context, addr sdk.ConsAddress, infractionHeight, p
 
 	// If this is a previously standalone chain and infraction happened before the changeover was completed,
 	// slash only on the standalone staking keeper.
-	if k.IsPrevStandaloneChain() && infractionHeight < k.FirstConsumerHeight(ctx) {
+	if k.IsPrevStandaloneChain(ctx) && infractionHeight < k.FirstConsumerHeight(ctx) {
 		k.standaloneStakingKeeper.Slash(ctx, addr, infractionHeight, power, slashFactor, stakingtypes.InfractionEmpty)
```
jtremback, line 115: We need to dig into how exactly consumer chains using the democracy module or other things will populate standaloneStakingKeeper

jtremback (approved):
LGTM

x/ccv/consumer/keeper/validators.go
mpoke: What do you mean? That's the staking module with the native token from when the chain was a standalone chain.

PR #833, opened 2023-04-10
Soft opt out

smarshall-spitzbart:
# Description

Please include a summary of the changes and the related issue. If the issue was ambiguous try to clarify it in this section.

## Linked issues

Closes: `#<issue>`

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

## New behavior tests

## Versioning Implications

## Targeting

- [ ] This PR is only relevant to main


smarshall-spitzbart:
TODO: setting the opt out parameter to `1` is technically allowed, but would cause the binary to panic with `"unreachable"`. We also need to make this work with e2e tests

x/ccv/consumer/keeper/validators.go
```diff
@@ -100,13 +102,74 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
...
+		powerSum = powerSum.Add(sdk.NewDecFromInt(sdk.NewInt(val.Power)))
+		// if powerSum / totalPower > SoftOptOutThreshold
+		if powerSum.Quo(totalPower).GT(sdk.MustNewDecFromStr(k.GetSoftOptOutThreshold(ctx))) {
+			// set largestSoftOptOutValidatorPower to the power of this validator
+			k.SetLargestSoftOptOutValidatorPower(ctx, uint64(val.Power))
+			k.Logger(ctx).Info("largest soft opt out validator power updated", "power", val.Power)
+			return
+		}
+	}
+	// This will be hit if the SoftOptOutThreshold param is equal to 1
```
mpoke, line 140: Why is this even allowed? The `SoftOptOutThreshold` param must be validated.

x/ccv/consumer/keeper/validators.go
```diff
@@ -100,13 +102,74 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
 	return stakingtypes.Validator{}
 }
 
+func (k Keeper) SetLargestSoftOptOutValidatorPower(ctx sdk.Context, power uint64) {
```
mpoke, line 105: The naming is very confusing: `SetLargestSoftOptOutValidatorPower` sets a uint64 under the key `SoftOptOutThresholdPowerKey`, while `GetSoftOptOutThreshold` returns a string param (representing a fraction). There is also `UpdateLargestSoftOptOutValidatorPower`, but then the getter is called `GetSoftOptOutThresholdPower`.

x/ccv/consumer/keeper/validators.go
```diff
@@ -100,13 +102,74 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
...
 // Slash queues a slashing request for the the provider chain
 // All queued slashing requests will be cleared in EndBlock
 func (k Keeper) Slash(ctx sdk.Context, addr sdk.ConsAddress, infractionHeight, power int64, _ sdk.Dec, infraction stakingtypes.InfractionType) {
 	if infraction == stakingtypes.InfractionEmpty {
 		return
 	}
 
+	// if this is a downtime infraction and the validator is allowed to
+	// soft opt out, do not queue a slash packet
+	if infraction == stakingtypes.Downtime {
```
mpoke, line 163: The `infractionHeight` for downtime misbehavior is the current height minus two, see https://github.com/cosmos/cosmos-sdk/blob/b05b6fe651514c11af3d4160f7c75fbaad92d5db/x/slashing/keeper/infractions.go#L89. This means that `SoftOptOutThresholdPower` should contain info from two height ago.

x/ccv/consumer/keeper/validators.go
mpoke, line 1: Could we add all these changes to a new file, e.g., `soft_opt_out.go`?

mpoke (changes requested):

testutil/keeper/unit_test_helpers.go
```diff
@@ -195,6 +195,18 @@ func GenPubKey() (crypto.PubKey, error) {
 	return cryptocodec.ToTmPubKeyInterface(privKey.PrivKey.PubKey())
 }
 
+// Generates a public key for unit tests (abiding by tricky interface implementations from tm/sdk) or panics
+func MustGenPubKey() crypto.PubKey {
```
smarshall-spitzbart, line 199: Can be replaced by other existing key utils

x/ccv/provider/keeper/proposal.go
```diff
@@ -309,6 +309,7 @@ func (k Keeper) MakeConsumerGenesis(
 		prop.ConsumerRedistributionFraction,
 		prop.HistoricalEntries,
 		prop.UnbondingPeriod,
+		"0.05",
```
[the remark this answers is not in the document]
smarshall-spitzbart: Long term yes we should have this param come from the consumer addition prop. I believe Jehan left that out in an effort to ship this feature quickly

x/ccv/consumer/keeper/validators.go
```diff
@@ -100,13 +102,74 @@ func (k Keeper) ValidatorByConsAddr(sdk.Context, sdk.ConsAddress) stakingtypes.V
...
+
+// UpdateLargestSoftOptOutValidatorPower sets the largest validator power that is allowed to soft opt out
+// This is the largest validator power such that the sum of the power of all validators with a lower or equal power
+// is less than 5% of the total power of all validators
+func (k Keeper) UpdateLargestSoftOptOutValidatorPower(ctx sdk.Context) {
+	// get all validators
+	valset := k.GetAllCCValidator(ctx)
+
+	// sort validators by power ascending
+	sort.Slice(valset, func(i, j int) bool {
```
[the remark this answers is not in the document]
smarshall-spitzbart: https://github.com/cosmos/interchain-security/commit/35068584d897e40bb6817430b507a6f09c7051ab

testutil/keeper/unit_test_helpers.go
smarshall-spitzbart: https://github.com/cosmos/interchain-security/commit/421bee2ff69b17d6b39a883ddc49dfc4fab5bdae

smarshall-spitzbart:
> Issue: As designed `UpdateSoftOptOutThresholdPower` is executed in beginblock. This method uses state that is updated in `ApplyCCValidatorChanges`, which runs on endblock
> 
> https://github.com/cosmos/interchain-security/blob/edac41ab4a085b0f77d33ac08bb0d2d71f139473/x/ccv/consumer/module.go#L202
> 
> Some tests currently panic from this ordering, since on the first block for a chain, `UpdateSoftOptOutThresholdPower` will not have the state it needs. We can solve this issue by moving both methods to beginblock, or both to endblock. If endblock, slashing would not work during the first block of a chain starting.

Update, we cannot add these methods to beginblock because we need to return updates to tm in endblock

x/ccv/consumer/keeper/validators.go
smarshall-spitzbart: It's a panic so not allowed, I've updated the comment https://github.com/cosmos/interchain-security/commit/df7ddaf2335ffe1e18c5d2018c8d2f13d8815212

x/ccv/consumer/keeper/validators.go
smarshall-spitzbart: Updated here https://github.com/cosmos/interchain-security/commit/059c75086c6a1ed611ea4c855ad6b2f02d3494e3

x/ccv/consumer/keeper/validators.go
smarshall-spitzbart: done in https://github.com/cosmos/interchain-security/commit/f86264268cc13816d06f574a5f88cf094b589dc9

x/ccv/consumer/keeper/soft_opt_out.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// UpdateSmallestNonOptOutPower updates the smallest validator power that cannot soft opt out.
+// This is the smallest validator power such that the sum of the power of all validators with a lower power
+// is less than [SoftOptOutThreshold] of the total power of all validators.
+func (k Keeper) UpdateSmallestNonOptOutPower(ctx sdk.Context) {
+	// get all validators
+	valset := k.GetAllCCValidator(ctx)
+
+	// sort validators by power ascending
```
smarshall-spitzbart, line 24: if valset empty we return. This only happens in tests, add warning log

x/ccv/consumer/types/params.go
```diff
@@ -161,3 +174,21 @@ func validateProviderFeePoolAddrStr(i interface{}) error {
...
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+	dec, err := sdktypes.NewDecFromStr(str)
+	if err != nil {
+		return err
+	}
+	if dec.IsNegative() {
+		return fmt.Errorf("param cannot be negative, got %s", str)
+	}
+	if dec.Sub(sdktypes.MustNewDecFromStr("0.33")).IsPositive() {
```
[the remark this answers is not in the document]
smarshall-spitzbart: https://github.com/cosmos/interchain-security/commit/171a573dde221272617bd7eb1bf44673ad2dc167

x/ccv/consumer/keeper/soft_opt_out.go
smarshall-spitzbart: https://github.com/cosmos/interchain-security/pull/833/commits/741da8a570eb6c0e7ee34f61280cd5dd4ce37758

x/ccv/consumer/types/params_test.go
```diff
@@ -19,47 +19,59 @@ func TestValidateParams(t *testing.T) {
...
 			"custom invalid params, negative unbonding period",
-			consumertypes.NewParams(true, 5, "", "", 5, 1005, "0.5", 1000, -24*21*time.Hour), false,
+			consumertypes.NewParams(true, 5, "", "", 5, 1005, "0.5", 1000, -24*21*time.Hour, "0.05"), false,
+		},
+		{
+			"custom invalid params, soft opt out threshold is negative",
+			consumertypes.NewParams(true, 5, "", "", 5, 1005, "0.5", 1000, 24*21*time.Hour, "-0.05"), false,
+		},
+		{
+			"custom invalid params, soft opt out threshold is over 0.33",
```
mpoke, line 69: nit
```suggestion
			"custom invalid params, soft opt out threshold is over 0.2",
```

x/ccv/consumer/types/params.go
```diff
@@ -161,3 +174,21 @@ func validateProviderFeePoolAddrStr(i interface{}) error {
...
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+	dec, err := sdktypes.NewDecFromStr(str)
+	if err != nil {
+		return err
+	}
+	if dec.IsNegative() {
+		return fmt.Errorf("soft opt out threshold cannot be negative, got %s", str)
+	}
+	if dec.Sub(sdktypes.MustNewDecFromStr("0.2")).IsPositive() {
```
mpoke, line 190: This doesn't allow `0.2`. 
```suggestion
	if !dec.Sub(sdktypes.MustNewDecFromStr("0.2")).IsNegative() {
```

mpoke (approved):

sainoe (approved):

x/ccv/consumer/keeper/validators.go
smarshall-spitzbart: @mpoke even with this order of begin blockers, wouldn't the soft opt out logic still be off by 1 block? (not 2)

mpoke: I don't think so. Consider the following example.

- At height 9, in ccv.EndBlock the consumer updates the valset in the CCV module using the powers received from the provider at height 9, see https://github.com/cosmos/interchain-security/blob/afcc4f558444e377d729a46cd4eaf2c278da5573/x/ccv/consumer/module.go#L202
- At height 10, in ccv.BeginBlock the consumer updates smallest validator power that cannot opt out using the valset in the CCV module that was updated at height 9, see https://github.com/cosmos/interchain-security/blob/afcc4f558444e377d729a46cd4eaf2c278da5573/x/ccv/consumer/module.go#L153
- At height 11, in slashing.BeginBlock the consumer calls ccv.Slash for a downtime infraction, which means that the infraction height is 9 (see https://github.com/cosmos/cosmos-sdk/blob/c0fe4f7da17b7ec17d9bea6fcb57b4644f044b7a/x/slashing/keeper/infractions.go#L85). Then, `ccv.Slash` compares the validator's power at height 9 with the smallest power that was updated at height 10 using powers from height 9.

smarshall-spitzbart: Ah makes sense, neat how that works out 👍

PR #844, opened 2023-04-12
fix: code and documentation mismatch

kogisin:
# Description

As I am going through the `ccv` module, I discover that there is a code-documentation mismatch. The [reward-distribution.md](https://github.com/cosmos/interchain-security/blob/main/docs/docs/features/reward-distribution.md#consumer_redistribution_fraction) provides an example of `consumer_redistribution_fraction` usage. 

> With its value set to 0.75, the consumer chain would send 75% of its block rewards and accumulated fees to the provider chain. 

However, when I read the [DistributeRewardsInternally](https://github.com/cosmos/interchain-security/blob/main/x/ccv/consumer/keeper/distribution.go#L53) function, it actually sends 75% of its block rewards and accumulated fees to consumer chain and the remaining 25% to the provider chain. Which one is right?

I fixed the document as I believe the document is incorrect.

## Type of change

## Targeting


mpoke (approved):
LGTM. Nice catch @kogisin. Thanks for pointing it out.

smarshall-spitzbart (approved):
Thank you 👍

PR #846, opened 2023-04-12
Correctly handle VSC packet with duplicate val updates on consumer 

smarshall-spitzbart:
# Description

Fixes a bug seen in the Neutron testnet where tendermint was sent duplicate val updates for the same validator, with different powers. This is a consumer side fix, and a corresponding provider side fix (involving key assignment), should be made eventually. 

## Linked issues

Closes: `#<issue>`

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

Existing tests for `OnRecvVSCPacket` like `TestOnRecvVSCPacket`

## New behavior tests

`TestOnRecvVSCPacketDuplicateUpdates`, which fails without the fix from `relay.go`

## Versioning Implications

## Targeting

- [x] This PR is relevant to main, and should also be back-ported to the v1.2.x branch!


mpoke (approved):
LGTM

x/ccv/consumer/keeper/relay.go
```diff
@@ -51,7 +51,8 @@ func (k Keeper) OnRecvVSCPacket(ctx sdk.Context, packet channeltypes.Packet, new
 	var pendingChanges []abci.ValidatorUpdate
 	currentChanges, exists := k.GetPendingChanges(ctx)
 	if !exists {
-		pendingChanges = newChanges.ValidatorUpdates
+		// Still accumulate changes if no pending changes exist. New changes must be deduplicated
+		pendingChanges = utils.AccumulateChanges([]abci.ValidatorUpdate{}, newChanges.ValidatorUpdates)
 	} else {
 		pendingChanges = utils.AccumulateChanges(currentChanges.ValidatorUpdates, newChanges.ValidatorUpdates)
 	}
```
glnro, line 58: What about:

```go
currentValUpdates := []abci.ValidatorUpdate{}
currentChanges, exists := k.GetPendingChanges(ctx)
if exists {
    currentValUpdates = currentChanges.ValidatorUpdates
}
pendingChanges = utils.AccumulateChanges(currentValUpdates, newChanges.ValidatorUpdates)
```

sainoe (approved):

PR #862, opened 2023-04-19
bump: ibc 4.3

smarshall-spitzbart:
# Description

https://github.com/cosmos/interchain-security/pull/823 upgraded us to IBC 4.3, but a final commit in https://github.com/cosmos/interchain-security/pull/805, downgraded us back to IBC 4.2. See https://github.com/cosmos/interchain-security/commit/31481fa7ed5f24a1b54decc3549916396c740c8f. 

@mpoke has added an additional repo restriction that we need an approver of the most recently pushed commit for any PR. Hopefully this new rule will prevent regressions like this.

## Linked issues

Closes: `#<issue>`

## Type of change

- [ ] `Testing`: Adds testing
- [x] `bump`

## Regression tests

Existing tests

## New behavior tests

No new app behavior from our repo

## Versioning Implications

## Targeting

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/confio/ics23/go v0.9.0
 	github.com/cosmos/cosmos-sdk v0.45.15
-	github.com/cosmos/ibc-go/v4 v4.2.0
+	github.com/cosmos/ibc-go/v4 v4.3.0
 	github.com/gogo/protobuf v1.3.3
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.3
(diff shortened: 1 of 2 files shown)


MSalopek (approved):

mpoke (approved):

faddat:
This is because it was explicitly requested that they be separated

mpoke:
> This is because it was explicitly requested that they be separated

@faddat **_Please avoid having a single PR that does multiple things._** For example, initially, https://github.com/cosmos/interchain-security/pull/805 was bumping both IBC to 4.3.0 and SDK to v0.45.15-ics. As mentioned in https://github.com/cosmos/interchain-security/pull/805#discussion_r1155699128, the bump to ibc-go 4.3.0 is state-machine breaning. Had such a PR been accepted, it would have been very difficult to have a non-breaking release with SDK 0.45.15-ics. There is no reason for these two changes to be part of the same PR.

smarshall-spitzbart:
> This is because it was explicitly requested that they be separated

I don't think the topic of targeted PRs is relevant here. IBC was bumped in main, then unintentionally downgraded, this PR fixes the issue

PR #865, opened 2023-04-20
update proto build system

faddat:
# Description

Update the build system for .proto files:

* use an image compatible with go 1.20
* ~rename the import path for proofs.proto to be compatible with that image~
* rebuild protobufs dependencies

I'd also like to add that I think that this PR is of hidden importance.  I think that we very much need to make sure that it works.  Right now we build ics with protos that don't match the versions of the software that we are using -- **in the present day**.

When we update to those newer .proto files, we find that the build fails.  I'll open a bug issue. 

## Type of change

- [ ] `Testing`: Adds testing

## Regression tests

## New behavior tests

## Versioning Implications

## Targeting

- [x] This PR is relevant to main, and should also be back-ported to v1.2.x (ex: v1.0.0 and v1.1.0)


jtremback:
Thanks! Are the tests supposed to be failing now? Seeing a message

```
--- FAIL: TestCoreSuite/TestAssumptionsSetup (0.00s)
        suite.go:87: test panicked: concrete type *types.ConsumerAdditionProposal has already been registered under typeURL /, cannot register *types.ConsumerRemovalProposal under same typeURL. This usually means that there are conflicting modules registering different concrete types for a same interface implementation
```

faddat:
hey I need to mark this as draft, and I don't know why it fails.

faddat:
much better

https://github.com/cosmos/gaia/pull/2432

faddat:
seems dangerous, sadly

PR #867, opened 2023-04-20
refactor: Run unit tests separately from e2e tests

faddat:
# Description

Closes https://github.com/cosmos/interchain-security/issues/908

* https://github.com/mvdan/github-actions-golang/blob/master/README.md

The venerable @mvdan has contributed many things to the golang ecosystem and it seems he has also made a nice matrix test setup.  Having this in CI makes ensuring compatibility easier and makes test results visible. 

This is a CI only change. 

Alongside the initial changes, I've decoupled the workflow for unit tests and e2e tests so that failures in unit tests do not block e2e tests.  This will make sure that failures are more visible in ci, making developers lives easier. 

## Type of change

- [x] `Testing`: Adds testing

## Regression tests

## New behavior tests

## Versioning Implications

## Targeting


jtremback:
Cool. So is this going to show compatibility of builds on different OSs?

faddat:
yep :) - also osx-windows-linux

jtremback:
Does this have any downsides? Does it slow down testing at all? Is it running the tests a bunch of different times?

faddat:
no

faddat:
I'd like to get this merged, so that these tests are run separately from the automated tests.  Right now the "automated tests" fail if the unit tests fail, and we don't get the feedback from the integration tests.

faddat:
This is a week old now, @mpoke are you invoking policy sir?

Screenshot 2023-04-29 at 7 14 25 PM

.github/workflows/go-tests.yml
mpoke, line 1: Rename file to `tests.yml`

.github/workflows/go-tests.yml
```diff
@@ -0,0 +1,15 @@
+on: [push, pull_request]
+name: Test
```
mpoke, line 2:
```suggestion
name: Tests
on: [push, pull_request] 
```

.github/workflows/go-tests.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.19.x, 1.20.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
mpoke, line 8: Why is it necessary to check the tests on different go versions and different OSs? [CometBFT](https://github.com/cometbft/cometbft/blob/main/.github/workflows/tests.yml), [SDK](https://github.com/cosmos/cosmos-sdk/blob/main/.github/workflows/test.yml#L113), and [IBC](https://github.com/cosmos/ibc-go/blob/main/.github/workflows/test.yml) are all checking only for ubuntu-latest and the go version in go.mod.

.github/workflows/automated-tests.yml
```diff
@@ -23,8 +23,6 @@ jobs:
         with:
           go-version: "1.19" # The Go version to download (if necessary) and use.
 
-      - name: Unit, integration and difference tests
-        run: go test ./...
-
       - name: E2E tests
-        run: make test-e2e-short
+        run: go run ./tests/e2e/... --happy-path-only
```
mpoke, line 27:
```suggestion
        run: make test-e2e-short
```

.github/workflows/automated-tests.yml
mpoke, line 1: Rename file to `e2e-tests.yml`

mpoke (changes requested):
Thanks @faddat for your contribution. See my comments below.

.github/workflows/automated-tests.yml
faddat: oh -- actually I'd meant to add --verbose

.github/workflows/go-tests.yml
faddat: because people may want to run chains on windows or mac

.github/workflows/automated-tests.yml
mpoke: Please use `make test-e2e-short`. Developers can always run the tests locally with `--verbose`.

mmulji-ic:
> Does this have any downsides? Does it slow down testing at all? Is it running the tests a bunch of different times?

We have a limited number of workers, Marko was complaining about this earlier, so we need to rationalise how many tests we run & how often. Ditto with Gaia, coz their already quite long, adding matrix testing will eat into our budget even more. Just FYI, we need to find a sweet spot - perhaps an exercise for later.

.github/workflows/automated-tests.yml
faddat: But we can't prove outcomes using CI, so that there's no debate, in the way that you recommend.

CI is a neutral and clearly defined build environment that can be used as proof of issues.

faddat:
@mmulji-ic so I've got to agree with you on this point. And the options are short of rough. I had thought about moving over to a gitlab that we sell host but this has other significant disadvantages.

I'm curious to learn about what we spend on GitHub, and how to reduce that spend.

At notional, we use only GitHub free features with the exception of copilot. 

When we accidentally enabled some kind of a corporate premium mode, we got a bill for 13500, I nearly had a heart attack.

.github/workflows/automated-tests.yml
faddat: We want proof.

tac0turtle:
@faddat id recommend implementing the approach in the sdk here. This way we can be consistent with all repos. Secondly im not sure the need to test on different OS a single os is fine as we target linux machine for production instead of things like windows and mac. I think after adopting the sdk approach this would be good to go 😄

faddat:
this log helped me to make a major step forward with 47ification:

https://github.com/cosmos/interchain-security/actions/runs/4903809348/jobs/8756471785?pr=869

the way things are displayed matters, and helps

human time is much more expensive than machine time

faddat:
I just want to say that I am comfortable asserting that this is the correct way for CI to work in this repository. You can close the pull request if you'd like to, but I believe that this is good.

smarshall-spitzbart:
The idea here is solid but we're going to take a different direction

Issue #731, opened 2023-02-10
Panics on failure to send IBC packets


andrey-kuprianov:
_Surfaced from @informalsystems audit on [Interchain Security](https://github.com/cosmos/interchain-security/), at commit 463ec20c_

# Problem

Currently, both provider and consumer chains would panic and halt if sending of IBC packets fails for the reason different than `ErrClientNotActive`. This problem is known to the team (see https://github.com/cosmos/interchain-security/issues/649), and no correcting actions are needed before the ICS 1.0 release, but the problem needs to be alleviated before ICS 1.1 release (before the first consumer chain goes live).

# Closing criteria

Do not panic on failure to send IBC packets.

# Problem details

In the implementation under audit, namely in the functions [SendVSCPacketsToChain()](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/provider/keeper/relay.go#L208) on provider, and [SendPackets()](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/consumer/keeper/relay.go#L209) on consumer, the code panics if sending of IBC packets fails. As these functions are called from EndBlockers, the respective chain will halt on error. This has indeed happened in a testnet, when sending of an IBC packet failed because of an expired LightClient (see issue [ICS#435](https://github.com/cosmos/interchain-security/issues/435)). A partial fix has been introduced, in which the packets are first queued, and only then sent, with errors about light client expiration are logged, and the unsent packets remain queued. The resulting code looks similar on both provider and consumer; here is the [relevant part of the provider code](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/provider/keeper/relay.go#L187-L209):

```go
		// send packet over IBC
		err := utils.SendIBCPacket(...)

		if err != nil {
			if clienttypes.ErrClientNotActive.Is(err) {
				// IBC client is expired!
				// leave the packet data stored to be sent once the client is upgraded
				// the client cannot expire during iteration (in the middle of a block)
				k.Logger(ctx).Debug("IBC client is expired, cannot send VSC, leaving packet data stored:", "chainID", chainID, "vscid", data.ValsetUpdateId)
				return
			}
			// TODO do not panic if the send fails
			// https://github.com/cosmos/interchain-security/issues/649
[...]

PR #876, opened 2023-04-24
fix!: Remove panics on failure to send IBC packets

mpoke:
## Description

Closes: https://github.com/cosmos/interchain-security/issues/731

On both the provider and consumer, if the sent fails, the packet is kept in the queue and a re-sent is attempt in the next block. 

- Replace the panic on the provider with `StopConsumerChain`. Note that most likely if the send if failing here, the consumer is malicious and it should be removed. 
- Replace the panic on the consumer with an error message. If packets are not eventually sent and relayed, the vscTimeout on the provider will be triggered and the consumer will be removed. 

---

### Author Checklist

I have...

  * [x] @smarshall-spitzbart could you please add the necessary tests (it seems that no tests were broken)
  * [x] @MSalopek Do we have docs on this behavior or it's too involved for the docs?

### Reviewers Checklist

I have...


x/ccv/consumer/keeper/relay.go
```diff
@@ -196,13 +196,17 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 		)
 		if err != nil {
 			if clienttypes.ErrClientNotActive.Is(err) {
-				k.Logger(ctx).Debug("IBC client is inactive, packet remains in queue", "type", p.Type.String())
+				// IBC client is expired!
 				// leave the packet data stored to be sent once the client is upgraded
+				k.Logger(ctx).Debug("IBC client is expired, cannot send IBC packet; leaving packet data stored:", "type", p.Type.String())
```
MSalopek, line 201: Can we promote this log line to `Info`?

It seems useful to know about inactive clients without turning on debug mode.

x/ccv/consumer/keeper/relay.go
```diff
@@ -196,13 +196,17 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
...
 				return
 			}
-			// something went wrong when sending the packet
-			// TODO do not panic if the send fails
-			panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
+			// Not able to send packet over IBC!
+			// Leave the packet data stored for the sent to be retried in the next block.
+			// Note that if VSCMaturedPackets are not sent for long enough, the provider
```
MSalopek, line 206: Do the stale packets get removed from queue if the consumer client expires?

x/ccv/consumer/keeper/relay.go
mpoke: It may be a good idea. It was like this before though that's why I decided not to change it.

x/ccv/consumer/keeper/relay.go
mpoke: What are stale packets?

MSalopek: I was talking about this step but did not communicate it clearly, my appologies.

My concern was about removing unsent packets intended for consumers with expired clients.
After further inspections, it can be seen that any pending/stale packets are demoved during `StopConsumer` subprotocol:
- https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/proposal.go#L193

x/ccv/provider/keeper/relay.go
```diff
@@ -194,12 +194,17 @@ func (k Keeper) SendVSCPacketsToChain(ctx sdk.Context, chainID, channelID string
...
-				k.Logger(ctx).Debug("IBC client is expired, cannot send VSC, leaving packet data stored:", "chainID", chainID, "vscid", data.ValsetUpdateId)
+				k.Logger(ctx).Info("IBC client is expired, cannot send VSC, leaving packet data stored:", "chainID", chainID, "vscid", data.ValsetUpdateId)
 				return
 			}
-			// TODO do not panic if the send fails
-			// https://github.com/cosmos/interchain-security/issues/649
-			panic(fmt.Errorf("packet could not be sent over IBC: %w", err))
+			// Not able to send packet over IBC!
+			k.Logger(ctx).Error("cannot send VSC, removing consumer:", "chainID", chainID, "vscid", data.ValsetUpdateId, "err", err.Error())
+			// If this happens, most likely the consumer is malicious; remove it
```
smarshall-spitzbart, line 202: @mpoke it's not intuitive to me why a failed IBC send would always indicate a malicious consumer. The error being returned could be caused by corrupted IBC state on the provider, where stoping a consumer could make such a scenario worse

x/ccv/consumer/keeper/relay.go
```diff
@@ -196,13 +196,17 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 		)
 		if err != nil {
 			if clienttypes.ErrClientNotActive.Is(err) {
```
smarshall-spitzbart, line 198: The check for `if clienttypes.ErrClientNotActive.Is(err)` is no longer needed since behavior is the same whether that statement is true or not. The type of error will be printed

smarshall-spitzbart:
@mpoke I've added a simple test for the consumer. Re provider i'll wait on writing the test until we chat, since if we remove the `StopConsumerChain` call, the behavior would be a lot simpler to test

x/ccv/provider/keeper/relay.go
mpoke: Here's my analysis on the reasons for `ccv.SendIBCPacket` to return error on the provider side:

- Channel not found – not possible as SendIBCPacket is called only for consumer with an established CCV channel and channels are not removed from the IBC module
- Module does not own channel capability – not possible as the capability was already claimed in OnChanOpenTry, before the CCV channel was marked as established
- The next sequence send is not found – if that happens, something is wrong with the IBC module store  
- The packet sent failed basic validation – if that happens, something is wrong with the code on the provider chain as the packet is constructed in SendIBCPacket
- The channel’s state is CLOSED – an opened channel can be closed only by one of the following functions
  - channelKeeper.ChanCloseInit, which is called on the provider only when the consumer chain is removed (note that OnChanCloseInit returns an error, thus user-initiated channel closing are not allowed)
  - channelKeeper.ChanCloseConfirm, which entails that the consumer endpoint of the channel is CLOSED; this can happen only if the consumer receives a negative ack for a packet sent to the provider
    - Attack: A malicious consumer could send an invalid SlashPacket that fails ValidateSlashPacket, this would lead to a nack sent to the consumer, which results in channelKeeper.ChanCloseInit being called on the consumer. Finally, this would enable a relayer to send a MsgChannelCloseConfirm to the provider, which would move the channel’s state to CLOSED without updating the ChainToChannel state. 
  - channelKeeper.TimeoutExecuted, which means that OnTimeoutPacket is first executed on the provider, which results in the consumer chain being removed (i.e., the ChainToChannel state is being updated).
- The packet’s destination channelD and portID do not match the ones of the counterparty endpoint – not possible due to construction of the packet
- There is no underlying connection – not possible as it was checked during the handshake
- There is no underlying client state – not possible as it was checked during the handshake
- The packet is already timed out on the consumer – not possible as the packet timeout is set in SendIBCPacket to the current block timestamp plus the timeout period and the latest known timestamp on the consumer cannot be larger than this packet timeout.

mpoke: To be more conservative, we could just log an error and leave the packets in the queue. Operators need to notice that `VSCPackets` are no longer being relayed, otherwise the unbondings in those packets will never complete. Note that the VSC timeout is set when the packet is sent, and not when it's queued. 

So we run the risk of nobody noticing for a while, which will affect users. Once we notice, we need to either do an emergency upgrade where we migrate the state (difficult) or we submit a ConsumerRemoval proposal, which would take 2 extra weeks before delegators that unbonded can get their funds back.

smarshall-spitzbart:
TODO changelog

smarshall-spitzbart:
@mpoke tests have been added, lmk if they seem correct

mpoke (commented):
LGTM

smarshall-spitzbart (approved):
Approval

x/ccv/provider/keeper/relay_test.go
```diff
@@ -676,3 +676,40 @@ func TestHandleVSCMaturedPacket(t *testing.T) {
...
+	// Append mocks for full consumer setup
+	mockCalls := testkeeper.GetMocksForSetConsumerChain(ctx, &mocks, "consumerChainID")
+
+	// Set 3 pending vsc packets
+	providerKeeper.AppendPendingVSCPackets(ctx, "consumerChainID", []ccv.ValidatorSetChangePacketData{{}, {}, {}}...)
+
+	// append mocks for the channel keeper to return an error
+	mockCalls = append(mockCalls,
+		mocks.MockChannelKeeper.EXPECT().GetChannel(ctx, ccv.ProviderPortID,
+			"CCVChannelID").Return(channeltypes.Channel{}, false).Times(1),
```
p-offtermatt, line 696: .Times(1) seems unnecessary
```suggestion
			"CCVChannelID").Return(channeltypes.Channel{}, false),
```

x/ccv/consumer/keeper/relay_test.go
```diff
@@ -280,3 +280,27 @@ func TestOnAcknowledgementPacket(t *testing.T) {
...
+
+	// Set some pending packets
+	consumerKeeper.SetPendingPackets(ctx, types.ConsumerPacketDataList{List: []types.ConsumerPacketData{
+		{}, {}, {},
+	}})
+
+	// Mock the channel keeper to return an error
+	gomock.InOrder(
+		mocks.MockChannelKeeper.EXPECT().GetChannel(ctx, types.ConsumerPortID,
+			"consumerCCVChannelID").Return(channeltypes.Channel{}, false).Times(1),
```
p-offtermatt, line 300: Times(1) seems unnecessary
```suggestion
			"consumerCCVChannelID").Return(channeltypes.Channel{}, false),
```

MSalopek (approved):
Approval.

So the behaviour is this:
- if the consumer **cannot send**, it will **retain** packets
- if the provider **cannot send VSC** packets it will:
    a) retain unsent packets for consumers with inactive clients
    b) delete all packets for consumers if the error is caused by anything other than an inactive client

**Re: docs**
This is very nuanced, it should probably go into a new section about relaying behaviour

p-offtermatt (approved):
LGTM, only very minor nitpicks. I looked most closely at the tests and I don't have an opinion on stopping consumer vs logging and ignoring

x/ccv/provider/keeper/relay_test.go
smarshall-spitzbart: Hmm I don't see the reason to remove the `Times(1)` since we indeed only want this method called once. Better to be explicit with the mocking package imo

Issue #896, opened 2023-04-27
feature: add a way to start a local testnet


p-offtermatt:
# Problem

Our testing infrastructure is relatively heavy-weight and hard to use outside the dedicated environment.
For example, we have scripts for starting test chains, but this is happening in docker containers, and is not easy to call by hand.
It would be beneficial to have a way to start a local testnet, to manually test changes quickly.
This is also particularly important with e.g. Hermes, which may be a bit of work to add into the existing docker setup.

I would expect any solution to allow starting a testnet with >1 validator and provider+consumer.
This base case should be set up with a single or at most a couple of commands, and this should be achievable by someone completely new to the repo.

The solution should be documented prominently in the README, since it is one of the most basic ways someone would want to interact with the repo.

# Closing criteria

There is an easy way to start a local testnet with a provider and consumer, with multiple validators for each.

p-offtermatt:
I made a PR that solves this issue with a bash script that is relatively hacky. The issue could alternatively be solved using existing tools, e.g. https://github.com/informalsystems/gm/ or https://github.com/hyphacoop/cosmos-ansible/tree/main/examples#start-a-local-replicated-security-testnet

p-offtermatt:
In #901 the point was made that instead of using a script and have a local testnet, we should reuse code from the e2e tests as much as possible, so this might be a better direction for this issue.

mpoke:
@p-offtermatt is this issue still relevant?

p-offtermatt:
This will be closed as part of #1334, as I will explain in the guidelines also how to use the e2e framework to have a local testnet to work with. I believe that should mean this stays open for now, let me adjust it.

PR #901, opened 2023-04-27
chore: add a way to start a local testnet

p-offtermatt:
## Description

Closes: #896 

The PR adds a file to start a simple, local testnet
with 3 validators, a provider chain, and a consumer chain.

This serves to help new or existing developers to manually test changes
and get quick feedback, to play with a small chain, and so on.

The script is far from perfect, but it's working as-is.
This is not production code, and merely supposed to help with some manual tests, so I think it brings value in its current state.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,35 @@ export PATH=$PATH:$(go env GOPATH)/bin
 
 Inspect the [Makefile](./Makefile) if curious.
 
+**Running a local testnet**
+    
+```bash
+# start a local testnet
+./scripts/local-testnet.sh
+```
+You can see the ports and home directories of the validators by checking the commands that started
+the validator processes, e.g. by checking processes with htop:
+
+```bash
+# check processes
+htop --filter interchain-security
+```
+
+The local testnet script has slightly non-deterministic behaviour due to non-determinism in
+network communication, so if you run into issues during the script run, try running the script again.
+
+To check that both chains are producing blocks, you can run:
+
+```bash
+# check that both chains are producing blocks
+interchain-security-cd query block --node http://127.0.0.1:29175
+interchain-security-pd query block --node http://127.0.0.1:29170
+```
+
+The argument to `--node` should be the `--rpc.laddr` of a node on the provider or consumer.
+You can search for `last_commit.height` in the output json to see the latest block height,
+and check that the height is increasing.
+
 ## Testing
 
 See [testing docs](./docs/testing.md).
--- a/scripts/local-testnet.sh
+++ b/scripts/local-testnet.sh
@@ -1,0 +1,512 @@
+#!/bin/bash
+set -eux 
+
+# User balance of stake tokens 
+USER_COINS="100000000000stake"
+# Amount of stake tokens staked
+STAKE="100000000stake"
+# Node IP address
+NODE_IP="127.0.0.1"
+
+# Home directory
+HOME_DIR=$HOME
+
+# Validator moniker
+MONIKERS=("coordinator" "alice" "bob")
+LEAD_VALIDATOR_MONIKER="coordinator"
+
+PROV_NODES_ROOT_DIR=${HOME_DIR}/nodes/provider
+CONS_NODES_ROOT_DIR=${HOME_DIR}/nodes/consumer
+
+# Base port. Ports assigned after these ports sequentially by nodes.
+RPC_LADDR_BASEPORT=29170
+P2P_LADDR_BASEPORT=29180
+GRPC_LADDR_BASEPORT=29190
+NODE_ADDRESS_BASEPORT=29200
+PPROF_LADDR_BASEPORT=29210
+CLIENT_BASEPORT=29220
+
+
+# Clean start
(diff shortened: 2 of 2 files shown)


smarshall-spitzbart (commented):
In general I'm not opposed to these changes and they look solid! Few questions:

- Is this script intended to be run in a docker container and/or should it be dockerized?
- How is this setup different from the e2e test setup? 
- Is it possible to reuse existing .sh code from `tests/e2e/testnet-scripts` to get the testnet up and running in a container, and have a way for a user to bring up this testnet without running any e2e "steps" against the two chains? Then provide instructions for how the user can shell into said container and execute txs as they please

p-offtermatt:
Thanks for the comments Shawn! Doing my best to address them below.

> In general I'm not opposed to these changes and they look solid! Few questions:
> 
>     * Is this script intended to be run in a docker container and/or should it be dockerized?

No, it wasn't my intention. Ideally I would like something that can be run without Docker here, to minimize dependencies and not have the added complexity of Docker. I can also see an argument for using Docker specifically to keep dependencies inside the Docker image and away from the host, however.

>     * How is this setup different from the e2e test setup?

The e2e test setup achieves a similar goal. Differences are: 1) e2e runs completely in Docker. 2) e2e relies on Go code and the e2e framework, since e.g. starting chains is a step in the e2e framework. To reuse anything major from the e2e framework, we'd have to modify Go code and add a new "manual-test" setup in the e2e framework.

>     * Is it possible to reuse existing .sh code from `tests/e2e/testnet-scripts` to get the testnet up and running in a container, and have a way for a user to bring up this testnet without running any e2e "steps" against the two chains? Then provide instructions for how the user can shell into said container and execute txs as they please

A bit redundant from the answer to the last comment: Yes it's possible, but it would probably involve modifying the e2e Go code. Starting a chain and adding a consumer chain are all done as steps in the e2e framework, so from the outside one would need to not interact with the testnet-script, but with the e2e Go entrypoint. It seems like a plausible way as well, it might not be worth the effort vs having this single script (though I am not very certain either way here, having the e2e framework provide this functionality might be desirable and easier to maintain long-term).

smarshall-spitzbart:
@p-offtermatt thanks for the responses! Yea, the main reason why I'm leaning toward using the e2e entrypoint for starting a local testnet is that this script you have here repeats a lot of the functionality that already exists in the `tests/e2e/testnet-scripts` directory. 

I'm sure there's some differences, but my bias is always to minimize the amount of code committed to the repo to achieve a goal. Happy to chat about this PR over a call tho!

jtremback:
I agree with Shawn here. Building your own shell script to start everything from scratch is an awesome exercise, and almost everyone on the team has done it at some point, but this is all stuff that the e2e tests are already doing. In fact, this e2e framework and many other e2e frameworks in Cosmos started with someone making a script like this. For fun, [here is the first commit of the script](https://github.com/althea-net/cosmos-gravity-bridge/commit/9010bbb1bbfc362858a01cd784c4255ff47bb53e) that is a distant ancestor of both our e2e testnet framework, and Strangelove's IBC-test (although they ported to Go).

I actually already use the e2e tests as a janky way to mess around on the command line by inserting a state check that I know will fail on one of steps. Then I go into the Docker and mess around. Seems to me that this sort of accomplishes what is being done here (correct me if I'm wrong on that).

Some of the pain points with my method are:

- Obviously it's sort of janky to get into the environment by making a test fail at a certain point. Maybe we can set up a test run that gets to a point that we want to get to and then hangs. If a test run completes, it tears down the Docker, so it needs to hang to prevent that.
- It's not documented how to get into the Docker etc, I just do it because I know how. It's not really an "official" technique. We could set up a documented command that uses stuff from the e2e framework to set up an environment that has a consumer and provider, and maybe the command could even shell into the docker.
- It's tough to remember how to query and and send txs in the e2e environment because you have to remember the right home folder and node ip, and keychain stuff, etc etc. Maybe we could write a file to the root of the docker using the config of the test run that would have reference commands for querying and sending txs from each node on each chain.

Not saying you should start on that project or anything right now, just my thoughts.

p-offtermatt:
I agree with the points made here.
In particular, I take the point that adding this in the repo adds an extra maintenance burden that we might not want.
There are a couple of ways that I prefer a local script for to a Docker image, e.g. easily including extra tools that are installed locally, without having to reproduce their installation in the Dockerfile.
However, I do agree that we should reuse the existing code as much as possible. This script was written because we needed it to test a specific thing that was hard to test otherwise, so I thought it would be valuable to preserve, but It's probably not worth the maintenance burden.

In hindsight, I agree that the ICS repo should probably not commit to maintaining this script, so I'll close this PR.
If it's ever useful, the script exists as a Gist here https://gist.github.com/p-offtermatt/553aa20cef0ea4ab6bbd95c67857c2c3

Reusing the e2e setup to also allow manual testing could be discussed more in [the related issue ](https://github.com/cosmos/interchain-security/issues/896) if necessary, it's likely a decent idea but seems out of the scope of this PR.

Issue #732, opened 2023-02-13
EPIC: Cryptographic verification of equivocation


jtremback:
# Problem

We would like the provider chain to be able to verify equivocation evidence on its own, instead of trusting consumers, or needing evidence to go through governance. For this, we can look at the code in Tendermint which does this, but it will need heavy changes.

# Problem details

To keep things simple at first, we should design this code to be used by an accuser who would like to cause an equivocating validator to be slashed by submitting a message to the chain containing all necessary evidence. Later, we can adapt this to be used automatically for packets coming from a consumer chain.

Fortunately, almost all the code used by Tendermint to verify equivocation evidence is in one file: https://github.com/cometbft/cometbft/blob/main/evidence/verify.go. A lot of this code should be possible to reuse, however, it depends on state stored by Tendermint nodes, most of which will not be available on the provider chain.

There are two basic types of equivocation: Double signing and light client attacks.

## Double signing

This is trivial, you can see the function that does it [here](https://github.com/cometbft/cometbft/blob/main/evidence/verify.go#L160). The only complicated thing is that it requires the validator set and powers at the height of the evidence. But we should be able to store these on the provider. Also, the evidence will have the height of the consumer which will need to be mapped to the height of the provider to get the validator powers at that height. But I think we already have code to do that.

## Light client attacks

This is much [more complicated](https://github.com/cometbft/cometbft/blob/main/evidence/verify.go#L57), and it requires more state. The additional state that is required is the true header of the block in question, known as the "trustedHeader" or the "commonHeader". In Tendermint, this is loaded from the node's state. On the provider chain, we will not be able to access this. Instead, it should be possible to have the accuser supply the trusted header along with the rest of the evidence. The provider chain can then use the record of validator powers from that block to check that the trusted header is real.

# Task list

```[tasklist]
### Must have
- [ ] https://github.com/cosmos/interchain-security/issues/1124
- [ ] https://github.com/cosmos/interchain-security/issues/825
- [ ] https://github.com/cosmos/interchain-security/issues/827
- [ ] https://github.com/cosmos/interchain-security/issues/1147
- [ ] https://github.com/cosmos/interchain-security/issues/1065
- [ ] https://github.com/cosmos/interchain-security/issues/1224
- [ ] https://github.com/cosmos/interchain-security/issues/1227
- [x] Fix consumer public keys retrieval in double voting evidence handling: #1261 #1254
[...]

PR #964, opened 2023-05-16
docs(adr): ADR-007 pause unbonding period during equivocation proposal

tbruyelle:
## Description

Link: #747 #791 

---

### Author Checklist

I have...

* [ ] Added `!` to the type prefix if API or client breaking change

### Reviewers Checklist

I have...


jtremback (approved):
Thanks!

albttx (approved):

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
```diff
@@ -0,0 +1,89 @@
...
+additional deposits. The hook is triggered however at arrival of a deposit, so
+a check to verify that the proposal is not already in voting period is
+required.
+
+### When unpause
+
+We can use a `gov` module's hook also here and it is
+`AfterProposalVotingPeriodEnded`.
+
+If the hook is triggered with an equivocation proposal, then for each
```
mpoke, line 61: This may lead to unpausing unbonding operations that occurred after the `AfterProposalDeposit` hook was called. As a result, those unbondings would no longer wait for the unbonding period on the consumer to elapse.

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
```diff
@@ -0,0 +1,89 @@
...
+and all consumer chains, it will remain paused.
+
+### When pause
+
+The unbonding period (if there is any unbonding) should be paused once an
+equivocation proposal enters the voting period. For that, the `gov` module's
+hook `AfterProposalDeposit` can be used. 
+
+If the hook is triggered with a an equivocation proposal in voting period, then
+for each equivocation of the proposal, the related unbonding operations of the
```
mpoke, line 48: > the related unbonding operations of the related validator must be paused.

This needs to be described more precisely. For example, unbonding operations that are no longer waiting for this particular consumer chain should probably not be put on hold. Also, if the infraction occurred after the unbonding operation was initiated, there is no reason to pause the unbonding.

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
```diff
@@ -0,0 +1,89 @@
...
+## Consequences
+
+### Positive
+
+- Validators subject to an equivocation proposal cannot finish unbonding
+  their tokens before the end of the voting period.
+
+### Negative
+
+- A malicious consumer chain could forge slash packets enabling submission of
```
mpoke, line 73: IMO, this is the main argument against this proposal. The main reason for pausing unbonding operations is to enable the consumer chains to have shorter unbonding periods and avoid unbonding operations to be delayed. However, reducing the unbonding period on a consumer chain would impact the trusting period of all light clients of that consumer chain (even clients on other chains). This may lead to light clients expiring, which would be a major disruption for the consumer chain.

mpoke: The  alternative to this proposal is to enable the provider to verify the evidence of equivocation on consumer chains, i.e., https://github.com/cosmos/interchain-security/issues/732.

mpoke (commented):
Thanks @tbruyelle. I left some comments below. In general, I'm not sure that this proposal solves the main problem: Maintain the correctness of PoS (and the slashing mechanism), while avoiding delays of unbonding operations (e.g., undelegations) on the Hub.

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+(Copied from [#747](https://github.com/cosmos/interchain-security/issues/747))
+
+Currently, if an equivocation slashing proposal is created after more than one
```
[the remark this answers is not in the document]
tbruyelle: Did you only change the text width ? I generally chose 80 characters for markdown but maybe you prefer something else ?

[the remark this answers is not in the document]
mpoke: Oh sorry. I just wanted to change "voting period is 3 weeks" to "voting period is 2 weeks". I didn't noticed that there are multiple lines.

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
jtremback: I'm not sure I follow. Currently (without this PR, or cryptographic verification), we are in a situation where the trusting period should be `(consumer unbonding - provider voting period) * ~0.75`. With a 20 day consumer unbonding period, that means the trusting period should be about 4-5 days. This is already pretty tough for consumer chains to deal with. If we took the consumer unbonding period down to 14 days like we want to, we'd need a trusting period of 0.

With this PR or cryptographic equivocation, we can go back to the trusting period being `consumer unbonding * 0.75`, or 10-11 days. So at least in this respect, it seems that this PR or cryptographic verification both mitigate this issue equivalently.

mpoke: @jtremback Yeah, that’s indeed confusing. Sorry for that.

There are actually two problems we are trying to address here. First, ensuring the correctness of the slashing mechanism of PoS. Second, avoiding delays in the completion of unbonding operations on the provider (which would affect the user experience). Note that the first is a safety concern, while the second is a liveness concern. 

For the first, we MUST make sure that the stake of a validator that double signed on a consumer chain is still locked (on the provider) after an equivocation proposal passes. Thus, the following condition MUST hold `consumerTrustingPeriod < consumerUnbondingPeriod - providerVotingPeriod`. In addition, we MUST leave time for the double signing to be detected, a SlashPacket to be relayed and an equivocation proposal to be submitted. Let’s denote this extra time period with `delta1`. Then,   `consumerTrustingPeriod < consumerUnbondingPeriod - delta1 - providerVotingPeriod`. 

For the second, the following condition SHOULD hold `consumerUnbondingPeriod < providerUnbondingPeriod`. In addition, we SHOULD  leave time to relay both `VSCPackets` and `VSCMaturedPackets`, and to account for potential downtimes of the consumer chain (as it happened on Neutron’s launch). Let’s denote this extra time period with `delta2`. Then, `consumerUnbondingPeriod + delta2 < providerUnbondingPeriod`. 

From these two inequalities, we get `consumerTrustingPeriod < providerUnbondingPeriod - delta2 - delta1 - providerVotingPeriod`. For the Hub, `providerUnbondingPeriod = 21 days`, `providerVotingPeriod = 14 days`, which means `consumerTrustingPeriod < 7 - delta2 - delta1`. Best case scenario, `delta1 = 1 day` and `delta2 = 1day`, which means  `consumerTrustingPeriod < 5 days`. This trusting period must be used for all light clients of the consumer chain, not only the ones on the provider. This means that the chances of a consumer light client expiring are quite high (which would clearly be a very disruptive event). Also, note that these values for the deltas are very low. For example, the network has ~7 days to detect a light client attack on a Hub client (i.e., delta1 ~ 7 days). Also, as we seen with Neutron, having `delta2=1day` may lead to delays of unbonding operations. 

The only way to deal with this problem, is to get rid of the `providerVotingPeriod`, i.e.,  `consumerTrustingPeriod < providerUnbondingPeriod - delta2 - delta1`.  This ADR is one way of doing that. The downside is that the suggested approach effectively increases the `consumerUnbondingPeriod` (when an equivocation proposal is submitted), which means that a malicious consumer can break liveness (i.e., delay unbondings). Given that such an event would result in that consumer being removed (via a ConsumerRemovalProposal), I think the risks are minimal. So, we could go ahead with this proposal for now until #732 is done.

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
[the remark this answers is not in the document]
tbruyelle: fixed c44afb3b

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
```diff
@@ -0,0 +1,89 @@
...
+## Changelog
+* 2023-05-16: Initial Draft
+
+## Status
+
+Proposed
+
+## Context
+
+(Copied from [#747](https://github.com/cosmos/interchain-security/issues/747))
```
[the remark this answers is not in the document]
tbruyelle: removed here b6f91753

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
tbruyelle: > For example, unbonding operations that are no longer waiting for this particular consumer chain should probably not be put on hold.

Actually, for this particular case, I wonder how this is possible. Considering the protocol, how an unbonding period can end on a consumer chain *before* the provider chain ? 

Or maybe when the unbonding period is deliberately shorter between consumer and provider ?

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
tbruyelle: Fixed dd353c29df93d8e7940098516bce5da9fe0cee64

docs/docs/adrs/adr-007-pause-unbonding-on-eqv-prop.md
tbruyelle: For 

> Also, if the infraction occurred after the unbonding operation was initiated,

Fixed dd353c29df93d8e7940098516bce5da9fe0cee64

mpoke (approved):

smarshall-spitzbart:
@tbruyelle we now got the two approvals 👍 could you update your fork and then we can merge this PR, thanks

tbruyelle:
@smarshall-spitzbart done!

I was wondering, shouldn't we change the ADR status before the merge ? Currently it's draft.

mpoke:
@tbruyelle let's discuss in a separate PR / issue / discussion if we want to go ahead and implement this solution. cc @jtremback

Issue #1001, opened 2023-06-08
Limit VSCMatured packets handled in endblocker


smarshall-spitzbart:
# Problem

This issue encapsulates what we've defined as "audit issue number 3" internally lol. We need to limit the rate that VSCMatured packets are handled in endblocker. 

# Closing criteria

Hardcode some number, ex 100 of vsc matured packets st. once you handle that many packets in a single endblocker, handling execution is stopped, and remaining queued packets will be handled on subsequent blocks. 

Make sure to include a comment that future improvements to throttling will make the short term solution obsolete!

# Problem details

Note this issue will be solved and a part of ICS v2. However, solving #713 will also handle the issue in a future release.

PR #1004, opened 2023-06-09
fix: limit vsc matured packets handled per endblocker

smarshall-spitzbart:
## Description

Closes: https://github.com/cosmos/interchain-security/issues/1001

Note this PR will need to be backported to the v2 release branch

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

* [x] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [x] confirmed this PR does not introduce changes requiring state migrations, OR confirmed migration code has been added to consumer and/or provider modules
* [x] confirmed all author checklist items have been addressed
* [x] reviewed state machine logic
* [x] reviewed tests and test coverage


x/ccv/provider/keeper/throttle.go
```diff
@@ -22,6 +22,13 @@ func (k Keeper) HandleThrottleQueues(ctx sdktypes.Context) {
 		return
 	}
 
+	// Return if vsc matured handle limit was already reached this block, during HandleLeadingVSCMaturedPackets.
+	// It makes no practical difference for throttling logic to execute next block.
+	// By doing this, we assure that all leading vsc matured packets were handled before any slash packets.
+	if k.VSCMaturedHandledLimitReached(ctx) {
```
smarshall-spitzbart, line 28: Not sure if this is strictly neccessary, but it will ensure that we change throttling behavior as little as possible

x/ccv/provider/keeper/throttle.go
```diff
@@ -85,15 +92,24 @@ func (k Keeper) HandlePacketDataForChain(ctx sdktypes.Context, consumerChainID s
...
 	}
-	for _, vscMData := range vscMaturedData {
+	for idx, vscMData := range vscMaturedData {
+		if k.VSCMaturedHandledLimitReached(ctx) {
+			// Break from for-loop, DeleteThrottledPacketData will still be called for this consumer
+			break
+		}
 		vscMaturedPacketHandler(ctx, consumerChainID, vscMData)
+		// Append seq num for this vsc matured packet
+		seqNumsHandled = append(seqNumsHandled, seqNums[idx+1]) // Note idx+1, since slash packet is at index 0
```
smarshall-spitzbart, line 108: I'm not a fan of assuming that `seqNums` will be constructed in s.t. the value at index 0 always corresponds to the slash packet, and all trailing indexes correspond to vsc matured. 

The tests around `GetSlashAndTrailingData` do validate the behavior I've described. However a better solution would be to refactor `GetSlashAndTrailingData` so that it returns structs that look something like..

```go
type slashPacketDataWithSeqNum struct {
	ibcSeqNum uint64
	data      ccvtypes.SlashPacketData
}

type vscMaturedPacketDataWithSeqNum struct {
	ibcSeqNum uint64
	data      ccvtypes.VSCMaturedPacketData
}
```

This is a trivial code change, but it'll likely make a large diff in `throttle_test.go`

x/ccv/provider/keeper/relay.go
```diff
@@ -267,6 +278,11 @@ func (k Keeper) EndBlockCIS(ctx sdk.Context) {
 	// - Marshaling and/or store corruption errors.
 	// - Setting invalid slash meter values (see SetSlashMeter).
 	k.CheckForSlashMeterReplenishment(ctx)
+
+	// Reset the vsc matured packet handler counter.
+	// Note this assumes vsc matured packets are only handled in the two methods following this call.
+	k.ResetVSCMaturedHandledThisBlock(ctx)
```
mpoke, line 284: All the VSCMaturedPackets are handled in EndBlock, more preciselly, in `EndBlockCIS`. Why do we need to store the number of handled packets in the store? We can just use a variable, right?

x/ccv/provider/keeper/relay.go
```diff
@@ -55,11 +55,19 @@ func (k Keeper) OnRecvVSCMaturedPacket(
...
+		for idx, data := range leadingVscMatured {
+			if k.VSCMaturedHandledLimitReached(ctx) {
+				// Break from inner for-loop, DeleteThrottledPacketData will still be called for this consumer
+				break
+			}
 			k.HandleVSCMaturedPacket(ctx, chain.ChainId, data)
+			ibcSeqNumsHandled = append(ibcSeqNumsHandled, ibcSeqNums[idx])
 		}
-		k.DeleteThrottledPacketData(ctx, chain.ChainId, ibcSeqNums...)
+		k.DeleteThrottledPacketData(ctx, chain.ChainId, ibcSeqNumsHandled...)
```
mpoke, line 70: Not really related to this PR, but why do we need the IBC packet seq number to store VSCMaturedPackets? We could just use the vscID, right?

x/ccv/provider/keeper/throttle.go
mpoke: What's the reasoning behind this check? If during HandleLeadingVSCMaturedPackets there were 100 packets handled, then without this line we'll handle one more in this block, right?

mpoke (changes requested):
Nice work @smarshall-spitzbart. The logic seems good. I think we don't need to store `VSCMaturedHandledThisBlock`.

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: You're right and esp because this feature will not be around for long 👍

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: We need ibc seq and not vscID because you could have a slash packet and vsc matured packet that would overwrite one another if they are relevant to the same vscID. ibcSeq is always unique

x/ccv/provider/keeper/relay.go
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/commit/cf5c452f53c21b404083cceeb3c1eb0eb7f6eebf

x/ccv/provider/keeper/throttle.go
smarshall-spitzbart: The behavior you've described sounds correct as far as what would happen. I added this check here because it'll keep the handling order as similar as possible to what existed previously

smarshall-spitzbart: ie. before this PR we _always_ handle _all_ VSC matured packets that're at the head of the queue, _before_ any slash packets. This line keeps that consistent

x/ccv/provider/keeper/throttle.go
```diff
@@ -82,18 +91,29 @@ func (k Keeper) GetEffectiveValPower(ctx sdktypes.Context,
...
 	slashPacketHandler func(sdktypes.Context, string, ccvtypes.SlashPacketData),
 	vscMaturedPacketHandler func(sdktypes.Context, string, ccvtypes.VSCMaturedPacketData),
+	vscMaturedHandledThisBlock int,
 ) {
 	// Get slash packet data and trailing vsc matured packet data, handle it all.
 	slashFound, slashData, vscMaturedData, seqNums := k.GetSlashAndTrailingData(ctx, consumerChainID)
+	seqNumsHandled := []uint64{}
 	if slashFound {
 		slashPacketHandler(ctx, consumerChainID, slashData)
+		// Slash packet should always be the first packet in the queue, so we can safely append the first seqNum.
```
MSalopek, line 101: Maybe link to parts of the code that make this more apparent.

smarshall-spitzbart: Good idea -> https://github.com/cosmos/interchain-security/commit/ba0d6abefa3b7179a49278d05580b494c617937c

MSalopek: Thanks, resolving

MSalopek (approved):
After a pair review session I agree with these changes.

The overall throttling logic can be refactored, but that's for another PR.

Approval

jtremback (approved):
LGTM

mpoke (approved):

Issue #1006, opened 2023-06-12
Add CometMock specific functionality: Advancing Time, Downtime & Advancing Blocks


p-offtermatt:
# Problem
CometMock should eventually provide specific functionality like advancing time, choosing signers and proposers,
running empty blocks, etc.
These custom endpoints should eventually be used in our e2e tests if we need more advanced testing behaviour.

This issue deals with three functionalities:
- [x] Advancing Time
- [x] Advancing Blocks
- [ ] Causing downtime

# Closing criteria

The end-to-end tests make use of CometMocks custom endpoints to advance time, advance blocks, and invoke downtime.

# Related issues
#938  the issue for implementing this functionality in CometMock

PR #1017, opened 2023-06-14
Add time and block advancement

p-offtermatt:
## Description

Closes: #1006 

CometMock now offers time advancement, block advancement and causing downtime via CometMock-specific RPC endpoints.

See https://github.com/informalsystems/CometMock/commit/4c231434b1b292bb657f4a6ea99744130f0c55c1 for changes on the CometMock side.

Changes should mostly be self-explanatory.

New baseline comparison with the time and block advancement is that CometBFT takes 4min50s for the short-happy-path,
CometMock takes 1min40s for it.

You can confirm this by running `go run ./tests/e2e --verbose --use-cometmock --use-gorelayer --short-happy-path --verbose` for CometMock and `go run ./tests/e2e --verbose --use-gorelayer --short-happy-path --verbose` for non-CometMock

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/e2e/state.go
```diff
@@ -199,6 +199,16 @@ func (tr TestRun) getBlockHeight(chain chainID) uint {
 }
 
 func (tr TestRun) waitBlocks(chain chainID, blocks uint, timeout time.Duration) {
```
jtremback, line 201: May want to rename this to newBlocks or something since it doesn't always wait

jtremback (approved):
LGTM

tests/e2e/actions.go
```diff
@@ -171,6 +172,17 @@ func (tr TestRun) startChain(
 		chain:     action.chain,
 		validator: action.validators[0].id,
 	}, verbose)
+
+	// store the fact that we started the chain
+	if tr.runningChains == nil {
```
smarshall-spitzbart, line 177: nit imo a more common pattern for this initialization would be a constructor for `TestRun` that initializes the map

tests/e2e/actions.go
```diff
@@ -1035,7 +1047,7 @@ func (tr TestRun) transferChannelComplete(
 	executeCommand(chanOpenConfirmCmd, "transferChanOpenConfirm")
 }
 
-func executeCommand(cmd *exec.Cmd, cmdName string) {
+func executeCommandWithVerbosity(cmd *exec.Cmd, cmdName string, verbose *bool) {
```
smarshall-spitzbart, line 1050: Nit for basic types
```suggestion
func executeCommandWithVerbosity(cmd *exec.Cmd, cmdName string, verbose bool) {
```

tests/e2e/actions.go
```diff
@@ -1063,6 +1075,10 @@ func executeCommand(cmd *exec.Cmd, cmdName string) {
 	}
 }
 
+func executeCommand(cmd *exec.Cmd, cmdName string) {
```
smarshall-spitzbart, line 1078:
```suggestion
// Executes a command with verbosity specified by CLI flag
func executeCommand(cmd *exec.Cmd, cmdName string) {
```

smarshall-spitzbart (approved):
Nice! LGTM

tests/e2e/actions.go
p-offtermatt: I put this into a seperate TestRun.initialize function. My problem with the constructor approach is that go doesn't support named arguments for funcs but for struct creation named arguments work. The calls where the TestRuns are constructed become huge blocks of 10 arguments which are only identifiable by their ordering if they go through a constructor, which I found worse than this solution

tests/e2e/state.go
p-offtermatt: I think it's better like this - even with CometMock, the name still gets at what it does, namely after this is done executing a number of blocks will have passed

p-offtermatt:
Thanks for the reviews, great suggestions!

PR #1024, opened 2023-06-15
feat!: throttle with retries, consumer changes

smarshall-spitzbart:
## Description

Implements the consumer-side changes from https://cosmos.github.io/interchain-security/adrs/adr-008-throttle-retries. See [this comment](https://github.com/cosmos/interchain-security/blob/25fc40a87de5cafc11a59c5f993962801896153c/x/ccv/consumer/keeper/throttle_retry.go#L12) for further context on how the behavior from the ADR was implemented in code.

Note the changes should be self documented, with an addition to the ADR to explain how upgrading would look like. Essentially, these changes should be merged separately and only affect consumers, with the provider throttling remaining unchanged. Then a future PR with provider changes can fully enable "throttling v2". 

Note e2e tests for new throttling functionality will be added to the PR introducing provider changes.

IMPORTANT: Note diff tests are currently commented out for this PR. We'll need a decision around what to do there.

Closes https://github.com/cosmos/interchain-security/issues/1101

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/consumer/keeper/relay.go
```diff
@@ -223,6 +236,40 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 // in conjunction with the ibc module's execution of "acknowledgePacket",
 // according to https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#processing-acknowledgements
 func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
+	if res := ack.GetResult(); res != nil {
+		if len(res) != 1 {
+			k.Logger(ctx).Error("recv invalid ack; expected length 1", "channel", packet.SourceChannel, "ack", res)
```
mpoke, line 241: Shouldn't we also return an error here?

x/ccv/consumer/keeper/relay.go
```diff
@@ -223,6 +236,40 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
...
+			k.ClearSlashRecord(ctx)           // Clears slash record state, unblocks sending of pending packets.
+			k.DeleteHeadOfPendingPackets(ctx) // Remove slash from head of queue. It's been handled.
+		case ccv.SlashPacketHandledResult[0]:
+			k.ClearSlashRecord(ctx)           // Clears slash record state, unblocks sending of pending packets.
+			k.DeleteHeadOfPendingPackets(ctx) // Remove slash from head of queue. It's been handled.
+		case ccv.SlashPacketBouncedResult[0]:
+			k.UpdateSlashRecordOnBounce(ctx)
+			// Note slash is still at head of queue and will now be retried after appropriate delay period.
+		default:
+			k.Logger(ctx).Error("recv invalid result ack; expected 1, 2, or 3", "channel", packet.SourceChannel, "ack", res)
```
mpoke, line 269: ditto: Shouldn't we also return an error here?

mpoke (changes requested):
Great work @smarshall-spitzbart. See my comments bellow.

tests/difference/core/driver/core_test.go
```diff
@@ -331,9 +330,9 @@ func (s *CoreSuite) TestTraces() {
 	fmt.Println("Shortest [traceIx, actionIx]:", shortest, shortestLen)
 }
 
-func TestCoreSuite(t *testing.T) {
-	suite.Run(t, new(CoreSuite))
-}
+// func TestCoreSuite(t *testing.T) {
```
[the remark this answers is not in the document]
insumity: It might make sense to either fully remove this if not needed, or if commented out (add a TODO comment) on what has to be checked/decided upon.

x/ccv/consumer/keeper/relay.go
```diff
@@ -213,6 +216,16 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
...
 		}
+		// If the packet that was just sent was a Slash packet, set the waiting on slash reply flag.
+		// This flag will be toggled false again when consumer hears back from provider. See OnAcknowledgementPacket below.
+		if p.Type == ccv.SlashPacket {
+			k.UpdateSlashRecordOnSend(ctx)
+			// Break so slash stays at head of queue
+			break
+		} else {
+			// Otherwise the vsc matured will be deleted
+			idxsForDeletion = append(idxsForDeletion, p.Idx)
```
insumity, line 227: Seems the `idxsForDeletion = append(idxsForDeletion, p.Idx)` is repeated here and below. Could this lead to cases where we try to delete a packet twice? Would it make sense to combine them to only `append` once?

x/ccv/consumer/keeper/relay.go
```diff
@@ -213,6 +216,16 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 			k.Logger(ctx).Error("cannot send IBC packet; leaving packet data stored:", "type", p.Type.String(), "err", err.Error())
 			break
 		}
+		// If the packet that was just sent was a Slash packet, set the waiting on slash reply flag.
+		// This flag will be toggled false again when consumer hears back from provider. See OnAcknowledgementPacket below.
+		if p.Type == ccv.SlashPacket {
+			k.UpdateSlashRecordOnSend(ctx)
+			// Break so slash stays at head of queue
```
insumity, line 223: Is the idea that we we want to `break` here, so that potential `VSCMature` packets that come after the `SlaskPacket` do **not** take precedence?
 It might be useful to add in the comment the exact reason we want the `SlashPacket` in the head of the queue.

x/ccv/consumer/keeper/relay.go
```diff
@@ -223,6 +236,40 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
...
+			return nil
+		}
+
+		// Otherwise we handle the result of the slash packet acknowledgement.
+		switch res[0] {
+		// We treat a v1 result as the provider successfully queuing the slash packet w/o need for retry.
+		case ccv.V1Result[0]:
+			k.ClearSlashRecord(ctx)           // Clears slash record state, unblocks sending of pending packets.
+			k.DeleteHeadOfPendingPackets(ctx) // Remove slash from head of queue. It's been handled.
+		case ccv.SlashPacketHandledResult[0]:
```
insumity, line 262: It seems that the `case ccv.SlashPacketHandledResult[0]:` and the `case ccv.V1Result[0]:` are handled in the same way. Could that potentially mean that we could get rid of `SlashPacketHandledResult = PacketAckResult([]byte{byte(2)})` in `ccv.go`?

x/ccv/consumer/keeper/relay.go
insumity: Is potentially the idea here that we do not return errors because the code that follows does `if err := ack.GetError(); err != "" {`?
Does this mean that we could have `ack.GetResult() != nil` being true and also `ack.GetError() != nil`? If this was the case, I guess then we shouldn't return `nil` later on in:
```
		if consumerPacket.Type == ccv.VscMaturedPacket {
			return nil
		}
```
because this would prevent checking for `ack.GetError()`?

Additionally, in the case where `ack.GetError() != nil` we close the underlying channel. Is this something we would like to do here as well?

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
+package keeper
+
+import (
+	"fmt"
+	"time"
+
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
```
insumity, line 7: nit: Not sure what's the naming convention here, but in `relay_test.go` we have:
```
```
Should we have `sdk` here as the name to be consistent?

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
...
+
+	consumertypes "github.com/cosmos/interchain-security/v3/x/ccv/consumer/types"
+)
+
+//
+// Throttling with retries follows a FSM design:
```
insumity, line 13: nit: Might help if this comment-description started with what the states of the state machine are and then describe the actions in each state. E.g., we have 2 states "No Slash" and "Standby" and in case we're ("No Slash", receive slash packet) => ("Standby"), ...

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
...
+		return true
+	}
+	if record.WaitingOnReply {
+		// We are waiting on a reply from provider, block sending
+		return false
+	}
+	// Retry delay period could be implemented as a param, but 1 hour is reasonable
+	retryDelayPeriod := time.Hour
+	timeSinceSend := ctx.BlockTime().Sub(record.SendTime)
+	// If retry delay period has elapsed, we can send again
```
insumity, line 58: I guess by param, you mean a general chain param (e.g., like one of [those](https://www.mintscan.io/cosmos/parameters))?

Would it still make sense to define this as a `const` in this file and then have `throttle_retry_test.go` use this constant? Because, if someone were to change this code now to have `retryDelayPeriod := 3 * time.Hour`, then the following test would fail:
```

	// Elapse retry delay period
	ctx = ctx.WithBlockTime(ctx.BlockTime().Add(2 * time.Hour))

	// Now packet sending is permitted again
	require.True(t, consumerKeeper.PacketSendingPermitted(ctx))
```

x/ccv/consumer/keeper/throttle_retry_test.go
```diff
@@ -0,0 +1,90 @@
...
+	require.False(t, consumerKeeper.PacketSendingPermitted(ctx))
+
+	// Elapse retry delay period
+	ctx = ctx.WithBlockTime(ctx.BlockTime().Add(2 * time.Hour))
+
+	// Now packet sending is permitted again
+	require.True(t, consumerKeeper.PacketSendingPermitted(ctx))
+}
+
+func TestThrottleRetryCRUD(t *testing.T) {
```
insumity, line 50: Would it make sense to break this test into multiple functions: `TestUpdateSlashRecordOnSend`, `TestUpdateSlashRecordOnBounce`, etc.?

x/ccv/consumer/keeper/relay.go
```diff
@@ -223,6 +236,40 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
...
+			k.ClearSlashRecord(ctx)           // Clears slash record state, unblocks sending of pending packets.
+			k.DeleteHeadOfPendingPackets(ctx) // Remove slash from head of queue. It's been handled.
+		case ccv.SlashPacketBouncedResult[0]:
+			k.UpdateSlashRecordOnBounce(ctx)
```
insumity, line 266: Just to verify I understand this correctly: If the provider chain considers itself getting overwhelmed with `SlashPacket`s it would acknowledge pack with a _bounce_ acknowledgment?

x/ccv/consumer/keeper/relay_test.go
```diff
@@ -210,10 +212,105 @@ func TestOnRecvVSCPacketDuplicateUpdates(t *testing.T) {
...
+	consumerKeeper.AppendPendingPacket(ctx, types.SlashPacket, &types.ConsumerPacketData_SlashPacketData{})
+	consumerKeeper.AppendPendingPacket(ctx, types.VscMaturedPacket, &types.ConsumerPacketData_VscMaturedPacketData{})
+
+	// Mock the channel keeper to return an error
+	gomock.InOrder(
+		mocks.MockChannelKeeper.EXPECT().GetChannel(ctx, types.ConsumerPortID,
+			"consumerCCVChannelID").Return(channeltypes.Channel{}, false).Times(1),
+	)
+
+	// No panic should occur, pending packets should not be cleared
```
insumity, line 234: nit: Maybe rename to something akin to `TestSendPacketsDoesNotSendWhenProblematicChannel`? Current naming of this test might be a bit confusing because I initially expected some kind of a failure (e.g. `err != nil`). I guess the assumption here (?) is that if you do not send the packets you consider to have failed. But in a similar vein, if you were to have 20 pending packets and only send the first 5 of those because the 6th is a `SlashPacket`, did the `SendPackets` fail or not?

x/ccv/consumer/keeper/relay_test.go
```diff
@@ -210,10 +212,105 @@ func TestOnRecvVSCPacketDuplicateUpdates(t *testing.T) {
...
+func TestSendPackets(t *testing.T) {
+	// Keeper setup
+	consumerKeeper, ctx, ctrl, mocks := testkeeper.GetConsumerKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	consumerKeeper.SetProviderChannel(ctx, "consumerCCVChannelID")
+	consumerKeeper.SetParams(ctx, consumertypes.DefaultParams())
+
+	// No slash record should exist
+	_, found := consumerKeeper.GetSlashRecord(ctx)
+	require.False(t, found)
+	require.True(t, consumerKeeper.PacketSendingPermitted(ctx))
```
insumity, line 248: nit: Could we remove this check here `// No slash record should exist` since tests on `PacketSendingPermitted` are already done in `throttle_retry_test.go`?

tests/integration/slashing.go
```diff
@@ -658,6 +682,7 @@ func (suite *CCVTestSuite) TestQueueAndSendSlashPacket() {
 	}
 
 	// send all pending packets - only slash packets should be queued in this test
+	// TODO: the following call can be removed
```
insumity, line 685: If we can remove, then why don't we do it here? Is it for reasons of clarity we want to keep it?

It might make sense to create issues for all TODOs we introduce in code and then in the TODO comment we can have something like:
```
// TODO (issue XYZ): Remove the following call
```
This way we can prevent cases were we forget to go back and fix the TODOs. (cc Not sure what's your take on that @mpoke)

x/ccv/provider/keeper/relay.go
```diff
@@ -44,7 +44,7 @@ func (k Keeper) OnRecvVSCMaturedPacket(
 		"vscID", data.ValsetUpdateId,
 	)
 
-	ack := channeltypes.NewResultAcknowledgement([]byte{byte(1)})
+	ack := channeltypes.NewResultAcknowledgement(ccv.V1Result)
```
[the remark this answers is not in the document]
smarshall-spitzbart: Returning `SlashPacketHandledResult` here would be state breaking for the provider. Note this PR is only intended to affect consumers

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
...
+// a consumer transitions from "No Slash" to "Standby". A slash record is created upon entry to this state,
+// and the consumer is now restricted from sending anymore packets.
+//
+// The slash packet remains at the head of the pending packets queue within the "Standby" state.
+//
+// - If the consumer receives a V1Result ack from the provider, the consumer checks for a slash record,
+// and if found, the consumer transitions from "Standby" to "No Slash". The slash record is cleared upon this transition,
+// and the slash packet is popped from the pending packets queue.
+//
+// - Else if the consumer receives a reply from the provider that the slash packet was successfully handled,
```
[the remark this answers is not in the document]
smarshall-spitzbart: Yea this is intentional, see [here](https://github.com/cosmos/interchain-security/blob/shawn/throttle-with-retries-consumer-changes/docs/docs/adrs/adr-008-throttle-retries.md#splitting-of-prs-and-upgrade-order). These consumer changes will be compatible with both the old and new provider throttling schemes

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
...
+// the consumer transitions from "Standby" to "No Slash". The slash record is cleared upon this transition,
+// and the slash packet is popped from the pending packets queue.
+//
+// - Else if the consumer receives a reply from the provider that the slash packet was bounced (not handled),
+// then SlashRecord.WaitingOnReply is set false, and the consumer retries sending the slash packet after a delay period.
+//
+// Once a retry is sent, the consumer enters a new cycle of the "Standby" state and the process repeats.
+//
+// Once the slash packet is successfully handled, the consumer transitions from "Standby" to "No Slash",
```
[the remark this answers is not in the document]
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/commit/81f47fbb21f9d9439f55ce0fff9df7125290a0ab, same description but I've removed some redundancies

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,109 @@
...
+	record, found := k.GetSlashRecord(ctx)
+	if !found {
+		// no slash record exists, send is permitted
+		return true
+	}
+	if record.WaitingOnReply {
+		// We are waiting on a reply from provider, block sending
+		return false
+	}
+	// Retry delay period could be implemented as a param, but 1 hour is reasonable
```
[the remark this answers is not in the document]
smarshall-spitzbart: This retry delay period doesn't matter at all as long as the period is:
- long enough that we're not retrying too often, potentially causing performance issues
- short enough that jailing isn't unnecessarily delayed 

1 hour fills these two criteria w/o introducing hundreds of lines of param boilerplate. We should prefer hardcoding when possible to reduce lines of code in the repo, but that's an opinion statement. Note if we did want to change params like these via gov prop, we'd still have to wait 2 weeks for that to take effect. If the param value was important enough to change ASAP, 2 weeks wouldn't be feasible.

x/ccv/consumer/types/throttle_retry.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import time "time"
+
+// NewSlashRecord creates a new slash record
+func NewSlashRecord(sendTime time.Time, waitingOnReply bool) (record SlashRecord) {
```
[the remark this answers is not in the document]
smarshall-spitzbart: No, just following golang patterns for constructors here

x/ccv/consumer/keeper/relay.go
```diff
@@ -188,8 +188,11 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 	pending := k.GetAllPendingPacketsWithIdx(ctx)
 	idxsForDeletion := []uint64{}
 	for _, p := range pending {
+		if !k.PacketSendingPermitted(ctx) {
```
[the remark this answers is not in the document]
smarshall-spitzbart: Not sure what you mean, `PacketSendingPermitted` could return true for multiple packets in the for-loop (all vsc matured for example). The code is intended to follow [this behavior](https://github.com/cosmos/interchain-security/blob/shawn/throttle-with-retries-consumer-changes/docs/docs/adrs/adr-008-throttle-retries.md#consumer-changes)

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: By checking ibc-go source code, it's impossible for `ack.GetResult()` and `ack.GetError()` to both return non-nil values for the same `ack`. 

Additionally, returning an error from this method [seems to just log again from ibc-go](https://github.com/cosmos/ibc-go/blob/951b3abd9f38af0697a6d65b95fd2b0921547acd/modules/core/keeper/msg_server.go#L680). I'm fine returning the error and logging again, just seems redundant. 

Re closing the underlying channel, [this issue is relevant ](https://github.com/cosmos/interchain-security/issues/547)

proto/interchain_security/ccv/consumer/v1/consumer.proto
```diff
@@ -89,3 +89,11 @@ message MaturingVSCPacket {
   google.protobuf.Timestamp maturity_time = 2
       [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
 }
+
+// A record storing the state of a slash packet sent to the provider chain
+// which maybe bounced back and forth until handled by the provider.
```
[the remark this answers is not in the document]
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/commit/130de0d330ad44dd6977a0c6f9a502472afa56f9

Re slash record vs slash packet data, these are two disparate types in that `SlashPacketData` is the protobuf type for bytes sent over the wire from consumer to provider, when a slash event happens. We should never change the schema of anything that's sent over the wire. 

`SlashRecord` is the protobuf type for state internal to the consumer neccessary to implement the finite state machine defined elsewhere in this PR

x/ccv/consumer/keeper/keeper.go
```diff
@@ -606,6 +606,18 @@ func (k Keeper) getAndIncrementPendingPacketsIdx(ctx sdk.Context) (toReturn uint
 	return toReturn
 }
 
+// DeleteHeadOfPendingPackets deletes the head of the pending packets queue.
+// TODO: UTs
```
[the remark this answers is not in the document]
smarshall-spitzbart: Unit tests, now added in https://github.com/cosmos/interchain-security/commit/8a3553318befd4e86e90bb6939e9c8d92e019381

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: Nice catch, this was either a mistake on my part or a faulty merge conflict resolution 

fixed in https://github.com/cosmos/interchain-security/commit/a972fd4cb2a0edb44651ba39bd1157ae050b6f8d

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: The idea is to follow [this design](https://cosmos.github.io/interchain-security/adrs/adr-008-throttle-retries#consumer-changes), we're breaking so that the sending of any other packet is blocked until the slash packet is handled

Hopefully [this change ](https://github.com/cosmos/interchain-security/commit/9d111a87c68ac85046d84b7364ac30b711cec144) can help clear things up

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/pull/1024#discussion_r1286268187, but after revisiting this issue, yes we could potentially consolidate `SlashPacketHandledResult` and `V1Result` into a single enum value, as long as that value is 0x1. However, that enum value has completely different meaning once we implement provider changes. Seems like bad practice to change the meaning of an existing enum sent over the wire. 

pro of current design:
- Not overloading multiple meanings to an enum sent over the wire

con of current design:
- slightly more code

x/ccv/consumer/keeper/throttle_retry.go
smarshall-spitzbart:
```
```

is probably used most often. I personally try to use the `sdktypes` alias as it's more descriptive of what path we're actually referencing

x/ccv/consumer/keeper/throttle_retry.go
smarshall-spitzbart: Yea generally if anyone uses the term "param" in a vague way, they're probably talking about the cosmos sdk type of general chain params you've linked 👍 

As far as defining a constant, seems reasonable to me. Note that I doubt this value will ever need to be changed. And if it's intended to be changed, it should be a proper "param". 

https://github.com/cosmos/interchain-security/commit/03daa2a2277c3a7f333f4e18102456dd2f1a2a3d

x/ccv/consumer/keeper/throttle_retry_test.go
smarshall-spitzbart: I see no problem with splitting this into multiple unit tests, but in this scenario I happened to be testing the overwrite functionality of `UpdateSlashRecordOnSend`, so grouping everything together worked out. 

Note these methods are well tested at higher levels as well

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: Yep thats correct 👍 

More info [here](https://cosmos.github.io/interchain-security/adrs/adr-002-throttle) and [here](https://cosmos.github.io/interchain-security/adrs/adr-008-throttle-retries)

x/ccv/consumer/keeper/relay_test.go
smarshall-spitzbart: Note this method is unchanged before and after this PR. I believe the diff is making it seem like a new test because I moved stuff around. 

Regarding your question akin to _what is this testing_, the test is just confirming that if `ChannelKeeper.GetChannel` returns found=false, the `SendPackets` method does not panic, and doesn't clear any packets. Your name seems good, but maybe we do that in a separate PR to not conflate things?

x/ccv/consumer/keeper/relay_test.go
smarshall-spitzbart: This check is not meant to test the `PacketSendingPermitted` method, it's just a quick sanity check that after setup, the state of the keeper is instantiated without a slash record. 

The check seemed relevant to both `TestPacketSendingPermitted`, and `TestSendPackets` because in test files, I prioritize readability and explicitness over code minimization

x/ccv/consumer/keeper/throttle_retry.go
smarshall-spitzbart: Good idea, I should draw up the state machine as well

tests/integration/slashing.go
smarshall-spitzbart: In this case, I think this was just a reminder for future shawn to remove the call, that I missed :) see https://github.com/cosmos/interchain-security/commit/569599a425d8fd9bdf8c5a1dd2fca313be8ce247

In general I agree that TODO comments are most useful when there's an issue attached to it

x/ccv/provider/keeper/relay.go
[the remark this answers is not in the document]
mpoke: It would not be if `SlashPacketHandledResult = V1Result`, right?

x/ccv/consumer/keeper/relay.go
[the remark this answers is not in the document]
mpoke: If the if condition is triggered in the middle of the iteration, then `k.DeletePendingDataPackets(ctx, idxsForDeletion...)` will not be called due to the return statement.

x/ccv/consumer/keeper/relay.go
mpoke: > Additionally, returning an error from this method [seems to just log again from ibc-go](https://github.com/cosmos/ibc-go/blob/951b3abd9f38af0697a6d65b95fd2b0921547acd/modules/core/keeper/msg_server.go#L680). I'm fine returning the error and logging again, just seems redundant.

If we don't return an error, then IBC will log `acknowledgement succeeded`, which will make debugging very annoying.

x/ccv/consumer/keeper/relay.go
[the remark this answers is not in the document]
smarshall-spitzbart: Ah great catch! From my understanding,`PacketSendingPermitted` could not return false in the middle of an iteration.

`PacketSendingPermitted` checks three pieces of state: 
- is slash record found? Note if a slash record is created or updated, the [iteration breaks anyways](https://github.com/cosmos/interchain-security/blob/18a86c359d219520e59bf6c1492780bfa212d8f6/x/ccv/consumer/keeper/relay.go#L222)
- is record.WaitingOnReply true? This flag is mutated from `OnAcknowledgementPacket`, which can't run in the middle of `SendPackets`. 
- Has retry delay period passed? `ctx.Blocktime()` will trivially not change in the middle of `SendPackets`

Regardless, it's possible I've missed something in my reasoning here. To protect against that possibility, we can simply `break` instead of `return` if `PacketSendingPermitted` returns false, ensuring the deletion line is always hit. I don't see any issues with that idea

Fixed in https://github.com/cosmos/interchain-security/commit/f57aa641139aa56b337bd64e516d841146e725f1

x/ccv/provider/keeper/relay.go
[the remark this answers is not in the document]
smarshall-spitzbart: Yea that's true, see https://github.com/cosmos/interchain-security/pull/1024#discussion_r1286374772, we could reuse an existing enum that's sent over the wire and give it a new meaning, but imo that would be bad practice

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: Ah valid point! Fixed in https://github.com/cosmos/interchain-security/commit/afbe04c03a6ba00b43b7f9bed6d658e307a2dac0

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: https://github.com/cosmos/interchain-security/commit/afbe04c03a6ba00b43b7f9bed6d658e307a2dac0

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: Unfortunately the above change caused an odd bug that only fails integration tests in CI, and not locally. I'm looking into it

x/ccv/consumer/keeper/relay.go
```diff
@@ -223,6 +236,40 @@ func (k Keeper) SendPackets(ctx sdk.Context) {
 // in conjunction with the ibc module's execution of "acknowledgePacket",
 // according to https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#processing-acknowledgements
 func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
+	if res := ack.GetResult(); res != nil {
+		if len(res) != 1 {
+			return fmt.Errorf("acknowledgement result length must be 1, got %d", len(res))
```
mpoke, line 241: Is this error message logged by IBC? If not, we should log it for better debugging.

mpoke (approved):

x/ccv/consumer/keeper/relay.go
insumity: >  Seems like bad practice to change the meaning of an existing enum sent over the wire.
Agreed! 👍

insumity (approved):

tests/difference/core/driver/core_test.go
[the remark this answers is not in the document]
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/commit/281527b2bc97069f544cd30a0a0e77f0754c25c1

x/ccv/consumer/keeper/throttle_retry.go
```diff
@@ -0,0 +1,102 @@
...
+// - Else if the consumer receives an ack from the provider that the slash packet was bounced (not handled),
+// then SlashRecord.WaitingOnReply is set false, and the consumer retries sending the slash packet after a delay period.
+//
+// Once a retry is sent, the consumer enters a new cycle of the "Standby" state and the process repeats.
+//
+// This design is implemented below, and in relay.go under SendPackets() and OnAcknowledgementPacket().
+//
+
+// Retry delay period could be implemented as a param, but 1 hour is reasonable
+const RETRY_DELAY_PERIOD = time.Hour
```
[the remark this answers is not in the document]
smarshall-spitzbart: Ah good catch, must've been reminiscing about C++ lol. Camel case is the const standard for the repo 👍 

https://github.com/cosmos/interchain-security/commit/9ed719517334bc6d9071d9980bac127d874e450a

x/ccv/consumer/keeper/relay.go
smarshall-spitzbart: Yea it seems to be logged here https://github.com/cosmos/ibc-go/blob/000e9d20b15b3c6939cefc0e26daa4b64c1be8fb/modules/core/keeper/msg_server.go#L680

x/ccv/consumer/keeper/throttle_retry.go
smarshall-spitzbart: https://github.com/cosmos/interchain-security/commit/a834a5bf668c396b2bcd4fb0de692184d754efff

Issue #1033, opened 2023-06-16
Improve storage of PendingPackets on consumer


smarshall-spitzbart:
## Problem

The consumer ccv module currently queues "pending packets" to be sent on each endblocker in [SendPackets](https://github.com/cosmos/interchain-security/blob/3bc4e7135066d848aac60b0787364c07157fd36d/x/ccv/consumer/keeper/relay.go#L178). These packets are queued using a protobuf list in `ConsumerPacketData`. For a single append operation, the entire list is deserialized, then a packet is appended to that list, and the list is serialized again, See [AppendPendingPacket](https://github.com/cosmos/interchain-security/blob/05c2dae7c6372b1252b9e97215d07c6aa7618f33/x/ccv/consumer/keeper/keeper.go#L606). That is, a single append operation has O(N) complexity, where N is the size of the list. This isn't a problem when the list is small. But with #713 being implemented, this list can potentially grow to the order of thousands of entries, in the scenario that a slash packet is bouncing.

## Proposed solution

We can improve the append time for this queue by converting it from a protobuf-esq list, to a queue implemented with sdk-esq code. 

My rough idea: The key schema for storing slash and VSC matured packets will look like `[prefix]-[blockheight]` and the value for each key will contain a protobuf list with any packets queued for that blockheight. Since we're keying by blockheight, an sdk iterator will iterate over the keys in FIFO order. That's it. We'll have to append to a protobuf list, but only for multiple packets that're queued during the same blockheight. Otherwise if most packets are queued from different block heights (which is surely the case for VSCMatured packets), appending to the queue is near constant time.

Note this should be solved in it's own PR separate from other throttle related changes

smarshall-spitzbart:
cc @jtremback since we chatted about this

smarshall-spitzbart:
Actually theres an even simpler way to approach this problem. 

Store a head/tail integer for the pending packets queue. Whenever you append to the queue, simply store the new entry with key looking like: [prefix-tail+1]. Then Increment tail. 

For access, leverage sdk's iterator, keys are already ordered by ascending bytes. 

Maybe there's already some queue boilerplate in sdk? I'll take a look on Monday

Issue #713, opened 2023-02-06
EPIC: Improve throttling mechanism with retries


shaspitz:
# Problem

Currently the throttling mechanism is designed so that provider logic (slash meter, etc.) dictates how many slash packets can be handled over time. Throttled slash packets are persisted on the provider, leading to issues such as #594. We can improve the throttling mechanism to instead queue/persist relevant data on each consumer, and have consumers retry slash requests as needed. CC @jtremback who came up with this idea

# Closing criteria

Relevant data is persisted on each consumer with retry system implemented to enforce throttling. 

# Problem details

- Throttling logic will stay on provider 
- Queuing/data persisting will now be on consumer
- Consumer will retry slash requests when they cannot be immediately handled by the provider. IBC acknowledgements are used to implement such an idea
- VSC matured packet delivery will likely need to be retried as well, as these packets are throttled in some cases 

# Task list

```[tasklist]
### Must have
- [ ] https://github.com/cosmos/interchain-security/issues/1023
- [ ] https://github.com/cosmos/interchain-security/issues/1033
- [ ] https://github.com/cosmos/interchain-security/issues/1101
- [x] Revert changes from https://github.com/cosmos/interchain-security/pull/1004 as they now should be implicitly handled
- [ ] https://github.com/cosmos/interchain-security/issues/609
- [ ] https://github.com/cosmos/interchain-security/issues/564
- [ ] https://github.com/cosmos/interchain-security/issues/1102
- [ ] https://github.com/cosmos/interchain-security/issues/1233
- [ ] https://github.com/cosmos/interchain-security/issues/1103
- [ ] https://github.com/cosmos/interchain-security/issues/1032
- [x] PR for final feature branch
```

```[tasklist]
### Nice to have
- [ ] https://github.com/cosmos/interchain-security/issues/566
- [ ] https://github.com/cosmos/interchain-security/issues/706
```

smarshall-spitzbart:
Note: implementing this improvement would close #594. We also need to consider the importance of this issue in relation to more general "untrusted consumer" features

mpoke:
This could considerably increase the cost of relaying.

smarshall-spitzbart:
Step 1 is writing an ADR

smarshall-spitzbart:
This PR should be based onto `feat/throttle-with-retries-complete`, but I'm not able to update that base branch to reflect merges to main.. This PR is ready to review regardless

PR #1019, opened 2023-06-14
feat!: upgrade ics sdk47 ibc7

MSalopek:
## Description

Closes: #852 (major epic)

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

* [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [x] Added `!` to the type prefix if API or client breaking change
* [ ] Confirmed this PR does not introduce changes requiring state migrations, OR migration code has been added to consumer and/or provider modules
* [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
* [x] Provided a link to the relevant issue or specification
* [ ] Followed the guidelines for [building SDK modules](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules)
* [x] Included the necessary unit and integration [tests](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#testing)
* [ ] Added a changelog entry to `CHANGELOG.md`
* [x] Included comments for [documenting Go code](https://blog.golang.org/godoc)
* [ ] Updated the relevant documentation or specification
* [x] Reviewed "Files changed" and left comments if necessary 
* [x] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

* [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ] confirmed `!` in the type prefix if API or client breaking change
* [ ] confirmed this PR does not introduce changes requiring state migrations, OR confirmed migration code has been added to consumer and/or provider modules
* [ ] confirmed all author checklist items have been addressed
* [ ] reviewed state machine logic
* [ ] reviewed API design and naming
* [ ] reviewed documentation is accurate
* [ ] reviewed tests and test coverage

PR #1037, opened 2023-06-19
feat!: optimize pending packets storage on consumer + migration

smarshall-spitzbart:
## Description

Closes: #1033 

This PR accomplishes two things:
- Reduces time complexity of appending "pending data packets" on the consumer. This optimization becomes important once #713 is implemented.
- Allows for #713 to easily delete some pending data packets, not all at once. 

**CONSUMER MIGRATION REQUIRED FOR THIS CHANGE**

Note: since this PR does not mutate `consumertypes.GenesisState`, no migration is required for the provider.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/v1/ccv.proto
```diff
@@ -58,17 +58,19 @@ message MaturedUnbondingOps {
...
-// ConsumerPacketData contains a consumer packet data and a type tag
+// ConsumerPacketData contains a consumer packet data, type tag, and index for storage.
 message ConsumerPacketData {
   ConsumerPacketDataType type = 1; 
 
   oneof data {
     SlashPacketData slashPacketData = 2;
     VSCMaturedPacketData vscMaturedPacketData = 3;
   }
+  uint64 idx = 4;
```
smarshall-spitzbart, line 69: Extra care should be taken in reviewing this PR, as it does mutate a proto file. Note this index field was added to the `ConsumerPacketData` type so that we can use that index in `PendingDataPacketsKey`, to implement a queue with constant append time. See `AppendPendingPacket`

x/ccv/consumer/keeper/genesis.go
```diff
@@ -89,9 +89,12 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state *consumertypes.GenesisState)
 			k.SetLastTransmissionBlockHeight(ctx, state.LastTransmissionBlockHeight)
 		}
 
-		// set pending consumer pending packets
+		// Set pending consumer packets, using the depreciated ConsumerPacketDataList type
+		// that exists for genesis.
 		// note that the list includes pending mature VSC packet only if the handshake is completed
-		k.AppendPendingPacket(ctx, state.PendingConsumerPackets.List...)
+		for _, packet := range state.PendingConsumerPackets.List {
```
smarshall-spitzbart, line 95: Genesis methods still expect the depreciated type `ConsumerPacketDataList` to reduce the complexity of how we'd actually deploy this PR

x/ccv/consumer/keeper/keeper.go
```diff
@@ -565,48 +565,78 @@ func (k Keeper) GetAllCCValidator(ctx sdk.Context) (validators []types.CrossChai
...
-	store := ctx.KVStore(k.storeKey)
-	bz, err := packets.Marshal()
-	if err != nil {
-		// This should never happen
-		panic(fmt.Errorf("failed to marshal ConsumerPacketDataList: %w", err))
-	}
-	store.Set(types.PendingDataPacketsKey(), bz)
+// Note: PendingDataPacketsBytePrefix is the correct prefix, NOT PendingDataPacketsByteKey.
+// See consistency with PendingDataPacketsKey().
+func PendingDataPacketsIterator(store sdk.KVStore) sdk.Iterator {
```
smarshall-spitzbart, line 570: The meat of this PR is in this file

smarshall-spitzbart:
Ready to review but I won't open up this PR until #1019 is merged and I can fix conflicts

docs/docs/adrs/adr-008-throttle-retries.md
```diff
@@ -46,6 +47,20 @@ With the behavior described, we maintain very similar behavior to the current th
 
 In the normal case, when no or a few slash packets are being sent, the VSCMaturedPackets will not be delayed, and hence unbonding will not be delayed.
 
+### Consumer pending packets storage optimization
+
+In addition to the mentioned consumer changes above. An optimization will need to be made to the consumer's pending packets storage to properly implement the feature from this ADR.
+
+The consumer ccv module previously queued "pending packets" to be sent on each endblocker in [SendPackets](https://github.com/cosmos/interchain-security/blob/3bc4e7135066d848aac60b0787364c07157fd36d/x/ccv/consumer/keeper/relay.go#L178). These packets are queued in state with a protobuf list of `ConsumerPacketData`. For a single append operation, the entire list is deserialized, then a packet is appended to that list, and the list is serialized again, See older version of [AppendPendingPacket](https://github.com/cosmos/interchain-security/blob/05c2dae7c6372b1252b9e97215d07c6aa7618f33/x/ccv/consumer/keeper/keeper.go#L606). That is, a single append operation has O(N) complexity, where N is the size of the list.
```
mpoke, line 54: typo:
```suggestion
The consumer ccv module previously queued "pending packets" to be sent on each endblocker in [SendPackets](https://github.com/cosmos/interchain-security/blob/3bc4e7135066d848aac60b0787364c07157fd36d/x/ccv/consumer/keeper/relay.go#L178). These packets are queued in state with a protobuf list of `ConsumerPacketData`. For a single append operation, the entire list is deserialized, then a packet is appended to that list, and the list is serialized again. See older version of [AppendPendingPacket](https://github.com/cosmos/interchain-security/blob/05c2dae7c6372b1252b9e97215d07c6aa7618f33/x/ccv/consumer/keeper/keeper.go#L606). That is, a single append operation has O(N) complexity, where N is the size of the list.
```

proto/interchain_security/ccv/v1/ccv.proto
```diff
@@ -56,17 +56,22 @@ message SlashPacketData {
...
 }
 
+
+// [Depreciated] favor using []ConsumerPacketData directly, which can be stored more efficiently.
+//
 // ConsumerPacketDataList is a list of consumer packet data packets.
+// It is only used for genesis to ensure backwards compatibility with older versions of ICS.
```
mpoke, line 74: Why is this needed for backwards compability? IIRC, this list is exported in `ExportGenesis`. If a chain runs `ExportGenesis` with ICS v2.0.0 for example, then it should run `InitGenesis` with the same version. I don't think we need to keep backwards compability here.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -592,48 +592,78 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
-func (k Keeper) SetPendingPackets(ctx sdk.Context, packets ccv.ConsumerPacketDataList) {
-	store := ctx.KVStore(k.storeKey)
-	bz, err := packets.Marshal()
-	if err != nil {
-		// This should never happen
-		panic(fmt.Errorf("failed to marshal ConsumerPacketDataList: %w", err))
-	}
-	store.Set(types.PendingDataPacketsKey(), bz)
+// Note: PendingDataPacketsBytePrefix is the correct prefix, NOT PendingDataPacketsByteKey.
+// See consistency with PendingDataPacketsKey().
```
mpoke, line 596: Remove this function as it's not really needed.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -592,48 +592,78 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
+// This index is used for implementing a FIFO queue of pending packets in the KV store.
+func (k Keeper) getAndIncrementPendingPacketsIdx(ctx sdk.Context) (toReturn uint64) {
 	store := ctx.KVStore(k.storeKey)
-	bz := store.Get(types.PendingDataPacketsKey())
+	bz := store.Get(types.PendingPacketsIndexKey())
 	if bz == nil {
-		return packets
+		toReturn = 0
+	} else {
+		toReturn = binary.BigEndian.Uint64(bz)
```
mpoke, line 609: Use `sdk.BigEndianToUint64(bz)`

x/ccv/consumer/keeper/keeper.go
```diff
@@ -592,48 +592,78 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
-	err := packets.Unmarshal(bz)
-	if err != nil {
-		// An error here would indicate something is very wrong,
-		// the PendingPackets are assumed to be correctly serialized in SetPendingPackets.
-		panic(fmt.Errorf("failed to unmarshal pending data packets: %w", err))
+// GetPendingPackets returns ALL the pending CCV packets from the store
+func (k Keeper) GetPendingPackets(ctx sdk.Context) []ccv.ConsumerPacketData {
+	var packets []ccv.ConsumerPacketData
+	store := ctx.KVStore(k.storeKey)
+	iterator := PendingDataPacketsIterator(store)
```
mpoke, line 620: Replace with `sdk.KVStorePrefixIterator(store, []byte{types.PendingDataPacketsBytePrefix})` as it's easier to read. Like this we need to jump to another function to see what prefix is iterated.

x/ccv/consumer/keeper/keeper.go
```diff
@@ -592,48 +592,78 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
+	store := ctx.KVStore(k.storeKey)
+	for _, idx := range idxs {
+		store.Delete(types.PendingDataPacketsKey(idx))
+	}
+}
+
+func (k Keeper) DeleteAllPendingDataPackets(ctx sdk.Context) {
 	store := ctx.KVStore(k.storeKey)
-	store.Delete(types.PendingDataPacketsKey())
+	iterator := PendingDataPacketsIterator(store)
```
mpoke, line 645: ditto

x/ccv/consumer/keeper/keeper.go
```diff
@@ -592,48 +592,78 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
-func (k Keeper) DeletePendingDataPackets(ctx sdk.Context) {
+// DeletePendingDataPackets deletes pending data packets with given indexes
+func (k Keeper) DeletePendingDataPackets(ctx sdk.Context, idxs ...uint64) {
+	store := ctx.KVStore(k.storeKey)
+	for _, idx := range idxs {
+		store.Delete(types.PendingDataPacketsKey(idx))
+	}
+}
+
+func (k Keeper) DeleteAllPendingDataPackets(ctx sdk.Context) {
```
mpoke, line 643: Not sure how safe it is to delete within an iterator. I think it's safer to just `GetPendingPackets` and then `DeletePendingDataPackets`.

x/ccv/consumer/keeper/genesis.go
mpoke: See my comment on ccv.proto

mpoke (changes requested):
I approve the general logic. I don't think we should care about backwards compatibility when it comes to export/import genesis.

x/ccv/consumer/keeper/keeper.go
smarshall-spitzbart: Nice catch, yea we should always delete outside the iterator

proto/interchain_security/ccv/v1/ccv.proto
smarshall-spitzbart: Your understanding is all correct 👍 however the issue becomes a little more complex when you consider that consumers query genesis state from the provider. 

If the provider is running v2 for example, and a consumer is running v4 with the PR merged, there would be manual changes we'd have to make to the consumer genesis JSON. Similar to Stride and adding the `PreCCV` field. Are we okay with manual changes being required? 

It's a tradeoff between keeping around some technical debt vs requiring some hackyness if the provider is on an old version of ICS

mpoke: That's a good point. I keep forgetting that the consumer CCV genesis state is part of the "wire".  Then I guess it's good to use `ConsumerPacketDataList` to export the genesis state. Please do not mark the struct as deprecated though. Just explain in the docstring that it's used only for exporting / importing the state.

smarshall-spitzbart: SGTM -> https://github.com/cosmos/interchain-security/commit/a42229a465fd6f8f3f4fc14dcb4a5285ba1c4b84

MSalopek:
We are not backporting this into v2/3?

smarshall-spitzbart:
> We are not backporting this into v2/3?

This is a feature that is needed for throttling improvements which will land in v4, these changes are not neccessary for v2/v3

smarshall-spitzbart:
@sainoe @MSalopek could I get a review on this when you have time

docs/docs/adrs/adr-008-throttle-retries.md
```diff
@@ -46,6 +47,21 @@ With the behavior described, we maintain very similar behavior to the current th
...
+
+This poor append performance isn't a problem when the pending packets list is small. But with this ADR being implemented, the pending packets list could potentially grow to the order of thousands of entries, in the scenario that a slash packet is bouncing.
+
+We can improve the append time for this queue by converting it from a protobuf-esq list, to a queue implemented with sdk-esq code. The idea is to persist an uint64 index that will be incremented each time you queue up a packet. You can think of this as storing the tail of the queue. Then, packet data will be keyed by that index, making the data naturally ordered byte-wite for sdk's iterator. The index will also be stored in the packet data value bytes, so that the index can later be used to delete certain packets from the queue.
```
sainoe, line 58: > making the data naturally ordered byte-wite for sdk's iterator
Wdym by "byte-wite"?

x/ccv/consumer/keeper/keeper.go
```diff
@@ -593,48 +593,80 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
+// This index is used for implementing a FIFO queue of pending packets in the KV store.
+func (k Keeper) getAndIncrementPendingPacketsIdx(ctx sdk.Context) (toReturn uint64) {
 	store := ctx.KVStore(k.storeKey)
-	bz, err := packets.Marshal()
-	if err != nil {
-		// This should never happen
-		panic(fmt.Errorf("failed to marshal ConsumerPacketDataList: %w", err))
+	bz := store.Get(types.PendingPacketsIndexKey())
+	if bz == nil {
+		toReturn = 0
```
sainoe, line 602: This condition could be simplified since `toReturn` is already equal to 0

x/ccv/consumer/keeper/keeper.go
```diff
@@ -593,48 +593,80 @@ func (k Keeper) GetAllValidators(ctx sdk.Context) (validators []stakingtypes.Val
...
+	} else {
+		toReturn = sdk.BigEndianToUint64(bz)
 	}
```
sainoe, line 605: This condition can be simplified as `toReturn` is already equal to 0.
```suggestion
	if bz == nil {
		toReturn = 0
	} else {
		toReturn = sdk.BigEndianToUint64(bz)
	}
```

sainoe (approved):
Great work @smarshall-spitzbart! Just two nits, looks pretty solid to me.

docs/docs/adrs/adr-008-throttle-retries.md
smarshall-spitzbart: my b haha I meant "byte-wise", as in the data is ordered by increasing byte value. Will change

Issue #801, opened 2023-03-24
EPIC: Split `go.mod` into multiple `go.mod`s


smarshall-spitzbart:
# Problem

Currently ICS has a single `go.mod` which has golang dependancies for any code utilized by the provider module, the consumer module, and all tests therein. Ex: this can cause issues when we need to bump a dependancy for the consumer module, but not for the provider module.

# Closing criteria

Create multiple `go.mod`s, one for the provider module, one for the consumer module, and one for everything else (tests, common helpers, etc.). That way consumer chains will only need to worry about the dependancies contained within the consumer go.mod.

# Problem details

It's notable that releasing and branching are topics that are relevant to this issue.

Once we split out the go.mods in this repo, we would then have the option to split up how we do versioning for the provider and consumer modules.

Currently, everything is very monorepo-esq, in that a change to the provider module, consumer module, or protocol API, all follow the same semver rules, and increment the same running version of ICS. 

We could setup a versioning system where a provider-only change only bumps the provider version, same with consumer, and a protocol/api change would affect both the provider and consumer versions. 

Release branching would get tricky if this split out release system were to be built out, although it could be as simple as adding a branch prefix for the consumer vs provider version

# Task list

```[tasklist]
### Must have
- [ ] https://github.com/cosmos/interchain-security/pull/890 serves as experimentation and some of that work can be salvaged
- [ ] Refactor /x/ccv/types s.t. it doesn't have external package dependancies from this repo. Make sure x/ccv/types only contains code that is shared by consumer/provider (including .pb files)
- [ ] Refactor x/ccv/provider and x/ccv/consumer s.t. they only rely on x/ccv/types, not one another, and not any other packages from this repo (not even testutils for example)
- [ ] Add go.mods to the three directories in x/ccv, import dependancies should match what's described above
- [ ] Add documentation and guidelines around how consumer/provider semver can be completely separate. Make note that if the shared /x/ccv/types folder is changed (and new release is cut), that's the only scenario that a new release will be cut for both consumer/provider. Also update PR template to reflect this workflow
- [ ] Split out `make proto-gen` into `make proto-gen-provider`, `make proto-gen-consumer`, and `make proto-gen-types`
- [ ] https://github.com/cosmos/interchain-security/issues/889
```

```[tasklist]
### Nice to have

```

smarshall-spitzbart:
[...]

PR #1071, opened 2023-06-22
refactor: clean protos for clarity

smarshall-spitzbart:
## Description

Closes: #1072 

Achieves:

- Properly organizing .proto files. 
- Rebuilding .pb.go files accordingly. 
- Changing package references accordingly.

Note this PR also contributes toward https://github.com/cosmos/interchain-security/issues/801, in that we're reducing the amount of cross dependancies between the consumer ccv module and the provider ccv module. The end goal is that all shared dependancies live within `x/ccv/types`, then `x/ccv/provider` and `/x/ccv/consumer` do no reference one another. This is surely possible, but not easy and a huge source of tech debt.

Notes:

- protobuf break check is expected to fail. 
- If this PR was implemented correctly, it should be a **pure refactor**, not changing any code functionality. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/v1/shared_consumer.proto
```diff
@@ -0,0 +1,146 @@
...
+import "tendermint/abci/types.proto";
+import "ibc/lightclients/tendermint/v1/tendermint.proto";
+import "ibc/core/channel/v1/channel.proto";
+import "google/protobuf/duration.proto";
+import "gogoproto/gogo.proto";
+import "interchain_security/ccv/v1/wire.proto";
+import "google/protobuf/timestamp.proto";
+
+//
+// Note any type defined in this file is referenced/persisted in both the consumer and provider CCV modules,
```
smarshall-spitzbart, line 16: TLDR most consumer protos are persisted on provider, so most ended up in this file

proto/interchain_security/ccv/v1/shared_consumer.proto
```diff
@@ -0,0 +1,146 @@
...
+// but not sent over the wire. These schemas could change, only with careful consideration of effects!
+//
+
+// TODO: rename to ConsumerParams (make issue)
```
smarshall-spitzbart, line 20: We'd need to rename the consumer `Params` type to `ConsumerParams`, since it's no longer in the x/ccv/consumer package. now it's in x/ccv

proto/interchain_security/ccv/v1/shared_consumer.proto
```diff
@@ -0,0 +1,146 @@
...
+    // Reward denoms. These are the denominations which are allowed to be sent to
+    // the provider as rewards.
+    repeated string reward_denoms = 11;
+  
+    // Provider-originated reward denoms. These are denoms coming from the
+    // provider which are allowed to be used as rewards. e.g. "uatom"
+    repeated string provider_reward_denoms = 12;
+}
+
+// TODO: rename to ConsumerGenesisState (make issue)
```
smarshall-spitzbart, line 82: Same for consumer genesis state

x/ccv/consumer/types/genesis_test.go
```diff
@@ -86,95 +85,95 @@ func TestValidateInitialGenesisState(t *testing.T) {
...
-				cs,
-				consensusState,
-				nil,
-				valUpdates,
-				nil,
-				nil,
-				ccv.ConsumerPacketDataList{},
-				types.LastTransmissionBlockHeight{},
-				false,
+				Params:                      params,
```
smarshall-spitzbart, line 88: I was required to add keys to struct fields in this file to appease linter

smarshall-spitzbart:
@mpoke @MSalopek does a change like this warrant increment the protobuf version number from v1? See proto package names like `interchain_security.ccv.consumer.v1`. Imo we should increment the version as this is a meaningful change to how protos are defined. The only possible issue that could arise is if the pb version is somehow serialized to metadata sent over the wire.

mpoke (changes requested):
Nice work @smarshall-spitzbart. I really like the idea of this. See my comments bellow.

x/ccv/provider/keeper/genesis.go
```diff
@@ -158,7 +158,7 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 		k.GetAllValsetUpdateBlockHeights(ctx),
 		consumerStates,
 		k.GetAllUnbondingOps(ctx),
-		&ccv.MaturedUnbondingOps{Ids: k.GetMaturedUnbondingOps(ctx)},
+		&types.MaturedUnbondingOps{Ids: k.GetMaturedUnbondingOps(ctx)},
```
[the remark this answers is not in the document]
smarshall-spitzbart: I fully agree that we should use the `ccvtypes` alias everywhere, instead of `types` or `ccv`. However I don't think that sort of change is appropriate for this PR as we're already at 4k lines of changes. 

Making the alias change in this file would add ~14 lines from the look of it, but other files have even more refs to the undesirable aliases

proto/interchain_security/ccv/v1/shared_consumer.proto
smarshall-spitzbart: I see where you're coming from, but from a pure dependency perspective, the consumer's genesis state struct is referenced by both the consumer and provider modules. Any type that's a field in that struct must also be referenced by both modules, even if nil. 

There's some ways we could untangle these cross dependancies, like creating a new struct that'd be spit out of `MakeConsumerGenesis` for the provider, and making that a field in the consumer's true GenesisState struct. But there's complexities in how the consumer genesis JSON would be constructed. 

This PR is mostly intended to illuminate dep issues rather than fix them all, and it should stay as a true refactor

x/ccv/consumer/ibc_module_test.go
```diff
@@ -262,7 +261,7 @@ func TestOnChanOpenAck(t *testing.T) {
 		{
 			"invalid: mismatched serialized version",
 			func(keeper *consumerkeeper.Keeper, params *params, mocks testkeeper.MockedKeepers) {
-				md := providertypes.HandshakeMetadata{
+				md := ccv.HandshakeMetadata{
```
insumity, line 264: Why do we have `ccv "github.com/cosmos/interchain-security/v3/x/ccv/types"` here and  not simply`"github.com/cosmos/interchain-security/v3/x/ccv/types"` as in `ibc_module.test.go`. Does this lead to inconsistencies between `ibc_module.go` and `ibc_module_test.go`, e.g., we have `ccv.HandshakeMetadata` here instead of `types.HandshakeMetadata` in `ibc_module.go`?

x/ccv/consumer/keeper/distribution_test.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	testkeeper "github.com/cosmos/interchain-security/v3/testutil/keeper"
 	"github.com/cosmos/interchain-security/v3/x/ccv/consumer/types"
+	ccvtypes "github.com/cosmos/interchain-security/v3/x/ccv/types"
```
insumity, line 15: Similarly to my previous [comment](https://github.com/cosmos/interchain-security/pull/1071/files#r1299958690) there seems the naming is consistent between `distribution.go` and `distribution_test.go`.

proto/interchain_security/ccv/v1/wire.proto
```diff
@@ -88,6 +77,12 @@ enum ConsumerPacketDataType {
       [ (gogoproto.enumvalue_customname) = "VscMaturedPacket" ];
 }
 
+// Note this type is used during IBC handshake methods for both the consumer and provider
```
insumity, line 80: nit: slight inconsistency in the usage of `Note` in the same file, e.g., below it's ` // NOTE: ccv.InfractionType to ...`.

proto/interchain_security/ccv/v1/wire.proto
```diff
@@ -88,6 +77,12 @@ enum ConsumerPacketDataType {
...
+message HandshakeMetadata {
+  string provider_fee_pool_addr = 1;
+  string version = 2;
+}
```
insumity, line 84: Is my understanding correct that the reward distribution is performed through a `MsgTransfer` [message](https://github.com/cosmos/interchain-security/blob/cca008d856e3ffc60ec1a486871d0faa702abe26/x/ccv/consumer/keeper/distribution.go#L126) and hence why we do not see something reward-related in this file? If yes, it might make sense to add a comment here to point to this.

proto/interchain_security/ccv/v1/shared_consumer.proto
```diff
@@ -0,0 +1,156 @@
...
+import "tendermint/abci/types.proto";
+import "ibc/lightclients/tendermint/v1/tendermint.proto";
+import "ibc/core/channel/v1/channel.proto";
+import "google/protobuf/duration.proto";
+import "gogoproto/gogo.proto";
+import "interchain_security/ccv/v1/wire.proto";
+import "google/protobuf/timestamp.proto";
+
+//
+// Note any type defined in this file is referenced/persisted in both the consumer and provider CCV modules,
```
insumity, line 16: nit: reading this `referenced/persisted in both the consumer and provider CCV modules,` I would have expected the file to be called `shared.proto`. My understanding is the file is called `shared**_consumer**.proto` because it contains _only_ consumer-related types used by the consumer but also referenced by the provider. If that's the case, might make sense to clarify in the comment.

x/ccv/consumer/ibc_module_test.go
smarshall-spitzbart: See https://github.com/cosmos/interchain-security/pull/1071#discussion_r1296133443, the type aliases in this repo are very inconsistent and this does lead to inconsistencies, but I don't think it's worth fixing all in this PR

x/ccv/consumer/keeper/distribution_test.go
smarshall-spitzbart: See above

proto/interchain_security/ccv/v1/wire.proto
smarshall-spitzbart: https://github.com/cosmos/interchain-security/pull/1071/commits/1674152eb664d5e0319ce6ba3ea820c54edae088

proto/interchain_security/ccv/v1/wire.proto
smarshall-spitzbart: Yea consumer rewards are sent through a distribution channel, whereas wire.proto is for schemas sent over a ccv channel, [this comment](https://github.com/cosmos/interchain-security/commit/124803a95e5246bba5134baaf2378411a5b0053e) can hopefully help

proto/interchain_security/ccv/v1/shared_consumer.proto
smarshall-spitzbart: In this scenario, the file contains only consumer-related types used by the consumer but also referenced by the provider. This happens to also be the same set of types that are more generally shared between consumer/provider.

I can add a comment with this sort of this thing, or maybe it's worth clarifying once it contains only consumer-related types used by the consumer but also referenced by the provider is completed

insumity: 👍 So, I guess if this file would radically change as part of https://github.com/cosmos/interchain-security/issues/1214, then we can skip any changes here.

x/ccv/consumer/ibc_module_test.go
insumity: 👍 I agree. We can create another PR for fixing naming inconsistencies across _all_ the files, just so we don't forget.

x/ccv/consumer/keeper/params.go
```diff
@@ -6,8 +6,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
-	"github.com/cosmos/interchain-security/v3/x/ccv/consumer/types"
-	ccvtypes "github.com/cosmos/interchain-security/v3/x/ccv/types"
+	types "github.com/cosmos/interchain-security/v3/x/ccv/types"
```
insumity, line 9: Related to the [comment](https://github.com/cosmos/interchain-security/pull/1071/files#r1300611106). I guess in these cases we can use `ccvtypes` even in this PR since the name is already changed?

x/ccv/consumer/types/genesis_test.go
```diff
@@ -16,8 +16,7 @@ import (
 	tmtypes "github.com/cometbft/cometbft/types"
 
 	"github.com/cosmos/interchain-security/v3/testutil/crypto"
-	"github.com/cosmos/interchain-security/v3/x/ccv/consumer/types"
-	ccv "github.com/cosmos/interchain-security/v3/x/ccv/types"
+	types "github.com/cosmos/interchain-security/v3/x/ccv/types"
```
insumity, line 19: Related to the [comment](https://github.com/cosmos/interchain-security/pull/1071/files#r1300611106). I guess in these cases we can use `ccvtypes` even in this PR since the name is already changed?

insumity (approved):
I left some additional nit comment, but in general looks good to me, so approving.
Thanks for working on this Shawn!

x/ccv/consumer/keeper/params.go
smarshall-spitzbart: Well the code that was previously in "github.com/cosmos/interchain-security/v3/x/ccv/consumer/types" is now in "github.com/cosmos/interchain-security/v3/x/ccv/types", so originally using the `types` alias here avoids a big diff. But I'll make the change since we're already talking about it https://github.com/cosmos/interchain-security/commit/065c2cd3ad0581aae39c8acd2ef0016ea6a44e8c

mpoke (approved):

Issue #612, opened 2022-12-19
InitTimeoutTimestamps and VscSendTimestamps not part of provider genesis state


mpoke:
# Problem

`InitTimeoutTimestamps` and `VscSendTimestamps` are not part of the provider's exported state.

# Closing criteria

Export / import `InitTimeoutTimestamps` and `VscSendTimestamps`.

# TODOs
- [ ] Labels have been added for issue
- [ ] Issue has been added to the ICS project

bolton00:
visit the official [COSMOS](https://cosmo-official.org/) page to report and review this error via [LIVE CHAT](https://cosmo-official.org/)

yaruwangway:
`InitTimeoutTimestamps` and `VscSendTimestamps` were not defined in [genesis proto](https://github.com/cosmos/interchain-security/blob/main/proto/interchain_security/ccv/provider/v1/genesis.proto). since they are already in store, can just direct add them into exported genesis, this is not state breaking ? 

another question is about VscSendTimestamps, all the  all VscId  will be dumped to genesis, will that be too much data ?

mpoke:
> InitTimeoutTimestamps and VscSendTimestamps were not defined in [genesis proto](https://github.com/cosmos/interchain-security/blob/main/proto/interchain_security/ccv/provider/v1/genesis.proto). since they are already in store, can just direct add them into exported genesis, this is not state breaking ?

Yeah, it shouldn't be. 

> another question is about VscSendTimestamps, all the VscId will be dumped to genesis, will that be too much data ?

It will be a lot of data, but export genesis is a very rare operation. Also dumping the entire state of the staking module is expensive (millions of delegations).

PR #1076, opened 2023-06-23
Fix:  export InitTimeoutTimestamps and VscSendTimestamps  to genesis

yaruwangway:
## Description

Closes: #612

 `InitTimeoutTimestamps` and `ExportedVscSendTimestamps` are exported to genesis.
`ExportedVscSendTimestamps` is `VscSendTimestamps` with chainID.
```go
type ExportedVscSendTimestamp struct {
	ChainId          string          
	VscSendTimestamp VscSendTimestamp 
}
```

exported Genesis:
```
 "provider": {
      "consumer_addition_proposals": [],
      "consumer_addrs_to_prune": [],
      "consumer_removal_proposals": [],
      "consumer_states": [],
      "exported_vsc_send_timestamps": [],
      "init_timeout_timestamps": [],
      "mature_unbonding_ops": {
        "ids": []
      },
...
}
```

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -216,8 +216,9 @@ message InitTimeoutTimestamp {
 }
 
 message VscSendTimestamp {
-  uint64 vsc_id = 1;
-  google.protobuf.Timestamp timestamp = 2
+  string chain_id = 1;
```
yaruwangway, line 219: a new field ChainId is added, so that when exporting VscSendTimestamp to genesis, the chainID info is carried.

MSalopek: Changing the order of fields inside a proto message is state breaking since this proto is used for storing state.

If you are adding a `chain_id` it should be done by adding it to the end (so it becomes `string chain_id = 3`).

I would advise against changing that structure, and instead change `GenesisState` of the provider like this:
```proto
type GenesisState struct {
	// strictly positive and set to 1 (DefaultValsetUpdateID) for a new chain
	ValsetUpdateId uint64 `protobuf:"varint,1,opt,name=valset_update_id,json=valsetUpdateId,proto3" json:"valset_update_id,omitempty"`
	// empty for a new chain
	ConsumerStates []ConsumerState `protobuf:"bytes,2,rep,name=consumer_states,json=consumerStates,proto3" json:"consumer_states" yaml:"consumer_states"`
	// empty for a new chain
	UnbondingOps []UnbondingOp `protobuf:"bytes,3,rep,name=unbonding_ops,json=unbondingOps,proto3" json:"unbonding_ops"`
	// empty for a new chain
	MatureUnbondingOps *types.MaturedUnbondingOps `protobuf:"bytes,4,opt,name=mature_unbonding_ops,json=matureUnbondingOps,proto3" json:"mature_unbonding_ops,omitempty"`
	// empty for a new chain
	ValsetUpdateIdToHeight []ValsetUpdateIdToHeight `protobuf:"bytes,5,rep,name=valset_update_id_to_height,json=valsetUpdateIdToHeight,proto3" json:"valset_update_id_to_height"`
	// empty for a new chain
	ConsumerAdditionProposals []ConsumerAdditionProposal `protobuf:"bytes,6,rep,name=consumer_addition_proposals,json=consumerAdditionProposals,proto3" json:"consumer_addition_proposals"`
	// empty for a new chain
	ConsumerRemovalProposals []ConsumerRemovalProposal `protobuf:"bytes,7,rep,name=consumer_removal_proposals,json=consumerRemovalProposals,proto3" json:"consumer_removal_proposals"`
	Params                   Params                    `protobuf:"bytes,8,opt,name=params,proto3" json:"params"`
	// empty for a new chain
	ValidatorConsumerPubkeys []ValidatorConsumerPubKey `protobuf:"bytes,9,rep,name=validator_consumer_pubkeys,json=validatorConsumerPubkeys,proto3" json:"validator_consumer_pubkeys"`
	// empty for a new chain
	ValidatorsByConsumerAddr []ValidatorByConsumerAddr `protobuf:"bytes,10,rep,name=validators_by_consumer_addr,json=validatorsByConsumerAddr,proto3" json:"validators_by_consumer_addr"`
	// empty for a new chain
	ConsumerAddrsToPrune []ConsumerAddrsToPrune `protobuf:"bytes,11,rep,name=consumer_addrs_to_prune,json=consumerAddrsToPrune,proto3" json:"consumer_addrs_to_prune"`
        InitTimeoutTImestamps ... // use the correct type
        VscSendTimestamps ... // use the correct type
}
```

That way we don't introduce state breaking changes, but have in mind that you would need to change the provider's Export genesis

```golang
// x/ccv/provider/keeper
func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {...}
```

yaruwangway: Hi, this way, we lost the chainID info for `InitTimeoutTImestamps` ?
how about create a new struct to wrap the  original `InitTimeoutTImestamps` with chainID for the genesis exporting.

mpoke: Yeah, unless it's necessary, try not to change the original structures as it will require state migration.

MSalopek: We don't lose it, it's stored as part of the key in the state.

You can have a structure like this when exporting/importing:
```
type GenesisState struct {
        InitTimeoutTImestamps ... // use the correct type
        VscSendTimestamps []ExportedVscSendTimestamps
```

Where `ExportedVscSendTimestamps` looks like:
```golang
struct ExportedVscSendTimestamps {
    ChainId string
    Timestamps []VscSendTimestamps
}
```

x/ccv/provider/keeper/genesis.go
```diff
@@ -90,6 +90,14 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
 		}
 	}
 
+	for _, item := range genState.InitTimeoutTimestamps {
+		k.SetInitTimeoutTimestamp(ctx, item.ChainId, item.Timestamp)
+	}
+
+	for _, item := range genState.ExportedVscSendTimestamps {
+		k.SetVscSendTimestamp(ctx, item.ChainId, item.VscSendTimestamp.VscId, item.VscSendTimestamp.Timestamp)
```
yaruwangway, line 98: no need to be sorted before write to store ?

x/ccv/provider/types/genesis.go
```diff
@@ -35,6 +37,8 @@ func NewGenesisState(
 		ValidatorConsumerPubkeys:  validatorConsumerPubkeys,
 		ValidatorsByConsumerAddr:  validatorsByConsumerAddr,
 		ConsumerAddrsToPrune:      consumerAddrsToPrune,
+		InitTimeoutTimestamps:     initTimeoutTimestamps,
+		ExportedVscSendTimestamps: exportedVscSendTimestamps,
```
yaruwangway, line 41: exported genesis:
```
 "provider": {
      "consumer_addition_proposals": [],
      "consumer_addrs_to_prune": [],
      "consumer_removal_proposals": [],
      "consumer_states": [],
      "exported_vsc_send_timestamps": [],
      "init_timeout_timestamps": [],
      "mature_unbonding_ops": {
        "ids": []
      },
...
```

x/ccv/provider/keeper/genesis.go
MSalopek: According to this code - no, they don't need to be sorted because they are never accessed in such a way where the ordering matters:
* https://github.com/cosmos/interchain-security/blob/07302ffb6692fe849f5f5d393de5db6395e74a53/x/ccv/provider/keeper/keeper.go#L905

proto/interchain_security/ccv/consumer/v1/genesis.proto
```diff
@@ -55,4 +55,4 @@ message HeightToValsetUpdateID {
 
 // OutstandingDowntime defines the genesis information for each validator
 // flagged with an outstanding downtime slashing.
-message OutstandingDowntime { string validator_consensus_address = 1; }
\ No newline at end of file
+message OutstandingDowntime { string validator_consensus_address = 1; }
```
smarshall-spitzbart, line 58: What changed here?

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -221,6 +221,13 @@ message VscSendTimestamp {
       [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
 }
 
+// ExportedVscSendTimestamps is VscSendTimestamp with chainID info for exporting to genesis
+message ExportedVscSendTimestamp {
```
smarshall-spitzbart, line 225: What's the reasoning behind introducing this new message instead of adding a chain_id field to the already existing VscSendTimestamp?

smarshall-spitzbart (commented):
Tests look solid 👍 imo it's cleaner to not introduce a new type, and add a chain ID field to `VscSendTimestamp` for import/export. Note I don't believe this would require migrations due to the fact that we only store the timestamp field in value bytes (see `SetVscSendTimestamp`). ChainID is already stored in key bytes.

proto/interchain_security/ccv/provider/v1/provider.proto
MSalopek: See this thread:
https://github.com/cosmos/interchain-security/pull/1076#discussion_r1239689182

proto/interchain_security/ccv/consumer/v1/genesis.proto
yaruwangway: nothing, prob my editor just formatting it.

proto/interchain_security/ccv/provider/v1/provider.proto
smarshall-spitzbart: > Changing the order of fields inside a proto message is state breaking since this proto is used for storing state.

This is usually true, but in this case the proto unmarshal/marshal methods are not called, and the only place the `VscSendTimestamp` type is used is in `GetAllVscSendTimestamps`. Considering this, imo the best design is to simply add a chain ID field to the `VscSendTimestamp` type, and adjust the related keeper methods. 

This wouldn't be state breaking and would prevent us from adding another pb type just for exporting.

yaruwangway: but this is state breaking ? if we use a new struct to wrap `VscSendTimestamp` and add chainID, this is not state-breaking.

yaruwangway: yeah, the chainID is in store, we want to export the `VscSendTimestamp` plus `chainID` to genesis. this newly defined
```
struct ExportedVscSendTimestamps {
    ChainId string
    Timestamps []VscSendTimestamps
}

smarshall-spitzbart: `If i dump VscSendTimestamp to genesis, there is not chainID info in the genesis.`

Correct, I'm suggesting to add a chainID field to `VscSendTimestamp` instead of creating a new `ExportedVscSendTimestamps` type. This would not be state breaking. 

Overall if @mpoke and @MSalopek suggested the other way of doing this, that's fine with me and would still work

yaruwangway: adding chainID to `VscSendTimestamp` is api breaking ?

anway, can you review this PR ? Thank you!

MSalopek: If possible, let's not reimplement this.

I'm suggesting to keep the changes already made. The benefit of the refactor is minimal.

smarshall-spitzbart: > adding chainID to VscSendTimestamp is api breaking ?

If API breaking means that the genesis JSON schema would change, then yes. That's not a problem tho

Re @MSalopek's comment, imo for tech debt reasons we should avoid adding shims/wrapper types unless they're absolutely needed to avoid a state migration. In this case there is no state migration to avoid.

I'll approve as to not slow everyone down, but since this is a `good first issue`, I just wanna make sure we understand the mechanics behind things being state breaking

smarshall-spitzbart (approved):

proto/interchain_security/ccv/provider/v1/provider.proto
yaruwangway: im just not sure about the impact of the api breaking change, so please let me know, i could change this PR if adding chainID to `VscSendTimestamp ` is better.

smarshall-spitzbart: People use "API breaking" in different ways, so I can't comment on impact of an api breaking change. But I can say it's ok to change the JSON schema for import/export genesis, and I believe any solution #612 would require we change that JSON schema

@MSalopek @mpoke lmk if I'm incorrect here

Re best solution for the PR, I'll leave that up to the other approver, as this thread has gotten long

MSalopek:
@yaruwangway Can you please resolve the conflicts and we can go over this once more?

MSalopek (approved):
Thanks!

All the comments are valid but would require re-writing the entire PR.

Since these changes are not breaking anything, I'm in favor of keeping them and moving on.

Issue #688, opened 2023-01-30
Confusing obsolete comments before setting EndBlockers order


ivan-gavran:
The [comment](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/app/provider/app.go#L555-L555) before setting  the order of EndBlockers is obsolete:
```
	//	- creating a new consumer chain requires the following order,
	//	CreateChildClient(), staking.EndBlock, provider.EndBlock;
	//	thus, gov.EndBlock must be executed before staking.EndBlock
```
A consumer client is actually created in the BeginBlock (and the `gov` module only puts proposals in the list of pending proposals)

PR #1081, opened 2023-06-26
docs: remove an obsolete comment before setting the order of EndBlockers

insumity:
This is work in progress. Please, do **not** review yet.

## Description

Closes: [#688](https://github.com/cosmos/interchain-security/issues/688)

Remove obsolete comment before setting the `EndBlockers` in [app.go](https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/app/provider/app.go#L555-L555).

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/app/provider/app.go
+++ b/app/provider/app.go
@@ -530,12 +530,7 @@ func New(
 		vestingtypes.ModuleName,
 		providertypes.ModuleName,
 	)
-	// Interchain Security requirements
-	// 	- provider.EndBlock gets validator updates from the staking module;
-	// 	thus, staking.EndBlock must be executed before provider.EndBlock;
-	//	- creating a new consumer chain requires the following order,
-	//	CreateChildClient(), staking.EndBlock, provider.EndBlock;
-	//	thus, gov.EndBlock must be executed before staking.EndBlock
+
 	app.MM.SetOrderEndBlockers(
 		crisistypes.ModuleName,
 		govtypes.ModuleName,


insumity:
As already mentioned by @ivan-gavran in [issue 688](https://github.com/cosmos/interchain-security/issues/688), the client is created in [BeginBlock](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/proposal.go#L372), so the existing comment does not really justify the order of `EndBlock`s.

Nevertheless, it remains true that:
> [...] gov.EndBlock must be executed before staking.EndBlock

because the `staking` module uses parameters that might be updated through government proposals. For instance, see [EndBlocker](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/abci.go#L23) that calls [BlockValidatorUpdates](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/val_state_change.go#L20), that calls [ApplyAndReturnValidatorSetUpdates](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/val_state_change.go#L30), that [reads](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/val_state_change.go#L134) the `MaxValidators` parameter.

app/provider/app.go
```diff
@@ -533,9 +533,8 @@ func New(
 	// Interchain Security requirements
 	// 	- provider.EndBlock gets validator updates from the staking module;
 	// 	thus, staking.EndBlock must be executed before provider.EndBlock;
-	//	- creating a new consumer chain requires the following order,
-	//	CreateChildClient(), staking.EndBlock, provider.EndBlock;
-	//	thus, gov.EndBlock must be executed before staking.EndBlock
+	// 	- staking.EndBlock gets parameter updates (e.g., `MaxValidators`) from the gov module;
+	// 	thus, gov.EndBlock must be executed before staking.EndBlock;
```
mpoke, line 537: Sure, but this is not a requirement of ICS, so it doesn't need to be added to "Interchain Security requirements".

mpoke (changes requested):

app/provider/app.go
insumity: That's true, it's not an _interchain security_ requirement. However, I believe the same could be said in regards to the first comment:
```
	// 	- provider.EndBlock gets validator updates from the staking module;
	// 	thus, staking.EndBlock must be executed before provider.EndBlock;
```
The order of the `staking.EndBlock` and `provider.EndBlock` could be reversed. In such a case, this would delay the creation of a potential `VSCPacket` by one block. For example, let's say we first execute `provider.EndBlock` and then `staking.EndBlock` and let's assume we have a block `A` at height `h`. Additionally, assume that block `A` contains unbonding operations that would lead to `ValidatorUpdate`s. If we execute `provider.EndBlock` before `staking.EndBlock`, then the _provider_ module won't see the validator updates (because those changes stem from the _staking_ module) and hence won't send a `VSCPacket`. However, in the next block at height `h + 1`, the `provider.EndBlock` would be able to pick up the `ValidatorUpdate`s from block `A` and send the corresponding `VSCPacket`.

In a similar fashion, if we do not execute the `gov.EndBlock` before the `staking.EndBlock` we would end up delaying the sending of a `VSCPacket` in the producer chain by one block. This could, for example, happen if `gov.EndBlock` introduces additional validators through an increase of `MaxValidators`.

I don't have strong feelings on this. We can remove the whole comment block altogether.

mpoke: > This could, for example, happen if gov.EndBlock introduces additional validators through an increase of MaxValidators.

Yeah, but like this, the provider valset would not be updated either. The idea is to keep the provider valset "in sync" with the consumer valset. Basically, this means that a VSCPacket is sent in every block when the valset is updated. Probably it's fine to delay with one block, but we'd need to look into that. I don't think it's worth the trouble. I would just remove the entire comment block.

mpoke (approved):

yaruwangway (approved):

Issue #720, opened 2023-02-07
Distribution Total will always be empty in emitted events


AdityaSripal:
fpTokens are retrieved to emit events after all funds are already sent out, so it should always be empty

https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/consumer/keeper/distribution.go#L21

This function sends all funds out of the FeeCollector ModuleAccount:

https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/consumer/keeper/distribution.go#L53

This then gets called after which should return empty coins:

https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/consumer/keeper/distribution.go#L131

Thus this event is always going to just emit empty coins on the DistributionTotal attribute:

https://github.com/cosmos/interchain-security/blob/463ec20c87423897009bd78c2dbf96d28bd8f150/x/ccv/consumer/keeper/distribution.go#L145

PR #1097, opened 2023-06-28
fix: `AttributeDistributionTotal` in event emit

yaruwangway:
## Description

Closes: #720 

fix the attribute `AttributeDistributionTotal` value in `FeeDistribution` event emit. `AttributeDistributionTotal` was always empty before.

Fix

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/consumer/keeper/distribution.go
```diff
@@ -153,7 +152,7 @@ func (k Keeper) SendRewardsToProvider(ctx sdk.Context) error {
 				sdk.NewAttribute(ccv.AttributeDistributionCurrentHeight, strconv.Itoa(int(currentHeight))),
 				sdk.NewAttribute(ccv.AttributeDistributionNextHeight, strconv.Itoa(int(currentHeight+k.GetBlocksPerDistributionTransmission(ctx)))),
 				sdk.NewAttribute(ccv.AttributeDistributionFraction, (k.GetConsumerRedistributionFrac(ctx))),
-				sdk.NewAttribute(ccv.AttributeDistributionTotal, fpTokens.String()),
+				sdk.NewAttribute(ccv.AttributeDistributionTotal, allBalances.String()),
```
smarshall-spitzbart, line 155: would something like `AttributeTotalFeePool` be a better name here? Likely not important tho

smarshall-spitzbart (approved):
Good stuff!

MSalopek (approved):
LGTM!

Backporting to `v2` and `v3` using labels.

x/ccv/consumer/keeper/distribution.go
yaruwangway: i can change it here, it is a var name, so not api breaking right?

smarshall-spitzbart: I'm not sure what API breaking would mean in this context, if you change the value string then the event will change (which seems ok? I don't think changing an event is state breaking but I could be wrong), if you just change the var naming then it's a pure refactor. Again not super important, so feel free to ignore my comment if we're not confident how the change would affect things

Issue #728, opened 2023-02-09
SlashAcks should not be persisted as string


smarshall-spitzbart:
# Problem

A "slash ack" is a string which is relayed between provider and consumer to reset the consumer's outstanding downtime flag. SlashAcks are *always* consumer consensus addresses, but we persist and send these slash acks as a string type. This design is error prone because:
1. It does not leverage type safety (see https://github.com/cosmos/interchain-security/issues/693)
2. There are multiple formats to convert a consensus address to string

# Closing criteria

Make slash acks persisted and ser/deser as pb types, not strings

# Problem details

A fix to this issue should be built on top of https://github.com/cosmos/interchain-security/pull/725. Since slash acks can be persisted as consumer cons address pb types

smarshall-spitzbart:
Note: this will likely take a state migration so I'm going push it off for a future ICS version

smarshall-spitzbart:
@yaruwangway, now that #725 is merged, this issue is able to be solved if you're interested! Lmk if you want more options for a first issue

ameya-deshmukh:
Hi @smarshall-spitzbart is this still available?

smarshall-spitzbart:
> Hi @smarshall-spitzbart is this still available?

@ameya-deshmukh sorry for the late response! Yes this issue is still open and noone is currently working on it. Are you interested in contributing?

MSalopek:
The issue has become state breaking in the meantime.
Any changes in those parts will require a state migration, I'm not sure how feasible this is.

smarshall-spitzbart:
A solution doesn't have to be state breaking, but it would naturally have to break the wire. Solution could be implemented in such a way that it is backwards compatible tho.

Now that ICS is in production I agree that this issue is most likely not feasible. Unfortunately it is a source of technical debt.

@yaruwangway it seems we could close this issue by simply making the API for `AppendSlashAck`, `GetSlashAcks`, `SetSlashAcks` type safe, and remove the TODOs. The wire can keep string types to avoid complications

PR #1163, opened 2023-07-20
Fix!: SlashAcks api change

yaruwangway:
## Description

Closes: #728 

Remove invalid todos.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -723,7 +723,6 @@ func (k Keeper) DeleteValsetUpdateBlockHeight(ctx sdk.Context, valsetUpdateId ui
 
 // SetSlashAcks sets the slash acks under the given chain ID
 //
-// TODO: SlashAcks should be persisted as a list of ConsumerConsAddr types, not strings.
 // See https://github.com/cosmos/interchain-security/issues/728
 func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []string) {
 	store := ctx.KVStore(k.storeKey)
@@ -742,7 +741,6 @@ func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []string) {
 
 // GetSlashAcks returns the slash acks stored under the given chain ID
 //
-// TODO: SlashAcks should be persisted as a list of ConsumerConsAddr types, not strings.
 // See https://github.com/cosmos/interchain-security/issues/728
 func (k Keeper) GetSlashAcks(ctx sdk.Context, chainID string) []string {
 	store := ctx.KVStore(k.storeKey)
@@ -779,7 +777,7 @@ func (k Keeper) DeleteSlashAcks(ctx sdk.Context, chainID string) {
 
 // AppendSlashAck appends the given slash ack to the given chain ID slash acks in store
 func (k Keeper) AppendSlashAck(ctx sdk.Context, chainID,
-	ack string, // TODO: consumer cons addr should be accepted here, see https://github.com/cosmos/interchain-security/issues/728
+	ack string,
 ) {
 	acks := k.GetSlashAcks(ctx, chainID)
 	acks = append(acks, ack)
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -722,7 +722,7 @@ func (k Keeper) DeleteValsetUpdateBlockHeight(ctx sdk.Context, valsetUpdateId ui
 }
 
 // SetSlashAcks sets the slash acks under the given chain ID
-func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []string) {
+func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []types.ConsumerConsAddress) {
 	store := ctx.KVStore(k.storeKey)
 
 	sa := types.SlashAcks{
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -765,7 +765,7 @@ func (k Keeper) ConsumeSlashAcks(ctx sdk.Context, chainID string) (acks []types.
 		return
 	}
 
-	acks, err := StrToConsumerConsAddresses(addresses)
+	acks, err := types.StrsToConsumerConsAddresses(addresses)
 	if err != nil {
 		// todo
 	}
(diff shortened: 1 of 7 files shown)


x/ccv/provider/keeper/keeper.go
```diff
@@ -779,7 +773,7 @@ func (k Keeper) DeleteSlashAcks(ctx sdk.Context, chainID string) {
 
 // AppendSlashAck appends the given slash ack to the given chain ID slash acks in store
 func (k Keeper) AppendSlashAck(ctx sdk.Context, chainID,
-	ack string, // TODO: consumer cons addr should be accepted here, see https://github.com/cosmos/interchain-security/issues/728
+	ack string,
```
smarshall-spitzbart, line 776: See https://github.com/cosmos/interchain-security/issues/728#issuecomment-1620706001 and comment that's being deleted
```suggestion
	ack types.ConsumerConsAddress, 
```

x/ccv/provider/keeper/keeper.go
```diff
@@ -722,9 +722,6 @@ func (k Keeper) DeleteValsetUpdateBlockHeight(ctx sdk.Context, valsetUpdateId ui
 }
 
 // SetSlashAcks sets the slash acks under the given chain ID
-//
-// TODO: SlashAcks should be persisted as a list of ConsumerConsAddr types, not strings.
-// See https://github.com/cosmos/interchain-security/issues/728
 func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []string) {
```
smarshall-spitzbart, line 725: See https://github.com/cosmos/interchain-security/issues/728#issuecomment-1620706001, we need to change the API but keep the serialized wire bytes the same
```suggestion
func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []types.ConsumerConsAddress) {
```

yaruwangway:
the api change will convert will deal with the covertion  `consumerConsAddr` <-> `consumerConsAddr.String()`
https://github.com/cosmos/interchain-security/blob/7484dbceb66ceafad05d34d46c6b4e4f145cb022/x/ccv/provider/keeper/relay.go#L456

x/ccv/provider/keeper/keeper.go
```diff
@@ -757,11 +755,21 @@ func (k Keeper) GetSlashAcks(ctx sdk.Context, chainID string) []string {
 		panic(fmt.Errorf("failed to unmarshal SlashAcks: %w", err))
 	}
 
-	return acks.GetAddresses()
+	slashAcks := make([]types.ConsumerConsAddress, len(acks.Addresses))
+	for _, ack := range acks.Addresses {
+		// the reverse of ConsumerConsAddress.String()
+		consAddr, err := sdk.ConsAddressFromBech32(ack)
```
yaruwangway, line 761: not very sure here is correct.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1066,3 +1075,17 @@ func (k Keeper) GetSlashLog(
 func (k Keeper) BondDenom(ctx sdk.Context) string {
 	return k.stakingKeeper.BondDenom(ctx)
 }
+
+// helper func
+func StrToConsumerConsAddresses(addresses []string) ([]types.ConsumerConsAddress, error) {
+	consAddresses := []types.ConsumerConsAddress{}
+	for _, address := range addresses {
+		// reverse of ConsumerConsAddress.String()
+		consAddr, err := sdk.ConsAddressFromBech32(address)
```
yaruwangway, line 1084: if someone can help confirm`sdk.ConsAddressFromBech32()` is the reverse of `ConsumerConsAddress.String()` ?

x/ccv/provider/keeper/keeper.go
```diff
@@ -761,12 +759,19 @@ func (k Keeper) GetSlashAcks(ctx sdk.Context, chainID string) []string {
...
-	if len(acks) < 1 {
+func (k Keeper) ConsumeSlashAcks(ctx sdk.Context, chainID string) (acks []types.ConsumerConsAddress) {
+	addresses := k.GetSlashAcks(ctx, chainID)
+	if len(addresses) < 1 {
 		return
 	}
+
+	acks, err := StrToConsumerConsAddresses(addresses)
+	if err != nil {
+		// todo
```
yaruwangway, line 770: shall return err ? if yes, where to deal with the err ? @smarshall-spitzbart

x/ccv/provider/keeper/keeper.go
```diff
@@ -761,12 +759,19 @@ func (k Keeper) GetSlashAcks(ctx sdk.Context, chainID string) []string {
 }
 
 // ConsumeSlashAcks empties and returns the slash acks for a given chain ID
-func (k Keeper) ConsumeSlashAcks(ctx sdk.Context, chainID string) (acks []string) {
-	acks = k.GetSlashAcks(ctx, chainID)
-	if len(acks) < 1 {
+func (k Keeper) ConsumeSlashAcks(ctx sdk.Context, chainID string) (acks []types.ConsumerConsAddress) {
```
smarshall-spitzbart, line 762: ConsumeSlashAcks is only used in one place in `relay.go` which accepts `[]string` anyways. Seems reasonable to return strings here to reduce complexity

x/ccv/provider/keeper/keeper.go
```diff
@@ -722,14 +722,15 @@ func (k Keeper) DeleteValsetUpdateBlockHeight(ctx sdk.Context, valsetUpdateId ui
 }
 
 // SetSlashAcks sets the slash acks under the given chain ID
-//
-// TODO: SlashAcks should be persisted as a list of ConsumerConsAddr types, not strings.
-// See https://github.com/cosmos/interchain-security/issues/728
-func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []string) {
+func (k Keeper) SetSlashAcks(ctx sdk.Context, chainID string, acks []types.ConsumerConsAddress) {
```
smarshall-spitzbart, line 725: probably fine to accept strings here, since it's only effectively called by the appender and genesis

x/ccv/provider/keeper/keeper.go
smarshall-spitzbart: see how consumer handles this https://github.com/cosmos/interchain-security/blob/46251c9184eda7289688e08740dcebe03bd4a50a/x/ccv/consumer/keeper/relay.go#L81. Although I'm not sure that we need to add the complexity of converting these types back and forth so much

smarshall-spitzbart (commented):
I'd say we should do one of the two following solutions:

- Only change API of the append function and keep everything else as strings. This is most simple, avoids state migration, but doesn't add much type safety. 
- Store slash acks as the cons addr type, adjust APIs accordingly, only convert to strings one time when the slash acks are about to be sent over wire. This would require migration as it's state breaking

We should avoid converting slash acks back and forth in-mem as that adds unneeded complexity

x/ccv/provider/keeper/keeper.go
smarshall-spitzbart: I'd say pattern match other places in the repo that panic or return upon error. See above on how consumer does this

smarshall-spitzbart:
Hey these changes have started to look like more work than we thought, lets just close the issue and call it good

Issue #362, opened 2022-09-22
Improve UTs for ibc_module.go


smarshall-spitzbart:
# Problem
Following #342 there will be better unit testing coverage of the IBC module interface implementations. However the following methods remain untested for the consumer AND provider.
- `OnRecvPacket`
- `OnAcknowledgementPacket`
- `OnTimeoutPacket`

# Closing criteria
Unit tests for these methods. Would higher level e2e tests be useful?

insumity:
There's no easy way to test the `OnRecvPacket`, `OnAcknowledgementPacket`, and `OnTimeoutPacket` methods in a _clean_ way as is.
For example, the already tested method [OnChanOpenConfirm](https://github.com/cosmos/interchain-security/blob/f636cfc4793e3518104bab001d1586441724b699/x/ccv/provider/ibc_module.go#L139) is tested in [ibc_module_test.go](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/ibc_module_test.go#L213) by checking different edge cases, e.g., "channel not found," "too many connection hops," etc. Therefore, the test of `OnChanOpenConfirm` is testing the underlying [SetConsumerChain](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/keeper.go#L321) method from the `keeper` package instead, and not the logic of the `OnChanOpenConfirm` in the `provider` package.

The reason testing is difficult for the aforementioned methods is because the methods operate on the underlying `providerKeeper`. For example, the [OnTimeoutPacket](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/ibc_module.go#L288) does:
```
if err := am.keeper.OnTimeoutPacket(ctx, packet); err != nil {
	return err
}
```
and to test this logic we would need to be able to mock the underlying `providerKeeper` (= `am.keeper`). This cannot be accomplished since the keeper is a `struct` and not an `interface` and hence there's no easy way to mock the `providerKeeper`. Although we could potentially change the `providerKeeper` to be an `interface` and hence make it easy to mock `providerKeeper` this would need a huge refactoring since app modules assume that keepers are `struct`s (see [here](https://github.com/cosmos/interchain-security/blob/main/app/provider/app.go#L432)). Such a refactoring would need to extensively modify the [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/) codebase as well.

[...]

PR #1169, opened 2023-07-31
test: introduce tests for OnTimeoutPacket and OnAcknowledgementPacket

insumity:
## Description

Closes: #362 

As described in this [comment](https://github.com/cosmos/interchain-security/issues/362#issuecomment-1678727931) we cannot test those methods in [ibc_module_test.go](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/ibc_module_test.go) so we test in [relay_test.go](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/relay_test.go) instead.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/testutil/keeper/unit_test_helpers.go
+++ b/testutil/keeper/unit_test_helpers.go
@@ -207,7 +207,10 @@ func GetNewVSCMaturedPacketData() types.VSCMaturedPacketData {
 }
 
 // SetupForStoppingConsumerChain registers expected mock calls and corresponding state setup
-// which asserts that a consumer chain was properly stopped from StopConsumerChain().
+// which assert that a consumer chain was properly setup to be later stopped from `StopConsumerChain`.
+// Note: This function only setups and tests that we correctly setup a consumer chain that we could later stop when
+// calling `StopConsumerChain` -- this does NOT necessarily mean that the consumer chain is stopped.
+// Also see `SetupForStoppingConsumerChainChannel` and `TestProviderStateIsCleanedAfterConsumerChainIsStopped`.
 func SetupForStoppingConsumerChain(t *testing.T, ctx sdk.Context,
 	providerKeeper *providerkeeper.Keeper, mocks MockedKeepers,
 ) {
@@ -215,7 +218,6 @@ func SetupForStoppingConsumerChain(t *testing.T, ctx sdk.Context,
 	expectations := GetMocksForCreateConsumerClient(ctx, &mocks,
 		"chainID", clienttypes.NewHeight(4, 5))
 	expectations = append(expectations, GetMocksForSetConsumerChain(ctx, &mocks, "chainID")...)
-	expectations = append(expectations, GetMocksForStopConsumerChain(ctx, &mocks)...)
 
 	gomock.InOrder(expectations...)
 
@@ -224,6 +226,50 @@ func SetupForStoppingConsumerChain(t *testing.T, ctx sdk.Context,
 	require.NoError(t, err)
 	err = providerKeeper.SetConsumerChain(ctx, "channelID")
 	require.NoError(t, err)
+}
+
+// SetupForStoppingConsumerChainChannel registers expected mock calls which assert that the channel to the consumer
+// chain is closed. To be used when we test that `StopConsumerChain` is called with `closeChan` set to `true`.
+func SetupForStoppingConsumerChainChannel(t *testing.T, ctx sdk.Context, mocks MockedKeepers) {
+	t.Helper()
+	gomock.InOrder(GetMocksForStopConsumerChain(ctx, &mocks)...)
+}
+
+// TestProviderStateIsCleanedAfterConsumerChainIsStopped executes test assertions for the provider's state being cleaned
+// after a stopped consumer chain.
+func TestProviderStateIsCleanedAfterConsumerChainIsStopped(t *testing.T, ctx sdk.Context, providerKeeper providerkeeper.Keeper,
+	expectedChainID, expectedChannelID string,
+) {
+	t.Helper()
+	_, found := providerKeeper.GetConsumerClientId(ctx, expectedChainID)
+	require.False(t, found)
+	_, found = providerKeeper.GetChainToChannel(ctx, expectedChainID)
+	require.False(t, found)
+	_, found = providerKeeper.GetChannelToChain(ctx, expectedChannelID)
+	require.False(t, found)
+	_, found = providerKeeper.GetInitChainHeight(ctx, expectedChainID)
+	require.False(t, found)
+	acks := providerKeeper.GetSlashAcks(ctx, expectedChainID)
+	require.Empty(t, acks)
+	_, found = providerKeeper.GetInitTimeoutTimestamp(ctx, expectedChainID)
+	require.False(t, found)
+
+	require.Empty(t, providerKeeper.GetAllVscSendTimestamps(ctx, expectedChainID))
+
+	// test key assignment state is cleaned
+	require.Empty(t, providerKeeper.GetAllValidatorConsumerPubKeys(ctx, &expectedChainID))
+	require.Empty(t, providerKeeper.GetAllValidatorsByConsumerAddr(ctx, &expectedChainID))
+	require.Empty(t, providerKeeper.GetAllKeyAssignmentReplacements(ctx, expectedChainID))
+	require.Empty(t, providerKeeper.GetAllConsumerAddrsToPrune(ctx, expectedChainID))
+
+	allGlobalEntries := providerKeeper.GetAllGlobalSlashEntries(ctx)
+	for _, entry := range allGlobalEntries {
+		require.NotEqual(t, expectedChainID, entry.ConsumerChainID)
+	}
+
+	slashPacketData, vscMaturedPacketData, _, _ := providerKeeper.GetAllThrottledPacketData(ctx, expectedChainID)
+	require.Empty(t, slashPacketData)
+	require.Empty(t, vscMaturedPacketData)
 }
 
 func GetTestConsumerAdditionProp() *providertypes.ConsumerAdditionProposal {
--- a/testutil/keeper/unit_test_helpers.go
+++ b/testutil/keeper/unit_test_helpers.go
@@ -210,7 +210,7 @@ func GetNewVSCMaturedPacketData() types.VSCMaturedPacketData {
 // which assert that a consumer chain was properly setup to be later stopped from `StopConsumerChain`.
 // Note: This function only setups and tests that we correctly setup a consumer chain that we could later stop when
 // calling `StopConsumerChain` -- this does NOT necessarily mean that the consumer chain is stopped.
-// Also see `SetupForStoppingConsumerChainChannel` and `TestProviderStateIsCleanedAfterConsumerChainIsStopped`.
+// Also see `TestProviderStateIsCleanedAfterConsumerChainIsStopped`.
 func SetupForStoppingConsumerChain(t *testing.T, ctx sdk.Context,
 	providerKeeper *providerkeeper.Keeper, mocks MockedKeepers,
 ) {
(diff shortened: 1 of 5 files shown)


testutil/keeper/unit_test_helpers.go
```diff
@@ -209,13 +209,18 @@ func GetNewVSCMaturedPacketData() types.VSCMaturedPacketData {
...
 ) {
 	t.Helper()
 	expectations := GetMocksForCreateConsumerClient(ctx, &mocks,
 		"chainID", clienttypes.NewHeight(4, 5))
 	expectations = append(expectations, GetMocksForSetConsumerChain(ctx, &mocks, "chainID")...)
-	expectations = append(expectations, GetMocksForStopConsumerChain(ctx, &mocks)...)
+
+	// StopConsumerChain receives a `closeChan` parameter. If the `closeChan` argument is true, then
+	// we register additional expected mock calls that capture the fact that the underlying channel is closed.
+	if closeChan {
```
smarshall-spitzbart, line 221: Note that when `closeChan` is false, we're not mocking anything for  `StopConsumerChain`, as the name (SetupForStoppingConsumerChain) suggests. This method becomes a more generic setup method which mocks and calls `CreateConsumerClient` and `SetConsumerChain`.  

Maybe we refactor to make naming more clear about what's going on under the hood.

x/ccv/provider/keeper/relay_test.go
```diff
@@ -715,3 +715,65 @@ func TestSendVSCPacketsToChainFailure(t *testing.T) {
...
+}
+
+// TestOnTimeoutPacketStopsChain tests that the chain is stopped in case of a timeout
+func TestOnTimeoutPacketStopsChain(t *testing.T) {
+	// Keeper setup
+	providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+	providerKeeper.SetParams(ctx, providertypes.DefaultParams())
+
+	testkeeper.SetupForStoppingConsumerChain(t, ctx, &providerKeeper, mocks, false)
```
smarshall-spitzbart, line 740: Since we're calling `SetupForStoppingConsumerChain` s.t. no mocks for `StopConsumerChain` are expected, this test is only asserting that `OnTimeoutPacket` doesn't return an error.

Ie. I could comment out the `StopConsumerChain` call in `OnTimeoutPacket` and this test would still pass. 

Therefore the comment is misleading

> tests that the chain is stopped in case of a timeout

x/ccv/provider/keeper/relay_test.go
```diff
@@ -715,3 +715,65 @@ func TestSendVSCPacketsToChainFailure(t *testing.T) {
...
+	providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+	providerKeeper.SetParams(ctx, providertypes.DefaultParams())
+
+	// test that `OnAcknowledgementPacket` returns an error if the ack contains an error and the channel is unknown
+	ackError := channeltypes.Acknowledgement{Response: &channeltypes.Acknowledgement_Error{Error: "some error"}}
+	err := providerKeeper.OnAcknowledgementPacket(ctx, channeltypes.Packet{}, ackError)
+	require.Error(t, err)
+
+	// test that we stop the consumer chain when `OnAcknowledgementPacket` returns an error and the chain is found
```
smarshall-spitzbart, line 772: Similar comment to above, I could comment out `StopConsumerChain` in `OnAcknowledgementPacket` and this test would still pass. We're not asserting that the consumer chain is being stopped

smarshall-spitzbart (commented):
I like the granularity of the individual tests, these are easy to follow 👍 

I've left some comments regarding intention about what we're actually testing. 

In the past there's been some tests in this repo that make it seem like a method is functioning as expected, but in fact the method had a bug that was hidden by vaguely  intentioned and overly generic tests (example: just checking that a method doesn't return an error, when there's more complexity to be tested under the hood). 

This is where I'm coming from on these seemingly nitpicky comments :)

smarshall-spitzbart:
See https://github.com/cosmos/interchain-security/pull/1207 on failing test

x/ccv/provider/keeper/relay_test.go
```diff
@@ -715,3 +715,65 @@ func TestSendVSCPacketsToChainFailure(t *testing.T) {
...
+}
+
+// TestOnAcknowledgementPacketWithNoAckError tests `OnAcknowledgementPacket` when the underlying ack contains no error
+func TestOnAcknowledgementPacketWithNoAckError(t *testing.T) {
+	// Keeper setup
+	providerKeeper, ctx, ctrl, _ := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+
+	ack := channeltypes.Acknowledgement{Response: &channeltypes.Acknowledgement_Result{Result: []byte{}}}
+	err := providerKeeper.OnAcknowledgementPacket(ctx, channeltypes.Packet{}, ack)
```
p-offtermatt, line 756: Is there any logic we should test for here outside of the NoError?

p-offtermatt (commented):
Thanks for the work!
I agree with Shawns comments, it's hard to tell whether the tests actually test meaningful behaviour. I'd recommend asserting stronger properties in the tests.

It may or may not be helpful to go back to the spec and see what the spec claims about the functionality here, e.g. see https://github.com/cosmos/ibc/blob/jk/ccvTLA/spec/app/ics-028-cross-chain-validation/methods.md#consumer-chain-removal

x/ccv/provider/keeper/relay_test.go
```diff
@@ -715,3 +715,65 @@ func TestSendVSCPacketsToChainFailure(t *testing.T) {
...
+	// Keeper setup
+	providerKeeper, ctx, ctrl, _ := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+
+	// We do not `SetChannelToChain` for "channelID" and therefore `OnTimeoutPacket` fails with no chain found
+	packet := channeltypes.Packet{
+		SourceChannel: "channelID",
+	}
+	err := providerKeeper.OnTimeoutPacket(ctx, packet)
+	require.Error(t, err)
```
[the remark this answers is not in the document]
insumity: Added a test for the type. Not clear if this is the idiomatic way to go about it, but found a similar usage in [handler_test.go](https://github.com/cosmos/interchain-security/blob/cca008d856e3ffc60ec1a486871d0faa702abe26/x/ccv/provider/handler_test.go#L30). Added an error-type test in `TestOnAcknowledgementPacketWithAckError` below as well.

x/ccv/provider/keeper/relay_test.go
insumity: There's not much logic if the `ack` contains no error, because in such a case `OnAcknowledgementPacket` just returns `nil`. This test is more added here to capture a potential regression.

x/ccv/provider/keeper/relay_test.go
insumity: Added a call to `TestProviderStateIsCleanedAfterConsumerChainIsStopped` to capture that the provider's state is cleaned and hence `StopConsumerChain` was called. This is not bullet proof and ideally we should be able to just verify that `StopConsumerChain` is called but since we cannot mock the provider keeper, this feels the best way to go about testing that the chain is stopped.

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -569,48 +572,12 @@ func TestStopConsumerChain(t *testing.T) {
 			require.NoError(t, err)
 		}
 
-		testProviderStateIsCleaned(t, ctx, providerKeeper, "chainID", "channelID")
+		testkeeper.TestProviderStateIsCleanedAfterConsumerChainIsStopped(t, ctx, providerKeeper, "chainID", "channelID")
```
insumity, line 575: Because this is used in `relay_test.go` as well now, the function is moved to `testutil/keeper/unit_test_helpers.go`.

testutil/keeper/unit_test_helpers.go
insumity: Good catch! I rewrote the godoc of this function to clearly state this.

> Maybe we refactor to split things out and/or make naming more accurate

I agree. It would be nice to have 2 functions here: i) `SetupForStoppingConsumerChain` that just sets up the consumer chain for stopping, ii) `TestChannelIsClosed` that asserts that the underlying channel is closed when we call `StopConsumerChain` with `closeChan` being `true`. Then the idea would be that in a test we do something like the following:
```
SetupForStoppingConsumerChain(...)
StopConsumerChain(...) // e.g., we call to stop the consumer chain through `OnTimeoutPacket`
TestChannelIsClosed(...)
TestProviderStateIsCleanedAfterConsumerChainIsStopped(...) // tests that the provider's chain is closed and hence is used as a proxy to verify that the chain was closed 
```
where `TestChannelIsClosed` would need to set the expectations from [GetMocksForStopConsumerChain](https://github.com/cosmos/interchain-security/blob/cca008d856e3ffc60ec1a486871d0faa702abe26/testutil/keeper/expectations.go#L85), so something like:
```
func TestChannelIsClosed(...) {
    ....
    mocks.MockChannelKeeper.EXPECT().ChanCloseInit(...).Times(1)
}
```
The issue with the above approach is that expectations are set after the `StopConsumerChain` call and hence when `StopConsumerChain` is called we would fail with something akin to `Unexpected call to ChanCloseInit([1])`.
To solve the issue, this PR introduces the `SetupForStoppingConsumerChainChannel` function and does  
```
SetupForStoppingConsumerChain(...)
SetupForStoppingConsumerChainChannel(...)
StopConsumerChain(...)
TestProviderStateIsCleanedAfterConsumerChainIsStopped(...)
```
The above approach might not be that beautiful since we do testing before (i.e., test that the channel would be closed) and after (i.e., that the producer's chain is cleaned) the `StopConsumerChain`. Nevertheless, this approachs feels the nicest way to circumvent the mock expectations issue I mentioned above.

x/ccv/provider/keeper/relay_test.go
insumity: If the consumer chain is stopped, then we expect that the provider's state is cleaned (see [spec methods](https://github.com/cosmos/ibc/blob/jk/ccvTLA/spec/app/ics-028-cross-chain-validation/methods.md#consumer-chain-removal)), so using `TestProviderStateIsCleanedAfterConsumerChainIsStopped` to test that the state is cleaned.

testutil/keeper/unit_test_helpers.go
```diff
@@ -226,6 +228,50 @@ func SetupForStoppingConsumerChain(t *testing.T, ctx sdk.Context,
 	require.NoError(t, err)
 }
 
+// SetupForStoppingConsumerChainChannel registers expected mock calls which assert that the channel to the consumer
+// chain is closed. To be used when we test that `StopConsumerChain` is called with `closeChan` set to `true`.
+func SetupForStoppingConsumerChainChannel(t *testing.T, ctx sdk.Context, mocks MockedKeepers) {
```
smarshall-spitzbart, line 233: Is there a reason why we don't just call `GetMocksForStopConsumerChain` inline to stay consistent with the rest of the repo? Ex: https://github.com/cosmos/interchain-security/blob/7986c99b3b49b7603590607e9a026a109dd5eecc/x/ccv/provider/keeper/proposal_test.go#L1077

Imo as a quick reader it's not intuitive that `SetupForStoppingConsumerChainChannel` would assert expected mocks for `StopConsumerChain`

insumity: Renamed to `GetMocksForStopConsumerChain` to`GetMocksForStopConsumerChainWithCloseChannel` because those mocks only make sense if you call `StopConsumerChan` with `closeChan = true`.

smarshall-spitzbart (approved):
LGTM thanks!

p-offtermatt (approved):
LGTM, thanks for the work!

Issue #1167, opened 2023-07-28
e2e tests fail with --verbose


p-offtermatt:
## Summary of Bug

When running the `happy-path`, tests run fine for me when running
```
go run ./tests/e2e/... --happy-path-only
```

but fail when running

```
go run ./tests/e2e/... --happy-path-only --verbose
```

The end of the log looks like this:

```
running default: step 2 == submitConsumerAdditionProposalAction 
running default: step 3 == assignConsumerPubKeyAction 
assignConsumerPubKey cmd: /opt/homebrew/bin/docker exec interchain-security-instance /bin/bash -c interchain-security-pd tx provider assign-consensus-key consu '{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}' --from validatorcarol --chain-id provi --home /provi/validatorcarol --node tcp://7.7.7.6:26658 --gas 90000 --keyring-backend test -y -o json
running default: step 4 == assignConsumerPubKeyAction 
assignConsumerPubKey cmd: /opt/homebrew/bin/docker exec interchain-security-instance /bin/bash -c interchain-security-pd tx provider assign-consensus-key consu '{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}' --from validatorcarol --chain-id provi --home /provi/validatorcarol --node tcp://7.7.7.6:26658 --gas 90000 --keyring-backend test -y -o json
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x105141f48]

goroutine 1 [running]:
main.TestRun.assignConsumerPubKey({{{0x105188ca0, 0x1d}, {0x1051871bd, 0x1c}, {0x1053c3c70, 0x1}, {0xc1290b3b0602cd98, 0x108a437, 0x105f69c60}}, 0x1400094bb60, ...}, ...)
        /Users/offtermatt/projects/interchain-security/tests/e2e/actions.go:1851 +0x448
main.(*TestRun).runStep(0x1400055ac60, {{0x10559afc0?, 0x1400023ef40?}, 0x140008864e0?}, 0x68?)
        /Users/offtermatt/projects/interchain-security/tests/e2e/main.go:171 +0x14c
[...]

PR #1186, opened 2023-08-07
test: Fix failing e2e test

bermuell:
## Description

Closes: #1167 

This PR fixes an error in the e2e test.
When running e2e tests
- tests did not check expected error condition in assignConsumerPubKey
- transaction command to assign consensus key did not return error until '--gas auto' option was provided

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/tests/e2e/actions.go
+++ b/tests/e2e/actions.go
@@ -1813,7 +1813,8 @@ type assignConsumerPubKeyAction struct {
 	// reconfigureNode will change keys the node uses and restart
 	reconfigureNode bool
 	// executing the action should raise an error
-	expectError bool
+	expectError   bool
+	expectedError string
 }
 
 func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbose bool) {
@@ -1848,8 +1849,8 @@ func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbos
 	}
 
 	if action.expectError {
-		if err == nil {
-			log.Fatalf("expected error not raised: %s", string(bz))
+		if err == nil || !strings.Contains(string(bz), action.expectedError) {
+			log.Fatalf("expected error not raised: expected: '%s', got '%s'", action.expectedError, (bz))
 		}
 
 		if verbose {
--- a/tests/e2e/actions.go
+++ b/tests/e2e/actions.go
@@ -1819,8 +1819,9 @@ type assignConsumerPubKeyAction struct {
 func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbose bool) {
 	valCfg := tr.validatorConfigs[action.validator]
 
+	// Note: to get error response reported back from this command '--gas auto' needs to be set.
 	assignKey := fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


tests/e2e/actions.go
```diff
@@ -1847,6 +1848,10 @@ func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbos
 	}
 
 	if action.expectError {
+		if err == nil {
+			log.Fatalf("expected error not raised: %s", string(bz))
+		}
+
 		if verbose {
 			fmt.Printf("got expected error during key assignment | err: %s \n", err.Error())
```
MSalopek, line 1856:
```suggestion
			fmt.Printf("got expected error during key assignment | err: %s | output: %s \n", err, string(bz)) 
```

MSalopek (approved):
Please apply the suggested comment - it will print out the err and the response.

Approve!

tests/e2e/actions.go
```diff
@@ -1847,8 +1848,12 @@ func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbos
 	}
 
 	if action.expectError {
+		if err == nil {
```
p-offtermatt, line 1851:
```suggestion
		if err == nil || !strings.Contains(string(bz), "a validator has assigned the consumer key already: consumer key is already in use by a validator") {
```

Just looking at the PR, my first instinct was that the error we are catching after putting `gas auto` just relates somehow to the gas change. What do you think about putting something like this here to make sure we don't get back an unrelated error?

tests/e2e/actions.go
```diff
@@ -1819,8 +1819,9 @@ type assignConsumerPubKeyAction struct {
 func (tr TestRun) assignConsumerPubKey(action assignConsumerPubKeyAction, verbose bool) {
 	valCfg := tr.validatorConfigs[action.validator]
 
+	// Note: to get error response reported back from this command '--gas auto' needs to be set.
```
p-offtermatt, line 1822: I don't find this comment satisfying, but it's definitely better than nothing.
Is it that the gas limit that was there before was somehow not enough?
Not sure if you want to dive in deeper to find out, if not I think the current comment is good enough to warn us to be careful about changing this.

p-offtermatt (approved):
Thanks for investigating this!
If you want, you could dive into why this needs `gas auto`, investigating this might be an admittedly rough way to dive deeper into the Cosmos SDK. I think it's not strictly necessary and I'd be happy with the current state, too.

tests/e2e/actions.go
bermuell: agree. gas-limit as root cause is an assumption we had but this is not yet identified/confirmed.
I'll keep it for now and update in case I find some time for  some bi-secting runs to narrow down the issue

tests/e2e/actions.go
bermuell: good point. changed a bit the proposal to be more generic so that the expected error is provided by the step expecting the failure.

p-offtermatt: Thanks, sounds even better

tests/e2e/actions.go
p-offtermatt: Sounds reasonable. I wouldn't fuss too much about it, definitely shouldn't block this pr imo

p-offtermatt:
Thanks for addressing my comments, feel free to merge

Issue #1190, opened 2023-08-09
Integrate CometMock for equivocation


p-offtermatt:
## Problem

Equivocation is hard to cause in the current tests, see https://github.com/cosmos/interchain-security/issues/1180.
This issue is concerned with integrating CometMock in the e2e framework to easily convince the app that a double sign infraction has occurred.

## Closing criteria

CometMock can be used in the end-to-end tests to cause double-sign infractions.

## Related issues
https://github.com/cosmos/interchain-security/issues/1180
https://github.com/informalsystems/CometMock/issues/5

PR #1191, opened 2023-08-09
Tests: 1175 ci refactor e2e

bermuell:
## Description

Closes: #1175

- Created github workflow e2e tests
- changed e2e tests for easier test selection (see -tc argument)
- dedicated jobs for each e2e test case on manual and nightly workflow
---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
- [x] Provided a link to the relevant issue or specification
- [x] Reviewed "Files changed" and left comments if necessary 
- [x] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] Confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Confirmed all author checklist items have been addressed
- [ ] Confirmed that this PR does not change production code

PR #1192, opened 2023-08-09
chore: Add integration for equivocation with CometMock

p-offtermatt:
## Description

Closes: #[1190](https://github.com/cosmos/interchain-security/issues/1190)

* Renames the short-happy-path to cometmock-happy-path and adds equivocation steps to it:
  * It is starting to not be significantly shorter as cometmock supports more operations
  * The previous name was confusing as became clear in some discussions
* Add a separate case for CometMock in equivocation
  * Refactor commonly used functionality that is needed for downtime and equivocation

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ test-e2e-short:
 # run only happy path E2E tests with cometmock
 # this set of traces does not test equivocation but it does check downtime
 test-e2e-short-cometmock:
-	go run ./tests/e2e/... --short-happy-path --use-cometmock --use-gorelayer
+	go run ./tests/e2e/... --cometmock-happy-path --use-cometmock --use-gorelayer
 
 # run full E2E tests in sequence (including multiconsumer)
 test-e2e-multi-consumer:
--- a/tests/e2e/actions.go
+++ b/tests/e2e/actions.go
@@ -1604,18 +1604,7 @@ func (tr TestRun) setValidatorDowntime(chain chainID, validator validatorID, dow
 
 	if tr.useCometmock {
 		// send set_signing_status either to down or up for validator
-		var validatorAddress string
-		if chain == chainID("provi") {
-			validatorAddress = tr.getValidatorKeyAddressFromString(tr.validatorConfigs[validator].privValidatorKey)
-		} else {
-			var valAddressString string
-			if tr.validatorConfigs[validator].useConsumerKey {
-				valAddressString = tr.validatorConfigs[validator].consumerPrivValidatorKey
-			} else {
-				valAddressString = tr.validatorConfigs[validator].privValidatorKey
-			}
-			validatorAddress = tr.getValidatorKeyAddressFromString(valAddressString)
-		}
+		validatorAddress := tr.GetValidatorAddress(chain, validator)
 
 		method := "set_signing_status"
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -36,7 +36,7 @@ test-e2e-short:
 # run only happy path E2E tests with cometmock
 # this set of traces does not test equivocation but it does check downtime
 test-e2e-short-cometmock:
-	go run ./tests/e2e/... --short-happy-path --use-cometmock --use-gorelayer
+	go run ./tests/e2e/... --cometmock-happy-path --use-cometmock --use-gorelayer
```
MSalopek, line 39: I think all of this is great and very exciting.

However a small suggestion, maybe we refactor the `main.go` flag parsing so that `--cometmock-happy-path` uses go relayer by default so the flag does not need to be specified.

The flag could just be `-cometmock-happy-path --use-cometmock`.

This can also be done as part of a larger refactor, I'm just asking for opinions

tests/e2e/actions.go
```diff
@@ -1795,15 +1800,28 @@ func (tr TestRun) invokeDoublesignSlash(
...
+			"/testnet-scripts/cause-doublesign.sh", chainConfig.binaryName, string(action.validator),
+			string(chainConfig.chainId), chainConfig.ipPrefix).CombinedOutput()
+		if err != nil {
+			log.Fatal(err, "\n", string(bz))
+		}
+		tr.waitBlocks("provi", 10, 2*time.Minute)
+	} else { // tr.useCometMock
+		validatorAddress := tr.GetValidatorAddress(action.chain, action.validator)
+
+		method := "cause_double_sign"
```
MSalopek, line 1816: Sweet!

MSalopek (approved):
Approve!

Makefile
smarshall-spitzbart: To me it seems implicit that if you're specifying `--cometmock-happy-path`, both `--use-cometmock` and `--use-gorelayer` would default to true

tests/e2e/actions.go
smarshall-spitzbart, line 1816: Curious where the script lives that actually handles `"cause_double_sign"`?

smarshall-spitzbart (approved):
Great stuff!

Makefile
bermuell: pls note that PR #1191 is changing the way tests can be selected.
Keeping selection of tests and infra seperated regarding parameters is better from a usablity perspective (full flexibility with reduced set of option).
I'd be in favor in not renaming the 'test case' and rather not having 'cometmock' in the name as it leads to assumptions that infra/test setup is using cometmock which is not the case. Running  --short-happy-path --use-cometmock (with or without implicit usage of go-relayer) is clearer IMO.

tests/e2e/actions.go
MSalopek: it's a cometmock RPC method, it does not live here. It's a feature of cometmock

Makefile
MSalopek: > assumptions that infra/test setup is using cometmock which is not the case. Running --short-happy-path --use-cometmock (with or without implicit usage of --use-gorelayer) is clearer IMO.

I think this is a good approach to take.

PR #1204, opened 2023-08-15
ci: add PR labeler

mpoke:
## Description

Closes: NA

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/.github/pr_labeler.yml
+++ b/.github/pr_labeler.yml
@@ -1,0 +1,29 @@
+"C:x/consumer":
+  - x/ccv/consumer/**/*
+"C:x/democracy":
+  - x/ccv/democracy/**/*
+"C:x/provider":
+  - x/ccv/provider/**/*
+"C:x/types":
+  - x/ccv/types/**/*
+"C:Docs":
+  - docs/docs/**/*
+"C:ADR":
+  - docs/docs/adrs/**/*
+"C:CI":
+  - .github/**/*.yml
+  - buf.work.yaml
+  - .mergify.yml
+  - .golangci.yml
+  - mlc_config.json
+  - sonar-project.properties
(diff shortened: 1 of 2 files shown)


.github/pr_labeler.yml
```diff
@@ -0,0 +1,29 @@
...
+  - buf.work.yaml
+  - .mergify.yml
+  - .golangci.yml
+  - mlc_config.json
+  - sonar-project.properties
+"C:Build":
+  - Makefile
+  - Dockerfile
+  - scripts/*
+"C:Testing":
```
insumity, line 24: Is the `C:Testing` the label applied when change under `app/**/*`, `cmd/**/*` etc. If yes, why introduce this label for changes under `app` and `cmd`? Also, would it be possible to add this label if modification in a `_test` file.

insumity (approved):
LGTM

.github/pr_labeler.yml
mpoke: Both app/* and cmd/* are only used for testing. Every chain will have its own. Maybe we can have a better split, maybe with a better label. I'll let @p-offtermatt and @MSalopek to suggest something. 

All PRs that change the production code should also change the _test files as they will have unit tests. I'm not sure how valuable is to add a special label for that.

smarshall-spitzbart (approved):
Seems we could also just look at what files are changed for a PR and determine these mental notes ourselves, but overall not opposed to these tags. LGTM

.github/pr_labeler.yml
p-offtermatt: I don't think I have an opinion on that yet, I would suggest we see how these labels feel and we can adjust later.

Issue #1205, opened 2023-08-15
CometMock tests fail randomly


p-offtermatt:
## Summary of Bug
Tests using CometMock seem to fail nondeterministically, with or without error messages.
Some failure logs are here:

https://gist.github.com/p-offtermatt/4294dd21755381ef0bc190bbeb208a7c

## Version

Various

## Steps to Reproduce

Run `go run ./tests/e2e/main.go --tc happy-path-short --use-cometmock --use-gorelayer --verbose` multiple times (~10 should do it, from empirical experience) until you get a test failure.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

p-offtermatt:
This should be fixed per https://github.com/informalsystems/CometMock/pull/35, see that PR for more context

PR #1211, opened 2023-08-18
chore: improve logging in getValPower

p-offtermatt:
## Description

Closes: N/A

Related: https://github.com/cosmos/interchain-security/issues/1205

The code this PR changes has unhelpful error messages. While debugging, I needed to add logging information here,
so I think it might be useful to have this information here with this PR.
This PR is not meant to fix all logging issues in the e2e tests, but just to improve this one particular part of the code where I encountered a problem.

* Make logging of fatal errors more informative in getValPowers

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/e2e/state.go
```diff
@@ -111,6 +111,7 @@ type Param struct {
 func (tr TestRun) getState(modelState State) State {
 	systemState := State{}
 	for k, modelState := range modelState {
+		log.Println("Getting model state for chain: ", k)
```
smarshall-spitzbart, line 114: Is it worth standardizing between `log.Println` and `fmt.Println` in e2e tests? I see both being used

p-offtermatt: Thanks, that's a good point. I think we should use log.Println whenever possible, my understanding is that it adds extra information like the timestamp. Not sure if it's worth refactoring existing stuff, but for new stuff it seems desirable.

tests/e2e/state.go
```diff
@@ -184,6 +185,8 @@ func (tr TestRun) getChainState(chain chainID, modelState ChainState) ChainState
 		chainState.RegisteredConsumerRewardDenoms = &registeredConsumerRewardDenoms
 	}
 
+	log.Println("Done getting chain state: ")
```
insumity, line 188: nit: do we expect some variable after `:` in this print?

tests/e2e/state.go
```diff
@@ -492,27 +495,31 @@ type ValPubKey struct {
...
-		log.Fatalf("error: %v", err)
+		log.Fatalf("encountered an error when executing command '%s': %v, output: %s", command.String(), err, string(bz))
 	}
 
 	valset := TmValidatorSetYaml{}
 
 	err = yaml.Unmarshal(bz, &valset)
 	if err != nil {
-		log.Fatalf("error: %v", err)
+		log.Fatalf("yaml.Unmarshal returned an error while unmarshalling validator set: %v, input: %s", err, string(bz))
```
insumity, line 517: nit: would it make sense to print the validator set here as well (similarly below).

insumity (approved):

tests/e2e/state.go
p-offtermatt: Yes, thanks!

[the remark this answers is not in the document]
p-offtermatt: Thanks, that's exactly what was meant to be here

tests/e2e/state.go
p-offtermatt: The validator set (valSet) is probably garbage since the unmarshalling failed, right? Or do you mean a different variiable?

Issue #1216, opened 2023-08-18
Separate releasing/semver for provider and consumer 


smarshall-spitzbart:
## Problem

We're going to ditch the idea of "go.mod split" in favor of keeping one go.mod and separating releasing/semver for consumer and provider with tags. See https://github.com/cosmos/interchain-security/issues/801#issuecomment-1683349298 

## Closing criteria

- New releasing procedure is documented
- PR template updated

smarshall-spitzbart:
Just tried some things out and it'd be better to slightly alter the release paradigm to look like `vX.Y.Z-provider` and `vX.Y.Z-consumer`, instead of prefixing the semver number, this is more consistent with tags we've released in the past, and I just confirmed both formats are importable by other repos

p-offtermatt:
I might have missed it, but should there also be an ADR for this to supercede https://github.com/cosmos/interchain-security/pull/897? It might be worth it to get a better summary of benefits/costs associated with this, and to get the context collected in one palce.

smarshall-spitzbart:
I can convert the [comment mentioned in this issue](https://github.com/cosmos/interchain-security/issues/801#issuecomment-1683349298) to an ADR if you'd like? This topic got to a point that it has been talked about all year (too much discussion imo) and never solved, so I skipped ADR :)

p-offtermatt:
I think that comment could almost verbatim become an ADR, yes!

smarshall-spitzbart:
> I think that comment could almost verbatim become an ADR, yes!

Sounds good 👍 https://github.com/cosmos/interchain-security/issues/1216#issuecomment-1687712480

PR #1217, opened 2023-08-18
chore: Separate semver

smarshall-spitzbart:
## Description

Closes: https://github.com/cosmos/interchain-security/issues/1216
---

### Author Checklist

I have...

### Reviewers Checklist

I have...


p-offtermatt (approved):
(Whether this ends up being the way we want to handle the separate versions is a separate issue, just commenting on the content of this PR.)

The PR looks good. It might be worth more explicitly writing how to tell whether two releases are mutually compatible.
I think all releases with the same major version should be compatible?

Before we merge this PR, we should check whether there are missing references to the versioning scheme that should be adjusted.
I checked the docs website and didn't find anything that needs adjusting, and no other place came to mind.

.github/PULL_REQUEST_TEMPLATE/production.md
```diff
@@ -30,6 +30,7 @@ I have...
 * [ ] Updated the relevant documentation or specification
 * [ ] Reviewed "Files changed" and left comments if necessary <!-- relevant if the changes are not obvious  -->
 * [ ] Confirmed all CI checks have passed
+* [ ] If this PR is library API breaking, bump the go.mod version string of the repo, and follow through on a new major release for both the consumer and provider
```
mpoke, line 33: Do we want to do this for every API breaking PR?

smarshall-spitzbart:
> It might be worth more explicitly writing how to tell whether two releases are mutually compatible. I think all releases with the same major version should be compatible?

See section on [backwards compatibility ](https://github.com/cosmos/interchain-security/blob/e9c178eb1a444308dc1866ea12e90cb7fe8d23d2/CONTRIBUTING.md#backwards-compatibility) which will remain unchanged by this PR

.github/PULL_REQUEST_TEMPLATE/production.md
smarshall-spitzbart: If we bump the major version of ICS, golang requires us to bump the version string. We can redefine what constitutes a major release, but that'd be steering away from how IBC does versioning

mpoke: My comment is more generally about the mapping between a single PR and a release. Yes, a release that contains a PR with library API breaking changes should have the major version bumped. But not every PR with library API breaking changes should result in a new release.

mpoke (approved):

Issue #1227, opened 2023-08-22
Add consumer double voting handling


sainoe:
## Problem

Add a new endpoint for consumer double voting. It should first verify the double voting evidence and then jail for eternity the Byzantine validator behind the infraction.

PR #1232, opened 2023-08-23
feat: add handler for consumer double voting

sainoe:
## Description

Closes: #1227

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -59,3 +62,19 @@ message MsgSubmitConsumerMisbehaviour {
 }
 
 message MsgSubmitConsumerMisbehaviourResponse {}
+
+
+// MsgSubmitConsumerDoubleVoting defines a message that reports an equivocation
+// observed on a consumer chain
+message MsgSubmitConsumerDoubleVoting {
```
sainoe, line 69: This definition stays unchanged bc the Hermes relayer is supposedly already using it.

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -59,3 +62,19 @@ message MsgSubmitConsumerMisbehaviour {
...
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  string submitter = 1;
+  // The equivocation of the consumer chain wrapping 
+  // an evidence of a validator that signed two conflicting votes
+  tendermint.types.DuplicateVoteEvidence duplicate_vote_evidence = 2;
+  // The light client header of the infraction block 
+  ibc.lightclients.tendermint.v1.Header infraction_block_header = 3;
```
sainoe, line 77: Only used for the consumer chain ID atm

x/ccv/provider/keeper/punish_validators.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// JailAndTombstoneValidator jails and tombstones the validator for the given validator consensus address
+func (k Keeper) JailAndTombstoneValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) {
+	logger := k.Logger(ctx)
+
+	// get validator
+	val, ok := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.ToSdkConsAddr())
+	if !ok || val.IsUnbonded() {
+		logger.Error("validator not found or is unbonded", providerAddr.String())
```
mpoke, line 16: nit: Use `k.Logger(ctx)` as bellow

x/ccv/provider/keeper/msg_server.go
```diff
@@ -146,3 +147,27 @@ func (k msgServer) SubmitConsumerMisbehaviour(goCtx context.Context, msg *types.
...
+
+func (k msgServer) SubmitConsumerDoubleVoting(goCtx context.Context, msg *types.MsgSubmitConsumerDoubleVoting) (*types.MsgSubmitConsumerDoubleVotingResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	evidence, err := tmtypes.DuplicateVoteEvidenceFromProto(msg.DuplicateVoteEvidence)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := k.Keeper.HandleConsumerDoubleVoting(ctx, evidence, msg.InfractionBlockHeader); err != nil {
```
mpoke, line 159: Pass directly the chainID instead of the entire header.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -146,3 +147,27 @@ func (k msgServer) SubmitConsumerMisbehaviour(goCtx context.Context, msg *types.
...
+		return &types.MsgSubmitConsumerDoubleVotingResponse{}, err
+	}
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			ccvtypes.EventTypeSubmitConsumerMisbehaviour,
+			sdk.NewAttribute(ccvtypes.AttributeConsumerDoubleVoting, msg.DuplicateVoteEvidence.String()),
+			sdk.NewAttribute(ccvtypes.AttributeInfractionBlockHeader, msg.InfractionBlockHeader.String()),
```
mpoke, line 167: This should just be the chainID.

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,109 @@
...
+
+// VerifyDoubleVotingEvidence verifies a double voting evidence
+// for a given chain id and validator public key
+func (k Keeper) VerifyDoubleVotingEvidence(
+	ctx sdk.Context,
+	evidence tmtypes.DuplicateVoteEvidence,
+	chainID string,
+	pubkey tmprotocrypto.PublicKey,
+) error {
+	// TODO: check the evidence age once we agreed on the infraction height mapping
```
mpoke, line 60: Use only the timestamp (see https://github.com/cosmos/cosmos-sdk/blob/5b02261cb5d52c1bb645efa559195a4e5b0a7945/x/evidence/keeper/infraction.go#L58), i.e.,

```
infractionTime := evidence.GetTime()
ageDuration := ctx.BlockHeader().Time.Sub(infractionTime)
// Reject evidence if the double-sign is too old. Evidence is considered stale
// if the difference in time is greater than the allowed parameters defined.
cp := ctx.ConsensusParams()
if cp != nil && cp.Evidence != nil {
    if ageDuration > cp.Evidence.MaxAgeDuration{
        ...
```

tests/integration/double_vote.go
```diff
@@ -0,0 +1,400 @@
...
+				// verifies that no jailing and tombstoning has occurred
+
+				s.Require().False(s.providerApp.GetTestStakingKeeper().IsValidatorJailed(s.providerCtx(), provAddr.ToSdkConsAddr()))
+				s.Require().False(s.providerApp.GetTestSlashingKeeper().IsTombstoned(s.providerCtx(), provAddr.ToSdkConsAddr()))
+			}
+		})
+	}
+}
+
+func (s *CCVTestSuite) TestVerifyDoubleVotingEvidence() {
```
mpoke, line 135: I see this more of a UT in double_vote_test.go. Would that be possible or it's too complicated?

mpoke (approved):
Nice work @sainoe. Approve the main logic. See my comments bellow.

tests/e2e/main.go
```diff
@@ -102,7 +102,7 @@ func (tr *TestRun) Run(steps []Step, localSdkPath string, useGaia bool, gaiaTag
 	tr.validateStringLiterals()
 	tr.startDocker()
 	tr.executeSteps(steps)
-	tr.teardownDocker()
+	// tr.teardownDocker()
```
MSalopek, line 105: Can you put this back?

tests/integration/double_vote.go
sainoe: Changed! Thanks for the input, `TestVerifyDoubleVotingEvidence` runs two times faster now.

x/ccv/provider/keeper/double_vote.go
sainoe: Done.

x/ccv/provider/keeper/msg_server.go
sainoe: Committed.

MSalopek (approved):

x/ccv/provider/keeper/punish_validators_test.go
```diff
@@ -0,0 +1,136 @@
...
+		providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(
+			t, testkeeper.NewInMemKeeperParams(t))
+
+		// Setup expected mock calls
+		gomock.InOrder(tc.expectedCalls(ctx, mocks, tc.provAddr)...)
+
+		// Execute method and assert expected mock calls
+		providerKeeper.JailAndTombstoneValidator(ctx, tc.provAddr)
+
+		ctrl.Finish()
```
insumity, line 134: nit: maybe `defer ctrl.Finish()` at the top after:
```
providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
``` because it seems to be the pattern mostly used in other tests.

File: x/ccv/provider/types/msg.go [hunk 10]
```diff
@@ -185,3 +188,47 @@ func (msg MsgSubmitConsumerMisbehaviour) GetSigners() []sdk.AccAddress {
...
+}
+
+// Type implements the sdk.Msg interface.
+func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
+	addr, err := sdk.AccAddressFromBech32(msg.Submitter)
+	if err != nil {
+		// same behavior as in cosmos-sdk
+		panic(err)
+	}
+	return []sdk.AccAddress{addr}
```
insumity, line 233: Is my understanding correct that `GetSigners` builds the addresses that need to have signed the `MsgSubmitConsumerDoubleVoting` in order for the message to be valid?
For example, in the [bank module](https://github.com/cosmos/cosmos-sdk/blob/147758fd179f334546d757794509cc2d1ad85b99/x/bank/types/msgs.go#L18), you need to verify that someone sending money from an account `from` has indeed signed the message, otherwise you could potentially steal tokens.

If the above is true, then in the case of `MsgSubmitConsumerDoubleVoting` why do we need the message to be signed by `addr`?

x/ccv/provider/keeper/punish_validators_test.go
sainoe: Actually `defer` would get `ctrl.Finish()` executed at the end of the test, i.e the enclosing function.
In a `for` loop, we instantiate and finish the `ctrl` for each iteration.

x/ccv/provider/keeper/punish_validators_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for _, tc := range testCases {
+		providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(
+			t, testkeeper.NewInMemKeeperParams(t))
+
+		// Setup expected mock calls
+		gomock.InOrder(tc.expectedCalls(ctx, mocks, tc.provAddr)...)
+
+		// Execute method and assert expected mock calls
+		providerKeeper.JailAndTombstoneValidator(ctx, tc.provAddr)
```
insumity, line 132: if a test fails would we be able to easily see the `tc.name` for the failed test?

x/ccv/provider/keeper/punish_validators.go
```diff
@@ -0,0 +1,35 @@
+package keeper
```
insumity, line 1: would it make sense to name the file `punish_validator.go` because it is used to "punish" once validator at a time?

x/ccv/provider/keeper/punish_validators_test.go
```diff
@@ -0,0 +1,136 @@
...
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	cryptotestutil "github.com/cosmos/interchain-security/v2/testutil/crypto"
+	testkeeper "github.com/cosmos/interchain-security/v2/testutil/keeper"
+	"github.com/cosmos/interchain-security/v2/x/ccv/provider/types"
+	"github.com/golang/mock/gomock"
+)
+
+func TestJailAndTombstoneValidator(t *testing.T) {
+	providerConsAddr := cryptotestutil.NewCryptoIdentityFromIntSeed(7842334).ProviderConsAddress()
+	testCases := []struct {
```
insumity, line 17: If my understanding is correct that those tests check the order of operations (i.e., method calls) taking place in `JailAndTombstoneValidator`. So, the first test checks that `GetValidatorByConsAddr` is called. Then, the third test checks that if `GetValidatorByConsAddr` passes and since we are not unbonded, then we call `IsTombstoned`, etc.

Maybe it would make sense for `JailAndTombstoneValidator` to return an `err`or on whether it succeeds or not and then test depending (on the mocks we have) on whether `JailAndTombstoneValidator` succeeds? Or is this not possible because the underlying `JailUntil and `Tombstone` from the `slashing` module do not return an `err`or?

x/ccv/provider/keeper/double_vote_test.go
```diff
@@ -0,0 +1,311 @@
...
+	val1 := tmtypes.NewValidator(signer1.PrivKey.PubKey(), 1)
+	val2 := tmtypes.NewValidator(signer2.PrivKey.PubKey(), 1)
+
+	valSet := tmtypes.NewValidatorSet([]*tmtypes.Validator{val1, val2})
+
+	blockID1 := testutil.MakeBlockID([]byte("blockhash"), 1000, []byte("partshash"))
+	blockID2 := testutil.MakeBlockID([]byte("blockhash2"), 1000, []byte("partshash"))
+
+	ctx = ctx.WithBlockTime(time.Now())
+	oldTime := ctx.BlockTime().Add(-505 * time.Hour)
```
insumity, line 33: Would it be possible to reuse some constant here (e.g., `ctx.ConsensusParams().Evidence.MaxAgeDuration`) instead of `-505` to make it clear that we go enough back in time for the evidence to be too old?

x/ccv/provider/keeper/double_vote_test.go
```diff
@@ -0,0 +1,311 @@
...
+					signer1,
+					chainID,
+				),
+			},
+			chainID,
+			valPubkey1,
+			false,
+		},
+		{
+			"given chain ID isn't the same that the one used to sign the votes - shouldn't pass",
```
insumity, line 145: nit: s/that the one/as the one

tests/integration/double_vote.go
```diff
@@ -0,0 +1,123 @@
...
+				ValidatorPower:   val.VotingPower,
+				TotalVotingPower: val.VotingPower,
+				Timestamp:        s.consumerCtx().BlockTime(),
+			},
+			s.consumerChain.ChainID,
+			false,
+		},
+		{
+			// In order to create an evidence for a consumer chain,
+			// we create two votes that only differs by their Block IDs and
```
insumity, line 82: nit: s/only differs/only differ

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,115 @@
...
+// for a given chain id and validator public key
+func (k Keeper) VerifyDoubleVotingEvidence(
+	ctx sdk.Context,
+	evidence tmtypes.DuplicateVoteEvidence,
+	chainID string,
+	pubkey tmprotocrypto.PublicKey,
+) error {
+	// check evidence age duration
+	maxAgeDuration := ctx.ConsensusParams().Evidence.MaxAgeDuration
+	ageDuration := ctx.BlockHeader().Time.Sub(evidence.Time())
```
insumity, line 58: `evidence.Time()` uses the `Timestamp` from `DuplicateVoteEvidence` that in my understanding could be easily faked? If that's the case, someone could send old evidence with a recent timestamp.
Would it make sense to get the time out of one of the votes instead?

insumity:
We talked about this offline but adding here for visibility. It seems that with the changes in this PR, a validator `V` could prevent getting slashed in the provider chain by doing the following:

1. `V` performs equivocation `E` in the provider chain
2. `V` quickly performs double signing in the consumer chain
3. `V` submits a `MsgSubmitConsumerDoubleVoting` to the provider chain
4. `V` gets jailed and tombstoned in the provider chain but not slashed
5. Now if equivocation `E` shows up in the provider chain, `V` is tombstoned and [won't be slashed](https://github.com/cosmos/cosmos-sdk/blob/4a536b21e2cb134d57ce20b83831c8a930d33702/x/evidence/keeper/infraction.go#L87)

The above scenario seems unlikely because `V` would have to hide equivocation `E` for some time, but I guess it is still be technically possible.

x/ccv/provider/types/msg.go
sainoe: Good question. AFAIU all messages must be signed by the submitter and should satisfy the [msg interface ](https://github.com/cosmos/cosmos-sdk/blob/v0.45.16/types/tx_msg.go#L9).

x/ccv/provider/keeper/punish_validators_test.go
sainoe: Not sure! Maybe it's worth finding a way to print the subtests' name like in the integration tests:
`func (*suite.Suite).Run(name string, subtest func()) bool`.

x/ccv/provider/keeper/punish_validators.go
sainoe: Agree.

x/ccv/provider/keeper/punish_validators_test.go
sainoe: The success is tested by the number of expected mocks per test case.
For instance, when the validator is unbonded, the test expects to have one unique call to 'GetValidatorByConsAddr` or would fail.

x/ccv/provider/keeper/double_vote.go
sainoe: Good catch! Changed to use the vote timestamp instead.

Issue #1239, opened 2023-08-25
Rewrite the TLA+ ICS model in Quint (see https://github.com/cosmos/ibc/pull/911)


p-offtermatt:
To make a step towards replacing the existing difftest model in model-based testing, we want to have a simpler model of ICS in Quint.
We chose to rewrite the existing TLA+ model in Quint, see https://github.com/cosmos/ibc/pull/911
The Quint model is not supposed to be an exact replica, but to capture roughly the same part of the protocol at the same abstraction level and complexity.

*Closing Criteria*
The TLA+ model has been rewritten in Quint.

PR #1242, opened 2023-08-25
chore: Add Quint model of Interchain Security

p-offtermatt:
## Description

Closes: #1239

This adds a quint model of interchain Security, based on the existing TLA+ model by @Kukovec.
There are a few notable differences:
* Initializing is removed as a status - this is mostly because my goal is to use the model for MBT, and the initializing state is probably something we would skip past in the tests (also since the interactions with the initializing state aren't very meaningful in the model)
* The actions the model exposes are slightly different, e.g. time advancement and chain status advancement are separate actions and don't happen simultaneously with other actions. This is because this makes the traces easier to test in Quint, i.e. more control over what happens when.
* I model endBlocks for both provider and consumer chains. This is because I think this is a useful extra level of implementation details that will make test traces more meaningful.

Other than that, some implementation details are different. Some are my preference, some are just due to what is more natural to write in Quint, some are to (hopefully) make traces that are easier to use.

When this PR is merged, the model should also be added to the spec repo.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+            id: int,
+            // the time, that when passed on the receiver chain, will mean the packet is considered timed out
+            timeout: Timestamp
+        }
+    
+    // MODEL PARAMETERS
+        
+
+    // set of identifiers of potential nodes
+    pure val Nodes: Set[Node] =
```
smarshall-spitzbart, line 45: Is this pretty much an example validator set? If so, why not define it as a `ValidatorSet`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+    
+    // MODEL PARAMETERS
+        
+
+    // set of identifiers of potential nodes
+    pure val Nodes: Set[Node] = 
+        Set("A", "B", "C", "D", "E", "F", "G", "H", "I", "J")
+
+    // the set of consumer chains
+    pure val ConsumerChains: Set[Chain] =
```
smarshall-spitzbart, line 49: Can these model parameters be changed? Ie. could we theoretically execute the model with 5 consumers for example

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+    // PROPERTIES
+
+    // Every validator set on any consumer chain MUST either be or
+    // have been a validator set on the provider chain.
+    val ValidatorSetReplication: bool = 
+        chains.forall(
+            chain => chain.getCurrentValidatorSet().wasValidatorSetOnProvider()
+        )
+
+    // TESTS
```
smarshall-spitzbart, line 573: Do tests over the model always have to be explicitly defined? My understanding is that quint somehow enables fuzzed or implicitly generated test cases

smarshall-spitzbart (commented):
Great work! Very useful and well documented. I haven't taken too deep of a dive into model logic yet. 

My two cents w.r.t the more broad conversation of using quint vs something like Rapid:
- This model definitely seems readable, so readability of .qnt vs .go isn't an issue. 
- I'm still having trouble understanding why quint as a language is more succinct for writing models compared to other programming languages that're more widely known. There are global variables, types, functions, actions etc. defined, all composed together to create what we're referring to as a "model". These abstractions are all possible to encode with golang from my knowledge.
- Can we use a debugger for .qnt like we could with .go files? 
- I'm not able to really validate the model logic until I take time to learn quint. This is mostly a me problem :)

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: This is supposed to be the set of "potential" full nodes that could be validators, i.e. the validator set will be a subset of these (because some may not have any power=bonded tokens).

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: They do not need to be defined, Quint allows doing PBT in-built (checking that given invariants always hold).
This manual test case is mostly there for my benefit while writing the model to gradually check things I introduced, and can complement random testing.

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: Yes, exactly, this can be changed.

p-offtermatt:
Thanks for taking the time to review!
>     * This model definitely seems readable, so readability of .qnt vs .go isn't an issue.
That's great to hear!
>     * I'm still having trouble understanding why quint as a language is more succinct for writing models compared to other programming languages that're more widely known. There are global variables, types, functions, actions etc. defined, all composed together to create what we're referring to as a "model". These abstractions are all possible to encode with golang from my knowledge.
In principle, you are right, this can also be written in Golang. One big advantage is that Quint can be model-checked, which we would have a hard time doing with Golang (since it has more fine-grained concepts of e.g. memory, the state space blows up much faster).
>     * Can we use a debugger for .qnt like we could with .go files?
Something like it, there is a REPL that can serve the same purpose, though it is a bit different in the exact usage.

Thanks again for commenting, it's very helpful to hear different perspectives on this!

tests/difference/core/quint_model/ccv.qnt
smarshall-spitzbart: Nice!

smarshall-spitzbart:
@p-offtermatt thanks for the responses! W.r.t `One big advantage is that Quint can be model-checked`, what do you mean by this? Is model checking the process of generating a large number of test traces over the described model? From my understanding Rapid has mechanisms which do this, although I admittedly do not have knowledge about the internals

p-offtermatt:
> @p-offtermatt thanks for the responses! W.r.t `One big advantage is that Quint can be model-checked`, what do you mean by this? Is model checking the process of generating a large number of test traces over the described model? From my understanding Rapid has mechanisms which do this, although I admittedly do not have knowledge about the internals

No, model-checking is more powerful. Intuitively, model-checking is an exhaustive way of checking *all* possible states of the model for given properties. Imagine for the Cosmos Hub, we want to ensure that token balances of wallets can never become negative. We could just try to show this e.g. via PBT by exhaustively generating test traces. Imagine there are 5 billion possible states reachable in 5 steps (this is a vast underestimation of the state space for the Cosmos Hub that can be reached in 5 steps), then exhaustively generating test traces can clearly never cover the whole state space. Model-checking in this context would mean checking all 5 billion possible states, and we could conclusively say that in none of them is a wallet balance negative.

The problem is that model-checking a "full programming language" is very hard and slow, but for Quint it's possible. The proposed workflow then is to write a Quint model -> Model-check the Quint model to confirm it fulfills desirable safety properties -> use model-based testing to confirm the code behaves like the model, thus the code fulfills the desirable properties.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+
+    // stores whether, in this step, the validator set considered to be changed.
+    // this is needed because the validator set might be considered to have changed, even though
+    // it is still technically identical at our level of abstraction, e.g. a validator power change on the provider
+    // might leave the validator set the same because a delegation and undelegation cancel each other out.
+    var providerValidatorSetChangedInThisBlock: bool
+
+
+    // UTILITY FUNCTIONS & ACTIONS
+    def wasValidatorSetOnProvider(validatorSet: ValidatorSet): bool = {
```
mpoke, line 157: Splitting the model into functional and SM would make these functions pure. For example, just define a type `VotingPowerHistories` and pass it to the function. I believe this would make the logic much easier to understand and maintain.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        votingPowerHistories.get(ProviderChain).toSet().exists(
+            historicalValSet => historicalValSet == validatorSet
+        )
+    }
+
+    def getCurrentValidatorSet(chain: Chain): ValidatorSet = 
+        votingPowerHistories.get(chain).head()
+
+    def getUpdatedValidatorSet(oldValidatorSet: ValidatorSet, validator: Node, newVotingPower: int): ValidatorSet =
```
mpoke, line 166: This should be a `pure` function.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        if (newVotingPower > 0) 
+        oldValidatorSet.put(validator, newVotingPower)
+        else
+        oldValidatorSet.mapRemove(validator)
+
+    // returns true if the consumer has timed out and should be dropped
+    def consumerTimedOut(consumer: Chain): bool =
+        any {
+            // either a package from provider to consumer has timed out
+            outstandingPacketsToConsumer.get(consumer).select(
```
mpoke, line 176: Wouldn't this be more efficient to just look at the oldest packet in the list? Or we do not remove packets once they arrive?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        (chain) => 
+        10
+    )
+
+    // the time until a packet times out
+    pure val PacketTimeout: int = 
+        5
+
+    // the time until a consumer chain is dropped by the provider due to inactivity
+    pure val InactivityTimeout: int =
```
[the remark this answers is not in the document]
mpoke: This is the VSCTimeout  Keep consistency between model and implementation.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+            val packetsWithoutResponse = sentVSCPackets.get(consumer).filter( // get packets without response
+                packet => 
+                not(receivedMaturations.exists(
+                    maturedPacket => maturedPacket.id == packet.id
+                ))
+            )
+            // among those, get packets where inactivity timeout has passed
+            packetsWithoutResponse.filter(
+                packet => 
+                val sentAt = curChainTimes.get(ProviderChain) - PacketTimeout // compute when the packet was sent
```
mpoke, line 194: Why was the packet sent at the current time on the provider minus the packet timeout?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+                val timesOutAt = sentAt + InactivityTimeout // compute when the packet times out
+                timesOutAt <= curChainTimes.get(ProviderChain)
+            ).size() > 0
+        }
+
+    // utility action that leaves all provider state untouched
+    action PROVIDER_NOOP(): bool = 
+        all {
+            receivedMaturations' = receivedMaturations,
```
mpoke, line 203: What about `sentVSCPackets` and `providerValidatorSetChangedInThisBlock`?

mpoke (commented):
Partial review before going into the model actions. See my comments below.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        Set("chain1", "chain2", "chain3")
+
+    // The singular provider chain.
+    pure val ProviderChain: Chain = 
+        "provider"
+        
+    pure val chains = ConsumerChains.union(Set(ProviderChain))
+
+    // length of the unbonding period on each chain
+    pure val UnbondingPeriod: Chain -> int = chains.mapBy(
```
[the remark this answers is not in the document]
p-offtermatt: This is already a map of Chains to unbonding periods, i.e. it can be different per chain. Putting this into the state machine only makes sense if we want to test changing the unbonding period during runtime. Do you think that is a concern?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        5
+
+    // the time until a consumer chain is dropped by the provider due to inactivity
+    pure val InactivityTimeout: int =
+        10
+
+    // consumer statuses
+    pure val STOPPED = "stopped"
+    pure val RUNNING = "running"
+    pure val UNUSED = "unused"
```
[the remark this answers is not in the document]
p-offtermatt: the default for chains that are not consumers and never were. "all the possible consumer chains"

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+    val RunningConsumers: Set[Chain] = 
+        ConsumerChains.filter(chain => consumerStatus.get(chain) == RUNNING)
+
+    // MODEL STATE
+    // --SHARED STATE
+
+    // Stores, for each chain, the list of voting powers that corresponded to voting powers 
+    // at blocks over its entire existence.
+    // Voting powers should be ordered by recency in descending order.
+    var votingPowerHistories: Chain -> List[ValidatorSet]
```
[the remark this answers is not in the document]
p-offtermatt: For the things that are in the model right now, the height is not actually needed. It would probably become important when we e.g. include slashing

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        votingPowerHistories' = votingPowerHistories.set(ProviderChain, getUpdatedHistory(ProviderChain)),
+        // the running validator set is now for sure the current validator set,
+        // so start with it in the next block
+        runningValidatorSet' = runningValidatorSet,
+        // send VSCPackets to consumers
+        val newOutstandingPackets = 
+            // if running validator set is considered to have changed
+            if (providerValidatorSetChangedInThisBlock)
+                // then send a packet to each running consumer
+                outstandingPacketsToConsumer.keys().mapBy(
```
mpoke, line 328: What if it's a new consumer chain without any outstanding packets? That would mean that `outstandingPacketsToConsumer` has no key with that consumer, right? 
Why not use `ConsumerChains.mapBy`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        )
+        all {
+            // remove matured packets from the maturation times
+            maturationTimes' = maturationTimes.set(consumer, maturationTimes.get(consumer).mapRemoveAll(maturedPackets)),
+            // send matured packets
+            outstandingPacketsToProvider' = outstandingPacketsToProvider.set(
+                consumer, 
+                // construct VSCMaturedPackets from the matured VSCPackets
+                outstandingPacketsToProvider.get(consumer).concat(
+                    maturedPackets.map(packet => {id: packet.id, timeout: 5}).toList()
```
mpoke, line 381: We shouldn't use hardcoded values.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+                )
+            )
+        },
+        PROVIDER_NOOP,
+        // no packets are sent to consumer or received by it
+        RegisterNewOutstandingPackets(outstandingPacketsToConsumer),
```
mpoke, line 387: `outstandingPacketsToConsumer` remains unchanged. Why calling `RegisterNewOutstandingPackets`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+        // chain times do not change
+        curChainTimes' = curChainTimes,
+        // the validator set was not changed by this action (but might have been changed before in this block)
+        // also, this is only a new block for a consumer, so the change variable shouldn't be reset
+        providerValidatorSetChangedInThisBlock' = providerValidatorSetChangedInThisBlock
+    }
+
+    // advance timestamps for maps nondeterministically
+    action AdvanceTime(): bool = 
+            val advanceAmounts = curChainTimes.keys().mapBy(
```
mpoke, line 399: Why not use `chains.mabBy()`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,619 @@
...
+            consumerStatus' = consumerStatus,
+            // the validator set was not changed by this action (but might have been changed before in this block)
+            providerValidatorSetChangedInThisBlock' = providerValidatorSetChangedInThisBlock
+        }
+
+    // each consumer chain may advance in the order
+    // some events may necessitate a transition, e.g. timeouts.
+    // shouldAdvance gives, for each consumer chain, whether it should advance if possible.
+    // if a chain has to advance, e.g. due to timeouts, or may not advance, the value will have no effect.
+    action AdvanceConsumers(shouldAdvance: Chain -> bool): bool =
```
mpoke, line 430: `shouldAdvance` is not clear.

mpoke (commented):
Some more comments.

p-offtermatt:
Notes to myself:
* curTimes are unnecessary - timestamps should be per block, and time advancement happens just in endBlock

p-offtermatt:
Just to keep it cleaner, I will close this PR and open a new one with the new model.
The model is rewritten from scratch anyways, so the old comments will not be very relevant.

Issue #1237, opened 2023-08-25
MsgTransfer is not validated before calling Transfer


mpoke:
## Problem

Before calling the [Transfer](https://github.com/cosmos/interchain-security/blob/45c7e360fd8ffc3d312230822f514077705668f6/x/ccv/consumer/keeper/distribution.go#L136) method of ICS-20, the [message](https://github.com/cosmos/interchain-security/blob/45c7e360fd8ffc3d312230822f514077705668f6/x/ccv/consumer/keeper/distribution.go#L126) is not being validated. Normally, Transfer is called as a result of submitting a message MsgTransfer to the chain, which calls `ValidateBasic`. However, the validation is skipped in case of calling the method directly. 

## Closing criteria

Validate MsgTransfer before calling Transfer.

mpoke:
Alternative solution (suggested by @crodriguezvega):

> Also, instead of calling directly Transfer I think the recommended approach is to route the message through the service router. For example, this is [what we do in ICA host when a message is received](https://github.com/cosmos/ibc-go/blob/d0f7773317daa4c242e10b7a4efe915fbf9c7e68/modules/apps/27-interchain-accounts/host/keeper/relay.go#L120-L128). The keeper of the host [is created with a ](https://github.com/cosmos/ibc-go/blob/d0f7773317daa4c242e10b7a4efe915fbf9c7e68/modules/apps/27-interchain-accounts/host/keeper/keeper.go#L39C21-L39C34)MessageRouter (see also [wiring in app.go](https://github.com/cosmos/ibc-go/blob/d0f7773317daa4c242e10b7a4efe915fbf9c7e68/testing/simapp/app.go#L486)). So maybe that could be a way to do it as well in interchain security, and then maybe you can remove the dependency to the transfer keeper. I believe in that case ValidateBasic will trigger for the message.

PR #1244, opened 2023-08-25
fix!: validate MsgTransfer before calling Transfer()

mpoke:
## Description

Closes: https://github.com/cosmos/interchain-security/issues/1237

Also, update the default consumer unbonding period to 2 weeks. 

---

### Author Checklist

I have...

* [ ] ~Updated the relevant documentation or specification~

### Reviewers Checklist

I have...


tests/integration/distribution.go
```diff
@@ -338,8 +340,156 @@ func (s *CCVTestSuite) TestEndBlockRD() {
 	}
 }
 
+// TestSendRewardsToProvider is effectively a unit test for TestSendRewardsToProvider(),
```
insumity, line 343: Maybe "a unit test for `SendRewardsToProvider`" instead of "a unit test for TestSendRewardsToProvider(),"?

tests/integration/distribution.go
```diff
@@ -338,8 +340,156 @@ func (s *CCVTestSuite) TestEndBlockRD() {
...
+	testCases := []struct {
+		name           string
+		setup          func(sdk.Context, *consumerkeeper.Keeper, icstestingutils.TestBankKeeper)
+		expError       bool
+		tokenTransfers int
+	}{
+		{
+			name: "successful token transfer",
+			setup: func(ctx sdk.Context, keeper *consumerkeeper.Keeper, bankKeeper icstestingutils.TestBankKeeper) {
+				// setup transfer channel
```
insumity, line 355: `// setup transfer channel` comment here and below could probably be removed.

x/ccv/consumer/keeper/distribution.go
```diff
@@ -103,62 +103,74 @@ func (k Keeper) shouldSendRewardsToProvider(ctx sdk.Context) bool {
...
-				if err != nil {
-					return err
-				}
-				sentCoins = sentCoins.Add(balance)
+	sourceChannelID := k.GetDistributionTransmissionChannel(ctx)
+	transferChannel, found := k.channelKeeper.GetChannel(ctx, transfertypes.PortID, sourceChannelID)
+	if !found || transferChannel.State != channeltypes.OPEN {
+		k.Logger(ctx).Info("WARNING: cannot send rewards to provider;",
+			"transmission channel not in OPEN state", "channelID", sourceChannelID)
+		return nil
```
insumity, line 111: Do we want to return an `err != nil` or maybe log an `Error` instead of `Info`. It seems that when we cannot [send the rewards](https://github.com/cosmos/interchain-security/blob/45c7e360fd8ffc3d312230822f514077705668f6/x/ccv/consumer/keeper/distribution.go#L34) we log a message with a level `Error`.

tests/integration/distribution.go
```diff
@@ -338,8 +340,156 @@ func (s *CCVTestSuite) TestEndBlockRD() {
...
+					consumertypes.ConsumerToSendToProviderName,
+					sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(100))),
+				)
+				s.Require().NoError(err)
+			},
+			expError:       false,
+			tokenTransfers: 1,
+		},
+		{
+			name: "no transfer channel",
```
insumity, line 376: I guess this has to do with when `SendRewardsToProvider` returns an error but reading this test's name I would've expected to see `expError: true` because with "no transfer channel" we cannot really send rewards (?) Similarly for the tests below.

tests/integration/distribution.go
```diff
@@ -338,8 +340,156 @@ func (s *CCVTestSuite) TestEndBlockRD() {
...
 }
 
+// TestSendRewardsToProvider is effectively a unit test for TestSendRewardsToProvider(),
+// but is written as an integration test to avoid excessive mocking.
+func (s *CCVTestSuite) TestSendRewardsToProvider() {
+	testCases := []struct {
+		name           string
+		setup          func(sdk.Context, *consumerkeeper.Keeper, icstestingutils.TestBankKeeper)
+		expError       bool
+		tokenTransfers int
```
insumity, line 350: Potentially this could be a `bool` because it would seem it only takes 2 values and then later on just check if `commitments` has one or zero elements?

insumity:
Because this PR introduces `ValidateBasic` I would assume we would like to have a test with a bogus `MsgTransfer` and verify that the `SendRewardsToProvider` fails in those cases?
But it would seem we cannot easily do this because `MsgTransfer` is created inside the `SendRewardsToProvider` method?

tests/integration/distribution.go
mpoke: Sure, but like this we avoid an extra if condition.

tests/integration/distribution.go
mpoke: Yeah, but SendRewardsToProvider returns nil when there is no transfer channel. That's the reason we have `tokenTransfers`

x/ccv/consumer/keeper/distribution.go
mpoke: This was the current behavior and I wouldn't want to change it as it works. I think it's fine to log a warning in case this happens. Since we don't return an error, it's weird to log an error.

mpoke:
> Because this PR introduces `ValidateBasic` I would assume we would like to have a test with a bogus `MsgTransfer` and verify that the `SendRewardsToProvider` fails in those cases?
> But it would seem we cannot easily do this because `MsgTransfer` is created inside the `SendRewardsToProvider` method?

The input to this function is not the `MsgTrasfer`. As you already mentioned, the `MsgTrasfer` is created inside the function. `TestSendRewardsToProvider` already tests the `SendRewardsToProvider` function for different potential inputs. Is there another scenario we should test for?

insumity:
> > Because this PR introduces `ValidateBasic` I would assume we would like to have a test with a bogus `MsgTransfer` and verify that the `SendRewardsToProvider` fails in those cases?
> > But it would seem we cannot easily do this because `MsgTransfer` is created inside the `SendRewardsToProvider` method?
> 
> The input to this function is not the `MsgTrasfer`. As you already mentioned, the `MsgTrasfer` is created inside the function. `TestSendRewardsToProvider` already tests the `SendRewardsToProvider` function for different potential inputs. Is there another scenario we should test for?

My understanding is that the only bogus `MsgTransfer` that is tested is for a "no recipient address," but `ValidateBasic` performs [additional checks](https://github.com/cosmos/ibc-go/blob/fe52d2a147592885c6aa189d9325b792c46c7b86/modules/apps/transfer/types/msgs.go#L88). I was  wondering if it makes sense to test for some of the other edge cases as well?

insumity (approved):
Thanks for your responses.

LGTM

p-offtermatt (approved):
LGTM. I find it a bit weird that things sometimes fail silently (i.e. we expect no error, but also expect no token transfers to be done), but doesn't seem too bad and also out of the scope of the PR

mpoke:
> LGTM. I find it a bit weird that things sometimes fail silently (i.e. we expect no error, but also expect no token transfers to be done), but doesn't seem too bad and also out of the scope of the PR

The reason is that it's not really possible to handle an error in an endblocker, besides by panicking and halting the chain. In this case, when the transfer channel is not set, the rewards will accumulate. There is no reason to shut down.

Issue #1248, opened 2023-08-30
Add CometMock specific functionality: Light Client Attacks


p-offtermatt:
This issue is for integrating the CometMock specific light client evidence functionality in CometMock, see
https://github.com/informalsystems/CometMock/issues/34 for context on the feature in CometMock

# Closing Criterion
in the e2e tests, there is an action that causes CometMock to put light client attack evidence on-chain via ABCI

Issue #1180, opened 2023-08-02
Cover light client equivocation attacks in E2E tests


sainoe:
## Problem

Currently, the E2E tests are only testing the ICS misbehaviour handling feature by triggering a "lunatic" light client attack.
However, this test case isn't sufficient since only the "equivocation" light client attacks are expected to be handled and lead to the jailing and tombstoning of validators.

## Closing criteria

Add a test case that generates an "equivocation" light client attack to the E2E tests.

## Problem details

The current way of generating a "lunatic" attack is by forking the consumer chain. This results in the malicious chain sending invalid client headers to the provider chain. Note that after observing the block produced by the main and forked chain, the only block header fields that aren't equal are: `time`, `app_hash`, and `last_commit_hash`. Maybe it's a point to start in order to simulate an "equivocation" attack.

p-offtermatt:
Hey, I took a look at this and it looks like CometMock could be useful here.
I added support for equivocation in CometMock - it can produce DuplicateVoteEvidence now.
Is this the type of evidence you were looking to produce? see https://github.com/cosmos/interchain-security/issues/1190

sainoe:
Hi. Great work in integrating the equivocation into CometMock!
Here the need is to get a [`LightClientAttackEvidence`](https://github.com/cometbft/cometbft/blob/3d1c36d1ecb27a8f9791fe661d4e2f15224a3796/proto/tendermint/types/evidence.proto#L28) that embeds a double signing infraction committed to a block. Does CometMock already support that?

Note that these E2E tests also involve a Hermes Relayer probing evidences committed on-chain.

p-offtermatt:
Ah, I misunderstood - I added DoubleSignEvidence to CometMock, LightClientAttackEvidence is not something I have looked at yet. I have looking into adding LightClientAttackEvidence to CometMock on the roadmap https://github.com/informalsystems/CometMock/issues/34, but if the test crucially relies on Hermes, CometMock won't be able to help at the moment, see https://github.com/cosmos/cosmos-sdk/issues/16277

If you think this could also be done with the go relayer, let me know and I'm happy to add LightClientAttackEvidence as soon as I can!

p-offtermatt:
Hey, light client attacks are now available in CometMock, see https://github.com/cosmos/interchain-security/pull/1249. Again, this does not work with Hermes, but just mentioning here in case this is useful.

sainoe:
Adressed in #1388 & #1249.

PR #1249, opened 2023-08-30
test: Add light client attack to e2e tests

p-offtermatt:
## Description

Closes: #1248 

Add three new actions that cause different types of LightClientAttackEvidence to surface on the given chain.

Also adds a new trace that makes use of this action. SInce this performs essentially exactly like double signing, I just repurpose the existing test case that works with CometMock.

When #1180 is done, it probably should add to these actions a way to do the same functionality without CometMock (see the issue for more context).

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/tests/e2e/actions.go
+++ b/tests/e2e/actions.go
@@ -1826,6 +1826,73 @@ func (tr TestRun) invokeDoublesignSlash(
 	}
 }
 
+// Cause light client attack evidence for a certain validator to appear on the given chain.
+// The evidence will look like the validator equivocated to a light client.
+type lightClientEquivocationAttackAction struct {
+	validator validatorID
+	chain     chainID
+}
+
+func (tr TestRun) lightClientEquivocationAttack(
+	action lightClientEquivocationAttackAction,
+	verbose bool,
+) {
+	tr.lightClientAttack(action.validator, action.chain, LightClientEquivocationAttack)
+}
+
+// Cause light client attack evidence for a certain validator to appear on the given chain.
+// The evidence will look like the validator tried to perform an amnesia attack.
+type lightClientAmnesiaAttackAction struct {
+	validator validatorID
+	chain     chainID
+}
+
+func (tr TestRun) lightClientAmnesiaAttack(
+	action lightClientAmnesiaAttackAction,
+	verbose bool,
+) {
+	tr.lightClientAttack(action.validator, action.chain, LightClientAmnesiaAttack)
+}
+
+type lightClientLunaticAttackAction struct {
+	validator validatorID
+	chain     chainID
+}
+
+func (tr TestRun) lightClientLunaticAttack(
+	action lightClientLunaticAttackAction,
+	verbose bool,
+) {
+	tr.lightClientAttack(action.validator, action.chain, LightClientLunaticAttack)
+}
+
+type LightClientAttackType string
+
+const (
+	LightClientEquivocationAttack LightClientAttackType = "Equivocation"
+	LightClientAmnesiaAttack      LightClientAttackType = "Amnesia"
+	LightClientLunaticAttack      LightClientAttackType = "Lunatic"
+)
+
+func (tr TestRun) lightClientAttack(
+	validator validatorID,
+	chain chainID,
+	attackType LightClientAttackType,
+) {
+	if !tr.useCometmock {
+		log.Fatal("light client attack is only supported with CometMock")
+	}
+	validatorAddress := tr.GetValidatorAddress(chain, validator)
+
+	method := "cause_light_client_attack"
+	params := fmt.Sprintf(`{"private_key_address":"%s", "misbehaviour_type": "%s"}`, validatorAddress, attackType)
+
+	address := tr.getQueryNodeRPCAddress(chain)
+
+	tr.curlJsonRPCRequest(method, params, address)
+	tr.waitBlocks(chain, 1, 10*time.Second)
+}
+
 type assignConsumerPubKeyAction struct {
 	chain          chainID
 	validator      validatorID
--- a/tests/e2e/actions.go
+++ b/tests/e2e/actions.go
@@ -1828,6 +1828,8 @@ func (tr TestRun) invokeDoublesignSlash(
 
 // Cause light client attack evidence for a certain validator to appear on the given chain.
 // The evidence will look like the validator equivocated to a light client.
+// See https://github.com/cometbft/cometbft/tree/main/spec/light-client/accountability
+// for more information about light client attacks.
 type lightClientEquivocationAttackAction struct {
 	validator validatorID
 	chain     chainID
@@ -1842,6 +1844,8 @@ func (tr TestRun) lightClientEquivocationAttack(
 
 // Cause light client attack evidence for a certain validator to appear on the given chain.
 // The evidence will look like the validator tried to perform an amnesia attack.
+// See https://github.com/cometbft/cometbft/tree/main/spec/light-client/accountability
+// for more information about light client attacks.
 type lightClientAmnesiaAttackAction struct {
 	validator validatorID
 	chain     chainID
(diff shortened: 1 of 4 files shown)


tests/e2e/actions.go
```diff
@@ -1826,6 +1826,73 @@ func (tr TestRun) invokeDoublesignSlash(
...
+}
+
+func (tr TestRun) lightClientAmnesiaAttack(
+	action lightClientAmnesiaAttackAction,
+	verbose bool,
+) {
+	tr.lightClientAttack(action.validator, action.chain, LightClientAmnesiaAttack)
+}
+
+type lightClientLunaticAttackAction struct {
```
insumity, line 1857: nit: previous 2 light client attack actions have a comment before them (e.g., `// The evidence will look like the validator tried to perform an amnesia attack.`) maybe add one here as well for consistency.

tests/e2e/actions.go
```diff
@@ -1826,6 +1826,73 @@ func (tr TestRun) invokeDoublesignSlash(
 	}
 }
 
+// Cause light client attack evidence for a certain validator to appear on the given chain.
+// The evidence will look like the validator equivocated to a light client.
```
insumity, line 1830: nit: maybe it makes sense to also include in the comment a pointer on what exactly is this attack about (e.g., to [CometBFT docs](https://docs.cometbft.com/main/spec/light-client/accountability/#the_misbehavior_of_faulty_validators)).

tests/e2e/actions.go
```diff
@@ -1826,6 +1826,73 @@ func (tr TestRun) invokeDoublesignSlash(
...
+	chain chainID,
+	attackType LightClientAttackType,
+) {
+	if !tr.useCometmock {
+		log.Fatal("light client attack is only supported with CometMock")
+	}
+	validatorAddress := tr.GetValidatorAddress(chain, validator)
+
+	method := "cause_light_client_attack"
+	params := fmt.Sprintf(`{"private_key_address":"%s", "misbehaviour_type": "%s"}`, validatorAddress, attackType)
```
insumity, line 1888: Just to verify: due to the `private_key_address` I would assume that the `validatorAddress` is created based on the private key (?) This would be unexpected to me because I assume addresses always stem from hashes of public keys but by looking at [getValidatorKeyAddressFromString](https://github.com/cosmos/interchain-security/blob/cca008d856e3ffc60ec1a486871d0faa702abe26/tests/e2e/actions.go#L1576) this is not clear to me.

tests/e2e/main.go
```diff
@@ -56,6 +56,12 @@ var (
 			description: `This is like the happy path, but skips steps
 that involve starting or stopping nodes for the same chain outside of the chain setup or teardown.
 This is suited for CometMock+Gorelayer testing`,
+		},
+		"light-client-attack": {
+			testRun: DefaultTestRun(), steps: lightClientAttackSteps,
+			description: `This is like the short happy path, but uses light client attacks instead of double sign slashing.
```
insumity, line 62: nit: maybe remove the `double sign slashing` wording for clarity, because as you have in the code above a light client attack could be an equivocation light client attack and hence `double sign`?

tests/e2e/steps_light_client_attack.go
```diff
@@ -0,0 +1,130 @@
...
+					},
+				},
+			},
+		},
+		{
+			// consumer double sign
+			// provider will only log the double sign slash
+			// stepsSubmitEquivocationProposal will cause the double sign slash to be executed
+			action: lightClientEquivocationAttackAction{
+				chain:     chainID("consu"),
```
insumity, line 60: nit: can we use `consumerName` instead of `"consu"` here?

tests/e2e/steps_light_client_attack.go
```diff
@@ -0,0 +1,130 @@
...
+						validatorID("bob"):   500,
+						validatorID("carol"): 0, // tombstoning visible on consumerName
+					},
+				},
+			},
+		},
+		{
+			// consumer double sign
+			// provider will only log the double sign slash
+			// stepsSubmitEquivocationProposal will cause the double sign slash to be executed
```
insumity, line 58: We do not seem to execute `stepsSubmitEquivocationProposal` in `lightClientAttackSteps`. Is this to be expected?

tests/e2e/steps_light_client_attack.go
```diff
@@ -0,0 +1,130 @@
...
+						validatorID("alice"): 509,
+						validatorID("bob"):   500,
+						validatorID("carol"): 0, // tombstoning visible on consumerName
+					},
+				},
+			},
+		},
+		{
+			// consumer double sign
+			// provider will only log the double sign slash
```
insumity, line 57: I do not fully understand the following steps. But by reading the comments I would have expected that we have a consumer double sign, then some packet relaying from consumer to provider that would lead to some power change on the provider. In the following steps we have no power change on neither the provider, nor the consumer. Is this to be expected?

tests/e2e/actions.go
p-offtermatt: Included, thanks for the suggestion

tests/e2e/actions.go
p-offtermatt: Done, thanks!

tests/e2e/actions.go
```diff
@@ -1826,6 +1826,73 @@ func (tr TestRun) invokeDoublesignSlash(
...
+	validator validatorID,
+	chain chainID,
+	attackType LightClientAttackType,
+) {
+	if !tr.useCometmock {
+		log.Fatal("light client attack is only supported with CometMock")
+	}
+	validatorAddress := tr.GetValidatorAddress(chain, validator)
+
+	method := "cause_light_client_attack"
```
MSalopek, line 1895: Can we somehow get the information about `cometmock` methods into the interchain-security repo?

Maybe just keep a .md with all the available methods?

I'm trying to think how we could check what the invoked method actually does and need some advice on how to go about it

tests/e2e/steps_light_client_attack.go
MSalopek: There will be no power changes until the `relay` step happens.

Notice that double signs and ligh client attacks are followed by a relay step. No action can be performed on the provider until it receives the correct packet

tests/e2e/actions.go
smarshall-spitzbart: +1 on this, an easily accessible doc outlining what's going on under the hood for each method would be nice

tests/e2e/actions.go
smarshall-spitzbart: The typical naming I've seen is `address`, `public key`, and `private key`, I think this is how sdk does things at least. 

Maybe we could avoid names containing both "key" and "address"?

p-offtermatt: To explain the reasoning behind the name, this is in the NODE_HOME/config/priv_validator_key.json file, and is the value of its "address" field. It is not an address of the validator, I renamed the variable to make it a bit clearer, but I think private_key_address most accurately describes what this is - the address field in the private key.

tests/e2e/actions.go
p-offtermatt: Great suggestion, though I am a bit worried about these things going out of date if they are not linked to anything. wdyt about linking the CometMock repo? It documents each method

tests/e2e/main.go
p-offtermatt: Thanks, made the wording here a bit sharper to address this

tests/e2e/steps_light_client_attack.go
p-offtermatt: Great catch, thanks!

tests/e2e/steps_light_client_attack.go
p-offtermatt: Good catch. This was also missing in the short-happy-path.
It didn't make the test cases wrong, just missed testing a part of the functionality.
It seems ok to add now, I'll push a change adding this

tests/e2e/actions.go
smarshall-spitzbart: Linking to the cometmock repo seems solid to me 👍 

Might be helpful to have a methods.md file or similar where you can link each method used in this repo to the appropriate md section in the cometmock repo

smarshall-spitzbart (approved):
great work!

PR #1262, opened 2023-09-05
fix: remove addr validation for provider fee pool addr param

NeverHappened:
Validation was incorrect: we validated provider address with rules of consumer chain.
(Param with key `ProviderFeePoolAddrStr`)

Need to remove this validation for testing purposes since we cannot change this param to any correct address now.

### DIFF
--- a/x/ccv/consumer/types/params_test.go
+++ b/x/ccv/consumer/types/params_test.go
@@ -30,10 +30,6 @@ func TestValidateParams(t *testing.T) {
 			ccvtypes.NewParams(true, 5, "badchannel/", "", 5, 1005, "0.5", 1000, 24*21*time.Hour, "0.05", []string{"untrn"}, []string{"uatom"}), false,
 		},
 		{
-			"custom invalid params, provider fee pool addr string",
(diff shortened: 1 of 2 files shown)


mpoke:
@NeverHappened Thanks for opening the PR. Could you please provide more context of why the current validation is incorrect? What do you mean by "with rules of consumer chain"?

NeverHappened:
I mean that this validation runs on the consumer chain code, for example on neutron chain. But the address is in the cosmoshub address format.
So if we try to validate any correct address in the form of "cosmos....", it will try to validate it by neutron chain rules (expects "neutron..." address form, and always fail.
For it see how function [AccAddressFromBech32](https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/types/address.go#L197) works. `GetConfig().GetBech32AccountAddrPrefix()` will return "neutron" in this case, as I understand that's just the limitation of cosmos-sdk, because it stores config for these things globally. Validation of address like "cosmos..." with expected "neutron..." prefix will always fail.

Hope it's clear!

mpoke (approved):
LGTM. Nice catch @NeverHappened.

p-offtermatt (approved):
LGTM,. fix looks good

Issue #1132, opened 2023-07-12
make e2e it's own module (de-couple from ICS lib and versioning scheme)


MSalopek:
More information is in the EPIC.

mpoke:
Closing as the objective should be to deprecate existing e2e tests in favor of the interchain tests (which are their own gomodule).

PR #1273, opened 2023-09-07
tests: separate e2e tests into a standalone module

MSalopek:
## Description

Closes: #1132 

e2e tests were moved outside `/tests` directory and made independent of the rest of the repository.

The `e2e` tests can still be used normally, no workflows were broken and there's no impact on the `Makefile` commands.

Some small refactors were made along the way.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


README.md
```diff
@@ -28,6 +28,8 @@ brew install jq
 
 **Installing and running binaries**
 
+This repository consists of multiple go modules. In order to have a smooth development experience, please add a `go.work` file. It is enough to copy and rename the `go.work.example` file.
```
p-offtermatt, line 31: The go.work file is also in this PR, this makes me think it should not. Should we remove the go.work file itself from the PR?

p-offtermatt (approved):
Nice work, did not think it would just work out of the box like this (well, as out-of-the-box as 44 files changed are). Just a small comment, but seems to work well.

README.md
MSalopek: Ideally, the `go.work` file should not be commited.

It is commited for now because the testing pipelines fail if you don't commit a go work file. The `go` tool sees that there are 2 modules but does not know how to build and execute normal tests without instructions from `go.work`.

This can also be fixed by some `make` magic, e.g.:
```bash
# create a go.work file on the fly by renaming go.work.example
mv ./go.work.example ./go.work
go test (...)
```

I opted to not do that now since it would be more confusing than just commiting the file directly.

Notice how none of the `make` commands are actually changed, they just point to a different path (`./e2e` instead of `./testing/e2e`).

Makefile
```diff
@@ -23,59 +23,59 @@ test-short:
...
 
 # run full E2E tests in sequence (including multiconsumer) using latest tagged gaia
 test-gaia-e2e:
-	go run ./tests/e2e/... --include-multi-consumer --use-gaia
+	go run ./e2e/... --include-multi-consumer --use-gaia
 
 # run only happy path E2E tests using latest tagged gaia
 test-gaia-e2e-short:
-	go run ./tests/e2e/... --tc happy-path --use-gaia
+	go run ./e2e/... --happy-path-only --use-gaia
```
bermuell, line 55:
```suggestion
	go run ./e2e/... --tc happy-path --use-gaia
```

Makefile
```diff
@@ -23,59 +23,59 @@ test-short:
...
 # usage: GAIA_TAG=v9.0.0 make test-gaia-e2e-tagged
 test-gaia-e2e-tagged:
-	go run ./tests/e2e/... --include-multi-consumer --use-gaia --gaia-tag $(GAIA_TAG)
+	go run ./e2e/... --include-multi-consumer --use-gaia --gaia-tag $(GAIA_TAG)
 
 # run only happy path E2E tests using latest tagged gaia
 # usage: GAIA_TAG=v9.0.0 make test-gaia-e2e-short-tagged
 test-gaia-e2e-short-tagged:
-	go run ./tests/e2e/... --tc happy-path --use-gaia --gaia-tag $(GAIA_TAG)
+	go run ./e2e/... --happy-path-only --use-gaia --gaia-tag $(GAIA_TAG)
```
bermuell, line 69: same here, what is the reason to go back to the 'old' flag and not using --tc
if I'm not mistaken that's not even supported anymore. did you test that?

MSalopek: That test has not been working for a while. Gaia is on v45, the ICS main is on v47.
Since e2e was coupled to the ICS version, some of the commands to execute against a gaia node are not valid and cannot be executed. Notice that most of that was intended for `gaia v9`.

README.md
p-offtermatt: Ah, makes sense. Then maybe let's change that line in the README to account for the fact that the go.work file is in the repo already and keep it like this for now?

Makefile
MSalopek: Changed, thanks!

bermuell (approved):

go.work
```diff
@@ -0,0 +1,6 @@
+go 1.20
```
smarshall-spitzbart, line 1: Aren't these in the gitignore?

smarshall-spitzbart: Nvm addressed in other comment

smarshall-spitzbart (approved):
LGTM

MSalopek:
Drafting until v50 is in to mitigate some risks

MSalopek:
Closing as the PR has decayed and takes a lot of work to get it working again.
It can be used as an example at a later time if we wish to proceed with the work.

Issue #1274, opened 2023-09-07
Implement slashing functionality on the provider (ADR-013)


insumity:
## Problem
Implement what is outlined in this [ADR](https://github.com/cosmos/interchain-security/pull/1252).

## Closing criteria
Slashing functionality implemented on the provider for consumer infractions. Also testing added as well.

PR #1275, opened 2023-09-07
feat: implement slashing functionality on the provider chain (ADR-013)

insumity:
## Description

Closes: #1274

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/integration/double_vote.go
```diff
@@ -159,6 +159,9 @@ func (s *CCVTestSuite) TestHandleConsumerDoubleVoting() {
 	consuAddr := types.NewConsumerConsAddress(sdk.ConsAddress(consuVal.Address.Bytes()))
 	provAddr := s.providerApp.GetProviderKeeper().GetProviderAddrFromConsumerAddr(s.providerCtx(), s.consumerChain.ChainID, consuAddr)
 
+	validator, _ := s.providerApp.GetTestStakingKeeper().GetValidator(s.providerCtx(), provAddr.ToSdkConsAddr().Bytes())
```
insumity, line 162: This is a simple test that some type of slashing takes place when we call `HandleConsumerDoubleVoting`.

Ideally, I would also like to verify that undelegations and redelegations are slashed as well with something similar to:
```
bondAmt := sdk.NewInt(10000)
delAddr := s.providerChain.SenderAccount.GetAddress()
_, _ = delegateAndUndelegate(s, delAddr, bondAmt, int64(20))

unbs := s.providerApp.GetTestStakingKeeper().GetUnbondingDelegationsFromValidator(provCtx, valAddress)
// check `unbs` have less tokens
```

However, I couldn't find a way to "clean" undelegations/redelegations for the next test case inside the `for _, tc := range testCases {` loop and this confused things a bit as we traverse through the test cases. I'll probably break the tests into 3 tests where i) all tests that have `expPass == false`, ii) "valid double voting evidence 1 - should pass", and iii) "valid double voting evidence 2 - should pass".

x/ccv/provider/keeper/double_vote.go
```diff
@@ -33,8 +34,18 @@ func (k Keeper) HandleConsumerDoubleVoting(
 		types.NewConsumerConsAddress(sdk.ConsAddress(evidence.VoteA.ValidatorAddress.Bytes())),
 	)
 
-	// execute the jailing
-	k.JailValidator(ctx, providerAddr)
+	k.SlashValidator(ctx, providerAddr)
+	k.JailAndTombstoneValidator(ctx, providerAddr)
+
+	// FIXME: verify the following values
```
insumity, line 40: We currently do not store double voting or misbehaviour evidence. It would be nice to store the evidence when we receive it. However, it's not yet clear how to go about it because `func (EvidenceKeeper) SetEvidence(ctx types.Context, evidence exported.Evidence)` expected evidence is of type `exported.Evidence`. We cannot use `DuplicateVodeEvidence` or `LightClientAttackEvidence` from [evidence.proto](https://github.com/cosmos/cosmos-sdk/blob/5eaa7b8d3c50eefc5dd95210c38c84c1e829509f/proto/tendermint/types/evidence.proto#L4) because they do not implement the `exported.Evidence` interface, e.g., 
```
Cannot use 'tmtypes.LightClientAttackEvidence{}' (type tmtypes.LightClientAttackEvidence) as the type exported.Evidence Type does not implement 'exported.Evidence' as some methods are missing: Reset() ProtoMessage() Route() string
```
We could do something similar to what is shown down below but it doesn't seem so useful. Even today, when querying:
```
gaiad query evidence
evidence:
- '@type': /cosmos.evidence.v1beta1.Equivocation
  consensus_address: cosmosvalcons18eg64zpqxmt6wru5705xg94hvqdh3hvv94e3cm
  height: "12842605"
  power: "225822"
  time: "2022-11-14T12:26:53.319800141Z"
- '@type': /cosmos.evidence.v1beta1.Equivocation
  consensus_address: cosmosvalcons16tnak7apushwznnd3wtku8gm0rt3xytz6ut006
  height: "16172865"
  power: "76511"
  time: "2023-07-17T11:48:22.657230386Z"
pagination:
  next_key: null
  total: "0"
```
the results do not seem that helpful because we do not know what votes led to the equivocation, etc.

Would it make sense to extend the `DuplicateVodeEvidence` or `LightClientAttackEvidence` so they implement the `exported.Evidence` interface?

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -3,26 +3,24 @@ package keeper
...
 	"github.com/cosmos/interchain-security/v2/x/ccv/provider/types"
 )
 
-// JailValidator jails the validator with the given provider consensus address
-// Note that the tombstoning is temporarily removed until we slash validator
-// for double signing on a consumer chain, see comment
-// https://github.com/cosmos/interchain-security/pull/1232#issuecomment-1693127641.
-func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) {
+// JailAndTombstoneValidator jails and tombstones the validator with the given provider consensus address
+func (k Keeper) JailAndTombstoneValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) {
```
insumity, line 11: Renamed to `JailAndTombstoneValidator` because this was what this method was doing [initially](https://github.com/cosmos/interchain-security/commit/f168b9b4a5d15f3641d9cb09182b923f7688aead). However, I intend to break it into 2 methods: `JailValidator` and `TombstoneValidator`.

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +29,68 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+	k.slashingKeeper.Tombstone(ctx, providerAddr.ToSdkConsAddr())
+}
+
+func (k Keeper) ComputePowerToSlash(undelegations []stakingtypes.UnbondingDelegation,
+	redelegations []stakingtypes.Redelegation, power int64, powerReduction sdk.Int) int64 {
+
+	// compute the total numbers of tokens currently being undelegated
+	undelegationsInTokens := sdk.NewInt(0)
+	for _, u := range undelegations {
+		for _, entry := range u.Entries {
```
mpoke, line 51: Must exclude entries that reached maturity, see https://github.com/cosmos/cosmos-sdk/blob/0af2f4da004cbea6414a8bad56e8bdcd45badf1e/x/staking/keeper/slash.go#L200. Maybe it would be better to just use the `SlashUnbondingDelegation` method from the staking module, but calling it from a cached context.

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +29,68 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+			undelegationsInTokens = undelegationsInTokens.Add(entry.InitialBalance)
+		}
+	}
+
+	// compute the total numbers of tokens currently being redelegated
+	redelegationsInTokens := sdk.NewInt(0)
+	for _, r := range redelegations {
+		for _, entry := range r.Entries {
```
mpoke, line 59: ditto

x/ccv/provider/keeper/double_vote.go
insumity: My assumption is that if we do not store evidence, then it would be difficult to explain to a validator for which infraction they got slashes. We would also be consistent with what Cosmos SDK does today [when handling an equivocation](https://github.com/cosmos/cosmos-sdk/blob/5eaa7b8d3c50eefc5dd95210c38c84c1e829509f/x/evidence/keeper/infraction.go#L150).

mpoke: The evidence can be query on the consumer chain as it will still be part of the consumer chain blocks and pass to the evidence module on the consumer.

insumity: Evidence would be retrieved from the _application state_ of the consumer chain, right? Assuming the application of the consumer chain is not trustworthy, it could have removed the evidence, etc.
Let me know if there are any concerns with the above scenario. For now, I'm removing the storing of evidence on the provider chain.

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -41,10 +42,22 @@ func (k Keeper) HandleConsumerMisbehaviour(ctx sdk.Context, misbehaviour ibctmty
 			misbehaviour.Header1.Header.ChainID,
 			types.NewConsumerConsAddress(sdk.ConsAddress(v.Address.Bytes())),
 		)
-		k.JailValidator(ctx, providerAddr)
+		k.SlashValidator(ctx, providerAddr)
+		k.JailAndTombstoneValidator(ctx, providerAddr)
 		provAddrs = append(provAddrs, providerAddr)
 	}
 
+	// FIXME: it would be nice to set the evidence but not sure what this would mean
```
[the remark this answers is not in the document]
insumity: Resolving this to keep discussion in one [place](https://github.com/cosmos/interchain-security/pull/1275/files#r1322045833).

tests/e2e/steps_consumer_misbehaviour.go
```diff
@@ -230,6 +230,10 @@ func stepsCauseConsumerMisbehaviour(consumerName string) []Step {
 						validatorID("alice"): 511,
 						validatorID("bob"):   20,
 					},
+					RepresentativePowers: &map[validatorID]uint{
```
insumity, line 233: Using `RepresentativePowers` was the only way I found to verify that slashing indeed takes place.
Using `ValPowers` was not useful because after a validator gets jailed, the validator's voting power is `0`. I also could not `unjail` and then check `ValPowers` similarly to what is done in [steps_multi_consumer_downtime.go](https://github.com/cosmos/interchain-security/blob/67bb558fcb841b17a4f54fd2dbd191210ecd444f/tests/e2e/steps_multi_consumer_downtime.go#L338) because in contrast to a downtime test, for misbehaviour & equivocation we also we also tombstone the validator.

x/ccv/provider/keeper/punish_validator.go
insumity: Good catch. Fixed it. Current approach uses the same condition to skip entries as the one in [Slash](https://github.com/cosmos/cosmos-sdk/blob/0af2f4da004cbea6414a8bad56e8bdcd45badf1e/x/staking/keeper/slash.go#L200).

I believe that calling `SlashUnbondingDelegation` directly might complicate things because when we call the staking's module `Slash` method,  `Slash` calls `SlashUnbondingDelegation` and `SlashRedelegation` so we would need to figure out a way to avoid slashing twice the same undelegations and redelegations.

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +31,76 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+
+	return power + undelegationsAndRedelegationsInPower
+}
+
+// SlashValidator slashes validator with `providerAddr`
+func (k Keeper) SlashValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) {
+	logger := k.Logger(ctx)
+
+	val, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.ToSdkConsAddr())
+	if !found {
```
sainoe, line 88: Why the case where the validator is unbonded isn't handled?
In this case, the slash method will panic, see:
 see https://github.com/cosmos/cosmos-sdk/blob/0af2f4da004cbea6414a8bad56e8bdcd45badf1e/x/staking/keeper/slash.go#L61-L63C1

insumity: Nice catch! Fixed it and added a test. Thanks.

docs/docs/adrs/adr-005-cryptographic-equivocation-verification.md
```diff
@@ -0,0 +1,105 @@
+---
```
insumity, line 1: cherry picked this file so markdown link checker doesn't complain.

sainoe (approved):

sainoe:
Nice work!

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +30,73 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+// `redelegations`, as well as the current `power` of the validator
+func (k Keeper) ComputePowerToSlash(ctx sdk.Context, validator stakingtypes.Validator, undelegations []stakingtypes.UnbondingDelegation,
+	redelegations []stakingtypes.Redelegation, power int64, powerReduction sdk.Int,
+) int64 {
+	// compute the total numbers of tokens currently being undelegated
+	undelegationsInTokens := sdk.NewInt(0)
+
+	cachedCtx, _ := ctx.CacheContext()
+	for _, u := range undelegations {
+		amountSlashed := k.stakingKeeper.SlashUnbondingDelegation(cachedCtx, u, 0, sdk.NewDec(1))
```
mpoke, line 55: Why is the slash factor `1`?

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +30,73 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+		undelegationsInTokens = undelegationsInTokens.Add(amountSlashed)
+	}
+
+	// compute the total numbers of tokens currently being redelegated
+	redelegationsInTokens := sdk.NewInt(0)
+	for _, r := range redelegations {
+		amountSlashed := k.stakingKeeper.SlashRedelegation(cachedCtx, validator, r, 0, sdk.NewDec(1))
```
mpoke, line 62: ditto: slashFactor == 1?

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +30,73 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+	undelegationsInTokens := sdk.NewInt(0)
+
+	cachedCtx, _ := ctx.CacheContext()
+	for _, u := range undelegations {
+		amountSlashed := k.stakingKeeper.SlashUnbondingDelegation(cachedCtx, u, 0, sdk.NewDec(1))
+		undelegationsInTokens = undelegationsInTokens.Add(amountSlashed)
+	}
+
+	// compute the total numbers of tokens currently being redelegated
+	redelegationsInTokens := sdk.NewInt(0)
```
mpoke, line 60: nit: We can use the same variable from above (e.g., `undelegationsInTokens`) and avoid the addition below.

x/ccv/provider/keeper/punish_validator.go
insumity: `SlashUnbondingDelegation` returns the [total slashed amount](https://github.com/cosmos/cosmos-sdk/blob/v0.47.0/x/staking/keeper/slash.go#L205) based on the provided `slashFactor`. When we compute the power we slash against, we use the [total power of all the undelegations](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-013-equivocation-slashing.md). We use a slash factor of `1.0` here to slash everything and hence get all the undelegation tokens.

x/ccv/provider/keeper/punish_validator.go
insumity: Same reason as on why we used `1` in `SlashUnbondingDelegation`s.

x/ccv/provider/keeper/punish_validator.go
mpoke: That makes sense. Thanks

mpoke (approved):
Nice work @insumity

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -31,8 +30,73 @@ func (k Keeper) JailValidator(ctx sdk.Context, providerAddr types.ProviderConsAd
...
+
+// ComputePowerToSlash computes the power to be slashed based on the tokens in non-matured `undelegations` and
+// `redelegations`, as well as the current `power` of the validator
+func (k Keeper) ComputePowerToSlash(ctx sdk.Context, validator stakingtypes.Validator, undelegations []stakingtypes.UnbondingDelegation,
+	redelegations []stakingtypes.Redelegation, power int64, powerReduction sdk.Int,
+) int64 {
+	// compute the total numbers of tokens currently being undelegated
+	undelegationsInTokens := sdk.NewInt(0)
+
+	cachedCtx, _ := ctx.CacheContext()
```
p-offtermatt, line 53: Maybe instead of using this cached context here (or in addition to this), we could have wraps for SlashUnbondingDelegation and SlashRedelegation that just call those on cached contexts and call the wraps e.g. ComputeUnbondingDelegationPower - it's very subtle with the cached context and the slash factor of 1 here

x/ccv/provider/keeper/punish_validator_test.go
```diff
@@ -125,8 +139,345 @@ func TestJailValidator(t *testing.T) {
...
+								continue
+							}
+							sum = sum.Add(sdk.NewInt(r.InitialBalance.Int64()))
+						}
+						return sum
+					}).AnyTimes(),
+		)
+
+		actualPower := providerKeeper.ComputePowerToSlash(ctx, validator,
+			tc.undelegations, tc.redelegations, tc.power, tc.powerReduction)
```
p-offtermatt, line 333: Could you check here that afterwards, the undelegations and redelegations have not been slashed? Just to make explicit that ComputePowerToSlash does not slash on its own?

tests/e2e/steps_consumer_misbehaviour.go
p-offtermatt: Seems ok to me. Could rename it to something like StakedTokens instead of RepresentativePowers (looking at the code, I think that's what this actually is, right?)

p-offtermatt (commented):
Partial review. I think the subtlety in ComputePowerToSlash should be addressed, this is a bit tricky imo and we can improve it with more comments and/or splitting out some logic.

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -3,26 +3,25 @@ package keeper
...
+// JailAndTombstoneValidator jails and tombstones the validator with the given provider consensus address
+func (k Keeper) JailAndTombstoneValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) {
 	logger := k.Logger(ctx)
 
 	// get validator
 	val, ok := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.ToSdkConsAddr())
 	if !ok || val.IsUnbonded() {
-		logger.Error("validator not found or is unbonded", providerAddr.String())
+		logger.Error("validator not found or is unbonded", "provider consensus address", providerAddr.String())
 		return
```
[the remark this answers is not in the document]
insumity: Good point. Fixed it and we return the error instead.

Found the relevant [ErrNoValidatorForAddress](https://github.com/cosmos/cosmos-sdk/blob/v0.47.0/x/slashing/types/errors.go#L9) error for this one. For others, I couldn't find a relevant error in `cosmos-sdk` I ended up doing `fmt.Errorf` to creat errors.

x/ccv/provider/keeper/punish_validator.go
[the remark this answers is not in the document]
insumity: Added comment in the godoc, as well as above `cachedCtx`.

x/ccv/provider/keeper/punish_validator_test.go
insumity: I think it's hard to check this in the unit test here.

We could have something like:
```
tokensBeforeCall := validator.GetTokens()
delegatorSharesBeforeCall := validator.GetDelegatorShares()

actualPower := providerKeeper.ComputePowerToSlash(ctx, validator,
	tc.undelegations, tc.redelegations, tc.power, tc.powerReduction)

// safeguard check that validator remains unmodified after a call to `ComputePowerToSlash`
// `ComputePowerToSlash` only computes the power and does not modify the state of the system in any way
require.Equal(t, tokensBeforeCall, validator.GetTokens())
require.Equal(t, delegatorSharesBeforeCall, validator.GetDelegatorShares())
```
but I do not feel it actually is that useful because we actually would like to check the underlying undelegations. 
We could build an integration test for this but I'm not sure it's worth the added complexity. Additionally, this is implicitly tested in the E2E tests through the `RepresentativePowers`. If somebody were to modify the `ComputePowerToSlash` so that it actually slashes as well, then the E2E test would fail.

p-offtermatt: Ok, I see. I thought this would be easy, but I see it's relatively complex.
Then I think it's ok to just have it in the e2e test.

p-offtermatt (approved):
Thanks, the changes look great! Approving

PR #1303, opened 2023-09-14
Fix `build` in makefile

sainoe:
`make build` attempted to write to the `/` root directory as
the current directory path was missing.


p-offtermatt (approved):
Thank you!

smarshall-spitzbart (approved):
LGTM and works on my machine, thanks!

PR #1265, opened 2023-09-06
test: Add trace input via JSON

p-offtermatt:
## Description

Closes: #XXXX

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Provided a link to the relevant issue or specification
- [ ] Reviewed "Files changed" and left comments if necessary 
- [ ] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] Confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Confirmed all author checklist items have been addressed
- [ ] Confirmed that this PR does not change production code

PR #1314, opened 2023-09-19
test: Add json marshal/unmarshal for test traces

p-offtermatt:
## Description

Closes: Part of #1265 

 The PR introduces a trace format that is just a JSON serialization of a step slice. This means it makes use of all the actions we already have.
  * tests/e2e/tracehandler_testdata/ just contains the serializations of a few existing steps. These can be regenerated by this test: https://github.com/cosmos/interchain-security/blob/e2175c206ac1cb8a641fa914b179414d609a51db/tests/e2e/trace_handlers_test.go#L84
  * The serialization has two complicated aspects: Actions and proposals
    * JSON does not add types, so marshalling/unmarshalling interfaces is hard because they can contain different types
    * See http://gregtrowbridge.com/golang-json-serialization-with-interfaces/
    * I solve this by, on marshal/unmarshal, adding/reading an extra type string that contains the type of the action/proposal. unfortunately, these have to be maintained when we add new actions, but I don't see any easier way.
    * This is the reason for these two massive switch-cases:
      * https://github.com/cosmos/interchain-security/blob/e2175c206ac1cb8a641fa914b179414d609a51db/tests/e2e/json_utils.go#L50
      * https://github.com/cosmos/interchain-security/blob/e2175c206ac1cb8a641fa914b179414d609a51db/tests/e2e/json_utils.go#L403

To review this...
* check that in the _rapid_test.go files, the structs are built exhaustively, i.e. I didn't forget to add any fields that aren't populated now
* check that the switch-cases are exhaustive, i.e. no actions/proposals that exist are missed by the switch-case in json_utils.go
* take a look at the tracehandlers_testdata folder and see the json format, check if you see any obvious defects. viewing these is much nicer after either formatting them, putting them into a json viewer, ...
* check out the meat of the changes: the actual ser/de is done in tests/e2e/json_parser.go, tests/e2e/json_writer.go and tests/e2e/json_utils.go

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/e2e/json_parser.go
```diff
@@ -0,0 +1,35 @@
...
+	ReadTraceFromFile(filepath string) ([]Step, error)
+}
+
+// JSONParser is a simple parser that reads steps by unmarshalling from a file.
+type JSONParser struct{}
+
+var GlobalJSONParser = JSONParser{}
+
+func (parser JSONParser) ReadTraceFromFile(path string) ([]Step, error) {
+	// Open the JSON file and read into a bite array
```
insumity, line 20: s/bite/byte

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+	var tmp struct {
+		ActionType string
+		Action     json.RawMessage
+		State      State
+	}
+	if err := json.Unmarshal(data, &tmp); err != nil {
+		return err
+	}
+
+	action, err := UnmarshalMapToActionType(tmp.Action, tmp.ActionType)
```
insumity, line 37: Just to note but it seems peculiar to me that we need a specific `UnmarshalMapToActionType` but when marshalling in `MarshalJSON()` we do not need any `marshalAction` function.

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+			return a, nil
+		}
+	case "main.startRelayerAction":
+		var a startRelayerAction
+		err := json.Unmarshal(rawAction, &a)
+		if err == nil {
+			return a, nil
+		}
+	case "main.slashThrottleDequeue":
+		var a slashThrottleDequeue
```
insumity, line 208: Can we refactor `slashThrottleDequeue` to be `slashThrottleDequeueAction` as well?

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+			return a, nil
+		}
+	case "main.lightClientLunaticAttackAction":
+		var a lightClientLunaticAttackAction
+		err := json.Unmarshal(rawAction, &a)
+		if err == nil {
+			return a, nil
+		}
+	default:
+		return nil, fmt.Errorf("unknown action name: %s", actionTypeString)
```
insumity, line 262: nit: s/action name/action type

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+		proposalsWithTypes := make(map[uint]ProposalAndType)
+		for k, v := range *c.Proposals {
+			proposalsWithTypes[k] = ProposalAndType{v, reflect.TypeOf(v).String()}
+		}
+		chainStateWithProposalTypes.Proposals = &proposalsWithTypes
+	}
+	return json.Marshal(chainStateWithProposalTypes)
+}
+
+// unmarshal the ChainStateWithProposalTypes into a ChainState by removing the type info from the proposals and getting back standard proposals
```
insumity, line 314: Use the name of the method as the first thing in the comment for godocs? Similarly above.

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+		RepresentativePowers           *map[ValidatorID]uint
+		Params                         *[]Param
+		Rewards                        *Rewards
+		ConsumerChains                 *map[ChainID]bool
+		AssignedKeys                   *map[ValidatorID]string
+		ProviderKeys                   *map[ValidatorID]string
+		ConsumerChainQueueSizes        *map[ChainID]uint
+		GlobalSlashQueueSize           *uint
+		RegisteredConsumerRewardDenoms *[]string
+		Proposals                      *map[uint]ProposalAndType // the only thing changed from the real ChainState
```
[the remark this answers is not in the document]
p-offtermatt: Exactly, I need to change the type of proposals. Happy about other suggestions, but I didn't find a good way

tests/e2e/json_writer.go
```diff
@@ -0,0 +1,30 @@
...
+	"encoding/json"
+	"os"
+)
+
+// TraceWriter is an interface for writers that write steps to files.
+type TraceWriter interface {
+	WriteTraceToFile(filepath string, trace []Step) error
+}
+
+// JSONWriter is a simple writer that simply marshals the array of Step objects.
```
insumity, line 13: nit: either "simple writer than marshals" or "a write that simply marshals"

tests/e2e/json_writer.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+)
+
+// TraceWriter is an interface for writers that write steps to files.
```
insumity, line 8: Do we need a comment on whether this would override the file with the steps, append the steps, etc.?

tests/e2e/json_utils.go
p-offtermatt: It is implicit: See how in the steps, I am modifying the definition of step to include a different action.
It might be doable to put the actionType inside the action, too, and modify the way actions are unmarshalled and marshalled, would you prefer that?

tests/e2e/json_writer.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// JSONWriter is a simple writer that simply marshals the array of Step objects.
+// To identify which type of action is being used, we add a field to the Step struct.
+type JSONWriter struct{}
+
+var GlobalJSONWriter = JSONWriter{}
+
+func (writer JSONWriter) WriteTraceToFile(filepath string, trace []Step) error {
+	// collect missing action types, if any. this way, we can provide a more helpful error message.
```
insumity, line 20: What does this comment refer to? 
nit: s/this way/**T**his way

tests/e2e/json_writer.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// workaround: we would keep a set, but go doesn't have sets.
```
insumity, line 22: I don't fully grasp this comment. This would be a workaround for what?
Could you use `map[X]bool` as a set?

tests/e2e/tracehandler_testdata/changeover.json
```diff
@@ -0,0 +1 @@
...s":null}}},{"ActionType":"main.unbondTokensAction","Action":{"Chain":"provi","Sender":"alice","UnbondFrom":"alice","Amount":1000000},"State":{"provi":{"ValBalances":null,"ValPowers":{"alice":510,"bob":500,"carol":500},"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null},"sover":{"ValBalances":null,"ValPowers":{"alice":511,"bob":500,"carol":500},"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null}}},{"ActionType":"main.relayPacketsAction","Action":{"ChainA":"provi","ChainB":"sover","Port":"provider","Channel":1},"State":{"sover":{"ValBalances":null,"ValPowers":{"alice":510,"bob":500,"carol":500},"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null}}}]
```
insumity, line 1: I formatted this in a text editors and skimmed through it. Didn't notice anything peculiar per se. However, I noticed that in some cases (e.g., the first 2 states) we have an `State` where `ValBalances != null` and in the next `State` we have `ValBalances == null`. Is this expected? I would have expected that if `ValBalances != null` for a state, then this would be the case for any state that follows.

tests/e2e/tracehandler_testdata/democracy.json
```diff
@@ -0,0 +1 @@
...sals":null}}},{"ActionType":"main.unjailValidatorAction","Action":{"Provider":"provi","Validator":"bob"},"State":{"democ":{"ValBalances":null,"ValPowers":{"alice":511,"bob":0,"carol":500},"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null},"provi":{"ValBalances":null,"ValPowers":{"alice":511,"bob":500,"carol":500},"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null}}},{"ActionType":"main.relayPacketsAction","Action":{"ChainA":"provi","ChainB":"democ","Port":"provider","Channel":0},"State":{"democ":{"ValBalances":null,"ValPowers":{"alice":511,"bob":500,"carol":500},"RepresentativePowers":{"alice":100500000,"bob":40000000},"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null}}}]
```
insumity, line 1: In here:
```
   {
        "ActionType": "main.relayPacketsAction",
        "Action":
        {
            "ChainA": "provi",
            "ChainB": "democ",
            "Port": "provider",
            "Channel": 0
        },
        "State":
        {
            "democ":
            {
                "ValBalances": null,
                "ValPowers":
                {
                    "alice": 511,
                    "bob": 500,
                    "carol": 500
                },
                "RepresentativePowers": null,
                "Params": null,
                "Rewards": null,
                "ConsumerChains": null,
                "AssignedKeys": null,
                "ProviderKeys": null,
                "ConsumerChainQueueSizes": null,
                "GlobalSlashQueueSize": null,
                "RegisteredConsumerRewardDenoms": null,
                "Proposals": null
            },
            "provi":
            {
                "ValBalances": null,
                "ValPowers":
                {
                    "alice": 511,
                    "bob": 0,
                    "carol": 500
                },
                "RepresentativePowers": null,
                "Params": null,
                "Rewards": null,
                "ConsumerChains": null,
                "AssignedKeys": null,
                "ProviderKeys": null,
                "ConsumerChainQueueSizes": null,
                "GlobalSlashQueueSize": null,
                "RegisteredConsumerRewardDenoms": null,
                "Proposals": null
            }
        }
    },
    {
        "ActionType": "main.relayPacketsAction",
        "Action":
        {
            "ChainA": "provi",
            "ChainB": "democ",
            "Port": "provider",
            "Channel": 0
        },
        "State":
        {
            "democ":
            {
                "ValBalances": null,
                "ValPowers":
                {
                    "alice": 511,
                    "bob": 0,
                    "carol": 500
                },
                "RepresentativePowers": null,
                "Params": null,
                "Rewards": null,
                "ConsumerChains": null,
                "AssignedKeys": null,
                "ProviderKeys": null,
                "ConsumerChainQueueSizes": null,
                "GlobalSlashQueueSize": null,
                "RegisteredConsumerRewardDenoms": null,
                "Proposals": null
            }
        }
    },
```
it seems you `relayPacketsAction` 2 steps in a row and nothing really changes. In the first case the `provi` state is stated as well but otherwise what's the difference?

tests/e2e/tracehandler_testdata/happyPath.json
```diff
@@ -0,0 +1 @@
...ctionType":"main.submitConsumerRemovalProposalAction","Action":{"Chain":"provi","From":"bob","Deposit":10000001,"ConsumerChain":"consu","StopTimeOffset":0},"State":{"provi":{"ValBalances":{"bob":9489999999},"ValPowers":null,"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":{"consu":true},"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":{"4":{"RawProposal":{"Deposit":10000001,"Chain":"consu","StopTime":0,"Status":"PROPOSAL_STATUS_VOTING_PERIOD"},"Type":"main.ConsumerRemovalProposal"}}}}},{"ActionType":"main.voteGovProposalAction","Action":{"Chain":"provi","From":["alice","bob","carol"],"Vote":["yes","yes","yes"],"PropNumber":4},"State":{"provi":{"ValBalances":{"bob":9500000000},"ValPowers":null,"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":{},"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":{"4":{"RawProposal":{"Deposit":10000001,"Chain":"consu","StopTime":0,"Status":"PROPOSAL_STATUS_PASSED"},"Type":"main.ConsumerRemovalProposal"}}}}}]
```
insumity, line 1: Maybe was in the previous file as well but noticed first in this one. Why are some actions capitalized after `main.` while others are not. For example `main.SendTokensAction` versus `main.delegateTokensAction`?

insumity: Also, there's a case with 2 `relayPacketsAction` in this file one after the other:
```
        "ActionType": "main.relayPacketsAction",
        "Action":
        {
            "ChainA": "provi",
            "ChainB": "consu",
            "Port": "provider",
            "Channel": 0
        },
```

tests/e2e/tracehandler_testdata/changeover.json
p-offtermatt: Important to note: null just means "do not check" for our e2e framework, so ValBalances != null followed by ValBalances == null just means "check this, and after that we don't care.
Wondering how that could be documented best, maybe once this format is stable, a PR should follow to create a README to document this?

insumity, line 1: Why not format the JSON files from the get-go?

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+	step.Action = action
+	step.State = tmp.State
+	return nil
+}
+
+// UnmarshalMapToActionType takes a JSON object and an action type and marshals into an object of the corresponding action.
+func UnmarshalMapToActionType(rawAction json.RawMessage, actionTypeString string) (interface{}, error) {
+	var err error
+	switch actionTypeString {
+	case "main.submitConsumerAdditionProposalAction":
```
insumity, line 51: I'm probably looking to an older local codebase on all the actions we have. I couldn't find any missing. To contrary, the following ones seem new:
```
main.submitParamChangeLegacyProposalAction
main.LegacyUpgradeProposalAction
main.waitUntilBlockAction
main.StartSovereignChainAction
main.lightClientEquivocationAttackAction
main.lightClientAmnesiaAttackAction
main.lightClientLunaticAttackAction
```

tests/e2e/state_rapid_test.go
```diff
@@ -0,0 +1,242 @@
...
+				Proposals:               &proposals,
+				ValPowers:               &valPowers,
+				RepresentativePowers:    &representativePowers,
+				Params:                  &params,
+				Rewards:                 &rewards,
+				ConsumerChains:          &consumerChains,
+				AssignedKeys:            &assignedKeys,
+				ProviderKeys:            &providerKeys,
+				ConsumerChainQueueSizes: &consumerChainQueueSizes,
+				GlobalSlashQueueSize:    &globalSlashQueueSize,
```
insumity, line 62: Do we need
```
RegisteredConsumerRewardDenoms *[]string
ClientsFrozenHeights           *map[string]clienttypes.Height
```
as well? I see `RegisteredConsumerRewardDenoms` exists in `json_utils.go`.

tests/e2e/action_rapid_test.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+
+	return nil
+}
+
+// This needs to be adjusted manually when new actions are added and should
+// include generators for all actions that are mentioned in main.go/runStep.
+func GetActionGen() *rapid.Generator[any] {
+	return rapid.OneOf(
+		GetStartSovereignChainActionGen().AsAny(),
```
insumity, line 59: What about `submitChangeRewardDenomsProposalAction`?

insumity:
Assume we add a new `Action` in the future. Based on this PR, I would assume we would need to modify a bunch of files to include this `Action`? What would be the process for that? Would it make sense to have a `README.md` file under `tests/e2e` to describe such a process (e.g., what files need to be changed, etc.)?

tests/e2e/tracehandler_testdata/happyPath.json
p-offtermatt: This is how they are written in the code, it just takes the struct names. They were inconsistent. If you think this should be fixed, we can make an issue, but I don't want to do it in this PR and make it even larger again

p-offtermatt:
> Assume we add a new `Action` in the future. Based on this PR, I would assume we would need to modify a bunch of files to include this `Action`? What would be the process for that? Would it make sense to have a `README.md` file under `tests/e2e` to describe such a process (e.g., what files need to be changed, etc.)?

yeah exactly. I'll draft up a README like that. I'll also put an issue in the backlog to look into setting up a static analyzer that could flag this. From all I can tell, the current solution is unfortunately the best I could find, so supporting it to be more smooth would probably be desirable.

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+	c.ConsumerChainQueueSizes = chainStateWithProposalTypes.ConsumerChainQueueSizes
+	c.GlobalSlashQueueSize = chainStateWithProposalTypes.GlobalSlashQueueSize
+	c.RegisteredConsumerRewardDenoms = chainStateWithProposalTypes.RegisteredConsumerRewardDenoms
+
+	if chainStateWithProposalTypes.Proposals != nil {
+		proposals := make(map[uint]Proposal)
+		for k, v := range *chainStateWithProposalTypes.Proposals {
+			proposal, err := UnmarshalProposalWithType(v.RawProposal, v.Type)
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
p-offtermatt: Good call, thanks

tests/e2e/json_utils.go
p-offtermatt: Thanks, good suggestion. Done, d1465053a4495a90af3a12c92842ecb2e3a5ec21

tests/e2e/json_writer.go
p-offtermatt: Thanks, 720962834ecb08ed3c124c8c4b1ce5dcb0cc8d5c

tests/e2e/json_writer.go
p-offtermatt: Sure, doesn't hurt I think
b4e35e35127bb3db0af7f780c72e05c37642be67

tests/e2e/json_writer.go
p-offtermatt: old comment that I did not remove, my bad! 55d268659

tests/e2e/json_writer.go
p-offtermatt: old comment that I did not remove!
55d268659

tests/e2e/json_utils.go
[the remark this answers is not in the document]
p-offtermatt: My feeling is that it is fair to have the chainState be arbitrary if this returns an error, in particular whatever it is at this point is fine.

tests/e2e/json_utils.go
p-offtermatt: Yes, that's an old codebase. This should be up-to-date with the actions here:
https://github.com/cosmos/interchain-security/blob/55d2686593b33384b065937ed5ed9ee5304f0536/tests/e2e/main.go#L193

tests/e2e/tracehandler_testdata/democracy.json
p-offtermatt: This is just the marshalled democracy steps object.
But to explain what happens, look closely at the democ state - in the first one it has "ValPowers":
                {
                    "alice": 511,
                    "bob": 500,
                    "carol": 500
                },
                
               in the second one 
               
               "ValPowers":
                {
                    "alice": 511,
                    "bob": 0,
                    "carol": 500
                },

tests/e2e/action_rapid_test.go
p-offtermatt: submitChangeRewardDenomsProposalAction yes, good catch.
startConsumerEvidenceDetectorAction no, this version of the codebase does not have it.

p-offtermatt: 1168acab2

tests/e2e/tracehandler_testdata/start_provider_chain.json
```diff
@@ -0,0 +1 @@
+[{"ActionType":"main.StartChainAction","Action":{"Chain":"provi","Validators":[{"Id":"bob","Allocation":10000000000,"Stake":500000000},{"Id":"alice","Allocation":10000000000,"Stake":500000000},{"Id":"carol","Allocation":10000000000,"Stake":500000000}],"GenesisChanges":"","SkipGentx":false},"State":{"provi":{"ValBalances":{"alice":9500000000,"bob":9500000000,"carol":9500000000},"ValPowers":null,"RepresentativePowers":null,"Params":null,"Rewards":null,"ConsumerChains":null,"AssignedKeys":null,"ProviderKeys":null,"ConsumerChainQueueSizes":null,"GlobalSlashQueueSize":null,"RegisteredConsumerRewardDenoms":null,"Proposals":null}}}]
```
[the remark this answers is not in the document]
p-offtermatt: Yeah, true. I've removed this, thanks! 72729a831

tests/e2e/tracehandler_testdata/changeover.json
p-offtermatt: Good call, I don't know how I didn't think of this 😅 
So much better pre-formatted! 90163279a

tests/e2e/json_utils.go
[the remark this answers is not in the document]
sainoe: @insumity the `ClientsFrozenHeights` chain state will be introduced by the `feat/ics-misbehaviour-handling` merge into `main`.

>Exactly, I need to change the type of proposals. 
@p-offtermatt I would check if field shadowing can solve this problem: https://stackoverflow.com/questions/56982842/how-to-selectively-marshal-json-for-a-struct.

tests/e2e/tracehandler_testdata/changeover.json
insumity: Thanks for the explanation. I think pointing out in a README should be fine but not necessary.

tests/e2e/tracehandler_testdata/democracy.json
insumity: I missed that. Thanks!

tests/e2e/tracehandler_testdata/happyPath.json
insumity: >  but I don't want to do it in this PR and make it even larger again.
Agreed.

>  but maybe setting a precedent to keep for the future would be good
That would be nice. I'm not sure what's the process of creating new minor issues like this, but we could have an issue so we won't forget.

insumity (approved):
Great work! Thanks for the changes.

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+		return nil, fmt.Errorf("unknown action type: %s", actionTypeString)
+	}
+	return nil, err
+}
+
+// custom marshal and unmarshal functions for the chainstate that convert proposals to/from the auxiliary type with type info
+
+// MarshalJSON transforms the ChainState into a ChainStateWithProposalTypes by adding type info to the proposals
+func (c ChainState) MarshalJSON() ([]byte, error) {
+	type ProposalAndType struct {
```
sainoe, line 271: Any reason why the `ProposalAndType` type isn't declared globally and duplicated in `UnmarshalJSON`?

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,412 @@
...
+		RawProposal interface{}
+		Type        string
+	}
+
+	type ChainStateWithProposalTypes struct {
```
sainoe, line 276: Same question

sainoe:
Great work @p-offtermatt! See my comments.

tests/e2e/tracehandler_testdata/happyPath.json
p-offtermatt: Done, https://github.com/cosmos/interchain-security/issues/1322

tests/e2e/json_utils.go
p-offtermatt: Note the difference: the raw proposal is an interface here, and a json.RawMessage in the other case.
This makes the Marshal easier, otherwise I would have to manually marshal the proposal into a RawMessage and fit it in

p-offtermatt: I will try to make this more universal

tests/e2e/json_utils.go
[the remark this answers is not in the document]
p-offtermatt: The problem with shadowing is that it will recursively try to marshal. Note that I am overriding the ChainState.Marshal/Unmarshal functions here. If I have a ChainState object as part of the ChainStateWithProposalTypes object, we have an endless loop. I think I could maybe use shadowing here, but I think that just shifts the complexity around to now getting the new ChainStateWithProposalType to marshal correctly.

tests/e2e/json_utils.go
p-offtermatt: Done. I pulled it out from the method. 05232d933

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,398 @@
...
+	RepresentativePowers           *map[ValidatorID]uint
+	Params                         *[]Param
+	Rewards                        *Rewards
+	ConsumerChains                 *map[ChainID]bool
+	AssignedKeys                   *map[ValidatorID]string
+	ProviderKeys                   *map[ValidatorID]string
+	ConsumerChainQueueSizes        *map[ChainID]uint
+	GlobalSlashQueueSize           *uint
+	RegisteredConsumerRewardDenoms *[]string
+	Proposals                      *map[uint]ProposalAndType // the only thing changed from the real ChainState
```
sainoe, line 28: Field shadowing might be a solution to have a `ChainState` while changing the Proposal field type, see https://stackoverflow.com/questions/56982842/how-to-selectively-marshal-json-for-a-struct

p-offtermatt: Thanks, amazing. This took a bit of effort but I managed to do it with shadowing. It needs a hack, since I need to avoid calling the same custom MarshalJSON method, so I have to define a copy of the ChainState struct, but I can just type alias it. See [f8c8329](https://github.com/cosmos/interchain-security/pull/1314/commits/f8c8329d4cf00040c965b4281b01250880ffa7fe)

tests/e2e/action_rapid_test.go
```diff
@@ -0,0 +1,501 @@
...
+		GetSlashThrottleDequeueActionGen().AsAny(),
+		GetCreateIbcClientsActionGen().AsAny(),
+		CreateCancelUnbondTokensActionGen().AsAny(),
+		CreateLightClientEquivocationAttackActionGen().AsAny(),
+		CreateLightClientAmnesiaAttackActionGen().AsAny(),
+		CreateLightClientLunaticAttackActionGen().AsAny(),
+	)
+}
+
+func CreateSubmitChangeRewardDenomsProposalActionGen() *rapid.Generator[submitChangeRewardDenomsProposalAction] {
```
smarshall-spitzbart, line 96: These funcs all share some boilerplate, so I messed around for a sec with using generics to make it more succicnt. Here's what I found:

You can define a generic GetAction func like so,
```go
func GetAction[T any](callback func(t *rapid.T) T) *rapid.Generator[T] {
	return rapid.Custom(callback)
}
```

and call it like so
```go
GetAction[submitChangeRewardDenomsProposalAction](
	func(t *rapid.T) submitChangeRewardDenomsProposalAction {
		return submitChangeRewardDenomsProposalAction{
			From:    GetValidatorIDGen().Draw(t, "From"),
			Deposit: rapid.Uint().Draw(t, "Deposit"),
			Denom:   rapid.String().Draw(t, "Denom"),
		}
	},
).AsAny(),
```

This reduces a couple lines of code for each action, in that you don't have to define a new func for each one. 

Nbd if you don't want to use this idea, just a bit of golang exploring on my part

tests/e2e/action_rapid_test.go
```diff
@@ -0,0 +1,501 @@
...
+		GetSubmitConsumerRemovalProposalActionGen().AsAny(),
+		GetSubmitParamChangeProposalActionGen().AsAny(),
+		GetSubmitEquivocationProposalActionGen().AsAny(),
+		GetVoteGovProposalActionGen().AsAny(),
+		GetStartConsumerChainActionGen().AsAny(),
+		GetAddChainToRelayerActionGen().AsAny(),
+		GetAddIbcConnectionActionGen().AsAny(),
+		GetAddIbcChannelActionGen().AsAny(),
+		GetStartRelayerActionGen().AsAny(),
+		GetTransferChannelCompleteActionGen().AsAny(),
```
smarshall-spitzbart, line 76: How important and neccessary is it that every action is included in this fuzz test? We'll surely add new actions as time goes on, and it may be easy for devs to forget to add a field here for a newly defined action.

Idea (only if it's important that every action is included in this test): define a top-level action set interface, where that interface has a field for every action type. This test then iterates over the fields in said interface to call the appropriate `Get___ActionGen` funcs.

At runtime, check that all actions used in the step files correspond to a field in the interface. If the field is missing, panic.

Just an idea here. Pro is that devs would have a safegaurd against forgetting to add neccessary boilerplate for a new e2e action. Con is increased complexity

tests/e2e/json_utils.go
```diff
@@ -0,0 +1,375 @@
...
+func UnmarshalMapToActionType(rawAction json.RawMessage, actionTypeString string) (interface{}, error) {
+	var err error
+	switch actionTypeString {
+	case "main.submitConsumerAdditionProposalAction":
+		var a submitConsumerAdditionProposalAction
+		err = json.Unmarshal(rawAction, &a)
+		if err == nil {
+			return a, nil
+		}
+	case "main.SendTokensAction":
```
smarshall-spitzbart, line 75: Similar interface idea could be used here for remembering to include a case for any new action

tests/e2e/state_rapid_test.go
```diff
@@ -0,0 +1,250 @@
...
+				AssignedKeys:                   &assignedKeys,
+				ProviderKeys:                   &providerKeys,
+				ConsumerChainQueueSizes:        &consumerChainQueueSizes,
+				GlobalSlashQueueSize:           &globalSlashQueueSize,
+				RegisteredConsumerRewardDenoms: &registeredConsumerRewardDenoms,
+			}
+		})
+}
+
+func GetRegisteredConsumerRewardDenomsGen() *rapid.Generator[[]string] {
```
smarshall-spitzbart, line 69: Could also use generics here to reduce boilerplate lines, but not a big deal

tests/e2e/trace_handlers_test.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Checks that writing, then parsing a trace results in the same trace.
+func TestWriterThenParser(t *testing.T) {
+	tests := map[string]struct {
+		trace []Step
+	}{
+		"proposalSubmission":    {proposalSubmissionSteps},
+		"proposalInState":       {proposalInStateSteps},
+		"start_provider_chain":  {stepStartProviderChain()},
```
smarshall-spitzbart, line 46: Nit, make `stepStartProviderChain` a var instead of func to be consistent with all the other fields

smarshall-spitzbart (approved):
Great work! Added some comments w.r.t boilerplate and maintenance moving forward

tests/e2e/json_utils.go
```diff
@@ -52,6 +47,7 @@ func (step *Step) UnmarshalJSON(data []byte) error {
 		Action     json.RawMessage
 		State      State
 	}
+	print(string(data))
```
sainoe, line 50: nit: Is this log still needed?

tests/e2e/json_utils.go
sainoe: Okay, got it. I knew it'd require a bit of hacking, but it's much cleaner now!
Solid work!

sainoe (approved):

tests/e2e/json_utils.go
p-offtermatt: Thanks, ccdd1173f

tests/e2e/trace_handlers_test.go
p-offtermatt: I think this should stay as-is - this is not a one-off function, it's called in many places.

tests/e2e/action_rapid_test.go
p-offtermatt: Thanks for exploring this! I think I personally like the current way a bit more, even if its slightly more verbose.

tests/e2e/json_utils.go
p-offtermatt: I think this is more or less ok: This will break non-silently when you try to use the action in a trace, so I think it's ok

tests/e2e/action_rapid_test.go
p-offtermatt: Neat idea, but I think the added complexity is not worth it. If an action is missing from this test, it's not too bad - if things break, you would notice if you tried to use a test trace with it, and it can be added here then.

Issue #1214, opened 2023-08-18
Split out consumer genesis state


smarshall-spitzbart:
## Problem

https://github.com/cosmos/interchain-security/pull/1071 has illuminated that the provider module relies on the consumer module with regard to consumer genesis. See [shared_consumer.proto](https://github.com/cosmos/interchain-security/blob/7986c99b3b49b7603590607e9a026a109dd5eecc/proto/interchain_security/ccv/v1/shared_consumer.proto). Note also that a lot of the fields in the consumer's genesis state struct are zeroed out when that genesis state is queried from provider. 

We should create a new struct which holds the info that a consumer actually needs from the provider (ie. no extraneous zeroed fields), this type will trivially be shared between consumer and provider. Then the consumer genesis state (only referenced by consumer) can have a field for the shared type. This allows us to make types like `HeightToValsetUpdateID`, `OutstandingDowntime` only relevant to the consumer. 

## Closing criteria

Make changes as described

maharsh312:
i'll like to work on this, can you assign it to me?

mpoke:
@maharsh312 I assigned it to you. Thanks for working on this. Please let us know if you need help.

mpoke:
@maharsh312 Could you please confirm that you're working on this issue? Do you have an ETA?

PR #1288, opened 2023-09-12
refactor: Vaguely named consumer structs

bermuell:
## Description
Rename consumer related data in shared types to better reflect consumer related context. 
Closes: #1206 

GenesisState and Param of consumer referenced by provider
---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

* [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ] Added `!` to the type prefix if state-machine breaking change (i.e., requires coordinated upgrade)
* [ ] Confirmed this PR does not introduce changes requiring state migrations, OR migration code has been added to consumer and/or provider modules
* [ ] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
* [ ] Provided a link to the relevant issue or specification
* [ ] Followed the guidelines for [building SDK modules](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules)
* [ ] Included the necessary unit and integration [tests](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#testing)
* [ ] Added a changelog entry to `CHANGELOG.md`
* [ ] Included comments for [documenting Go code](https://blog.golang.org/godoc)
* [ ] Updated the relevant documentation or specification
* [ ] Reviewed "Files changed" and left comments if necessary 
* [ ] Confirmed all CI checks have passed
* [ ] If this PR is library API breaking, bump the go.mod version string of the repo, and follow through on a new major release for both the consumer and provider

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

* [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ] confirmed `!` in the type prefix if API or client breaking change
* [ ] confirmed this PR does not introduce changes requiring state migrations, OR confirmed migration code has been added to consumer and/or provider modules
* [ ] confirmed all author checklist items have been addressed
* [ ] reviewed state machine logic
* [ ] reviewed API design and naming
* [ ] reviewed documentation is accurate
* [ ] reviewed tests and test coverage

PR #1324, opened 2023-09-22
refactor!: Split out consumer genesis state

bermuell:
## Description

Closes: #1214

Refacoring shared data between provider and consumer.

Note: This breaks json format used by augmenting Genesis files of consumer chains with consumer genesis content exported from provider chain. Consumer Genesis content exported from a provider chain using major version 1,2 or3 of the provider module needs to be transformed with the transformation command introduced by this PR:

```
Transform the consumer genesis file from a provider version v1,v2 or v3 to a version supported by this consumer. Result is printed to STDOUT.

Example:
$ <appd> transform /path/to/ccv_consumer_genesis.json

Usage:
  interchain-security-cd genesis transform [genesis-file] [flags]
```

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


MSalopek:
The refactor looks good, thanks for the awesome work!

What is a bit unclear to me is if we should maintain backward compatibility with regards to responses from the `consumer-genesis` provider query:
```shell
interchain-security-pd query provider consumer-genesis <chain-id>
```

The structure returned by this query is an integral part of starting a consumer chain, so maybe we should address the issue of provider-consumer version compatibility.

This is the scenario I'm trying to solve:
1. consumer is on the newest version of ICS (ICS@latest)
2. provider is on an old version of ICS (lCS@v3.1)
3. after `spawn_time` the `consumer_genesis` becomes available on the provider
4. the consumer cannot parse the `consumer_genesis` in the format sent by the provider chain when doing `interchain-security-pd query provider consumer-genesis <chain-id>` since the provider is sending an old version of the genesis

It seems to me that this issue can be solved by parsing the "old" genesis into the "new" genesis:
```shell
interchain-security-cd parse-genesis v1 <path_to_v1_genesis>

# output of this cmd is a genesis that the consumer can process (ICS@latest)
```

If the provider and consumer are on the same ICS version there are no compatibility issues.

bermuell:
thanks @MSalopek for your comment. 
Your scenario is the exactly the one we need to provide a way to deal with it.
There is already such a CLI (similar to what you suggested) to deal with that:
```
 interchain-security-cd genesis migrate -h
Migrate the source genesis into the target version and print to STDOUT.

Usage:
  interchain-security-cd genesis migrate [target-version] [genesis-file] [flags]
```

I'll look into that and provide migration code for my changes

smarshall-spitzbart:
Matija's suggestion of creating a new parsing function sounds solid to me with appropriate docs for the procedure 👍

proto/interchain_security/ccv/consumer/v1/genesis.proto
```diff
@@ -0,0 +1,84 @@
...
+import "google/protobuf/timestamp.proto";
+
+
+
+// GenesisState defines the CCV consumer genesis state
+//
+// Note: this type is only used on consumer side and references shared types with
+// provider
+message GenesisState {
+  reserved 5, 6, 8;
```
smarshall-spitzbart, line 19: What  is this `reserved` call used for ?

smarshall-spitzbart: Got it, great stuff!

proto/interchain_security/ccv/consumer/v1/genesis.proto
```diff
@@ -0,0 +1,92 @@
...
+import "tendermint/abci/types.proto";
+
+
+
+
+// GenesisState defines the CCV consumer genesis state
+//
+// Note: this type is only used on consumer side and references shared types with
+// provider
+message GenesisState {
```
shaspitz, line 22: In https://github.com/cosmos/interchain-security/pull/1288 we renamed the consumer's GenesisState struct to `ConsumerGenesisState`, is there a reason why we change it back in this PR?

app/consumer/genesis.go
```diff
@@ -15,7 +28,130 @@ import (
...
+type (
+	// Transformation callback for a consumer genesis 'section' exported from a
+	// specific version of a provider
+	TransformationCallback func([]byte, client.Context) (json.RawMessage, error)
+
+	// TransformationMap defines the mapping from a version to a transformation callback
+	TransformationMap map[string]TransformationCallback
+)
+
+// Migration of consumer genesis content as it is exported from a provider version v2
```
shaspitz, line 40: Prob need some clarification on the use of "v2" and "v3" here, it seems like the term is applied to both consumer and provider. Are we referring to ICS major versions?

proto/interchain_security/ccv/consumer/v1/genesis.proto
bermuell: We changed it in the shared type namespace to make it clear that it's consumer related. This type is the result of the split and resides  in the consumer namespaces so no need to repeat that in the name. The shared one kept it's name

app/consumer/genesis.go
bermuell: I was looking for a way how a user could find out what  module version is used on a provider chain for the provider implementation. I found the command `gaiad query upgrade module_versions` suitable for that. If you have another approach in mind pls let me know.
The above command returned the following on a real node:
```
...
{
  "name": "provider",
  "version": "2"
}
...
```

MSalopek:
@bermuell docs are needed for the change to CLI.

app/consumer/genesis.go
```diff
@@ -15,7 +28,130 @@ import (
...
+
+	newJson, err := ctx.Codec.MarshalJSON(&newGenesis)
+	if err != nil {
+		return nil, fmt.Errorf("failed marshalling data to new type: %s", err)
+	}
+	return newJson, nil
+}
+
+// Mapping of provider module version to related transformation function
+var transformationMap = TransformationMap{
```
MSalopek, line 82: I think we can remove this and simply comment what the CLI `transform` does.

The `transform` should only do the following:
* transform consumer genesis output coming from providers running ICSv1/2/3 into the latest consumer genesis version

MSalopek (approved):
LGTM!

app/consumer/genesis.go
```diff
@@ -15,7 +26,96 @@ import (
...
 	return ModuleBasics.DefaultGenesis(cdc)
 }
+
+// GetConsumerGenesisTransformCmd transforms Consumer Genesis JSON content exported from a
+// provider version v1,v2 or v3 to a JSON format supported by this consumer version.
+func GetConsumerGenesisTransformCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "transform [genesis-file]",
+		Short: "Transform CCV consumer genesis from an older provider version not supporting current format",
+		Long: fmt.Sprintf(`Transform the consumer genesis file from a provider version v1,v2 or v3 to a version supported by this consumer. Result is printed to STDOUT.
```
sainoe, line 112: nit: perhaps add a `strings.TrimSpace()` call here

sainoe (approved):
LGTM! See my comment.

PR #1333, opened 2023-09-27
refactor: RepresentativePowers to StakedTokens

insumity:
## Description
This is a minor-refactor PR and hence no issue is created for it.

Based on @p-offtermatt [comment](https://github.com/cosmos/interchain-security/pull/1275#discussion_r1336804594
):
> Seems ok to me. Could rename it to something like StakedTokens instead of RepresentativePowers (looking at the code, I think that's what this actually is, right?)

We rename `RepresentativePowers` to `StakedTokens`. We can see that `StakedTokens` is a better name if we look at what `getRepresentativePower` is computing, which is just the number of tokens:

```
func (tr TestConfig) getRepresentativePower(chain ChainID, validator ValidatorID) uint {
	//#nosec G204 -- Bypass linter warning for spawning subprocess with cmd arguments.
	bz, err := exec.Command("docker", "exec", tr.containerConfig.InstanceName, tr.chainConfigs[chain].BinaryName,

		"query", "staking", "validator",
		tr.validatorConfigs[validator].ValoperAddress,
		`--node`, tr.getQueryNode(chain),
		`-o`, `json`,
	).CombinedOutput()
	if err != nil {
	}
	amount := gjson.Get(string(bz), `tokens`)
	return uint(amount.Uint())
}
```

### DIFF
--- a/tests/e2e/state.go
+++ b/tests/e2e/state.go
@@ -20,7 +20,7 @@ type ChainState struct {
 	ValBalances                    *map[ValidatorID]uint
 	Proposals                      *map[uint]Proposal
 	ValPowers                      *map[ValidatorID]uint
-	RepresentativePowers           *map[ValidatorID]uint
+	StakedTokens                   *map[ValidatorID]uint
 	Params                         *[]Param
 	Rewards                        *Rewards
 	ConsumerChains                 *map[ChainID]bool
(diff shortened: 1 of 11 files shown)


p-offtermatt (approved):
LGTM, thanks for the refactor! 🥇 just a nit, the prefix for the pr should probably just be refactor
https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json
not a big deal but this ensures that this is e.g. easily evaluated via machine

mpoke (approved):

PR #1336, opened 2023-09-29
Add Quint model of Interchain Security

p-offtermatt:
Closes: https://github.com/cosmos/interchain-security/issues/1239

This adds a quint model of interchain Security, based on the existing TLA+ model by @Kukovec.
There are a few notable differences:

* Initializing is removed as a status - this is mostly because my goal is to use the model for MBT, and the initializing state is probably something we would skip past in the tests (also since the interactions with the initializing state aren't very meaningful in the model)
* The actions the model exposes are slightly different, e.g. time advancement and chain status advancement happen only during EndBlocks/BeginBlocks and don't happen simultaneously with other actions. This is because this makes the traces easier to test in Quint, i.e. more control over what happens when.
* I model endBlocks for both provider and consumer chains. This is because I think this is a useful extra level of implementation details that will make test traces more meaningful.

Other than that, some implementation details are different. Some are my preference, some are just due to what is more natural to write in Quint, some are to (hopefully) make traces that are easier to use.

How to review:
* Read README.md for some cursory explanations
* Run tests and sanity checks using the commands from the README. Running the invariant checks is probably too time consuming and not necessary.
* Read through ccv.qnt. Reading top-down is probably fine. What to pay attention to here: Does the code match the intention of the CCV protocol as specified here https://github.com/cosmos/ibc/tree/main/spec/app/ics-028-cross-chain-validation?
* Read through ccv_test.qnt. What to pay attention to here: Do the test traces make sense and provide fine coverage? These are supposed to be unit tests.
* Read through ccv_statemachine.qnt. This needs to not be looked at as thoroughly as the core logic in ccv.qnt, except: the INVARIANT CHECKS are important to look at. What to pay attention to: Are there more invariants that you think should be expressed here? Just generally, this would be any properties that you assume hold true for the interchain security protocol when you reason about it.


tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
+This folder contains a Quint model for the core logic of Cross-Chain Validation (CCV).
+
+### File structure
+The files are as follows:
+- ccv.qnt: Contains the type definitions, data structures, functional logic for CCV.
+The core of the protocol.
+- ccv_statemachine.qnt: Contains the state machine layer for CCV. Very roughly speaking, this could be seen as "e2e tests".
+Also contains invariants.
+- ccv_test.qnt: Contains unit tests for the functional layer of CCV.
+- libararies/*: Libraries that don't belong to CCV, but are used by it.
```
insumity, line 10:
```suggestion
- libraries/*: Libraries that don't belong to CCV, but are used by it.
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Model details
+
+To see the data structures used in the model, see the ProtocolState type in ccv.qnt.
```
insumity, line 14:
```suggestion
To see the data structures used in the model, see the `ProtocolState` type in ccv.qnt.
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The "public" endpoints of the model are those functions that take as an input the protocol state, and return a Result.
```
insumity, line 16:
```suggestion
The "public" endpoints of the model are those functions that take as an input the protocol state, and return a `Result`.
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+Other functions are for utility.
+
+The parameters of the protocol are defined as consts in ccv.qnt.
```
insumity, line 19:
```suggestion
The parameters of the protocol are defined as consts in [ccv.qnt](ccv.qnt).
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Tests
+
+To run unit tests, run 
+```
+quint test ccv_test.qnt --main CCVTest
+```
+and 
+```
+quint test ccv_statemachine.qnt --main CCVDefaultStateMachine
```
insumity, line 29: This one fails on my machine:
```
quint test ccv_statemachine.qnt --main CCVDefaultStateMachine

quint test <input>

Run tests against a Quint specification

Options:
  --help         Show help                                             [boolean]
  --version      Show version number                                   [boolean]
  --out          output file (suppresses all console output)            [string]
  --main         name of the main module (by default, computed from filename)
                                                                        [string]
  --output       write a trace for every test, e.g., out{#}{}.itf.json
                 {} is the name of a test, {#} is the test sequence number
                                                                        [string]
  --max-samples  the maximum number of successful runs to try for every
                 randomized test                       [number] [default: 10000]
  --seed         random seed to use for non-deterministic choice        [string]
  --verbosity    control how much output is produced (0 to 5)
                                                           [number] [default: 2]
  --match        a string or regex that selects names to use as tests   [string]

Error: Couldn't find any result for id 53 while Inferring effect for ConsumerChains
    at EffectInferrer.fetchResult (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:262:19)
    at EffectInferrer.effectForName (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:278:25)
    at EffectInferrer.exitName (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:99:41)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:272:25)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:38
    at Array.forEach (<anonymous>)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:23)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:38
    at Array.forEach (<anonymous>)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:23)
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+### Invariants
+
+The invariants to check are in ccv_statemachine.qnt.
+Check a single invariant by running
+`quint run --invariant INVARIANT_NAME ccv_statemachine.qnt --main CCVDefaultStateMachine`,
```
insumity, line 36: Running something like `quint run --invariant ValidatorUpdatesArePropagated ccv_statemachine.qnt --main CCVDefaultStateMachine` fails similar to the above:
```
Error: Couldn't find any result for id 53 while Inferring effect for ConsumerChains
    at EffectInferrer.fetchResult (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:262:19)
    at EffectInferrer.effectForName (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:278:25)
    at EffectInferrer.exitName (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/effects/inferrer.js:99:41)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:272:25)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:38
    at Array.forEach (<anonymous>)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:23)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:38
    at Array.forEach (<anonymous>)
    at walkExpression (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/IRVisitor.js:289:23```

File: tests/difference/core/quint_model/README.md [hunk 7]
```diff
@@ -0,0 +1,72 @@
...
+- [X] MatureOnTimeInv: For every ValidatorSetChangePacket received by a consumer chain at 
+time t, a MaturedVSCPacket is sent back to the provider in the first block 
+with a timestamp >= t + UnbondingPeriod on that consumer.
+- [X] EventuallyMatureOnProviderInv: If we send a VSCPacket, this is eventually responded to by all consumers
+that were running at the time the packet was sent (and are still running).
+
+Invariants can also be model-checked by Apalache, using this command:
+```
+quint verify --invariant ValidatorUpdatesArePropagated,ValidatorSetHasExistedInv,SameVSCPacketsInv,MatureOnTimeInv,EventuallyMatureOnProviderInv --main CCVDefaultState
+Machine ccv_statemachine.qnt
```
insumity, line 56: Fails on my machine with:
```
Not enough non-option arguments: got 0, need at least 1
Machine: Can't find any plists for Machine```

File: tests/difference/core/quint_model/README.md [hunk 8]
```diff
@@ -0,0 +1,72 @@
...
+```
+
+### Sanity Checks
+
+Sanity checks verify that certain patterns of behaviour can appear in the model.
+In detail, they are invariant checks that we expect to fail.
+They usually negate the appearance of some behaviour, i.e. not(DesirableBehaviour).
```
insumity, line 63:
```suggestion
They usually negate the appearance of some behaviour, i.e. `not(DesirableBehaviour)`.
```

tests/difference/core/quint_model/README.md
p-offtermatt: Could you check that your quint is up-to-date? 
```
$ quint --version 
0.14.4
```

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
...
+### Invariants
+
+The invariants to check are in ccv_statemachine.qnt.
+Check a single invariant by running
+`quint run --invariant INVARIANT_NAME ccv_statemachine.qnt --main CCVDefaultStateMachine`,
+or all invariants one after another with the help of the script `run_invariants.sh`.
+Each invariant takes about a minute to run.
+
+Invariants are as follows:
+- [X] ValidatorUpdatesArePropagated: When a validator power update is comitted on chain, a packet containing that change in voting power is sent to all running consumers.
```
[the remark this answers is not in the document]
p-offtermatt: Yes, good point!

tests/difference/core/quint_model/README.md
p-offtermatt: Also here, can you make sure Quint is up-to-date?

tests/difference/core/quint_model/README.md
p-offtermatt: Thanks, it's the rogue breakline that messes it up

tests/difference/core/quint_model/README.md
insumity: Thanks Philip. This was a personal setup issue on my side. Reinstalled quint and it works.

tests/difference/core/quint_model/libraries/extraSpells.qnt
```diff
@@ -0,0 +1,259 @@
...
+    }
+    run ValuesTest =
+    all {
+        assert(Values(Map(1 -> 2, 3 -> 4)) == Set(2, 4)),
+        assert(Values(Map()) == Set())
+    }
+
+    // Returns the maximal element of the set.
+    // If the set is empty, the function call will fail at runtime.
+    pure def Max(__set: Set[int]): int = {
```
insumity, line 192: 1. Why` are `Max`, `Min` capitalized in comparison to other `def`s?
2. (nit) You could use `MaxBy` to define `Max` with something like:
``` pure def Max(__set: Set[int]): int = MaxBy(__set, a => a)```

File: tests/difference/core/quint_model/libraries/extraSpells.qnt [hunk 11]
```diff
@@ -0,0 +1,259 @@
...
+    pure def Min(__set: Set[int]): int = {
+        __set.fold(oneOf(__set), (__m, __e) => if(__m < __e) __m else __e)
+    }
+
+    run MinTest =
+    all {
+        assert(Min(Set(1, 2, 3)) == 1),
+    }
+
+    pure def HasSubsequence(__this: List[a], __other: List[a]): bool = {
```
insumity, line 212: nit: add comment to this one.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+                message: msg
+            }
+        }
+    }
+
+    // possible consumer statuses
+    pure val STOPPED = "stopped" // the chain was once a consumer chain, but has been voluntarily dropped by the provider.
+    pure val TIMEDOUT = "timedout" // the chain has timed out and was dropped by the provider. This is only used for involuntary drops.
+    pure val RUNNING = "running" // the chain is currently a consumer chain. Running chains are those that get sent VSCPackets.
+    pure val UNUSED = "unused" // the chain has never been a consumer chain, and is available to become one.
```
insumity, line 198: nit: Not sure how I feel about the naming, especially for `UNUSED`, in the sense that consumer chains already run.  Although, I see that the [UNUSED](https://github.com/cosmos/ibc/pull/911/files#diff-ed9c0aa72ff4479e19c87de5d812a7e9c41151d0a73f908070930145bd916b4fR130) stems from the TLA+ spec. I cannot think of any succinct name but would it make sense to have something like `NOT_CONSUMNG_SECURITY`, etc.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+            },
+            error: {
+                message: msg
+            }
+        }
+    }
+
+    // possible consumer statuses
+    pure val STOPPED = "stopped" // the chain was once a consumer chain, but has been voluntarily dropped by the provider.
+    pure val TIMEDOUT = "timedout" // the chain has timed out and was dropped by the provider. This is only used for involuntary drops.
```
insumity, line 196: Do we _win_ anything by making this distinction on whether a chain was dropped by the consumer or it was voluntarily stopped? It feels we could put both under the umbrella of stopped for simplicity?

If you leave this as is, then it might make sense to explain what "timed out" means here -- that the provider has not received a matured packet  from the chain the last `CcvTimeout` time since the packet was sent.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+    pure def deliverPacketToProvider(currentState: ProtocolState, sender: Chain): (Result, bool) = {
+        if (not(isCurrentlyConsumer(sender, currentState.providerState))) {
+            (Err("Sender is not currently a consumer - must have 'running' status!"), false)
+        } else if (length(currentState.consumerStates.get(sender).outstandingPacketsToProvider) == 0) {
+            (Err("No outstanding packets to deliver"), false)
+        } else {
+            val packet = currentState.consumerStates.get(sender).outstandingPacketsToProvider.head()
+            // if the packet has timed out, drop the consumer. its state doesn't matter anymore
+            val timeout = CcvTimeout.get(sender)
+            if(packet.sendingTime + CcvTimeout.get(sender) < currentState.providerState.chainState.lastTimestamp) {
```
insumity, line 286: If I understand this correctly, the `packet.sendingTime` corresponds to time set by the consumer chain, but here we compare against the provider's time. I guess this is done for simplicity? Because otherwise, a provider `p` would need to store the latest time `p` received a packet from a consumer and then go through all those latest times to timeout consumer chains?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+    // The second return is an error string: If it is not equal to "",
+    // it contains an error message, and the first return should be ignored.
+    pure def stopConsumers(
+        currentConsumerStatusMap: Chain -> str,
+        consumersToStop: Set[Chain],
+        consumersToTimeout: Set[Chain]): (Chain -> str, str) = {
+            val runningConsumers = currentConsumerStatusMap.keys().filter(
+                chain => currentConsumerStatusMap.get(chain) == RUNNING
+            )
+            // if a consumer is both started and stopped, this is an error
```
insumity, line 552: nit: fix indentation? Also in `startConsumers` below.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+        val furthestConsumer = getRunningConsumers(currentState.providerState).MaxBy(
+            consumer => abs(
+                currentState.providerState.chainState.lastTimestamp -
+                currentState.consumerStates.get(consumer).chainState.lastTimestamp
+            )
+        )
+        furthestConsumer
+    }
+
+    pure def getRunningConsumersFromMap(consumerStatus: Chain -> str): Set[Chain] = {
```
insumity, line 528: Maybe I missed this, but is this `def` and `getStoppedConsumersFromMap` used somewhere?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+                    (false, "")
+                }
+            } else {
+                // no packet has been sent yet, so no timeout
+                (false, "")
+            }
+        }
+
+    // ===================
+    // ASSUMPTIONS ON MODEL PARAMETERS
```
insumity, line 852: Is there a way to force those tests to run when somebody `import`s `CCV` and binds values to the corresponding constant? If not, maybe it would make sense to do the testing by extending `run_invariants.sh`.
My concern is that if those are not tested, then someone could update `ccv_statemachine.qnt` with wrong values and hence those tests do not add much value.

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+        currentState.with(
+            "providerState", 
+                currentState.providerState.with(
+                    "chainState", newChainState
+            )
+        )
+    }
+
+    // Returns true if the given chain is currently a running consumer, false otherwise.
+    pure def isCurrentlyConsumer(chain: Chain, providerState: ProviderState): bool = {
```
insumity, line 804: maybe rename to `isRunningConsumer`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+            val newState = currentState.with(
+                "providerState", newProviderState
+            )
+            Ok(newState)
+        }
+    }
+
+    // removes the oldest outstanding packet from the consumer. on-chain, this would happen when the packet is acknowledged.
+    // only the oldest packet can be removed, since we model ordered channels.
+    pure def removeOutstandingPacketFromConsumer(currentState: ProtocolState, sender: Chain): Result = {
```
insumity, line 752: Because `removeOutstandingPacketFromConsumer` never returns an error, does it make sense to just have it return the `Result's underlying `ProtocolState`?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+        Ok(newState)
+    }
+
+    // Updates the given oldValidatorSet by setting the validator to newVotingPower.
+    // If newVotingPower is zero, the validator is removed.
+    pure def getUpdatedValidatorSet(oldValidatorSet: ValidatorSet, validator: Node, newVotingPower: int): ValidatorSet =
+        if (newVotingPower > 0) 
+        oldValidatorSet.put(validator, newVotingPower)
+        else
+        oldValidatorSet.mapRemove(validator)
```
insumity, line 787: Similar to a [previous comment](https://github.com/cosmos/interchain-security/pull/1336/files#r1348633598). It feels we might unnecessarily complicate our lives by removing validators and we shouldn't consider this case?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+        }
+
+    // ===================
+    // ASSUMPTIONS ON MODEL PARAMETERS
+    // ===================
+
+    run UnbondingPeriodPositiveTest =
+        UnbondingPeriodPerChain.keys().forall(chain => UnbondingPeriodPerChain.get(chain) > 0)
+
+    run VscTimeoutPositiveTest =
```
insumity, line 858: nit: naming discrepancies when it comes to `Vsc` (here) versus `VSC` (e.g., in `TimeoutDueToVSCTimeout`).

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+
+    pure def endAndBeginBlockForConsumer(
+        currentState: ProtocolState,
+        chain: Chain,
+        // by how much the timestamp of the chain should be advanced for the next block
+        timeAdvancement: Time): Result = {
+            if (not(currentState.consumerStates.keys().contains(chain))) {
+                Err("chain is not a consumer")
+            } else {
+                // if the chain is not a consumer, return an error
```
insumity, line 479: Should this comment move 2 lines above to correspond to the `if` case?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+    // Defines the current state of the provider chain. Essentially, all information here is stored by the provider on-chain (or could be derived purely by information that is on-chain).
+    type ProviderState =
+        {
+            // the state that each chain needs to store
+            chainState: ChainState,
+
+            // Stores, for each consumer chain, the list of packets that have been sent to the consumer chain
+            // and have not been received yet.
+            // In the implementation, this would roughly be the unacknowledged packets on an ibc channel.
+            outstandingPacketsToConsumer: Chain -> List[VSCPacket],
```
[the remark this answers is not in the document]
p-offtermatt: I tried to make it clearer. Outstanding are not yet received by the consumer, sent are ones that we have not yet received a maturation for.
If you have a suggestion, this could be simplified, but at least my intuition is that any solution should use two separate lists to do this

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+            val consumerState = currentState.consumerStates.get(consumer)
+
+            // has a packet been sent on the provider more than VSCTimeout ago, but we have not received an answer since then?
+            val sentVSCPackets = providerState.sentVSCPackets.get(consumer)
+            if(sentVSCPackets.length() > 0) {
+                val oldestSentVSCPacket = sentVSCPackets.head() // if length is 0, this is undefined, but we check for this before we use it
+                if(oldestSentVSCPacket.sendingTime + VscTimeout < providerState.chainState.lastTimestamp) {
+                    (true, "")
+                } else {
+                    // no timeout yet, it has not been VscTimeout since that packet was sent
```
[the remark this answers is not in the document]
p-offtermatt: An empty string here means there is no error.
False just means there have been no timeouts, see the comment above the method description.
Returning a string here would indicate an error

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: Changed everything to Vsc, thanks

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+
+    // ===================
+    // PROTOCOL LOGIC contains the meat of the protocol
+    // functions here roughly correspond to API calls that can be triggered from external sources
+    // ===================
+    
+    // the power of a validator on the provider chain is changed to the given amount. We do not care how this happens,
+    // e.g. via undelegations, or delegations, ...
+    pure def votingPowerChange(currentState: ProtocolState, validator: Node, amount: int): Result = {
+            if (amount < 0) {
```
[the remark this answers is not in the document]
p-offtermatt: Yup, thanks!

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: I think setting to 0 should be allowed, since its the way we would expect validators to get removed from the set.
Whether removing or setting to 0 is better is the right question imo

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: True

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: I think it makes it more human-understandable.
It also allows me to write a sanity check that ensures that packets can also time out,
so it allows more fine-grained testing.

I think it makes the model more explicit and the traces easier to understand,
and I think the tradeoff of slightly more complexity is worth it

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: Let me comment on this:
> in the sense that consumer chains already run.
This part is not true - UNUSED consumers might already run, or they might not. 

But you are right that the current name is not great. I'll call it `NOT_CONSUMER`

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: I think this is correct and not a simplification -
the provider receives a packet.
packet.sendingTime + CcvTimeout.get(sender) is at what timestamp on the receiving chain the packet will time out.

Notice this is *not* the VscTimeout - the CcvTimeout is related to the timeout timestamp written into the ibc packets.

If you disagree or this doesn't make sense, let me know and let's have a sync

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: It's not, you are right :) removed!

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: I will add it to `run_invariants.sh`
What you asked is not possible right now, but the Quint team has it on the radar :) https://github.com/informalsystems/quint/issues/1182

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: Good call!

tests/difference/core/quint_model/libraries/extraSpells.qnt
p-offtermatt: Unified this, thanks

tests/difference/core/quint_model/libraries/extraSpells.qnt
p-offtermatt: This was not needed anymore anyways, so I removed it

tests/difference/core/quint_model/ccv.qnt
insumity: Yes, that makes sense. Thanks for the explanation. (I'm resolving this).

tests/difference/core/quint_model/ccv.qnt
insumity: I don't have strong feelings on this.

>Whether removing from the set or treating 0 are removed is better is the right question imo

Good point. My take is that this currently tests that a validator gets removed when voting power is 0 but this does  not seem to be part of the CCV protocol. It seems to me that it is testing more  the `staking` module instead.
We mainly care that `VSCPacket`s are correctly propagated to the consumer chain -- why would it make a difference if the validator shows up with voting power 0 or actually removed?

Additionally, if we believe that removing nodes tests something more, would it make sense to be able to add nodes as well?

tests/difference/core/quint_model/ccv.qnt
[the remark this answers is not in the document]
insumity: I see. Missed that. Thanks for the clarification. (Resolving this).

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: Ok, that convinces me, this is not really part of CCV. I will just allow 0 by setting the validator to 0 for simplicity. We can imagine it means the val is unbonded, but no special behaviour

tests/difference/core/quint_model/ccv.qnt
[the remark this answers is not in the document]
insumity: So, `outstandingPacketsToConsumer` is just a way to put a packet in a send queue that would be eventually be picked up and delivered to consumer's `receivedVSCPackets`?

The idea of keeping the `sentVSCPackets` is to check if we haven't received a `VSCMaturedPacket` since sending the `VSCPacket` and this cannot be done with `outstandingPacketsToConsumer` because you pop packets from `outstandingPacketsToConsumer` after sending them?

It might make sense to rename `sentVSCPackets` to `sentVSCPacketsToConsumer` similarly to `outstandingPacketsToConsumer` in such a case?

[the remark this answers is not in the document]
p-offtermatt: correct. Changed according to the suggestion, thanks

tests/difference/core/quint_model/ccv_test.qnt
```diff
@@ -0,0 +1,255 @@
...
+        val result = DeliverPacketToProviderHappyPathTest_resultAndTimeout._1
+        val timeout = DeliverPacketToProviderHappyPathTest_resultAndTimeout._2
+        not(result.hasError) and
+        not(timeout)
+    }
+
+    run StateModificationOK_DeliverPacketToProviderHappyPathTest =
+    {
+        val result = DeliverPacketToProviderHappyPathTest_resultAndTimeout._1
+        val timedOut = DeliverPacketToProviderHappyPathTest_resultAndTimeout._2
```
insumity, line 148:
```suggestion
```
Doesn't seem to be used. If used maybe use consistent naming with `timeout` above.

tests/difference/core/quint_model/ccv_test.qnt
```diff
@@ -0,0 +1,255 @@
...
+        val sender = "sender"
+        val providerState = currentState.providerState
+        val consumerState = GetEmptyConsumerState
+        // add the consumer to the consumerStates
+        val consumerStates = currentState.consumerStates.put(sender, consumerState)
+        val providerState2 = providerState.with(
+            "consumerStatus", providerState.consumerStatus.put(sender, RUNNING)
+        )
+        val providerState3 = providerState2.with(
+            "outstandingPacketsToConsumer", providerState2.outstandingPacketsToConsumer.put(sender, List({
```
insumity, line 99: This was initially a bit confusing because the test state is named `_DeliverPacketToProvider_` but the initial state is having the provider ready to send a packet to the consumer.

tests/difference/core/quint_model/ccv_test.qnt
```diff
@@ -0,0 +1,255 @@
...
+        "consumerStates", _DeliverPacketToProvider_TestState.consumerStates.put(
+            "sender", _DeliverPacketToProvider_TestState.consumerStates.get("sender").with(
+                "outstandingPacketsToProvider", List({
+                    id: 0,
+                    sendingTime: 0
+                })
+            )
+        )
+    ).with(
+        // put an entry into sentVscPacket on the provider that corresponds to the packet we put on the consumer
```
insumity, line 122:
```suggestion
        // put an entry into sentVscPacketsToConsumer on the provider that corresponds to the packet we put on the consumer
```
I guess the packet put on the consumer is the maturation packet?

p-offtermatt: the VSC packet is sent to the consumer, the consumer responds with a VSCMatured packet

tests/difference/core/quint_model/ccv_test.qnt
```diff
@@ -0,0 +1,255 @@
...
+    run ConsumerStatusMapAlreadyStoppedTest =
+    {
+        val currentConsumerStatusMap = Map(
+            "chain1" -> NOT_CONSUMER,
+            "chain2" -> RUNNING,
+            "chain3" -> STOPPED
+        )
+        val res = StartStopConsumers(
+            currentConsumerStatusMap,
+            Set("chain1"),
```
insumity, line 233: Do we need this here? We could have `Set()` since we do not check anything for `chain1` in this specific test?

tests/difference/core/quint_model/ccv_test.qnt
```diff
@@ -0,0 +1,255 @@
...
+            Set("chain1"),
+            Set()
```
insumity, line 251: Is the timing out of chains something we would like to test as well?

tests/difference/core/quint_model/ccv_statemachine.qnt
```diff
@@ -0,0 +1,506 @@
...
+    pure val nodes = Set("node1", "node2", "node3", "node4", "node5", "node6", "node7", "node8", "node9", "node10")
+    pure val InitialValidatorSet = nodes.mapBy(node => 100)
+
+    import ccv(VscTimeout = 5 * Week, CcvTimeout = ccvTimeouts, UnbondingPeriodPerChain = unbondingPeriods, ConsumerChains = consumerChains).* from "./ccv"
+
+
+    var currentState: ProtocolState
+
+
+    pure def ACTIONS = Set(
```
insumity, line 22: I couldn't find where this is being used. If not, we can remove or have as comment instead on what the possible actions are?

tests/difference/core/quint_model/ccv_statemachine.qnt
```diff
@@ -0,0 +1,506 @@
...
+        val consumerStatus = currentState.providerState.consumerStatus
+        nondet consumersToStart = oneOf(nonConsumers.powerset())
+        nondet consumersToStop = oneOf(runningConsumers.powerset())
+        nondet timeAdvancement = oneOf(timeAdvancements)
+        EndAndBeginBlockForProvider(timeAdvancement, consumersToStart, consumersToStop),
+
+        nondet node = oneOf(nodes)
+        // very restricted set of voting powers. exact values are not important,
+        // and this keeps the state space smaller.
+        // 0 for getting a validator out of the validator set, and two non-zero values
```
insumity, line 88: > 0 for getting a validator out of the validator set,
We do not kick the validator out of the set anymore from what I understand. So, would it make sense to remove this comment?

tests/difference/core/quint_model/ccv_statemachine.qnt
insumity, line 22: Is this used somewhere? If not, we can remove?

tests/difference/core/quint_model/ccv_statemachine.qnt
```diff
@@ -0,0 +1,506 @@
...
+                        packetsBetween1 == packetsBetween2
+                    }
+                }
+            )
+        )
+
+    // For every ValidatorSetChangePacket received by a consumer chain at 
+    // time t, a MaturedVscPacket is sent back to the provider in the first block 
+    // with a timestamp >= t + UnbondingPeriod
+    // NOTE: because we remove the maturationTimes entry when we send the packets,
```
insumity, line 259: Thanks, that was a really useful comment 👍
Otherwise, someone might wonder how this invariant is not violated.

insumity (approved):
Great work!

+ It might make sense to ask someone from the Quint team to also review this for Quint-style, etc.

tests/difference/core/quint_model/ccv_statemachine.qnt
p-offtermatt: Thanks, a remnant of something I tried and forgot to remove!

tests/difference/core/quint_model/ccv_statemachine.qnt
p-offtermatt: Yes, thanks!

tests/difference/core/quint_model/ccv_test.qnt
p-offtermatt: I think it's okay if it's confusing for the first few lines :D

tests/difference/core/quint_model/ccv_test.qnt
p-offtermatt: We could, but I think it's not worth it - it shares a lot of code with the stopped option anyways

tests/difference/core/quint_model/ccv_statemachine.qnt
```diff
@@ -0,0 +1,506 @@
...
+        .then(
+            // change voting power
+            VotingPowerChange("node1", 50)
+        )
+        .then(
+            // send packets
+            EndAndBeginBlockForProvider(1 * Second, Set(), Set())
+        )
+        .then(
+            // advance time on provider by VscTimeout + 1 Second
```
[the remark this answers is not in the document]
p-offtermatt: No, 1 second is good enough. I've changed it back to 1 second, thanks!

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,828 @@
...
+
+    type Error = {
+        message: str
+    }
+
+    // we return either a result or an error.
+    // if hasError is true, newState may be arbitrary, but the error will be meaningful.
+    // if hasError is false, error may be arbitrary, but newState will be meaningful.
+    type Result = {
+        hasError: bool,
```
konnov, line 154: `hasError` is a bit redundant. I normally have just an `error` field of type `str`. If `error == ""`, then there is no error.

p-offtermatt: Good point. I still like the explicitness of calling hasError, but I will put it just as a method of the result type and remove the field

tests/difference/core/quint_model/ccv_test.qnt
p-offtermatt: Yeah, true, but I think this doesn't really matter. Unless you think this makes it much harder to understand

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,827 @@
...
+
+    // ===================
+    // PROTOCOL LOGIC contains the meat of the protocol
+    // functions here roughly correspond to API calls that can be triggered from external sources
+    // ===================
+    
+    // the power of a validator on the provider chain is changed to the given amount. We do not care how this happens,
+    // e.g. via undelegations, or delegations, ...
+    pure def votingPowerChange(currentState: ProtocolState, validator: Node, amount: int): Result = {
+            if (amount < 0) {
```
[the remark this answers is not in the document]
p-offtermatt: Yes. The reason is that composing actions is annoying (two actions that touch the same state variables cannot easily be composed), while composing functions is extremely easy.
While this spec doesn't have compositions that are too difficult, this approach of keeping actions minimal and putting logic into functions is more scalable and easier to understand imo. But happy to discuss this more if you feel differently

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,827 @@
...
+    // that indicates whether the consumer timed out or not.
+    // If the result has an error, the second return should be ignored.
+    pure def deliverPacketToProvider(currentState: ProtocolState, sender: Chain): (Result, bool) = {
+        if (not(isRunningConsumer(sender, currentState.providerState))) {
+            (Err("Sender is not currently a consumer - must have 'running' status!"), false)
+        } else if (length(currentState.consumerStates.get(sender).outstandingPacketsToProvider) == 0) {
+            (Err("No outstanding packets to deliver"), false)
+        } else {
+            val packet = currentState.consumerStates.get(sender).outstandingPacketsToProvider.head()
+            // if the packet has timed out, drop the consumer. its state doesn't matter anymore
```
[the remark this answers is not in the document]
p-offtermatt: Done two lines above 
```
length(currentState.consumerStates.get(sender).outstandingPacketsToProvider) == 0
```

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,827 @@
...
+                    "providerState", newProviderState
+                )
+                pure val newState = setProviderValidatorSet(tmpState, newValidatorSet)
+                Ok(newState)
+            }
+    }
+
+    // Delivers the next queued VscMaturedPacket from a consumer chain to the provider chain.
+    // Only argument is the consumer chain, from which the packet will be delivered.
+    // If this packet will time out on the provider on delivery,
```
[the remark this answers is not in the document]
p-offtermatt: Thanks, done

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,827 @@
...
+    ): (Chain -> str, str) = {
+        // check if any consumer is both started and stopped
+        if (consumersToStart.intersect(consumersToStop).size() > 0) {
+            (currentConsumerStatusMap, "Cannot start and stop a consumer at the same time")
+        } else {
+            val res1 = currentConsumerStatusMap.startConsumers(consumersToStart)
+            val newConsumerStatus = res1._1
+            val err1 = res1._2
+            val res2 = newConsumerStatus.stopConsumers(consumersToStop, consumersToTimeout)
+            val err2 = res2._2
```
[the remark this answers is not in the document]
p-offtermatt: Technically not, no, but it doesn't hurt and I think it's slightly more readable like this. (Performance is not really an issue for models, we should likely radically prefer readability unless we have evidence certain things need to be faster). If you think it's currently less readable than the alternative, lmk and I will adjust

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,877 @@
...
+module CCVTypes {
+    import Time.* from "./libraries/Time"
+
+    type Node = str
+    type Chain = str
+    type Power = int
+    type VSCId = int
+    type ValidatorSet = Node -> Power
+    type Height = int
+    // a list of validator sets per blocks, ordered by recency
```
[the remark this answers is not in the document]
p-offtermatt: I think the order is not used in the model (or rather, only in places where the dependence doesn't really matter), but I want this to be ordered like this also for traces and just human readability of the traces. The explainability of traces is much better if there's an order to these imo

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: No, it should be the timestamp on the consumer.

Here is the explanation in more detail:
The consumer chain sends the packet at packet.sendingTime.
In the real code, IBC packets store the time *on the destination chain* at which they are meant to time out.
This is typically the sending time *on the sender* + a certain delay (the CcvTimeout in our case).
In our case, the timeout time is therefore packet.sendingTime + CcvTimeout.get(sender).
Remember, this is the timestamp *on the destination chain* at which the packet times out.
So we need to compare packet.sendingTime + CcvTimeout.get(sender) with the timestamp on the destination chain (so provider.time, or currentState.providerState.chainState.lastTimestamp, in our case).

References: https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#definitions for the fields of packets,
https://github.com/cosmos/ibc/blob/jk/ccvTLA/spec/app/ics-028-cross-chain-validation/data_structures.md#ccv-state for the ccv timeout

Does that make sense? Do you think there's anything I could add to the code to make it easier to understand?

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,825 @@
...
+                List({
+                        id: providerState.runningVscId,
+                        validatorSet: providerState.chainState.currentValidatorSet,
+                          sendingTime: providerState.chainState.lastTimestamp
+                })
+            } else {
+                List()
+            }
+        )
+        providerState.with(
```
konnov, line 623: would not it be nicer to introduce values for the new syntax and then use the `{ ... }` syntax, e.g., `{ ...providerState, outstandingPacketsToConsumer: newOPC, sentVscPacketsToConsumer: newSPC }` ?

tests/difference/core/quint_model/ccv_statemachine.qnt
```diff
@@ -0,0 +1,496 @@
...
+            // ... for each consumer that is running then ...
+            runningConsumers.forall(
+                // ...the validator set is in a sent packet
+                consumer => currentState.providerState.sentVscPacketsToConsumer.get(consumer).toSet().exists(
+                    packet => packet.validatorSet == providerValSetInCurBlock
+                )
+            )
+        } else {
+            true
+        }
```
konnov, line 212: I am just curious, why would not you use `A implies B` here? Is it because it would be harder to explain?

konnov (commented):
This looks nice! Obviously, I did not try to understand the logic of CCV, but only looked at the Quint side of things.

tests/difference/core/quint_model/README.md
```diff
@@ -0,0 +1,72 @@
+This folder contains a Quint model for the core logic of Cross-Chain Validation (CCV).
+
+### File structure
+The files are as follows:
+- ccv.qnt: Contains the type definitions, data structures, functional logic for CCV.
+The core of the protocol.
+- ccv_statemachine.qnt: Contains the state machine layer for CCV. Very roughly speaking, this could be seen as "e2e tests".
```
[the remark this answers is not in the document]
p-offtermatt: I guess this makes sense in the context of blockchain. I'll rename to model :)

tests/difference/core/quint_model/ccv.qnt
p-offtermatt: I changed it here and in a few other places where the .with semantics was used a bit too much. I agree the ... way is usually nicer, I just didn't think of it while writing this :)

tests/difference/core/quint_model/ccv.qnt
```diff
@@ -0,0 +1,825 @@
...
+    // gets a protocol state that is initialized minimally.
+    pure def GetEmptyProtocolState(): ProtocolState = 
+        {
+            providerState: GetEmptyProviderState,
+            consumerStates: Map(),
+        }
+
+    type Error = {
+        message: str
+    }
```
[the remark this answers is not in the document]
p-offtermatt: I introduced it to potentially expand the error field later with more info that doesn't fit in the string field
That wasn't necessary after all and I just didn't change it. I'll adjust this and just use strings though

tests/difference/core/quint_model/ccv.qnt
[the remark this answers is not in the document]
bermuell: Not just performance but alos readability. With constructs like this it means to me that other steps have to be executed until you return, but maybe it's just me. Up to you as it's not an issue.

tests/difference/core/quint_model/ccv_statemachine.qnt
p-offtermatt: I just missed it! I changed it to an implication now

bermuell (approved):

PR #1294, opened 2023-09-13
refactor: remove equivocation proposal

insumity:
## Description

Removes the `EquivocationProposal` that was original added [here](https://github.com/cosmos/interchain-security/issues/703) by @tbruyelle.

---

### Author Checklist
I have...

* [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title -- it's not a clear `revert` and hence put it under a `refactor`
* [x] Added `!` to the type prefix if state-machine breaking change (i.e., requires coordinated upgrade) -- it seems that we cannot have validators with this changes and others without it because then some might attempt to propose an equivocation
* [x] Confirmed this PR does not introduce changes requiring state migrations, OR migration code has been added to consumer and/or provider modules
* [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
* [ ] Provided a link to the relevant issue or specification
* [ ] Followed the guidelines for [building SDK modules](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules)
* [ ] Included the necessary unit and integration [tests](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#testing)
* [ ] Added a changelog entry to `CHANGELOG.md` -- not sure if we want to do this here or later when we merge everything under `feat/ics-misbehaviour-handling`
* [ ] Included comments for [documenting Go code](https://blog.golang.org/godoc)
* [x] Updated the relevant documentation or specification
* [x] Reviewed "Files changed" and left comments if necessary 
* [ ] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

* [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ] confirmed `!` in the type prefix if API or client breaking change
* [ ] confirmed this PR does not introduce changes requiring state migrations, OR confirmed migration code has been added to consumer and/or provider modules
* [ ] confirmed all author checklist items have been addressed
* [ ] reviewed state machine logic
* [ ] reviewed API design and naming
* [ ] reviewed documentation is accurate
* [ ] reviewed tests and test coverage

PR #1340, opened 2023-10-03
feat!: add cryptographic equivocation

sainoe:
## Description

Closes: #732

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN go mod tidy
 RUN make install
 
 # Get Hermes build
-FROM ghcr.io/informalsystems/hermes:1.4.1 AS hermes-builder
+FROM otacrew/hermes-ics:evidence-cmd AS hermes-builder
```
sainoe, line 32: Will be replaced with a released version of Hermes.

tests/e2e/actions.go
```diff
@@ -718,8 +645,9 @@ func (tr TestConfig) startChangeover(
 }
 
 type addChainToRelayerAction struct {
-	Chain     ChainID
-	Validator ValidatorID
+	Chain      ChainID
+	Validator  ValidatorID
+	IsConsumer bool
```
p-offtermatt, line 650: Could you also add this in the action_rapid_test Gen for AddChainToRelayer?

tests/e2e/testnet-scripts/hermes-config.toml
```diff
@@ -1,2 +1,18 @@
...
+
+[mode]
+
+[mode.clients]
+enabled = true
+refresh = true
+misbehaviour = true
+
+[mode.connections]
+enabled = false
```
p-offtermatt, line 12: is it worth briefly explaining why these are false/true here?

x/ccv/provider/keeper/hooks.go
```diff
@@ -167,3 +167,7 @@ func (h Hooks) AfterValidatorBeginUnbonding(_ sdk.Context, _ sdk.ConsAddress, _
 func (h Hooks) BeforeDelegationRemoved(_ sdk.Context, _ sdk.AccAddress, _ sdk.ValAddress) error {
 	return nil
 }
+
+func (h Hooks) BeforeTokenizeShareRecordRemoved(_ sdk.Context, _ uint64) error {
```
p-offtermatt, line 171: This is not mentioned anywhere else in the code base - I saw the same is true for some other hooks, too.
Just double-checking and for my understanding: is this for compatibility with something?

p-offtermatt (approved):
Thanks. Commented on some minor things. I went over all changes, but did not go into the most detail for things I recognized from previous rounds of reviews. I approved, but I do think some of the comments need to be addressed before merging.

Should this docs website be adjusted? https://github.com/cosmos/interchain-security/blob/0da6fac7837eb7c6829b9d2563867a2a3984bd15/docs/docs/frequently-asked-questions.md#L75

The status ADR should probably be set to something like "deprecated"? Or better "superseded by ADR 005 [link]" https://github.com/cosmos/interchain-security/blob/0da6fac7837eb7c6829b9d2563867a2a3984bd15/docs/docs/adrs/adr-003-equivocation-gov-proposal.md#L11 

Same with this, since it seems irrelevant now https://github.com/cosmos/interchain-security/blob/0da6fac7837eb7c6829b9d2563867a2a3984bd15/docs/docs/adrs/adr-003-equivocation-gov-proposal.md#L11

Also this should be updated https://github.com/cosmos/interchain-security/blob/0da6fac7837eb7c6829b9d2563867a2a3984bd15/docs/docs/adrs/intro.md#L37

x/ccv/provider/keeper/hooks.go
MSalopek: Hooks are needed to support the staking keeper interface. Most of them are no-ops.

I think we override them to be sure that we use the method specific to the `providerKeeper` since we cannot control what the `sdk.StakingKeeper` hooks do.

CHANGELOG.md
```diff
@@ -12,6 +13,9 @@ Add an entry to the unreleased provider section whenever merging a PR to main th
 * (deps!) [#1196](https://github.com/cosmos/interchain-security/pull/1196) Bump [ibc-go](https://github.com/cosmos/ibc-go) to [v7.2.0](https://github.com/cosmos/ibc-go/releases/tag/v7.2.0).
 * `[x/ccv/provider]` (fix) [#1076](https://github.com/cosmos/interchain-security/pull/1076) Add `InitTimeoutTimestamps` and `ExportedVscSendTimestamps` to exported genesis.
 
+### Cryptographic verification of equivocation
```
insumity, line 16: The linked EPIC's task list also contains slashing issues, so it might make sense to put it in the title here as well?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 Add an entry to the unreleased provider section whenever merging a PR to main that is not targeted at a specific release. These entries will eventually be included in a provider release.
 
+* (refactor) [#1294](https://github.com/cosmos/interchain-security/pull/1294) Remove the equivocation proposal handler.
```
insumity, line 7: Not clear to me when this file gets updated. But seems peculiar we point to this #1294 and not https://github.com/cosmos/interchain-security/pull/1275.

docs/docs/features/slashing.md
```diff
@@ -20,17 +20,10 @@ Instead of slashing, the provider will only jail offending validator for the dur
 Slash throttling (sometimes called jail throttling) mechanism ensures that only a fraction of the validator set can be jailed at any one time to prevent malicious consumer chains from harming the provider.
 :::
 
-## Double-signing (equivocation)
-infractions are not acted upon immediately.
+# Cryptographic verification of equivocation and slashing
+The Cryptographic verification of equivocation allows external agents to submit evidences of light client and double signing attack observed on a consumer chain. When a valid evidence is received, the malicious validators will be slashed, jailed, and tombstoned on the provider.
```
insumity, line 24:
```suggestion
The Cryptographic verification of equivocation allows external agents to submit evidences of light client and double signing attacks observed on a consumer chain. When valid evidence is received, the malicious validators will be slashed, jailed, and tombstoned on the provider.
```

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -28,3 +32,33 @@ message MsgAssignConsumerKey {
 }
 
 message MsgAssignConsumerKeyResponse {}
+
+
+// MsgSubmitConsumerMisbehaviour defines a message that reports a light client attack,
+//  also known as a misbehaviour, observed on a consumer chain
```
insumity, line 38:
```suggestion
// also known as a misbehaviour, observed on a consumer chain
```

x/ccv/provider/client/cli/tx.go
```diff
@@ -65,3 +72,110 @@ func NewAssignConsumerKeyCmd() *cobra.Command {
 
 	return cmd
 }
+
+func NewSubmitConsumerMisbehaviourCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "submit-consumer-misbehaviour [misbehaviour]",
+		Short: "submit an IBC misbehaviour for a consumer chain",
+		Long: strings.TrimSpace(
```
insumity, line 80: nit: is `strings.TrimSpace` needed here? My understanding is that it would just remove leading and trailing spaces but I do not see the underlying string to have any?

x/ccv/provider/client/cli/tx.go
```diff
@@ -65,3 +72,110 @@ func NewAssignConsumerKeyCmd() *cobra.Command {
...
+		Use:   "submit-consumer-double-voting [evidence] [infraction_header]",
+		Short: "submit a double voting evidence for a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Submit a Tendermint duplicated vote evidence detected on a consumer chain with
+ the IBC light client header for the infraction height.
+ The DuplicateVoteEvidence type definition can be found in the Tendermint messages,
+ , see cometbft/proto/tendermint/types/evidence.proto and the IBC header
+ definition can be found in the IBC messages, see ibc-go/proto/ibc/lightclients/tendermint/v1/tendermint.proto.
+
+Examples:
```
insumity, line 137:
```suggestion
Example:
```

x/ccv/provider/client/proposal_handler.go
```diff
@@ -536,33 +440,8 @@ func postConsumerRemovalProposalHandlerFn(clientCtx client.Context) http.Handler
...
-
-		if rest.CheckBadRequestError(w, msg.ValidateBasic()) {
-			return
-		}
-
-		tx.WriteGeneratedTxResponse(clientCtx, w, req.BaseReq, msg)
-	}
-}
-
+func CheckPropUnbondingPeriod(clientCtx client.Context, propUnbondingPeriod time.Duration) {
```
insumity, line 443: This seems like was added by accident and being part of a huge comment section. The `CheckPropUnbondingPeriod` function appears earlier from what I see.

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,110 @@
...
+	pubkey cryptotypes.PubKey,
+) error {
+	if pubkey == nil {
+		return fmt.Errorf("validator public key cannot be empty")
+	}
+
+	// Note that since we're only jailing validators for double voting on a consumer chain,
+	// the age of the evidence is irrelevant and therefore isn't checked.
+
+	// H/R/S must be the same
```
insumity, line 68: Why do we use `S` here in `H/R/S`. I guess `H` stands for `Height`, `R` for `Round`, but `S` doesn't correspond to `Type`?

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	tmtypes "github.com/cometbft/cometbft/types"
+
+	"github.com/cosmos/interchain-security/v3/x/ccv/provider/types"
+)
+
+// HandleConsumerMisbehaviour checks if the given IBC misbehaviour corresponds to an equivocation light client attack,
+// and in this case, jails the Byzantine validators
```
insumity, line 19:
```suggestion
// and in this case, slashes, jails, and tombstones the Byzantine validators
```

x/ccv/types/errors.go
```diff
@@ -6,23 +6,24 @@ import (
...
+	ErrInvalidVSCMaturedTime       = errorsmod.Register(ModuleName, 12, "invalid maturity time for VSC packet")
+	ErrInvalidConsumerState        = errorsmod.Register(ModuleName, 13, "provider chain has invalid state for consumer chain")
+	ErrInvalidConsumerClient       = errorsmod.Register(ModuleName, 14, "ccv channel is not built on correct client")
+	ErrInvalidProposal             = errorsmod.Register(ModuleName, 15, "invalid proposal")
+	ErrInvalidHandshakeMetadata    = errorsmod.Register(ModuleName, 16, "invalid provider handshake metadata")
+	ErrChannelNotFound             = errorsmod.Register(ModuleName, 17, "channel not found")
+	ErrClientNotFound              = errorsmod.Register(ModuleName, 18, "client not found")
+	ErrDuplicateConsumerChain      = errorsmod.Register(ModuleName, 19, "consumer chain already exists")
+	ErrConsumerChainNotFound       = errorsmod.Register(ModuleName, 20, "consumer chain not found")
+	ErrInvalidDoubleVotingEvidence = errorsmod.Register(ModuleName, 21, "invalid consumer double voting evidence")
```
insumity, line 28: I'm wondering if it would anything if we had a similar error type for invalid misbehaviour evidence?

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+	clientStore := k.clientKeeper.ClientStore(ctx, misbehaviour.ClientId)
+
+	// Check that the headers are at the same height to ensure that
+	// the misbehaviour is for a light client attack and not a time violation,
+	// see ibc-go/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
+	if !misbehaviour.Header1.GetHeight().EQ(misbehaviour.Header2.GetHeight()) {
+		return errorsmod.Wrap(ibcclienttypes.ErrInvalidMisbehaviour, "headers are not at same height")
+	}
+
+	// CheckMisbehaviourAndUpdateState verifies the misbehaviour against the trusted consensus states
```
insumity, line 156: Do we need to update the name of the used method in these comments?

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+		return fmt.Errorf("validator is tombstoned. provider consensus address: %s", providerAddr.String())
+	}
+
+	// jail validator if not already
+	if !validator.IsJailed() {
+		k.stakingKeeper.Jail(ctx, providerAddr.ToSdkConsAddr())
+	}
+
+	// Jail the validator to trigger the unbonding of the validator
+	// (see cosmos/cosmos-sdk/blob/v0.45.16-ics-lsm/x/staking/keeper/val_state_change.go#L192).
```
insumity, line 38: Do we still want to refer to `v0.45.16-ics-lsm` or just `v0.47.5` here and in other places?

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+		return fmt.Errorf("validator is unbonded. provider consensus address: %s", providerAddr.String())
+	}
+
+	if k.slashingKeeper.IsTombstoned(ctx, providerAddr.ToSdkConsAddr()) {
+		return fmt.Errorf("validator is tombstoned. provider consensus address: %s", providerAddr.String())
+	}
+
+	// jail validator if not already
+	if !validator.IsJailed() {
+		k.stakingKeeper.Jail(ctx, providerAddr.ToSdkConsAddr())
```
insumity, line 34: Rereading this, I believe just this `Jail` call is enough and we do not need a `JailUntil` call. This is because [Jail](https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/x/staking/keeper/slash.go) calls [jailValidator](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/val_state_change.go#L319) that removes the validator `ByPowerIndex` and therefore later on this validator would start the unbonding [here](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/keeper/val_state_change.go#L149).

Additionally, if we were to just call `Jail` here we still wouldn't be able to unjail in the future because the validator also gets tombstoned (see [here](https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/x/slashing/keeper/unjail.go#L51)).

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+// JailAndTombstoneValidator jails and tombstones the validator with the given provider consensus address
+func (k Keeper) JailAndTombstoneValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) error {
+	validator, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.ToSdkConsAddr())
+	if !found {
+		return errorsmod.Wrapf(slashingtypes.ErrNoValidatorForAddress, "provider consensus address: %s", providerAddr.String())
+	}
+
+	if validator.IsUnbonded() {
```
insumity, line 24: Maybe we could add a comment on why this check is needed:
```suggestion
        // we check if the validator is unbonded to prevent panicking when slashing (see cosmos/cosmos-sdk/blob/v0.47.5/x/staking/keeper/slash.go#L61)
	if validator.IsUnbonded() {
```

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+	if !validator.IsJailed() {
+		k.stakingKeeper.Jail(ctx, providerAddr.ToSdkConsAddr())
+	}
+
+	// Jail the validator to trigger the unbonding of the validator
+	// (see cosmos/cosmos-sdk/blob/v0.45.16-ics-lsm/x/staking/keeper/val_state_change.go#L192).
+	k.slashingKeeper.JailUntil(ctx, providerAddr.ToSdkConsAddr(), evidencetypes.DoubleSignJailEndTime)
+
+	// Tombstone the validator so that we cannot slash the validator more than once
+	// (see cosmos/cosmos-sdk/blob/v0.45.16-ics-lsm/x/evidence/keeper/infraction.go#L81).
```
insumity, line 42: Rereading this, I realize this comment is slightly confusing. We check if the validator is already tombstoned in our own `SlashValidator` function. Therefore, there's no reason to point to some other method here.
```suggestion
```

x/ccv/provider/keeper/punish_validator.go
insumity, line 24: It would seem to me that this check is not needed here because we already perform the check in `SlashValidator` and we always call `SlashValidator` before calling `JailAndTombstoneValidator`. 

Nothing wrong in living it here but then we could maybe add a comment like:
```
        // we check if the validator is unbonded to prevent panicking when slashing (see cosmos/cosmos-sdk/blob/v0.47.5/x/staking/keeper/slash.go#L61)
        // note that we already check this in `SlashValidator` but we check here as well as a potential future safe guard
        // in case someone calls `JailAndTombstoneValidator` before `SlashValidator`
```

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// SlashValidator slashes validator with `providerAddr`
+func (k Keeper) SlashValidator(ctx sdk.Context, providerAddr types.ProviderConsAddress) error {
+	validator, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.ToSdkConsAddr())
+	if !found {
+		return errorsmod.Wrapf(slashingtypes.ErrNoValidatorForAddress, "provider consensus address: %s", providerAddr.String())
+	}
+
+	if validator.IsUnbonded() {
```
insumity, line 89:
```suggestion
        // we check if the validator is unbonded to prevent panicking when slashing (see cosmos/cosmos-sdk/blob/v0.47.5/x/staking/keeper/slash.go#L61)
	if validator.IsUnbonded() {
```

x/ccv/provider/keeper/punish_validator.go
```diff
@@ -0,0 +1,106 @@
...
+		return fmt.Errorf("validator is unbonded. provider consensus address: %s", providerAddr.String())
+	}
+
+	if k.slashingKeeper.IsTombstoned(ctx, providerAddr.ToSdkConsAddr()) {
```
insumity, line 93:
```suggestion
        // check if the validator is already tombstoned to avoid slashing a validator more than once
	if k.slashingKeeper.IsTombstoned(ctx, providerAddr.ToSdkConsAddr()) {
```

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+
+	var errors []error
+	// slash, jail, and tombstone the Byzantine validators
+	for _, v := range byzantineValidators {
+		providerAddr := k.GetProviderAddrFromConsumerAddr(
+			ctx,
+			misbehaviour.Header1.Header.ChainID,
+			types.NewConsumerConsAddress(sdk.ConsAddress(v.Address.Bytes())),
+		)
+		err := k.SlashValidator(ctx, providerAddr)
```
insumity, line 50: Rereading this, it feels we're not really consistent in the way we act when `SlashValidator` fails. In the double-voting case, we return an `err`or, while here we might go ahead and jail and tombstone the validator without having successfully slashed the validator. However, I do not believe this is a problem because both `SlashValidator` and `JailAndTombstoneValidator` return an `err`or in the exact same cases, namely when i) the provider address is not found, ii) the validator is unbonded, or iii) the validator is tombstoned. With this in mind, if `SlashValidator` fails, then `JailAndTombstoneValidator` should fail as well. Because of this, I do not believe we can have a case where we slash a validator, fail to jail and tombstone but `SlashValidator` still returns `nil` and hence we could keep slashing the validator until it has no more tokens.


Does this sound reasonable? If yes, I'll add a suggestion/comment here to explain this. This would also imply that we should **not** remove the unbonded check from `JailAndTombstoneValidator`.

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+		err = k.JailAndTombstoneValidator(ctx, providerAddr)
+		if err != nil {
+			errors = append(errors, err)
+		}
+		provAddrs = append(provAddrs, providerAddr)
+	}
+
+	logger.Info(
+		"confirmed equivocation light client attack",
+		"byzantine validators", provAddrs,
```
insumity, line 63: Let's either do

```suggestion
		"byzantine validators that got slashes, jailed, and tombstoned", provAddrs,
```

or

```suggestion
		"byzantine validators", byzantineValidators,
```

because `provAddrs` might not be the same as `byzantineValidators`.

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// VerifyDoubleVotingEvidence verifies a double voting evidence
+// for a given chain id and a validator public key
+func (k Keeper) VerifyDoubleVotingEvidence(
+	ctx sdk.Context,
+	evidence tmtypes.DuplicateVoteEvidence,
+	chainID string,
+	pubkey cryptotypes.PubKey,
+) error {
```
insumity, line 60: Unless I'm missing something, it seems to me that we need a check on whether the provided `pubKey` can generate the provided `validatorAddress` inside `evidence`. Is this done somewhere?
If we do not have this check, can we guarantee that someone doesn't simply sign two blocks with an arbitrary public key but with a correct `validatorAddress` and has a validator get slashed without really having equivocated?

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/cosmos/interchain-security/v3/x/ccv/provider/types"
+)
+
+// HandleConsumerMisbehaviour checks if the given IBC misbehaviour corresponds to an equivocation light client attack,
+// and in this case, jails the Byzantine validators
+func (k Keeper) HandleConsumerMisbehaviour(ctx sdk.Context, misbehaviour ibctmtypes.Misbehaviour) error {
+	logger := k.Logger(ctx)
+
+	// Check that the misbehaviour is valid and that the client consensus states at trusted heights are within trusting period
+	if err := k.CheckMisbehaviour(ctx, misbehaviour); err != nil {
```
insumity, line 24: I would have expected that `CheckMisbehaviour` would check that singatures are valid based on some public keys, etc. under the hoods, similarly to what we do for double voting, but couldn't find something like this.

What guarantees us that we won't accept 2 arbitrary headers for the sam eheight with real validator addresses and arbitrary signatures that would lead to slashing?

x/ccv/provider/types/msg.go
```diff
@@ -93,3 +109,99 @@ func ParseConsumerKeyFromJson(jsonStr string) (pkType, key string, err error) {
...
+// Route implements the sdk.Msg interface.
+func (msg MsgSubmitConsumerDoubleVoting) Route() string { return RouterKey }
+
+// Type implements the sdk.Msg interface.
+func (msg MsgSubmitConsumerDoubleVoting) Type() string {
+	return TypeMsgSubmitConsumerDoubleVoting
+}
+
+// Type implements the sdk.Msg interface.
+func (msg MsgSubmitConsumerDoubleVoting) ValidateBasic() error {
```
insumity, line 166: How thorough do we have to be with these checks?
Could a malicious actor that provides `nil` to a message field that we do not check against halt the chain?


E..g, Should we check that `VoteA` and `VoteB` are also there and that each have a `BlockID` that is not `nil` and so on and so forth?

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	provAddrs := make([]types.ProviderConsAddress, len(byzantineValidators))
+
+	var errors []error
+	// slash, jail, and tombstone the Byzantine validators
+	for _, v := range byzantineValidators {
+		providerAddr := k.GetProviderAddrFromConsumerAddr(
+			ctx,
+			misbehaviour.Header1.Header.ChainID,
```
insumity, line 47: Do we have a check somewhere that both `Header1` and `Header2` are for the same `ChainID`?
If not, could we have a case where headers are used from different chains?

x/ccv/provider/keeper/punish_validator.go
sainoe: I agree with the reasoning in general but I think it's safer to follow the [SDK evidence handler logic](https://github.com/cosmos/cosmos-sdk/blob/main/x/evidence/keeper/infraction.go#L141).

x/ccv/provider/types/msg.go
sainoe: It can be argued to have a stricter validation. However, any panic will be recovered and will just reject the tx without halting the chain.

x/ccv/provider/client/cli/tx.go
sainoe: Idnk the exact reason but It's kind of a convention used in the SDK cli.

x/ccv/provider/keeper/double_vote.go
sainoe: Comment taken from [CometBFT](https://github.com/cometbft/cometbft/blob/v0.37.2/evidence/verify.go#L169), I guess `S` stands for `SignMsgType`.

x/ccv/types/errors.go
sainoe: Because IBC already defines an [`ErrInvalidMisbehaviour`](https://github.com/cometbft/cometbft/blob/v0.37.2/evidence/verify.go#L169).

x/ccv/provider/keeper/punish_validator.go
sainoe: Agree we can remove these references.

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	logger.Info(
+		"confirmed equivocation light client attack",
+		"byzantine validators", provAddrs,
+	)
+
+	// If we fail to slash all validators we return an error. However, if we only fail to slash some validators
+	// we just log an error to avoid having the whole `MsgSubmitMisbehaviour` failing and reverting the partial slashing.
+	if len(errors) == len(byzantineValidators) {
```
insumity, line 68: Here, it should've been:
```suggestion
	if len(errors) == 2 * len(byzantineValidators) {
```
because an error is appended once when `SlashValidator` fails and a second time when `JailAndTombstoneValidator` fails. My mistake.

Technically, `len(errors)` should always be an even number, otherwise something went wrong (according to my previous [comment](https://github.com/cosmos/interchain-security/pull/1340/files#r1356584667)). Maybe we can log something if that's not the case -- it's not clear to me who's even looking at the logs though? Is the assumption here that when we log something, eventually a validator would ping us about them?

x/ccv/provider/keeper/double_vote.go
insumity: I see. It still might make sense to just have `Height/Round/Type` here  even if CometBFT has it otherwise.

x/ccv/provider/types/msg.go
insumity: Thanks. I wasn't aware of this. Resolving it.

x/ccv/types/errors.go
insumity: Maybe it would add a bit value having a comment in this `errors.go` file above (or below) `ErrInvalidDoubleVotingEvidence` on this that `ErrInvalidMisbehaviour` pre-exists in IBC.

x/ccv/provider/keeper/misbehaviour.go
sainoe: Solved in https://github.com/cosmos/interchain-security/pull/1401.

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,118 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	tmtypes "github.com/cometbft/cometbft/types"
+
+	"github.com/cosmos/interchain-security/v3/x/ccv/provider/types"
+	ccvtypes "github.com/cosmos/interchain-security/v3/x/ccv/types"
+)
+
+// HandleConsumerDoubleVoting verifies a double voting evidence for a given a consumer chain ID
+// and a public key and, if successful, executes the jailing of the malicious validator.
```
insumity, line 19:
```suggestion
// and a public key and, if successful, executes the slashing, jailing, and tombstoning of the malicious validator.
```

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,118 @@
...
+	// check that the validator address in the evidence is derived from the provided public key
+	if !bytes.Equal(pubkey.Address(), evidence.VoteA.ValidatorAddress) {
+		return errorsmod.Wrapf(
+			ccvtypes.ErrInvalidDoubleVotingEvidence,
+			"public key %s doesn't correspond to the validator address %s in double vote evidence",
+			pubkey.String(), evidence.VoteA.ValidatorAddress.String(),
+		)
+	}
+
+	// Note that since we're only jailing validators for double voting on a consumer chain,
```
insumity, line 73: Remove this comment since we do not "only jail" now.

x/ccv/provider/keeper/double_vote.go
```diff
@@ -0,0 +1,118 @@
...
+	// the age of the evidence is irrelevant and therefore isn't checked.
+
+	// height/round/type must be the same
+	if evidence.VoteA.Height != evidence.VoteB.Height ||
+		evidence.VoteA.Round != evidence.VoteB.Round ||
+		evidence.VoteA.Type != evidence.VoteB.Type {
+		return errorsmod.Wrapf(
+			ccvtypes.ErrInvalidDoubleVotingEvidence,
+			"h/r/s does not match: %d/%d/%v vs %d/%d/%v",
```
insumity, line 82:
```suggestion
			"height/round/type are not the same: %d/%d/%v vs %d/%d/%v",
```

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,233 @@
...
+		return fmt.Errorf("incorrect misbehaviour: expected client ID for consumer chain %s is %s got %s",
+			misbehaviour.Header1.Header.ChainID,
+			clientId,
+			misbehaviour.ClientId,
+		)
+	}
+
+	clientState, found := k.clientKeeper.GetClientState(ctx, clientId)
+	if !found {
+		return errorsmod.Wrapf(ibcclienttypes.ErrClientNotFound, "cannot check misbehaviour for client with ID %s", misbehaviour.ClientId)
```
insumity, line 172: Maybe make the specific error more specific:
```suggestion
		return errorsmod.Wrapf(ibcclienttypes.ErrClientNotFound, "cannot find client state for client with ID %s", clientId)
```

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	clientStore := k.clientKeeper.ClientStore(ctx, misbehaviour.ClientId)
```
insumity, line 175:
```suggestion
	clientStore := k.clientKeeper.ClientStore(ctx, clientId)
```

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	return &tmtypes.LightBlock{
+		SignedHeader: sh,
+		ValidatorSet: vs,
+	}, nil
+}
+
+// CheckMisbehaviour checks that headers in the given misbehaviour forms
+// a valid light client attack from an ICS consumer chain and that the light client isn't expired
```
insumity, line 156: > [...] and that the light client isn't expired

Where's this check performed?
I think we can remove this expiration message from the comment. Additionally, maybe we can say "[...] a valid light client attack on a light client that tracks an ICS consumer chain" instead of "[...]  a valid light client attack from an ICS consumer chain"

x/ccv/provider/types/proposal.go
```diff
@@ -204,6 +204,7 @@ func (sccp *ConsumerRemovalProposal) ValidateBasic() error {
 }
 
 // NewEquivocationProposal creates a new equivocation proposal.
+// [DEPRECATED]: do not use
```
insumity, line 207: I guess this message is related to this https://github.com/cosmos/interchain-security/pull/1394? I would add a message on why it's deprecated and why it's still here nevertheless.

I was wondering, do we have to keep this piece of code and the corresponding equivocation proposal protos or could we get rid of them when we release a new _MAJOR_ version?

x/ccv/types/events.go
```diff
@@ -2,17 +2,42 @@ package types
...
-
-	AttributeChainID          = "chain_id"
-	AttributeValidatorAddress = "validator_address"
-	AttributeInfractionType   = "infraction_type"
-
-	AttributeValSetUpdateID = "valset_update_id"
+	AttributeKeyAckSuccess            = "success"
+	AttributeKeyAck                   = "acknowledgement"
+	AttributeKeyAckError              = "error"
+	AttributeValidatorConsumerAddress = "validator_consumer_address"
```
insumity, line 23: Not related to this PR per se but I do not think `AttributeValidatorConsumerAddress` is used anywhere (?) in which case we can remove.

x/ccv/provider/client/cli/tx.go
```diff
@@ -65,3 +72,110 @@ func NewAssignConsumerKeyCmd() *cobra.Command {
...
+
+	return cmd
+}
+
+func NewSubmitConsumerDoubleVotingCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "submit-consumer-double-voting [evidence] [infraction_header]",
+		Short: "submit a double voting evidence for a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Submit a Tendermint duplicated vote evidence detected on a consumer chain with
```
insumity, line 131:
```suggestion
			fmt.Sprintf(`Submit a Tendermint duplicate vote evidence detected on a consumer chain with
```

x/ccv/provider/client/cli/tx.go
```diff
@@ -65,3 +72,110 @@ func NewAssignConsumerKeyCmd() *cobra.Command {
...
+ the IBC light client header for the infraction height.
+ The DuplicateVoteEvidence type definition can be found in the Tendermint messages,
+ , see cometbft/proto/tendermint/types/evidence.proto and the IBC header
```
insumity, line 134:
```suggestion
 see cometbft/proto/tendermint/types/evidence.proto and the IBC header
```

.changelog/unreleased/features/provider/1340-cryptographic-equivocation-feature.md
```diff
@@ -0,0 +1,3 @@
+- Add the cryptographic verification of equivocation feature to the provider
+  (cf. ADR-005 & ADR-013). ([\#1340](https://github.com/cosmos/interchain-
```
insumity, line 2: Link doesn't seem to show correctly.
Screenshot 2023-11-16 at 11 14 04
Probably you need:
```suggestion
  (cf. ADR-005 & ADR-013). ([\#1340](https://github.com/cosmos/interchain-security/pull/1340))
```

.changelog/unreleased/state-breaking/provider/1340-cryptographic-equivocation-feature.md
```diff
@@ -0,0 +1,3 @@
+- Add the cryptographic verification of equivocation feature to the provider
+  (cf. ADR-005 & ADR-013). ([\#1340](https://github.com/cosmos/interchain-
```
insumity, line 2: Link has the same issue as [above](https://github.com/cosmos/interchain-security/pull/1340/files#r1395465459).

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -106,10 +106,13 @@ message ConsumerRemovalProposal {
 // punish a validator for equivocation on a consumer chain.
 // 
 // This type is only used internally to the consumer CCV module.  
+// WARNING: This message is deprecated now that equivocations can be submitted
+// and verified automatically on the provider. (see SubmitConsumerDoubleVoting in proto/interchain-security/ccv/provider/v1/tx.proto).
```
insumity, line 110: Can we remove those at some point? With a next release or so? If yes, then it would make sense to create an issue so we don't forget to do so.

tests/e2e/steps.go
```diff
@@ -106,3 +100,18 @@ var changeoverSteps = concatSteps(
...
+	// start provider and consumer chain
+	stepsStartChainsWithSoftOptOut("consu"),
+	// make a consumer validator to misbehave and get jailed
+	stepsCauseConsumerMisbehaviour("consu"),
+)
+
+var consumerDoubleSignSteps = concatSteps(
+	// start provider and consumer chain
+	stepsStartChains([]string{"consu"}, false),
+
```
insumity, line 114: nit: Why do we have a new line here while in  `consumerMisbehaviourSteps` we do not have a line between `stepsStartChainsWithSoftOptOut("consu")` and `stepsCauseConsumerMisbehaviour("consu")`.

tests/integration/double_vote.go
```diff
@@ -0,0 +1,350 @@
...
+
+	tmcrypto "github.com/cometbft/cometbft/crypto"
+	tmtypes "github.com/cometbft/cometbft/types"
+
+	testutil "github.com/cosmos/interchain-security/v3/testutil/crypto"
+	"github.com/cosmos/interchain-security/v3/x/ccv/provider/types"
+)
+
+// TestHandleConsumerDoubleVoting verifies that handling a double voting evidence
+// of a consumer chain results in the expected tombstoning and jailing the misbehaved validator
```
insumity, line 15:
```suggestion
// of a consumer chain results in the expected tombstoning, jailing, and slashing of the misbehaved validator
```

tests/integration/double_vote.go
```diff
@@ -0,0 +1,350 @@
...
+	s.Require().NoError(err)
+
+	provVal := provValSet.Validators[0]
+	provSigner := s.providerChain.Signers[provVal.Address.String()]
+
+	blockID1 := testutil.MakeBlockID([]byte("blockhash"), 1000, []byte("partshash"))
+	blockID2 := testutil.MakeBlockID([]byte("blockhash2"), 1000, []byte("partshash"))
+
+	// Note that votes are signed along with the chain ID
+	// see VoteSignBytes in https://github.com/cometbft/cometbft/blob/main/types/vote.go#L139
```
insumity, line 41: Use a specific release reference here because line `139` at `vote.go` might show something else in the future.

x/ccv/types/events.go
sainoe: Good catch. Removed.

x/ccv/provider/types/proposal.go
sainoe: Added more info on why it's deprecated. Responded below why it's not simple to remove it.

insumity (approved):

proto/interchain_security/ccv/provider/v1/provider.proto
p-offtermatt: Meaning we should do this if we have a release coming up that needs a migration anyways?

tests/integration/double_vote.go
```diff
@@ -307,12 +336,15 @@ func (s *CCVTestSuite) TestHandleConsumerDoubleVotingSlashesUndelegations() {
...
 		// check undelegations are slashed
-		ubds, _ := s.providerApp.GetTestStakingKeeper().GetUnbondingDelegation(s.providerCtx(), delAddr, validator.GetOperator())
+		ubds, _ = s.providerApp.GetTestStakingKeeper().GetUnbondingDelegation(s.providerCtx(), delAddr, validator.GetOperator())
 		s.Require().True(len(ubds.Entries) > 0)
 		for _, unb := range ubds.Entries {
 			initialBalance := sdk.NewDecFromInt(unb.InitialBalance)
 			currentBalance := sdk.NewDecFromInt(unb.Balance)
 			s.Require().True(initialBalance.Sub(initialBalance.Mul(slashFraction)).Equal(currentBalance))
 		}
+
```
p-offtermatt, line 346:
```suggestion
// check that redelegations are slashed
```

That's what the next 2 lines do, right?

x/ccv/provider/types/proposal.go
```diff
@@ -198,6 +203,41 @@ func (sccp *ConsumerRemovalProposal) ValidateBasic() error {
 	return nil
 }
 
+// NewEquivocationProposal creates a new equivocation proposal.
+// [DEPRECATED]: do not use since that equivocations can be submitted
```
p-offtermatt, line 207:
```suggestion
// [DEPRECATED]: do not use because equivocations can be submitted
```
I don't get the comment, is the suggestion what it's supposed to say?

p-offtermatt (approved):
I reviewed the diff from my previous review.
I didn't see anything I found concerning. Just a few minor suggestions,
mostly for comments.

Issue #1241, opened 2023-08-25
Write a shim/driver that handles Quint-generated traces


p-offtermatt:
We likely want the trace format we want the e2e tests to consume to be much lower-level than the traces generated by the Quint model.
Thus, we need to have a shim/driver between them, translating the high-level Quint trace into a low-level "e2e" trace.
This should be implemented as a simple utility in Golang. It should have both ways to be used as a CLI tool to translate and dump the "e2e-level" traces, as well as a way to be called in-memory to consume Quint-level traces and construct the "e2e" trace without ever dumping it to a file.

*Closing Criterion*
The shim/driver is implemented as specified above.

p-offtermatt:
It was decided to, instead of targetting the e2e tests for this, first target in-memory tests with this, i.e.
using the same underlying tech as our integration tests.

PR #1369, opened 2023-10-19
test: Replace difftest model with Quint=>MBT

p-offtermatt:
## Description

Closes: #[1241](https://github.com/cosmos/interchain-security/issues/1241)

Changes:
* Changes to the Quint model: bug fixes, things that make it easier to use to generate traces, ...
* Removes the difference tests in tests/difference
* Adds tests/mbt
* Adds a driver, which uses ibctesting to take Quint traces and run them against provider and consumer apps
* Adds infrastructure to generate traces from the Quint model
* Adds the mbt tests to the CI
* Adds a README explaining the MBT approach, how to use, etc
* Made some changes to the Quint model. There are 2 types of changes: Making the traces more useful for MBT (doing a change amount instead of new voting power for VotingPowerChange, keeping the parameter values in a variable, ...) and making the model actually correct (splitting timestamps into running and last timestamps, ...)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


app/consumer-democracy/proposals_whitelisting_test.go
```diff
@@ -12,7 +12,8 @@ import (
 )
 
 func TestDemocracyGovernanceWhitelistingKeys(t *testing.T) {
-	_, valUpdates, _ := testutil.CreateValidators(t, 4)
+	_, valUpdates, _, err := testutil.CreateValidators(4)
```
p-offtermatt, line 15: I changed the signature of the CreateValidators function to better suit my needs, so I had to adjust this here

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,418 @@
...
+func (s *Driver) delegator() sdk.AccAddress {
+	return s.providerChain().SenderAccount.GetAddress()
+}
+
+// validator returns the address for the validator with id (ix) i
+func (s *Driver) validator(i int64) sdk.ValAddress {
+	return sdk.ValAddress(s.validators[i].Address)
+}
+
+// consumerPower returns the power on the consumer chain chain for
```
insumity, line 115:
```suggestion
// consumerPower returns the power on the consumer chain for
```
Would it make sense to just write "index" instead of "id (ix)"?

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,418 @@
...
+		state.WriteString(fmt.Sprintf("Chain %s\n", chain))
+		state.WriteString(s.getChainStateString(chain))
+		state.WriteString("\n")
+	}
+
+	return state.String()
+}
+
+func (s *Driver) isProviderChain(chain ChainId) bool {
+	return chain == "provider"
```
insumity, line 216: Could also do `chain == P` to reuse `P`.

tests/mbt/driver/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	invariantConfig := InvariantConfig{
+		invariant:  *invariant,
+		numSamples: *numSamples,
+		numSteps:   *numSteps,
+	}
+
+	// Print the parsed values
+	fmt.Println("Model Config:")
```
insumity, line 56: nit: add space after `:`

tests/mbt/model/ccv.qnt
```diff
@@ -385,63 +406,69 @@ module ccv {
...
         )
         
-        // check for Vsc timeouts
-        val timedOutConsumers = getRunningConsumers(providerStateAfterTimeAdvancement).filter(
-            consumer => 
-            val res = TimeoutDueToVscTimeout(tmpState, consumer)
-            res._1
-        )
+
+        // send vsc packets
```
insumity, line 410: nit: in previous comment `Vsc` is used instead of `vsc`

tests/mbt/driver/main.go
p-offtermatt: Not needed, see how this prints:

```
Model Config:
Model Path: ../model/ccv_boundeddrift.qnt
Init: init
Step: stepBoundedDrift
Invariant Config:
Invariant: CanTimeoutConsumer
Num Samples: 200
Num Steps: 200
Trace Folder: traces/bound_timeout
Num Traces: 20
```

tests/mbt/model/ccv_model.qnt
```diff
@@ -176,9 +171,118 @@ module ccv_model {
...
+            runningConsumers.size() > 0, // ensure there is a running consumer, otherwise this action does not make sense
+            nondet sender = oneOf(runningConsumers)
+            DeliverVscMaturedPacket(sender),
+        },
+
+        // again, we could filter by chains that can actually receive,
+        // but it's probably not much faster than just trying and failing.
+        all {
+            runningConsumers.size() > 0, // ensure there is a running consumer, otherwise this action does not make sense
+            nondet recciver = oneOf(runningConsumers)
```
insumity, line 198:
```suggestion
            nondet receiver = oneOf(runningConsumers)
            DeliverVscPacket(receiver),
```

tests/mbt/model/ccv_model.qnt
```diff
@@ -176,9 +171,118 @@ module ccv_model {
...
+            validatorSet: Map(),
+            sendingTime: -9999999999 * Second,
+            timeoutTime: -9999999999 * Second,
+        }
+        packets.fold(
+            earliestPossiblePacket,
+            (res, pack) => if(res.sendingTime >= pack.sendingTime) { res } else { pack }
+        )
+
+    // run maxByTest =
```
insumity, line 252: remove `maxByTest` if not used

tests/mbt/model/ccv_model.qnt
```diff
@@ -176,9 +171,118 @@ module ccv_model {
...
+        EndAndBeginBlockForProvider(timeAdvancement, consumersToStart, consumersToStop),
+
+        step_common
+    }
+
+    // ==================
+    // UTILITY FUNCTIONS
+    // ==================
+
+    pure def earliest(packets: Set[VscPacket]): VscPacket =
```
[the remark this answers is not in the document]
p-offtermatt: They used something that was not supported in Apalache, see this: https://github.com/informalsystems/quint/issues/1235

tests/mbt/model/ccv_boundeddrift.qnt
```diff
@@ -0,0 +1,97 @@
...
+
+    def GetRunningChainStates(): Set[Ccvt::ChainState] = 
+        val runCons = ccv::getRunningConsumers(currentState.providerState)
+        val consumerChainStates = runCons.map(c => currentState.consumerStates.get(c).chainState)
+        consumerChainStates.union(Set(currentState.providerState.chainState))
+
+    // step will advance time for all chains at the same rate,
+    // thus the clock times are always in sync.
+    // This is useful to test happy paths.
+    action stepBoundedDrift = any {
```
insumity, line 56: nit: camel case (`stepBoundedDrift`) versus snake case (`step_common`)

tests/mbt/model/ccv_boundeddrift.qnt
```diff
@@ -0,0 +1,97 @@
...
+        
+        // advance a block for the provider
+        val maxAdv = findMaxTimeAdvancement(GetChainState(Ccvt::PROVIDER_CHAIN), GetOtherChainStates(Ccvt::PROVIDER_CHAIN), maxDrift)
+        val possibleAdvancements = timeAdvancements.filter(t => t <= maxAdv)
+        all {
+            possibleAdvancements.size() > 0, // ensure there is a possible advancement, otherwise this action does not make sense
+            // advance a block for the provider
+            val consumerStatus = currentState.providerState.consumerStatus
+            nondet consumersToStart = oneOf(nonConsumers.powerset())
+            // make it so we stop consumers only with small likelyhood:
```
insumity, line 80: nit: s/likelyhood/likelihood

tests/mbt/model/ccv_sync.qnt
```diff
@@ -0,0 +1,67 @@
...
+                (acc, consumer) => acc.append((consumer, 1 * Second))
+            ),
+        }
+
+
+
+    // step will advance time for all chains at the same rate,
+    // thus the clock times are always in sync.
+    // This is useful to test happy paths.
+    action stepHappy = any {
```
insumity, line 41: nit: I wonder if a name `stepInSync` or something in this vein would be slightly better because the steps lead to happy paths but the steps themselves are not happy.

tests/mbt/model/ccv_sync.qnt
```diff
@@ -0,0 +1,67 @@
...
+        all {
+            QueuedEndBlocks.length() > 0,
+            val pair = QueuedEndBlocks.head()
+            val chain = pair._1
+            val timeAdv = pair._2
+            EndAndBeginBlockForConsumer(chain, timeAdv),
+            QueuedEndBlocks' = QueuedEndBlocks.tail(),
+        },
+
+        all {
```
insumity, line 61: Is my understanding correct that this `all`{ ... }` is added here so that packets get delivered, etc. through `step_common`?

tests/mbt/model/ccv_model.qnt
```diff
@@ -225,8 +328,8 @@ module ccv_model {
                     if (commonPackets.size() == 0) {
                         true // they don't share any packets, so nothing to check
                     } else {
-                        val newestCommonPacket = commonPackets.maxBy(packet => packet.sendingTime)
-                        val oldestCommonPacket = commonPackets.minBy(packet => packet.sendingTime)
+                        val newestCommonPacket = commonPackets.latest()
```
insumity, line 331: nit: maybe `latest()` could be renamed to `newest` and similarly `earliest()` to `oldest` because, for example, saying `latest` might confuse someone to think of the `oldest`.

tests/mbt/model/ccv_model.qnt
```diff
@@ -359,50 +469,71 @@ module ccv_model {
...
-                    assert(currentState.consumerStates.get("consumer1").maturationTimes.keys().size() == 1),
+                    assert(currentState.consumerStates.get("consumer1").maturationTimes.length() == 1),
                     // the validator set was put as the current validator set
-                    assert(currentState.consumerStates.get("consumer1").chainState.currentValidatorSet == InitialValidatorSet.put("node1", 50)),
-                    // advance time on provider until the unbonding period is over
-                    EndAndBeginBlockForProvider(UnbondingPeriodPerChain.get("consumer1"), Set(), Set()),
+                    assert(currentState.consumerStates.get("consumer1").chainState.currentValidatorSet == InitialValidatorSet.put("node1", 200)),
+                    // advance time on provider and consumer until unbonding period is over - ensure that the consumer and provider
+                    // stay in sync relative to each other
+                    4.reps(
```
insumity, line 511: What's the benefit of advancing `UnbondingPeriod / 4` 4 times instead of advancing `UnbondingPeriod` at once?

tests/mbt/model/ccv_model.qnt
```diff
@@ -482,6 +613,10 @@ module ccv_model {
             // advance time on provider by VscTimeout + 1 Second
             EndAndBeginBlockForProvider(VscTimeout + 1 * Second, Set(), Set())
         )
+        .then(
+            // eenter the advanced time on chain
```
insumity, line 617:
```suggestion
            // enter the advanced time on chain
```

tests/mbt/model/ccv_model.qnt
```diff
@@ -5,16 +5,36 @@ module ccv_model {
...
+        VscTimeout: Time,
+        CcvTimeout: Chain -> Time,
+        UnbondingPeriodPerChain: Chain -> Time,
+        TrustingPeriodPerChain: Chain -> Time,
+        ConsumerChains: Set[Chain],
+        Nodes: Set[Node],
+        InitialValidatorSet: Node -> int,
+    }
 
+    var params: Parameters
```
insumity, line 37: Not fully clear how `params` are utilized. Might be helpful to add a comment here :)

tests/mbt/model/ccv_sync.qnt
p-offtermatt: 😢 <- the steps

Jokes aside, good feedback! The Happy name was from an older version, but now with 2 different modules doing different levels of happy things, the change is probably warranted

tests/mbt/model/ccv_model.qnt
p-offtermatt: Added comment, please resolve if that clears it up

tests/mbt/model/ccv_model.qnt
insumity: This makes sense. Thanks for the explanation.

tests/mbt/model/ccv_model.qnt
```diff
@@ -508,6 +512,8 @@ module ccv_model {
                     assert(currentState.consumerStates.get("consumer1").chainState.currentValidatorSet == InitialValidatorSet.put("node1", 200)),
                     // advance time on provider and consumer until unbonding period is over - ensure that the consumer and provider
                     // stay in sync relative to each other
+                    // We want to make sure the steps we make are shorter the trusting period,
```
insumity, line 515:
```suggestion
                    // We want to make sure the steps we make are shorter than the trusting period,
```

tests/mbt/model/ccv_model.qnt
insumity: 👍 Makes sense. Thanks!

tests/mbt/driver/generate_traces.sh
```diff
@@ -0,0 +1,10 @@
+echo "Generating bounded drift traces with timeouts"
+go run ./... -modelPath=../model/ccv_boundeddrift.qnt -step stepBoundedDrift  -invariant CanTimeoutConsumer -traceFolder traces/bound_timeout -numTraces 1 -numSteps 200 -numSamples 200
+echo "Generating long bounded drift traces without invariants"
+go run ./... -modelPath=../model/ccv_boundeddrift.qnt -step stepBoundedDrift -traceFolder traces/bound_noinv -numTraces 1 -numSteps 500 -numSamples 1
+echo "Generating bounded drift traces with maturations"
+go run ./... -modelPath=../model/ccv_boundeddrift.qnt -step stepBoundedDrift -invariant CanReceiveMaturations -traceFolder traces/bound_mat -numTraces 1 -numSteps 100 -numSamples 20
+echo "Generating synced traces with maturations"
+go run ./... -modelPath=../model/ccv_sync.qnt -init initHappy -step stepSync -invariant CanReceiveMaturations -traceFolder traces/sync_mat -numTraces 1 -numSteps 300 -numSamples 20
```
insumity, line 8: s/initHappy/initSync here and below?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+		TrustingPeriodPerChain:  trustingPeriodPerChain,
+	}
+
+	valExprs := params["Nodes"].Value.(itf.ListExprType)
+	valNames := make([]string, len(valExprs))
+	for i, val := range valExprs {
+		valNames[i] = val.Value.(string)
+	}
+
+	// initialValSet has the right vals, but not yet the right powers
```
insumity, line 143: Why doesn't it have the right voting powers?
Don't we get the right voting powers earlier when we do:
```
		initialValSet[val] = power.Value.(int64)
```

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+	timeOffset := driver.runningTime("provider")
+
+	t.Log("Started chains")
+
+	t.Log("Reading the trace...")
+
+	for index, state := range trace.States {
+		t.Logf("Reading state %v of trace %v", index, path)
+
+		// modelState := state.VarValues["currentState"]
```
insumity, line 170: nit: clean up if not used.

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+			consumersToStop := lastAction["consumersToStop"].Value.(itf.ListExprType)
+			t.Log("EndAndBeginBlockForProvider", timeAdvancement, consumersToStart, consumersToStop)
+
+			stats.numStartedChains += len(consumersToStart)
+			stats.numStops += len(consumersToStop)
+
+			// we need 2 blocks, because for a packet sent at height H, the receiving chain
+			// needs a header of height H+1 to accept the packet
+			// so we do one time advancement with a very small increment,
+			// and then increment the rest of the time]
```
insumity, line 210:
```suggestion
			// and then increment the rest of the time
```

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+			// their times were messed up by the coordinator
+			for consumer, timestamp := range lastTimestamps {
+				driver.setTime(consumer, timestamp)
+			}
+
+			// for all connected consumers, update the clients...
+			// unless it was the last consumer to be started, in which case it already has the header
+			for _, consumer := range driver.runningConsumers() {
+				if len(consumersToStart) > 0 && consumer.ChainId == consumersToStart[len(consumersToStart)-1].Value.(string) {
+					continue
```
insumity, line 267: It's not clear to me why the last consumer to be started already has the provider's header?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+
+	t.Log("Reading the trace...")
+
+	for index, state := range trace.States {
+		t.Logf("Reading state %v of trace %v", index, path)
+
+		// modelState := state.VarValues["currentState"]
+		trace := state.VarValues["trace"].Value.(itf.ListExprType)
+		// fmt.Println(modelState)
+		lastAction := trace[len(trace)-1].Value.(itf.MapExprType)
```
insumity, line 173: Might be useful to have a comment on why we only care of the `lastAction`. If my understanding is correct, subsequent `state`s always contain all the actions of all previous `state`s.

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+		CompareSentPacketsOnProvider(driver, currentModelState, timeOffset)
+
+		stats.EnterStats(driver)
+
+		t.Logf("State %v of trace %v is ok!", index, path)
+	}
+	t.Log("🟢 Trace is ok!")
+}
+
+func CompareValidatorSets(t *testing.T, driver *Driver, currentModelState map[string]itf.Expr, consumers []string, index int) {
```
insumity, line 384: remove `index` if not used

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+			if !found {
+				providerConsAddr = providertypes.NewProviderConsAddress(consAddr.Address)
+			}
+
+			providerVal, found := driver.providerStakingKeeper().GetValidatorByConsAddr(driver.providerCtx(), providerConsAddr.Address)
+			require.True(t, found, "Error getting provider validator")
+
+			consumerCurValSet[providerVal.GetMoniker()] = val.Power
+		}
+		for val, power := range modelValSet {
```
insumity, line 419: Is this `for` loop needed?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+
+			providerConsAddr, found := driver.providerKeeper().GetValidatorByConsumerAddr(driver.providerCtx(), consumer, consAddr)
+			if !found {
+				providerConsAddr = providertypes.NewProviderConsAddress(consAddr.Address)
+			}
+
+			providerVal, found := driver.providerStakingKeeper().GetValidatorByConsAddr(driver.providerCtx(), providerConsAddr.Address)
+			require.True(t, found, "Error getting provider validator")
+
+			consumerCurValSet[providerVal.GetMoniker()] = val.Power
```
insumity, line 417: Why is `providerVal` added to the `consumerCurValSet`?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+
+		t.Logf("State %v of trace %v is ok!", index, path)
+	}
+	t.Log("🟢 Trace is ok!")
+}
+
+func CompareValidatorSets(t *testing.T, driver *Driver, currentModelState map[string]itf.Expr, consumers []string, index int) {
+	t.Helper()
+	modelValSet := ValidatorSet(currentModelState, "provider")
+	curValSet := driver.providerValidatorSet()
```
insumity, line 387: Maybe just have:
```
actualValSet := make(map[string]int64, len(driver.providerValidatorSet()))
```
`curValSet` might be confused to be related with `currentModelState` due to `cur`?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+			intPow := power.Value.(int64)
+			_ = intPow
+		}
+		require.NoError(t, CompareValSet(modelValSet, consumerCurValSet), "Validator sets do not match for consumer %v", consumer)
+	}
+}
+
+// ComparePacketQueues compares the packet queues in the model to the packet queues in the system.
+// It compares both incoming (provider->consumer) and outgoing (consumer->provider) packets.
+// It only takes the number of packets into account, not the contents.
```
insumity, line 430: "not the contents" => it seems to me from `ComparePacketQueue` that also the timeout timestamps inside the packets are taken into account. Is that not the case?

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,577 @@
...
+			sender,
+			receiver)
+	}
+}
+
+// CompareTimes compares the block times in the model to the block times in the system.
+// It takes into account the timeOffsets, which should be the starting times
+// of the chains in the system after the system has been initialized.
+// We only compare down to seconds, because the model and system will differ
+// on the order of nanoseconds.
```
insumity, line 483: I guess rare but still possible that one time would be `second X nanonseconds 999999999 ` and the other `second X + 1 nanoseconds 0`. In such a case the comparison would fail, right? Would it make sense to compare their absolute value instead and verify that it's `<= 100ms` or something in this vein?

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,414 @@
...
+// undelegate undelegates amt tokens from validator val
+func (s *Driver) undelegate(val, amt int64) {
+	providerStaking := s.providerStakingKeeper()
+	server := stakingkeeper.NewMsgServerImpl(&providerStaking)
+	coin := sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(amt))
+	d := s.delegator()
+	v := s.validator(val)
+	msg := stakingtypes.NewMsgUndelegate(d, v, coin)
+	server.Undelegate(sdk.WrapSDKContext(s.ctx(P)), msg)
+	providerStaking.GetAllDelegations(s.ctx(P))
```
insumity, line 164: Is the `providerStaking.GetAllDelegations(s.ctx(P))` line nedeed?

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// packetQueue returns the queued packets from sender to receiver,
+// where either sender or receiver must be the provider.
+func (s *Driver) packetQueue(sender, receiver ChainId) []simibc.Packet {
+	var path *simibc.RelayedPath
+	if sender == P {
+		path = s.path(receiver)
+	} else {
```
insumity, line 173: Do we want to have an assert here that `receiver` is the provider in case we queue packets erroneously?

tests/mbt/driver/model_viewer.go
```diff
@@ -0,0 +1,103 @@
+package main
+
+import (
+	"github.com/informalsystems/itf-go/itf"
+)
+
+// This file contains logic to process
+// and access parts of the current state of the Quint trace.
+
+const Provider = "provider"
```
insumity, line 10: nit: Maybe make the same as the one in `common.go`:
```
const (
	P = "provider"
)
```
to be consistent?

insumity (approved):
Amazing work! Thanks Philip

I looked at everything but focused my review on the Quint model and `mbt_test.go`. It would beneficial to have someone look more at the `driver` components.

tests/mbt/driver/mbt_test.go
p-offtermatt: Woops! Yup, that indeed does not do anything anymore

tests/mbt/driver/mbt_test.go
p-offtermatt: Added a comment

tests/mbt/driver/mbt_test.go
p-offtermatt: I think I like it better as-is - I want to avoid the impression that it's just that they differ on the order of hundreds of milliseconds.
I need to increment timestamps between blocks by nanoseconds more than the model, but only in some specific cases. Rounding down is always more "correct" than checking some arbitrary difference, because it will *always* be that the model has time `second X` and the actual system has time `second X nanoseconds Y` until it spills over into seconds.

Hope that makes some amount of sense? I wrote a bit of the explanation into the comment

insumity: I see. Sounds good. Thanks for the explanation. 😃

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,417 @@
...
+func (s *Driver) consumerPower(i int64, chain ChainId) (int64, error) {
+	v, found := s.consumerKeeper(chain).GetCCValidator(s.ctx(chain), s.validator(i))
+	if !found {
+		return 0, fmt.Errorf("validator %v not found", s.validator(i))
+	}
+	return v.Power, nil
+}
+
+// consumerTokens returns the number of tokens that the validator with
+// id (ix) i has delegated to it in total on the provider chain
```
MSalopek, line 127: Nit: this docstring reads a bit weird.

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,417 @@
...
+			path.Outboxes.Commit(string(chain))
+		}
+	}
+
+	simibc.BeginBlock(testChain, timeAdvancement)
+	return header
+}
+
+func (s *Driver) runningConsumers() []providertypes.Chain {
+	// consumers that are still consumers on the provider
```
MSalopek, line 348: Nit: not sure docstring is necessary here

tests/mbt/driver/README.md
```diff
@@ -0,0 +1,134 @@
...
+### Generating traces
+
+To generate traces, run
+```
+./generate_traces.sh
+```
+
+This will generate several "families" of traces in the `traces` folder.
+
+Here is what one of the lines from `./generate traces` looks like:
```
sainoe, line 18:
```suggestion
Here is what one of the lines from `./generate_traces.sh` looks like:
```

tests/mbt/run_invariants.sh
```diff
@@ -1,4 +1,5 @@
 #! /bin/sh
```
sainoe, line 1:
```suggestion
#!/bin/bash
```

tests/mbt/driver/generate_traces.sh
```diff
@@ -0,0 +1,10 @@
+echo "Generating bounded drift traces with timeouts"
```
sainoe, line 1:
```suggestion
#!/bin/bash

echo "Generating bounded drift traces with timeouts"
```

tests/mbt/driver/core.go
```diff
@@ -0,0 +1,417 @@
...
+			log.Fatalf("either receiver '%v' or sender '%v' should be provider '%v', but neither is", sender, receiver, PROVIDER)
+		}
+		path = s.path(sender)
+	}
+	outboxes := path.Outboxes
+	outboxPackets := outboxes.OutboxPackets
+	res, ok := outboxPackets[string(sender)]
+	if !ok {
+		return []simibc.Packet{}
+	} else {
```
sainoe, line 184: nit: unnecessary `else` condition.

tests/mbt/driver/model_viewer.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func GetSendingTimeForPacket(packetExpr itf.MapExprType) int64 {
+	return packetExpr["sendingTime"].Value.(int64)
+}
+
+func VscSendTimestamps(curStateExpr itf.MapExprType, consumer string) []int64 {
+	sentVscPackets := ProviderState(curStateExpr)["sentVscPacketsToConsumer"].Value.(itf.MapExprType)[consumer].Value.(itf.ListExprType)
+
+	res := make([]int64, 0)
```
sainoe, line 88: nit:

```suggestion
	res := make([]int64, len(sentVscPackets))
```

testutil/simibc/relayed_path.go
```diff
@@ -30,39 +31,94 @@ type RelayedPath struct {
...
+	return &RelayedPath{
 		t:             t,
 		clientHeaders: map[string][]*ibctmtypes.Header{},
-		path:          path,
+		Path:          path,
 		Outboxes:      MakeOrderedOutbox(),
 	}
 }
 
+// PacketBelongs returns true if the packet belongs to this relayed path.
```
sainoe, line 44:
```suggestion
// PacketSentByChain returns true if the packet belongs to this relayed path.
```

tests/mbt/driver/mbt_test.go
```diff
@@ -0,0 +1,587 @@
...
+			require.True(t, err == nil, "Error updating client from %v on provider: %v", consumerChain, err)
+
+		case "DeliverVscPacket":
+			consumerChain := lastAction["consumerChain"].Value.(string)
+			t.Log("DeliverVscPacket", consumerChain)
+
+			// delivering the packet should give an error
+			// if the consumer is timed out in the model
+			var expectError bool
+			if ConsumerStatus(currentModelState, consumerChain) == "timedout" {
```
sainoe, line 302: Perhaps it'd be helpful to declare the consumer status as constant.

sainoe (approved):
Thank you for this big work @p-offtermatt. My review of the Quint code doesn't carry much weight as I'm not yet very familiar with the model yet. Apart from tiny nits, the driver code LGTM and is easier to read than the diff testing!

PR #1381, opened 2023-10-31
chore: update rosetta import

boojamya:
Updating this import resolves ambiguous import errors we were seeing while trying to add ics support to [interchaintest v7](https://github.com/strangelove-ventures/interchaintest/tree/v7)

To be specific, in interchaintest I'm importing `v3.1.0` not sure of your backport policy or if I should redirect to release branch.


p-offtermatt:
Thanks for the contribution!
Out of curiosity, do you have any more details about the errors you are running into that this fixes?
Would like to get the error message in here so it comes up if someone in the future searches for a similar error.

I'm not sure we need to backport this to the 3.2.x-consumer release, too, but afaict doesn't hurt if we include it when we cut a new release and could avoid similar problems in the future.

boojamya:
Thanks @p-offtermatt!
Sure, to be more specific...
 I'm seeing this error:
```
error while importing cosmossdk.io/tools/rosetta/lib/errors: ambiguous import: found package github.com/coinbase/rosetta-sdk-go/types in multiple modules:
	github.com/coinbase/rosetta-sdk-go v0.7.9 (...go/pkg/mod/github.com/coinbase/rosetta-sdk-go@v0.7.9/types)
	github.com/coinbase/rosetta-sdk-go/types v1.0.0 (...go/pkg/mod/github.com/coinbase/rosetta-sdk-go/types@v1.0.0)
```
The cosmos sdk is also importing this, but a different version of this. So this PR should align the versions.

And just to be clear, all I did was remove that import from the mod file, and then re-run mod tidy and it correctly updated the import.

Also, now that I think about it, backporting probably won't work because of the need to run a go mod tidy. Will follow up with specific PR's.

p-offtermatt (approved):
LGTM, thanks for the change and the comment!
Confirmed from my end that I end up with the same changes when I remove the entry from go.mod and run go mod tidy

I merged the backport PRs in the release branches already

MSalopek (approved):
Thanks for the contribution!

Approve.

Issue #1387, opened 2023-11-02
Drop Amnesia attacks


sainoe:
## Problem

A [security issue](https://docs.google.com/document/d/1lfo1rafDhazcHZweT1MSmG0s42g2alMt2UnDze1oONA/edit?pli=1#heading=h.pbd62d7k2d8a) was found in the last version of the [cryptographic equivocation feature of ICS](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-005-cryptographic-equivocation-verification.md). Simply put, the handling of light client amnesia attacks could potentially lead to the slashing of correct validators on the hub.

## Closing criteria

Drop consumer misbehaviours that correspond to an amnesia attack.

Credits to @insumity for finding the issue.

PR #1388, opened 2023-11-02
fix: drop amnesia attacks in consumer misbehaviour handling

sainoe:
## Description

Closes: #1387 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/ccv/provider/keeper/misbehaviour.go
+++ b/x/ccv/provider/keeper/misbehaviour.go
@@ -75,18 +75,26 @@ func (k Keeper) HandleConsumerMisbehaviour(ctx sdk.Context, misbehaviour ibctmty
 // GetByzantineValidators returns the validators that signed both headers.
 // If the misbehavior is an equivocation light client attack, then these
 // validators are the Byzantine validators.
-func (k Keeper) GetByzantineValidators(ctx sdk.Context, misbehaviour ibctmtypes.Misbehaviour) ([]*tmtypes.Validator, error) {
+func (k Keeper) GetByzantineValidators(ctx sdk.Context, misbehaviour ibctmtypes.Misbehaviour) (validators []*tmtypes.Validator, err error) {
 	// construct the trusted and conflicted light blocks
 	lightBlock1, err := headerToLightBlock(*misbehaviour.Header1)
 	if err != nil {
-		return nil, err
+		return
 	}
 	lightBlock2, err := headerToLightBlock(*misbehaviour.Header2)
 	if err != nil {
-		return nil, err
+		return
 	}
 
-	var validators []*tmtypes.Validator
+	// Check if the misbehaviour corresponds to an Amnesia attack,
+	// meaning that the conflicting headers have both valid state transitions
+	// and different commit rounds. In this case, we return no validators as we can't identify the byzantine validators.
+	//
+	// Note that we cannot differentiate which of the headers is trusted or malicious.
+	ev := &tmtypes.LightClientAttackEvidence{ConflictingBlock: lightBlock1}
+	if !ev.ConflictingHeaderIsInvalid(lightBlock2.Header) && lightBlock1.Commit.Round != lightBlock2.Commit.Round {
+		return
+	}
 
 	// compare the signatures of the headers
 	// and return the intersection of validators who signed both
--- a/x/ccv/provider/keeper/misbehaviour.go
+++ b/x/ccv/provider/keeper/misbehaviour.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/cosmos/interchain-security/v2/x/ccv/provider/types"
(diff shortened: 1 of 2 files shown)


x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -160,3 +169,14 @@ func (k Keeper) CheckMisbehaviour(ctx sdk.Context, misbehaviour ibctmtypes.Misbe
 
 	return nil
 }
+
+// Check if the given block headers have conflicting state transitions.
+// Note that this method was copied from ConflictingHeaderIsInvalid in CometBFT,
+// see https://github.com/cometbft/cometbft/blob/v0.34.27/types/evidence.go#L285
+func headersStateTransitionsAreConflicting(h1, h2 tmtypes.Header) bool {
```
mpoke, line 176: Why not just use the CometBFT method?

mpoke (approved):
LGTM

x/ccv/provider/keeper/misbehaviour.go
```diff
@@ -75,18 +76,26 @@ func (k Keeper) HandleConsumerMisbehaviour(ctx sdk.Context, misbehaviour ibctmty
...
 	}
 
-	var validators []*tmtypes.Validator
+	// Check if the misbehaviour corresponds to an Amnesia attack,
+	// meaning that the conflicting headers have both valid state transitions
+	// and different commit rounds. In this case, we return no validators as
+	// we can't identify the byzantine validators.
+	//
+	// Note that we cannot differentiate which of the headers is trusted or malicious,
+	if !headersStateTransitionsAreConflicting(*lightBlock1.Header, *lightBlock2.Header) && lightBlock1.Commit.Round != lightBlock2.Commit.Round {
```
insumity, line 96: It's not clear to me why those header fields (i.e., `ValidatorsHash`, `NextValidatorsHash` etc.) must be the same?
I'm not sure what I'm missing but my understanding is that for this to be misbehaviour the `blockID`s have to be different (e.g., see [here](https://github.com/cosmos/ibc-go/blob/v7.3.0/modules/light-clients/07-tendermint/misbehaviour_handle.go#L69)) and `blockID` is a hash that incorporates among others the above fields (?)

sainoe: These fields must be the same because, [in all amnesia scenarios](https://github.com/cometbft/cometbft/blob/main/spec/light-client/accountability/README.md#flip-flopping-amnesia-based-attacks), it's assumed that correct validators validate both headers.

The BlockID hash contains other fields that are only related to the current block e.g. the time or the proposer.

x/ccv/provider/keeper/misbehaviour.go
sainoe: Bc it'd require creating a `LightClientAttackEvidence` type struct, which can be confusing imo.

x/ccv/provider/keeper/misbehaviour.go
insumity: > These fields must be the same because, [in all amnesia scenarios](https://github.com/cometbft/cometbft/blob/main/spec/light-client/accountability/README.md#flip-flopping-amnesia-based-attacks), it's assumed that correct validators validate both headers.

I'm not sure I fully understand this. Indeed, validators would need to validate both headers but why do we need to verify anything about `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`, `AppHash`,  `LastResultHash` here?

I believe that the signed headers would need to be simply verified (see [here](https://github.com/cometbft/cometbft/blob/bd3acaa455751eb7029380fd1e764747e860dc29/types/light.go#L157)). This verification checks that the hash of the header (see [here](https://github.com/cometbft/cometbft/blob/d4708b4d7f697e629818d42a2ac1a9c43929ea85/types/block.go#L458) -- that includes `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`, `AppHash`,  `LastResultHash`, etc.) is the same as the one in the `commit.BlockID.Hash`.
So with the check it seems we're checking that we got two headers that have a different `BlockID`s but have the exact same `ValidatorsHash`, `NextValidatorsHash`, `ConsensusHash`, `AppHash` and  `LastResultHash` and that is indeed possible (e.g., the headers have different `ProposerAddress` as you mention).

By definition, in an amnesia attack, we would have two headers with different `BlockID`s. From my point of view, for an amnesia attack to make sense, someone would try to introduce a bogus `AppHash` or `NextValidatorsHash` (see [here](https://github.com/cosmos/ibc-go/blob/fec07ec55ca3e35f49a62bc6b1243c27cdf7799d/modules/light-clients/07-tendermint/header.go#L19)) to have an impact on what would be stored in the light client's `ConsensusState`s.
With the current check, I think we could have an amnesia attack with two headers that only have a different `AppHash`. Because of this `!bytes.Equal(h1.AppHash, h2.AppHash)` would return `true`, hence `headersStateTransitionsAreConflicting` would return `true` and hence `if !headersStateTransitionsAreConflicting(*lightBlock1.Header, *lightBlock2.Header) && lightBlock1.Commit.Round != lightBlock2.Commit.Round {` would fail. Therefore, we would not `return` at this stage and we would try to find the byzantine validators, even in the case of an amnesia attack. Isn't that possible?

insumity: Deleted my previous comment. After the offline discussion we had yesterday this makes sense and the above code checks what CometBFT already checks in [_GetByzantineValidators_](https://github.com/cometbft/cometbft/blob/5989a7314fea2b47efcd8747e9f9c9b6ebd1f6e0/types/evidence.go#L233).
Thanks @sainoe !

Just for posterity: If we have 2 headers at the same height `H` with `h1.AppHash != h2.AppHash` then this is a _lunatic attack_ because `AppHash` is computed  as the "[state after txs from the previous block](https://github.com/cometbft/cometbft/blob/d12edd17c5c3c7f86e72e6d80e98f7c79f56c5da/types/block.go#L342)" `H - 1`. Only a "lunatic" validator would have 2 different views of the `AppHash` when applying the transactions of a previous block. The same is the case for all the fields that are based on the previous block which are the ones checked against in `headersStateTransitionsAreConflicting`. In case of a lunatic attack we can still slash.

Issue #1133, opened 2023-07-12
update e2e initialization


MSalopek:
Simplify e2e framework initialization procedure to make it more accessible to:
* developers
* CI/CD integrators
* testers

More context is in https://github.com/cosmos/interchain-security/issues/877

PR #1486, opened 2023-12-05
test: e2e initialization

bermuell:
## Description

Closes: #1133 

Infra work on e2e tests to support testruns against different versions of provider and consumer
Proposed solution allows to define provider version and consumer version to be  build and runs local test cases
on the constructed target.
Note: dependeing on the selected versions the tests might fail due to required changes as transformations and not supported checks. A separate PR will be opened targeting compatiblity tests. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


Makefile
```diff
@@ -57,6 +57,11 @@ test-e2e-multi-consumer:
 test-e2e-parallel:
 	go run ./tests/e2e/... --include-multi-consumer --parallel
 
+# run E2E compatibility tests for v3.x
+# Note: e54799020 is the rev for v3.1.x candidate as build is broken for  v3.1.0
+test-e2e-compatibility-v3:
+	go run ./tests/e2e/... -parallel -tc compatibility -pv v3.2.0-rc1 -pv e54799020 -pv v3.0.0 -cv v3.2.0-rc1 -cv e54799020 -cv v3.0.0
```
bermuell, line 63: unfortunately this stopped working since rebase on latest main:
`Error: unknown command "list-proposed-consumer-chains" for "provider"`

tests/e2e/actions.go
```diff
@@ -480,30 +484,79 @@ type StartConsumerChainAction struct {
...
 
-	consumerGenesis := ".app_state.ccvconsumer = " + string(bz)
+	// only needed when consumer is running v3.3.x and later
+	if tr.transformGenesis {
+		return string(tr.transformConsumerGenesis(consumerChain, bz, target))
+	}
+	return string(bz)
+}
+
+func (tr *TestConfig) startConsumerChain(
```
bermuell, line 553: move up below type StartConsumerChain

tests/e2e/compatibility_tests.go
```diff
@@ -0,0 +1,20 @@
+package main
```
bermuell, line 1: remove this: compatibility tests will go to separate PR

tests/e2e/main.go
```diff
@@ -48,6 +59,14 @@ var (
 	gaiaTag = flag.String("gaia-tag", "", "gaia tag to use - default is latest")
 )
 
+var (
+	//useConsumerVersion = flag.String("consumer-version", "", "ICS tag to specify the consumer version to test the provider against")
+	//useProviderVersion = flag.String("provider-version", "", "ICS tag to specify the provider version to test the consumer against")
```
bermuell, line 64: remove

tests/e2e/actions.go
```diff
@@ -1770,13 +1809,15 @@ type DoublesignSlashAction struct {
...
 	action DoublesignSlashAction,
+	target ExecutionTarget,
 	verbose bool,
 ) {
 	if !tr.useCometmock {
 		chainConfig := tr.chainConfigs[action.Chain]
-		//#nosec G204 -- Bypass linter warning for spawning subprocess with cmd arguments.
-		bz, err := exec.Command("docker", "exec", tr.containerConfig.InstanceName, "/bin/bash",
-			"/testnet-scripts/cause-doublesign.sh", chainConfig.BinaryName, string(action.Validator),
+		isConsumer := false
```
p-offtermatt, line 1817: inline this? it's used only once, right below

p-offtermatt: unless you like it better like this for clarity, I can see an argument for that, too

tests/e2e/builder.go
```diff
@@ -0,0 +1,121 @@
...
+	"bytes"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// setupWorkSpace checks out given revision in a tmp directory
+// and returns path where the workspace is located
```
p-offtermatt, line 13:
```suggestion
// and returns the path where the workspace is located
```

tests/e2e/builder.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	fmt.Printf("Setting up worktree in '%s'\n", workSpace)
+
+	cmd := exec.Command("git", "worktree", "add",
+		"--checkout", workSpace, revision)
+	var errbuf bytes.Buffer
+	cmd.Stderr = &errbuf
+	fmt.Printf("Running: %s", cmd.String())
+	if err := cmd.Start(); err != nil {
```
p-offtermatt, line 27: nit: this is just doing Start and then Wait, can probably be combined into one call, right? cmd.CombinedOutput, maybe?

tests/e2e/main.go
```diff
@@ -48,6 +59,12 @@ var (
 	gaiaTag = flag.String("gaia-tag", "", "gaia tag to use - default is latest")
 )
 
+var (
+	consumerVersions VersionSet
+	providerVersions VersionSet
+	transformGenesis = flag.Bool("transform-genesis", false, "do consumer genesis transformation for newer clients. Needed when provider chain is on an older version")
```
p-offtermatt, line 65: I think this help string needs a bit of attention. The message is not clear to me - what should I put when checking multiple provider versions, some older and some newer? is any version != latest an older version?

tests/e2e/main.go
```diff
@@ -208,6 +200,10 @@ func parseArguments() (err error) {
 
 	flag.Var(&selectedTestfiles, "test-file",
 		getTestFileUsageString())
+
+	flag.Var(&consumerVersions, "cv", "Consumer version")
```
p-offtermatt, line 204: could you document the default behaviour for these and make the help strings more descriptive?

tests/e2e/main.go
```diff
@@ -287,226 +283,121 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet) (tests []t
...
+	for _, target := range targets {
+		for _, tc := range testCases {
+			tr := TestRunner{
+				config:  tc.testRun,
+				steps:   tc.steps,
+				target:  target,
+				verbose: *verbose,
 			}
-			gaiaTag = tr.gaiaTag
+			//TODO: refactor this target specific setting
```
p-offtermatt, line 333: this TODO will be addressed in a different PR, right? we can leave it, I'm just a bit concerned about it never getting addressed

tests/e2e/main.go
```diff
@@ -287,226 +283,121 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet) (tests []t
...
 }
 
-// remove docker container to reduce resource usage
-// otherwise the chain will keep running in the background
-func (tr *TestConfig) teardownDocker() {
-	fmt.Printf("===============  tearing down %s testRun ===============\n", tr.name)
-	//#nosec G204 -- Bypass linter warning for spawning subprocess with cmd arguments.
-	cmd := exec.Command("docker", "kill", tr.containerConfig.InstanceName)
+// runs E2E tests
+// all docker containers are built sequentially to avoid race conditions when using local cosmos-sdk
```
p-offtermatt, line 373: Not sure I get this - would parallel builds really have a race condition? Could you explain a bit more why this is the case?

tests/e2e/test_driver.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"fmt"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/kylelemons/godebug/pretty"
+)
+
+// TestCaseDriver knows how different TC can be executed
```
p-offtermatt, line 12:
```suggestion
// TestCaseDriver knows how different testcases can be executed
```
no need for acronym here imo

tests/e2e/test_target.go
```diff
@@ -0,0 +1,221 @@
...
+	"os/exec"
+	"strings"
+)
+
+type ExecutionTarget interface {
+	GetTargetType() string
+	GetLogs(path string) []byte
+	GetTestScriptPath(isConsumer bool, script string) string
+	ExecCommand(name string, arg ...string) *exec.Cmd
+	ExecDetachedCommand(name string, args ...string) *exec.Cmd
```
p-offtermatt, line 17: I understand, and maybe it's obvious, but could you comment here a sentence about what you expect from ExecCommand (return only after the command has been executed to completion) and ExecDetachedCommand (start the command, then return immediately)?

p-offtermatt (commented):
Nice work, I really like particularly the way the test target abstracts away so much of the underlying logic.

Some comments, none too major, mostly about adding a bit of documentation.
Could you also please update https://github.com/cosmos/interchain-security/blob/bernd/1133-e2e-compatiblity-tests/tests/e2e/README.md#L14 to match the changed setup? Nothing major, but at least adjust file names

tests/e2e/main.go
bermuell: agree as this a complicated subject. referencing to the release notes to clarify when this is needed.
pls note that this is subject to be changed by [#1559](https://github.com/cosmos/interchain-security/issues/1559) which should facilitate this step

p-offtermatt (approved):

p-offtermatt:
Hey, there's one more Gosec error, could you please fix it? :) https://github.com/cosmos/interchain-security/actions/runs/7489943947/job/20387679246?pr=1486#step:4:536

bermuell:
> Hey, there's one more Gosec error, could you please fix it? :) https://github.com/cosmos/interchain-security/actions/runs/7489943947/job/20387679246?pr=1486#step:4:536
yes, just pushed the fix + improvement on worktree creation (in case the revision is already checked out) and docker image creation to deal with issues realted to caching.

tests/e2e/main.go
```diff
@@ -33,6 +27,23 @@ func (t *TestSet) String() string {
 	return fmt.Sprint(*t)
 }
 
+type VersionSet []string
+
+func (c *VersionSet) Set(value string) error {
```
sainoe, line 32: It seems that this function never returns an error. Also, a trick to create sets is to use `map[string]struct{}{}`. It allows to filter out duplicates in 0(1). [Here](https://github.com/sainoe/interchain-security/blob/087a9fe65be483e08447768854c7cc71c8629b4e/tests/e2e/main.go#L190) is an example in ICS.

tests/e2e/main.go
```diff
@@ -33,6 +27,23 @@ func (t *TestSet) String() string {
...
+	for _, v := range *c {
+		if v == value {
+			return nil
+		}
+	}
+	*c = append(*c, value)
+	return nil
+}
+
+func (t *VersionSet) String() string {
```
sainoe, line 43: nit: perhaps `vs` would be easier to understand than `t` as a short name for `VersionSet`.

tests/e2e/main.go
```diff
@@ -287,226 +283,132 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet) (tests []t
...
-// after building docker containers, all tests are run in parallel using their respective docker containers
-func main() {
-	if err := parseArguments(); err != nil {
-		flag.Usage()
-		log.Fatalf("Error parsing command arguments %s\n", err)
-	}
-
-	testCases := getTestCases(selectedTests, selectedTestfiles)
+func deleteTargets(targets []ExecutionTarget) error {
+	var rc error = nil
```
sainoe, line 287: Nit: can be simplified to:
```suggestion

func deleteTargets(targets []ExecutionTarget) (rc error) {
```

tests/e2e/main.go
```diff
@@ -287,226 +283,132 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet) (tests []t
...
+	for _, target := range targets {
+		if err := target.Delete(); err != nil {
+			rc = err
```
sainoe, line 290: Is the expected behaviour to only log the errors?
This function only returns an error when the last execution target deletion fails.

tests/e2e/main.go
```diff
@@ -287,226 +283,132 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet) (tests []t
...
+			wg.Add(1)
+			go func(runner TestRunner) {
+				defer wg.Done()
+				result := runner.Run()
+				if result != nil {
+					log.Printf("Test '%s' failed", runner.config.name)
+				}
+				if err == nil {
+					err = result
+				}
```
sainoe, line 368: This section is a bit confusing. What should be the error returned?
It seems that once `result != nil`, `err` will never be updated and returned.

tests/e2e/test_driver.go
```diff
@@ -0,0 +1,150 @@
...
+		fmt.Printf("=============== %s FAILED ===============\n", pr.testCfg.name)
+		fmt.Println("FAILED action", reflect.TypeOf(step.Action).Name())
+		pretty.Print("actual state", actualState)
+		pretty.Print("model state", modelState)
+		log.Fatal(`actual state (-) not equal to model state (+): ` + pretty.Compare(actualState, modelState))
+	}
+	return nil
+}
+
+func (td *DefaultDriver) runAction(action interface{}) error {
```
sainoe, line 68: nit:

```suggestion
func (pr *DefaultDriver) runAction(action interface{}) error {
```

tests/e2e/test_target.go
```diff
@@ -0,0 +1,231 @@
...
+	images       []string //images needed to build the target container
+	ImageName    string
+}
+
+func (dc *DockerContainer) GetTargetType() string {
+	return "docker"
+}
+
+func (dc *DockerContainer) GetLogs(path string) []byte {
+	logs := []byte{}
```
sainoe, line 38: Is it expected that the `path` arg isn't used?

tests/e2e/main.go
sainoe, line 287: Nit: can be simplified to:
```suggestion
func deleteTargets(targets []ExecutionTarget) (rc error) {
```

MSalopek (approved):
LGTM!

I would like to have an option to use pre-built docker images that can just be pulled from a registry. That way we could avoid extraneous build steps an using git worktree.

The released versions don't change so it's safe to push docker images (eg v3.3.0 will never change, all changes will exist in subsequent versions such as v3.3.1, v3.4.0, v4.x.y etc)

Issue #1492, opened 2023-12-08
Bypass Interchain Security Disabled Modules Messages


mpoke:
## Summary of Bug

Consumer chains disable messages for both the Slashing and Evidence modules, i.e., https://github.com/cosmos/interchain-security/blob/a9b090f98c212018e063f707e300a912401e6e05/app/consumer/ante_handler.go#L46
These mechanism can be bypass by submitting messages wrapped within an `authz` `MsgExec` transaction. 

This is also the case for [Stride](https://github.com/Stride-Labs/stride/blob/32fd245d898acc9a1aa7604b942e3ca7d481b56d/app/ante_handler.go#L44) and [Neutron](https://github.com/neutron-org/neutron/blob/168815d3703a803f97566f2ddaf0954ac8bda203/app/ante_handler.go#L64). 

## Version

https://github.com/cosmos/interchain-security/commit/a9b090f98c212018e063f707e300a912401e6e05

## Steps to Reproduce

1. Wrap a /cosmos.evidence or /cosmos.slashing transaction within a authz MsgExec transaction
2. Broadcast this transaction to the chain
3. bypass the blocked messages

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

sontrinh16:
hi @mpoke, correct me if i'm wrong but we can check the for authz message in the consumer ante handler and check the msgs array in MsgExec for any message type that corresponded to the disable message types. What do you think

sontrinh16:
spin up some rough idea PR, hope this be any help

PR #1500, opened 2023-12-12
fix: Prevent Authz MsgExec bypass Interchain Security Disabled Modules Messages

sontrinh16:
## Description

Closes: #1492 

---

## Changes
- Add `nestedAuthzMsgExecCheck` func for filtering `/cosmos.authz.v1beta1.MsgExec` msgs and check all the wrapped msgs with `hasDisabledModuleMsgs` func. If there is a MsgExec among the wrapped msgs, repeat the process again until all the wrapped msgs is check or when found a disabled module msgs.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

* [x] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [x] confirmed `!` in the type prefix if API or client breaking change

### DIFF
--- a/app/consumer/ante/disabled_modules_ante.go
+++ b/app/consumer/ante/disabled_modules_ante.go
@@ -25,7 +25,7 @@ func (dmd DisabledModulesDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
 
 		// Check if there is an atempt to bypass disabled module msg
 		// with authz MsgExec
-		if nestedCheck(msg, dmd.prefixes...) {
+		if nestedAuthzMsgExecCheck(msg, dmd.prefixes...) {
 			return ctx, fmt.Errorf("tx contains message types from unsupported modules at height %d", currHeight)
 		}
 	}
@@ -33,7 +33,7 @@ func (dmd DisabledModulesDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
 	return next(ctx, tx, simulate)
 }
 
-func nestedCheck(msg sdk.Msg, prefixes ...string) bool {
+func nestedAuthzMsgExecCheck(msg sdk.Msg, prefixes ...string) bool {
 	check := false
 	msgExec, ok := msg.(*authz.MsgExec)
 
@@ -54,7 +54,7 @@ func nestedCheck(msg sdk.Msg, prefixes ...string) bool {
 
 		// Check for nested authz msgExec
 		if hasAuthzMsgExec(msg) {
-			check = nestedCheck(msg, prefixes...)
+			check = nestedAuthzMsgExecCheck(msg, prefixes...)
 			if check {
 				break
 			}
--- a/app/consumer/ante/disabled_modules_ante.go
+++ b/app/consumer/ante/disabled_modules_ante.go
@@ -34,34 +34,32 @@ func (dmd DisabledModulesDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
 }
 
 func nestedAuthzMsgExecCheck(msg sdk.Msg, prefixes ...string) bool {
-	check := false
 	msgExec, ok := msg.(*authz.MsgExec)
 
 	if !ok {
-		return check
+		return false
 	}
 
 	sdkMsgs, err := msgExec.GetMessages()
 	if err != nil {
-		return check
+		return false
 	}
 
 	for _, msg := range sdkMsgs {
 		if hasDisabledModuleMsgs(msg, prefixes...) {
-			check = true
-			break
+			return true
 		}
 
 		// Check for nested authz msgExec
-		if hasAuthzMsgExec(msg) {
-			check = nestedAuthzMsgExecCheck(msg, prefixes...)
-			if check {
-				break
+		_, ok = msg.(*authz.MsgExec)
+		if ok {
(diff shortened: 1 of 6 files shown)


p-offtermatt:
I think this change requires a coordinated upgrade, right? Otherwise, some nodes may filter out the newly disabled types, while other nodes would still process them. Does that seem right to you?

sontrinh16:
> I think this change requires a coordinated upgrade, right? Otherwise, some nodes may filter out the newly disabled types, while other nodes would still process them. Does that seem right to you?

yesss i think so

app/consumer/ante_handler.go
```diff
@@ -42,7 +42,7 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 	anteDecorators := []sdk.AnteDecorator{
 		ante.NewSetUpContextDecorator(),
 		ante.NewExtensionOptionsDecorator(nil),
-		consumerante.NewMsgFilterDecorator(options.ConsumerKeeper),
+		consumerante.NewMsgFilterDecorator(options.ConsumerKeeper, "/cosmos.evidence", "/cosmos.slashing"),
```
mpoke, line 45: Adding the filtering logic in the `MsgFilterDecorator` would require all the consumer chains to update their application, i.e., adding this line to their `ante_handler.go`. An alternative is to add the logic directly to the `DisabledModulesDecorator`, which would require consumer chains to only upgrade their version of ICS.

app/consumer/ante/msg_filter_ante.go
```diff
@@ -54,3 +64,43 @@ func hasValidMsgsPreCCV(msgs []sdk.Msg) bool {
...
+
+			wrappedMsgs = append(wrappedMsgs, sdkMsgs...)
+
+		}
+	}
+
+	return wrappedMsgs
+}
+
+func hasDisabledModuleMsgs(msgs []sdk.Msg, prefixes ...string) bool {
```
mpoke, line 94: Moving this logic to the `DisabledModulesDecorator` would also avoid duplicating this code as it's already there.

app/consumer/ante_handler.go
sontrinh16: hmmm make sense, i will move the logic to `DisabledModulesDecorator`

sontrinh16: updated as request

mpoke (commented):
Thanks @sontrinh16. See my comments bellow (nits). My only concern is about nested authz exec messages. For example, the following message should be filtered out, but it is not:
```go
	authzMsgExecEvidence := authz.NewMsgExec(sdk.AccAddress{}, []sdk.Msg{&evidencetypes.MsgSubmitEvidence{}})
	nestedAuthzMsgExecSlashing := authz.NewMsgExec(
		sdk.AccAddress{},
		[]sdk.Msg{&authzMsgExecEvidence},
	)
```

app/consumer/ante/disabled_modules_ante.go
```diff
@@ -20,12 +21,52 @@ func (dmd DisabledModulesDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
...
+	}
+
+	sdkMsgs, err := msgExec.GetMessages()
+	if err != nil {
+		return []sdk.Msg{}
+	}
+
+	wrappedMsgs = append(wrappedMsgs, sdkMsgs...)
+
+	return wrappedMsgs
```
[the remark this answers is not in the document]
sontrinh16: oh yeah, missed this one, since last time this func loop through msg array

sontrinh16:
> Thanks @sontrinh16. See my comments bellow (nits). My only concern is about nested authz exec messages. For example, the following message should be filtered out, but it is not:
> 
> ```go
> 	authzMsgExecEvidence := authz.NewMsgExec(sdk.AccAddress{}, []sdk.Msg{&evidencetypes.MsgSubmitEvidence{}})
> 	nestedAuthzMsgExecSlashing := authz.NewMsgExec(
> 		sdk.AccAddress{},
> 		[]sdk.Msg{&authzMsgExecEvidence},
> 	)
> ```

hey i have added logic to cover the nested case

MSalopek:
> I think this change requires a coordinated upgrade, right? Otherwise, some nodes may filter out the newly disabled types, while other nodes would still process them. Does that seem right to you?

No, everything in `/app` is for our simapps/testing. We recommend that the consumer teams implement these methods following our examples. 

Only the modules in `x/*` are actually importable library code.
 
Do you feel that is being communicated clearly?

p-offtermatt:
> > I think this change requires a coordinated upgrade, right? Otherwise, some nodes may filter out the newly disabled types, while other nodes would still process them. Does that seem right to you?
> 
> No, everything in `/app` is for our simapps/testing. We recommend that the consumer teams implement these methods following our examples.
> 
> Only the modules in `x/*` are actually importable library code.
> 
> Do you feel that is being communicated clearly?

Makes sense to me - I didn't pay enough attention to where this file was, I think, but this is subtle, so I don't think we should worry too much about it in this particular case.

MSalopek (approved):
Thanks for taking this on!

LGTM

mpoke (approved):
LGTM. Thanks @sontrinh16

PR #1501, opened 2023-12-12
chore: Fix definition of breaking in production template

p-offtermatt:
## Description

Closes: N/A

The PR template is inconsistently using client breaking, state breaking, library breaking.
This unifies it with the updated definitions from the RELEASE document

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/production.md
+++ b/.github/PULL_REQUEST_TEMPLATE/production.md
@@ -19,7 +19,7 @@ please add links to any relevant follow up issues.*
 I have...
 
 * [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
-* [ ] Added `!` to the type prefix if state-machine breaking change (i.e., requires coordinated upgrade)
+* [ ] Added `!` to the type prefix if the change is library breaking, state-machine breaking, or node API breaking (see more info [here](https://github.com/cosmos/interchain-security/blob/main/RELEASES.md#breaking-changes
 * [ ] Confirmed this PR does not introduce changes requiring state migrations, OR migration code has been added to consumer and/or provider modules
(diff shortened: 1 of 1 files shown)


mpoke (changes requested):
I think this complicates the process. I believe it's better to keep it simple and add `!` only if the changes are state breaking. IMO, this is the most important aspect of blockchains as it requires a coordinated upgrade of the network. 

- library breaking should result in the bump of the major version and this should be properly communicated
- node API breaking will not change the version number, but it should be properly communicated as well

p-offtermatt:
My change doesn't change the semantics. Note that the original version calls for ! in all three cases that I outlined: 

```
Added `!` to the type prefix if state-machine breaking change
```
and
```
confirmed `!` in the type prefix if API or client breaking change
```
are both in the original.

.github/PULL_REQUEST_TEMPLATE/production.md
```diff
@@ -19,7 +19,7 @@ please add links to any relevant follow up issues.*
 I have...
 
 * [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
-* [ ] Added `!` to the type prefix if state-machine breaking change (i.e., requires coordinated upgrade)
+* [ ] Added `!` to the type prefix if the change is state-machine breaking and therefore requires a coordinated upgrade (see further [here](https://github.com/cosmos/interchain-security/blob/main/RELEASES.md#breaking-changes)
```
mpoke, line 22: Keep it concise
```suggestion
* [ ] Added `!` to the type prefix if the change is [state-machine breaking](https://github.com/cosmos/interchain-security/blob/main/RELEASES.md#breaking-changes)
```

mpoke (approved):

Issue #1512, opened 2023-12-14
Support CCV consumer genesis backward transformation


bermuell:
## Problem

Exported CCV consumer-data from a provider chain running v3.4.x can not be used to patch a genesis file of a consumer chain running on an older version e.g. 2.4.x. Existing transformation CLI only supports transformations only from old->new version.

## Closing criteria

CVV consumer data exported from a provider chain running v3.4.x can be patched to the genesis file of a consumer chain running older version of ICS e.g. 2.4.x

## Problem details

Refactoring of consumer genesis data brought the need to transform ccv consumer-data exported from provider chains running older versions.
A genesis transform cli was introduced to  do the transformation of old verions to latest of exported ccv consumer-data but there is no support to transform a new-version to the old format. This would be needed for provider chains running on latest (supporting already the new data format) to consumer chains running on an older version (supporting only the old format)

PR #1534, opened 2023-12-20
test: Add backtransform functionality to genesis transform cli

bermuell:
## Description

Closes: #1512

Changes will allow to transform exported consumer genesis data from version v2.x, v3.x, v4.x to a specified target version 
v2.x, v3.x, v4.x . With that newer version of provider chain exports can be transformed for older consumer chains e.g. 2.x

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/app/consumer/genesis.go
+++ b/app/consumer/genesis.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -146,7 +147,8 @@ func transformToV33(jsonRaw []byte, ctx client.Context) ([]byte, error) {
 
 // Transformation of consumer genesis content as it is exported from current provider version
 // to a format readable by consumer implementation of version v2.x
-func transformToV2(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) {
+// Use removePreHashKey to remove prehash_key_before_comparison from result.
+func transformToV2(jsonRaw []byte, ctx client.Context, removePreHashKey bool) (json.RawMessage, error) {
 
 	// populate deprecated fields of GenesisState used by version v2.x
 	srcConGen := consumerTypes.GenesisState{}
@@ -218,7 +220,7 @@ func transformToV2(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) 
 	}
 
 	// patch .provider_client_state from provider.client_state if needed
-	if srcConGen.Provider.ConsensusState != nil {
+	if srcConGen.Provider.ClientState != nil {
 		clientState, exists := providerMap["client_state"]
 		if !exists {
 			return nil, fmt.Errorf("'client_state' not found in provider data")
@@ -237,6 +239,12 @@ func transformToV2(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) 
 	if err != nil {
 		return nil, fmt.Errorf("marshalling transformation result failed: %v", err)
 	}
+
+	if removePreHashKey {
+		// remove all `prehash_key_before_comparison` entries not supported in v2.x (see ics23)
+		re := regexp.MustCompile(`,\s*"prehash_key_before_comparison"\s*:\s*(false|true)`)
+		result = re.ReplaceAll(result, []byte{})
+	}
 	return result, nil
 }
 
@@ -247,8 +255,12 @@ func transformGenesis(ctx client.Context, targetVersion IcsVersion, jsonRaw []by
 	var err error = nil
 
 	switch targetVersion {
-	case v2_x, v3_0_x, v3_1_x:
-		newConsumerGenesis, err = transformToV2(jsonRaw, ctx)
+	// v2.x, v3.0-v3.2 share same consumer genesis type
+	case v2_x:
+		newConsumerGenesis, err = transformToV2(jsonRaw, ctx, true)
+	case v3_0_x, v3_1_x, v3_2_x:
+		// same as v2 replacement without need of `prehash_key_before_comparison` removal
+		newConsumerGenesis, err = transformToV2(jsonRaw, ctx, false)
 	case v3_3_x:
 		newConsumerGenesis, err = transformToV33(jsonRaw, ctx)
 	case v4_x_x:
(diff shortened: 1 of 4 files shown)


MSalopek (approved):
LGTM.

We need docs on this. Consider adding them in this or a different PR.

app/consumer/genesis.go
```diff
@@ -74,19 +76,195 @@ func transform(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) {
...
+	result, err = json.Marshal(genState)
+	if err != nil {
+		return nil, fmt.Errorf("marshalling transformation result failed: %v", err)
+	}
+	return result, nil
+}
+
+// transformGenesis tries to transform provided data to ccv consumer genesis data
+// either to a format supported by current consumer impelementation or to the format
+// supported by consumer of version v2.x
```
p-offtermatt, line 223: This comment doesn't seem consistent with the actual code to me, could you dounle-check?
The code allows also transforming to v3.3.x and v4.x. My intuition is to just not list the possible versions in the docstring

app/consumer/genesis.go
```diff
@@ -74,19 +76,195 @@ func transform(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) {
...
+func transformGenesis(ctx client.Context, targetVersion string, jsonRaw []byte) (json.RawMessage, error) {
+	var newConsumerGenesis json.RawMessage = nil
+	var err error = nil
+
+	switch targetVersion {
+	case "v2.x":
```
p-offtermatt, line 229: should we use an enum here instead of plain strings?

app/consumer/genesis.go
```diff
@@ -114,17 +292,23 @@ func NewDefaultGenesisState(cdc codec.JSONCodec) GenesisState {
 // provider version v1,v2 or v3 to a JSON format supported by this consumer version.
 func GetConsumerGenesisTransformCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "transform [genesis-file]",
-		Short: "Transform CCV consumer genesis from an older provider version not supporting current format",
+		Use:   "transform [-to version] genesis-file",
+		Short: "Transform CCV consumer genesis data exported to a specific target format v2.x, v3.3.x, v4.x",
```
p-offtermatt, line 296: I would recommend populating this and the supported versions strings below automatically from a slice we define outside of here, so that it's impossible to change this and the strings below in an inconsistent way

app/consumer/genesis_test.go
```diff
@@ -170,52 +424,247 @@ func getGenesisTransformCmd() (*cobra.Command, error) {
...
+	filePath := createConsumerDataGenesisFile(t, version)
+	defer os.Remove(filePath)
+	resultGenesis := consumerTypes.GenesisState{}
+	result, err := transformConsumerGenesis(filePath, nil)
+	require.NoError(t, err)
+	err = ctx.Codec.UnmarshalJSON(result, &resultGenesis)
 	require.NoError(t, err)
+
+	// Some basic sanity checks on the content.
+	require.NotNil(t, resultGenesis.Provider.ClientState)
```
p-offtermatt, line 473: It feels a bit brittle to me, since it's very hard to see that this actually makes the v2 genesis into a genesis for the current version - is there a way you could parse the transformed genesis similar to how it actually would be parsed?

app/consumer/genesis_test.go
```diff
@@ -170,52 +424,247 @@ func getGenesisTransformCmd() (*cobra.Command, error) {
...
-	_, err = cmd.ExecuteC()
+
+	resultGenesis := consumerTypes.GenesisState{}
+	err = ctx.Codec.UnmarshalJSON(result, &resultGenesis)
 	require.NoError(t, err)
 
-	var oldConsumerGenesis map[string]interface{}
-	err = json.Unmarshal([]byte(consumerGenesisStates[version]), &oldConsumerGenesis)
-	require.NoError(t, err, "Error parsing old version of ccv genesis content for consumer")
+	require.Equal(t, srcGenesis.Params, resultGenesis.Params)
```
p-offtermatt, line 527: Similar here - I think here it's probably harder and might not be worth it.
Technically, I think we could provide old binaries inside the repo and use those to parse the genesis to check that it actually works with the old version, but probably it's too much effort and what you are doing is fine

p-offtermatt (approved):
LGTM, added some comments.

I think this is already in the docs, right?
https://cosmos.github.io/interchain-security/consumer-development/consumer-genesis-transformation
Could you double-check that everything there is still consistent with what this PR adapts?

app/consumer/genesis_test.go
bermuell: agree that this is hard to figure out as one would expect v2 types being use here as target version.
pls note that this is due to the problem not being able to import the v2 types due to module conflicts... which would be needed to properly test this.
also for encoding reasons I can't just create custom generic target structs faking v2 - which would also be brittle/tradeoff

app/consumer/genesis_test.go
bermuell: same as above... tried to do so but failed

app/consumer/genesis.go
bermuell: updated

app/consumer/genesis_test.go
p-offtermatt: Ok, I expected that and I think the solution you came up with is exactly a good tradeoff :)

p-offtermatt:
Thanks, changes look good to me!

app/consumer/genesis.go
```diff
@@ -74,19 +99,205 @@ func transform(jsonRaw []byte, ctx client.Context) (json.RawMessage, error) {
...
+	// delete .provider entry (introduced in v3.3.x)
+	delete(genState, "provider")
+
+	// Marshall final result
+	result, err = json.Marshal(genState)
+	if err != nil {
+		return nil, fmt.Errorf("marshalling transformation result failed: %v", err)
+	}
+
+	// remove all prehash_key_before_comparison entries not supported by ics23 versions used by v2.x
```
bermuell, line 242: going the shortcut here bypassing json result manipulation as element is deeply nested in consumer genesis type.

Issue #1517, opened 2023-12-15
Soft opt out implementation may lead to incorrect jailing 


mpoke:
## Summary of Bug

Due to the soft opt out feature, validators in the bottom `SoftOptOutThreshold` (default: 5%) can opt out of validating consumer chains without getting jailed. However, if one of these validators gets into the top e.g., 95%, then it get's jailed immediately. This is because in the [downtime logic](https://github.com/cosmos/cosmos-sdk/blob/03d578b9de5136ffc658f2b0c07434373b07b7da/x/slashing/keeper/infractions.go#L13) of the Slashing module on the consumer chains the missed blocks are counted even for validators in the bottom 5%. The soft opt-out logic just stops the consumer CCV module from sending SlashPackets for these validators (see [here](https://github.com/cosmos/interchain-security/blob/f1a61297251c4e14f66b70e81d03434d54f29c6c/x/ccv/consumer/keeper/validators.go#L138)). 

Thanks to @freak12techno for notifying us of this unexpected behavior. 

## Version

https://github.com/cosmos/interchain-security/commit/f1a61297251c4e14f66b70e81d03434d54f29c6c

## Steps to Reproduce

- have a validator in the bottom 5% that doesn’t run a consumer node
- after a while (more than the signing window on the consumer chain), delegate enough tokens to the validator to get it in the top 95%
- the validator will get immediately jailed for downtime

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

mpoke:
A solution is to have a flag for every CCV validator in the consumer module. This flag indicates whether a validator is part of the 95% or not. We can update these flags [here](https://github.com/cosmos/interchain-security/blob/812c64743c1ceb96c6a8705938043e6c99fcb02d/x/ccv/consumer/module.go#L142) when we update `SmallestNonOptOutPower`. When a validator becomes part of the 95%, then we set its `signInfo.StartHeight` to the current height using the `SetValidatorSigningInfo` method from the `slashingKeeper`. Clearly this is state breaking for the consumers.

freak12techno:
Is it possible to not update the signing blocks bit array if a validator is not a part of the set that is required to sign blocks on this consumer chain (as in, it's in the bottom 5%)?

[...]

PR #1549, opened 2024-01-05
fix!: Avoid immediately jailing validators that are no longer opted-out

mpoke:
## Description

Closes: #1517 

This PR implements the solution from [this comment](https://github.com/cosmos/interchain-security/issues/1517#issuecomment-1857930734). Mainly, when a validator can no longer opt out, the `StartHeight` fields in its `SigningInfo` struct (in the slashing module) is set to the current height. As a result, the validator cannot be jailed for downtime for at least the sliding window for downtime slashing (`SignedBlocksWindow`). 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/integration/common.go
```diff
@@ -597,20 +597,25 @@ func (suite *CCVTestSuite) GetConsumerEndpointClientAndConsState(
 
 // setupValidatorPowers delegates from the sender account to give all
 // validators on the provider chain 1000 power.
```
sainoe, line 599:
```suggestion
// validators on the provider chain the given voting powers.
```

tests/integration/soft_opt_out.go
```diff
@@ -0,0 +1,247 @@
...
+
+	testCases := []struct {
+		name            string
+		downtimeFunc    func(*consumerKeeper.Keeper, *slashingkeeper.Keeper, []byte, int)
+		targetValidator int
+		expJailed       bool
+		expSlashPacket  bool
+	}{
+		{
+			"donwtime top 95%",
```
sainoe, line 31:
```suggestion
			"downtime top 95%",
```

p-offtermatt (commented):
Curious about this: why reset the start height instead of the missed blocks counter?

afaict, the cleanest solution to me seems to be to, for every opt-out validator, set the missed blocks counter to zero in each block, before it is entered on chain.

This has a few advantages:
* No need for an extra field on validators (though it's still state machine breaking)
* Works out-of-the-box with existing downtime checkers (they will just not see the missed blocks)

I might be missing something though. I think the current solution is also fine, maybe not worth changing this for.

mpoke:
> Curious about this: why reset the start height instead of the missed blocks counter?

@p-offtermatt I didn't want to mess with the downtime logic too much. If you set the missed blocks counter to zero, you risk having the counter negative and potentially never trigger a downtime event. See https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/slashing/keeper/infractions.go#L41-L52. 

> afaict, the cleanest solution to me seems to be to, for every opt-out validator, set the missed blocks counter to zero in each block, before it is entered on chain.

That would be a lot of store writes and I don't see the advantage of doing it like this. 
 
> This has a few advantages:
> 
> * No need for an extra field on validators (though it's still state machine breaking)

Why is the extra field an issue? 

> * Works out-of-the-box with existing downtime checkers (they will just not see the missed blocks)

See my comment above. I think reseting the missed blocks counter would mess with the downtime logic.

p-offtermatt:
Makes sense, thanks. I didn't take a close look at the missed-blocks logic in the slashing module (which is more intricate than I thought), and indeed I think your solution seems good under the constraints.

p-offtermatt (approved):

x/ccv/consumer/keeper/soft_opt_out.go
```diff
@@ -73,3 +83,42 @@ func (k Keeper) GetSmallestNonOptOutPower(ctx sdk.Context) int64 {
...
 	return int64(binary.BigEndian.Uint64(bz))
 }
+
+func (k Keeper) UpdateSlashingSigningInfo(ctx sdk.Context) {
+	smallestNonOptOutPower := k.GetSmallestNonOptOutPower(ctx)
+
+	// Update SigningInfo for opted out validators
+	valset := k.GetAllCCValidator(ctx)
+	// sort validators by power ascending
+	sort.SliceStable(valset, func(i, j int) bool {
```
bermuell, line 93: valset should be ordered already by validator address (as claimed by SDK)
pls check if the re-ordering according to power is still needed as it comes with an overhead.

mpoke: Nice catch. See https://github.com/cosmos/interchain-security/pull/1549/commits/b00c874094af6b725fe1d0d97e5ee0a89e3e50eb

sainoe (approved):

Issue #1569, opened 2024-01-17
Validation of SlashAcks fails due to marshaling to Bech32 


mpoke:
## Summary of Bug

`VSCPackets` contain a list of `SlashAcks`, which are addresses of validators for which SlashPackets were received. These addresses are marshaled to string using Bech32. On the consumer, the `VSCPackets` fail validation as the consumer has a different Bech32Prefix than the provider. 

Note: this bug was not caught by the e2e tests as in the tests all chains use the same Bech32Prefix. 

Thanks to @pr0n00gler for bringing this issue to our attention. 

## Version

- (main) https://github.com/cosmos/interchain-security/commit/bd5b1ef2c4e3053b18083382c2dd853b0042c85e
- https://github.com/cosmos/interchain-security/tree/v3.3.0

## Steps to Reproduce

- Start a testnet with a provider and a consumer
- On the consumer, one validator doesn't sign, leasing to a `SlashPacket` being sent to the provider
- The first `VSCPacket` the provider sends to the consumer  after receiving the `SlashPacket` will fail validation resulting in the packet being dropped. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

Issue #1568, opened 2024-01-17
e2e tests with gaia fail


mpoke:
## Summary of Bug

The e2e tests with Gaia fail. First it’s an issue with `Dockerfile.gaia` as it needs to build Gaia with go 1.20, but ICS requires go 1.21. Then, it fails because gaiad doesn’t have the genesis command, i.e., `Error: unknown command "genesis" for "gaiad"`. 

## Version

https://github.com/cosmos/interchain-security/commit/bd5b1ef2c4e3053b18083382c2dd853b0042c85e

## Steps to Reproduce

Run `make test-gaia-e2e-short`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

mpoke:
The most likely issue is that the genesis root command was introduced with SDK v0.47, while the current version of Gaia (v14.1.0) uses SDK 0.45. Thanks @glnro for pointing this out.

p-offtermatt:
Hey @bermuell this sounds like we might be able to fix with the infra you are building for the compatibility tests, wdyt?

bermuell:
> Hey @bermuell this sounds like we might be able to fix with the infra you are building for the compatibility tests, wdyt?
need to look into that in detail but I'd say that should be possible with the new infra by adding a new test driver for that version.

mpoke:
Closing as the objective should be to deprecate existing e2e tests in favor of the interchain tests.

PR #1570, opened 2024-01-17
fix!: Validation of SlashAcks fails due to marshaling to Bech32 

mpoke:
## Description

Closes: #1569

- On the consumer, when validating `VSCPackets`, remove validation of `SlashAcks`. If a SlashAck cannot be converted to a consensus address, log an error message and move on.  
- Enable e2e tests to use different Bech32Prefix for consumers and provider. 

TODO:

- [x] changeover e2e tests fails due to wrong signature, `go run ./tests/e2e/... --tc changeover`
- [x] e2e tests with Gaia fail, which means we cannot check if the e2e tests work when Bech32Prefix on the provider is "cosmos"; see https://github.com/cosmos/interchain-security/issues/1568
  - Just use "cosmos" as Bech32Prefix on the provider
- [x] add e2e test where a validator get jailed on the consumer, gets unjailed on the provider, and then get again jailed on the consumer 
- [x] potential migration needed on the consumer (deleting all the state with prefix OutstandingDowntimeBytePrefix for validators with `SigningInfo.IndexOffset = 0`) 
  - Just add to UPGRADING.md how to update state on consumer upgrade handler

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


p-offtermatt:
Both test-cometmock and test-trace pass locally now, just small issues in both cases :)

.github/workflows/nightly-e2e.yml
```diff
@@ -1,9 +1,7 @@
 # Run integration tests nightly on main
-
-# !! Relevant changes to this file should be propagated manual-integration.yml
-
 name: nightly-e2e-main
 on:
+  workflow_dispatch:
```
mpoke, line 4: @MSalopek this should remove the need for manual-e2e.yml, right?

.github/workflows/nightly-e2e.yml
```diff
@@ -114,6 +112,54 @@ jobs:
           go-version: "1.21" # The Go version to download (if necessary) and use.
       - name: E2E multi-consumer tests
         run: go run ./tests/e2e/... --tc multiconsumer
+  consumer-misbehaviour-test:
```
mpoke, line 115: @sainoe Any reasons this was not added before?

.github/workflows/nightly-e2e.yml
```diff
@@ -114,6 +112,54 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Checkout LFS objects
+        run: git lfs checkout
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.21" # The Go version to download (if necessary) and use.
+      - name: E2E consumer-misbehaviour tests
+        run: go run ./tests/e2e/... --tc consumer-misbehaviour
+  consumer-double-sign-test:
```
mpoke, line 131: ditto: @sainoe Any reasons this was not added before?

tests/e2e/main.go
```diff
@@ -128,6 +128,12 @@ var stepChoices = map[string]StepChoice{
 		description: "consumer double signing tests",
 		testConfig:  DefaultTestConfig(),
 	},
+	"consumer-double-downtime": {
```
mpoke, line 131: @p-offtermatt is this the right way to add a new e2e test?

p-offtermatt: LGTM. Optionally, add the test case in https://github.com/cosmos/interchain-security/blob/e78593d8b9a2a071fd6dae79d91eeb035796c21e/tests/e2e/main.go#238if you want it to be a default test case

.github/workflows/nightly-e2e.yml
MSalopek: Yes

tests/e2e/actions.go
```diff
@@ -717,9 +737,17 @@ func (tr TestConfig) addChainToGorelayer(
 	ChainId := tr.chainConfigs[action.Chain].ChainId
 	rpcAddr := "http://" + queryNodeIP + ":26658"
 
+	var accountPrefix string
+	if ChainId == ChainID("provi") {
```
p-offtermatt, line 741: wdyt about having this in the ChainConfig https://github.com/cosmos/interchain-security/blob/f0b9d7b3b6139181687fcada146a74e88fcec18e/tests/e2e/config.go#L76 instead of in this "ad-hoc" way?

i.e. have a field in the chainConfig called "AccountPrefix", then here calling
```
accountPrefix := tr.chainConfigs[action.Chain].AccountPrefix
```

or something like that

tests/e2e/config.go
```diff
@@ -15,22 +15,57 @@ type (
...
+	// with Bech32Prefix = "cosmos"
+	ConsumerDelAddressOnProvider string
+	// Validator operator address on consumer marshaled to string using Bech32
+	// with Bech32Prefix = "consumer"
+	ConsumerValoperAddress string
+	// Validator operator address on consumer marshaled to string using Bech32
+	// with Bech32Prefix = "cosmos"
+	ConsumerValoperAddressOnProvider string
+	// Validator consensus address on consumer marshaled to string using Bech32
+	// with Bech32Prefix = "consumer". It matches the PrivValidatorKey bellow.
```
p-offtermatt, line 62:
```suggestion
	// with Bech32Prefix = "consumer". It matches the PrivValidatorKey below.
```

tests/e2e/config.go
```diff
@@ -15,22 +15,57 @@ type (
...
+	// with Bech32Prefix = "consumer"
+	DelAddressOnConsumer string
+	// Validator operator address on provider marshaled to string using Bech32
+	// with Bech32Prefix = "cosmos"
+	ValoperAddress string
+	// Validator operator address on provider marshaled to string using Bech32
+	// with Bech32Prefix = "consumer"
+	ValoperAddressOnConsumer string
+	// Validator consensus address on provider marshaled to string using Bech32
+	// with Bech32Prefix = "cosmos". It matches the PrivValidatorKey bellow.
```
p-offtermatt, line 33:
```suggestion
	// with Bech32Prefix = "cosmos". It matches the PrivValidatorKey below.
```

tests/e2e/state.go
```diff
@@ -519,28 +534,48 @@ func (tr TestConfig) getValPower(chain ChainID, validator ValidatorID) uint {
...
+		return uint(votingPower)
 	}
 
 	// Validator not in set, its validator power is zero.
 	return 0
 }
 
 func (tr TestConfig) getValStakedTokens(chain ChainID, validator ValidatorID) uint {
+	valoperAddress := tr.validatorConfigs[validator].ValoperAddress
+	if chain != ChainID("provi") {
```
p-offtermatt, line 564: I wonder if it makes sense for this repetitive logic to be in a util function. I know it's often slightly different, and it seems annoying to extract a core logic (maybe it's more like 4 different core logics, but then it might not make sense to extract at all). Just wondering out loud, not sure I have a strong feeling either way

p-offtermatt (approved):
Some nits. I focussed on the e2e tests in the review, so encourage others to focus on unit tests/the actual logic.

tests/e2e/actions.go
mpoke: Good idea. See https://github.com/cosmos/interchain-security/pull/1570/commits/7a8a5d55cc9f09df83c0e34a2ca0ba1a899b52af

tests/e2e/state.go
mpoke: @p-offtermatt I could add a struct that looks like this
```golang
type Bech32Address struct {
	// Provider address marshaled to string using Bech32 with ProviderAccountPrefix
	ProviderAddress string
	// Provider address marshaled to string using Bech32 with ConsumerAccountPrefix
	ProviderAddressOnConsumer string
	// Consumer address marshaled to string using Bech32 with ConsumerAccountPrefix
	ConsumerAddress string
	// Consumer address marshaled to string using Bech32 with ProviderAccountPrefix
	ConsumerAddressOnProvider string
}
```
Then have in the ValidatorConfig struct the following fields
```golang
	// Validator account address
	AccountAddress Bech32Address
	// Validator operator address
	ValoperAddress Bech32Address
	// Validator consensus address
	ValconsAddress Bech32Address
```
And move the logic to a method like this
```golang
func (bech32Addr Bech32Address) ChooseAddress(chain ChainID, useConsumerKey bool) string {
	addr := bech32Addr.ProviderAddress
	if chain != ChainID("provi") {
		// use binary with Bech32Prefix set to ConsumerAccountPrefix
		if useConsumerKey {
			addr = bech32Addr.ConsumerAddress
		} else {
			addr = bech32Addr.ProviderAddressOnConsumer
		}
	}
	return addr
}
```

However, this would refactor a lot of the code and make the reviewing harder. In case you think it's a good idea, I would prefer to do that in a separate PR.

mpoke: It is used in a different context. Look for `ConsumerValconsAddressOnProvider` in this PR.

MSalopek (approved):
LGTM!

Actual logic changes seem sound. There's a lot of testing code that neede to be changed so the PR looks really long.

Issue #1529, opened 2023-12-20
MBT: Add key assignment


p-offtermatt:
This is an issue that tracks the progress on adding key assignment: https://cosmos.github.io/interchain-security/features/key-assignment

Several main tasks should be done:
1) Adjust the Quint model to include the feature
2) Add tests in Quint to test the feature
3) Make sure that the generated traces use the feature in non-trivial ways (or add new generation methods, i.e. invariants, that do)
4) Expand the driver to understand/utilize the new feature

PR #1573, opened 2024-01-19
test: add key assignment to model and driver

p-offtermatt:
## Description

Closes: #1529 

* Adds key assignment to the Quint model
* Adds key assignment to the MBT driver

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/mbt/model/README.md
```diff
@@ -94,6 +106,13 @@ with a timestamp >= t + UnbondingPeriod on that consumer.
 - [X] EventuallyMatureOnProviderInv: If we send a VscPacket, this is eventually responded to by all consumers
 that were running at the time the packet was sent (and are still running).
 
+Invariants only relevant when running with key assignment (`--step stepKeyAssignment`):
+- [X] ValidatorSetHasExistedKeyAssignmentInv: Replaces ValidatorSetHasExistedInv. Validator sets are checked for equality under key assignment when checking whether they have existed.
```
insumity, line 110: nit: States that it replaces `ValidatorSetHasExistedInv` but `ValidatorSetHasExistedInv` seems to still exist.

tests/mbt/model/ccv.qnt
```diff
@@ -192,6 +227,87 @@ module ccv_types {
     // given as a pure val so that we can switch cases based on
     // whether a chain is the provider or not
     pure val PROVIDER_CHAIN = "provider"
+
+    // Takes the current provider state and validator set and returns 
+    // the validator set under the current the key assignments for the given consumer, as stored in the provider state.
```
insumity, line 232:
```suggestion
    // the validator set under the current key assignments for the given consumer, as stored in the provider state.
```

tests/mbt/model/README.md
p-offtermatt: it is meant to replace ValidatorSetHasExistedInv when running with --step stepKeyAssignment
I will make this clearer

tests/mbt/model/ccv.qnt
```diff
@@ -93,6 +99,29 @@ module ccv_types {
...
+
+            // For every consumer chain, stores the consumer address assigned by each validator.
+            validatorConsumerPubKey: Chain -> (Node -> ConsumerAddr),
+
+            // the history of validator sets on the provider, but with the key assignments applied.
+            // This is needed to check invariants about the validator set when key assignments are in play.
+            keyAssignedValSetHistory: Chain -> VotingPowerHistory,
+
+            // For every consumer chain, holds the provider validator for each assigned consumer key.
+            validatorByConsumerAddr: Chain -> (ConsumerAddr -> Node),
```
insumity, line 111: This seems to be the reverse of `validatorConsumerPubKey` so it might make sense to put under `validatorConsumerPubKey`. Also, this is called `ConsumerAddr` versus `ConsumerPubKey` above. Might make sense to have them consistent (I personally prefer the name with the `Addr` because `PubKey` feels to implementation specific).

tests/mbt/model/ccv.qnt
```diff
@@ -93,6 +99,29 @@ module ccv_types {
...
+
+            // Stores the key assignments that need to be replaced in the current block.
+            // Needed to apply the key assignments received in a block to the validator updates sent to the consumer chains.
+            // This is necessary because
+            // a key assignment itself triggers validator updates
+            //(old key to 0 power, new key to current power)
+            keyAssignmentReplacements: Chain -> (Node -> (ConsumerAddr, Power)),
+
+            // Stores the mapping from VSC ids to consumer validators addresses. Needed for pruning ValidatorByConsumerAddr.
```
insumity, line 120:
```suggestion
            // Stores the mapping from VSC ids to consumer validators addresses. Needed for pruning validatorByConsumerAddr.
```

tests/mbt/model/ccv.qnt
```diff
@@ -192,6 +227,87 @@ module ccv_types {
...
+        // (set old key to 0, new key to current power)
+        pure val newValSet = valSet.keys().map(
+            (node) => 
+            pure val power = valSet.get(node)
+            pure val keyAssignmentReplacements = providerState.keyAssignmentReplacements.getOrElse(consumer, Map())
+            if(keyAssignmentReplacements.keys().contains(node)) {
+                // there is a KeyAssignmentReplacement
+                pure val replacement = keyAssignmentReplacements.get(node)
+                pure val oldConsAddr = replacement._1
+                // set old key to 0
```
insumity, line 250: Why is this called **old**`ConsAddr` because the way I understood `keyAssignmentReplacements` is that it provides the new consumer address?

tests/mbt/model/ccv.qnt
```diff
@@ -192,6 +227,87 @@ module ccv_types {
...
+                pure val oldToZero = (oldConsAddr, 0)
+                // set new key to current power
+                pure val newToPower = (providerState.validatorConsumerPubKey.getOrElse(consumer, Map()).getOrElse(node, node), power)
+                Set(oldToZero, newToPower)
+            } else {
+                // no replacement
+                // check if the validator has a key assigned
+                pure val validatorConsumerPubKey = providerState.validatorConsumerPubKey.getOrElse(consumer, Map())
+                if (validatorConsumerPubKey.keys().contains(node)) {
```
insumity, line 259: Why doesn't `keyAssignmentReplacements` have the replacement but `validatorConsumerPubKey` does have a key? What exactly is the relation between `keyAssignmentReplacements` and `validatorConsumerPubKey`?

tests/mbt/model/ccv.qnt
sainoe: IIUC, `keyAssignmentReplacements` contains the validator addresses that need to be updated 
 (i.e. set old key to 0) in the current provider validator, before being sent to a consumer chain.
The new consumer addresses are stored in `providerState.validatorConsumerPubKey`.

tests/mbt/model/ccv_model.qnt
```diff
@@ -229,6 +238,17 @@ module ccv_model {
     // UTILITY FUNCTIONS
     // ==================
 
+    pure def removeZeroPowers(valSet: Node -> int): Node -> int =
```
insumity, line 241: Isn't possible to have this as `removeZeroPowers(valSet: ValidatorSet): ValidatorSet`?

tests/mbt/model/ccv.qnt
p-offtermatt: Exactly what Simon said, yeah

mpoke (commented):
Great work @p-offtermatt. Partial review without ccv_model.qnt. See my comments below. In general, try to either comment large blocks of quint or split the blocks -- it's very hard for me (Quint beginner) to read it.

tests/mbt/model/ccv.qnt
p-offtermatt: keyAssignmentReplacements has an entry only when the key was replaced in *this block*

testutil/integration/validators.go
```diff
@@ -8,6 +8,12 @@ import (
 	tmtypes "github.com/cometbft/cometbft/types"
 )
 
+// creates n signers. does not return a validator set, just only the PrivValidators by their addresses
+func CreateSigners(n int) (map[string]tmtypes.PrivValidator, error) {
```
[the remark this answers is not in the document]
p-offtermatt: To make the intention clearer, but I'm not attached. Will change.

tests/mbt/model/ccv.qnt
```diff
@@ -192,6 +227,87 @@ module ccv_types {
...
+        ).fold(
+            Map(),
+            (acc, pair) => acc.put(pair._1, pair._2)
+        )
+
+        // for each node in the valset, reverse its key assignment
+        valSetWithAssignment.keys().map(
+            (node) => 
+            pure val power = valSetWithAssignment.get(node)
+            pure val consAddr = reverseAssignment.getOrElse(node, node)
```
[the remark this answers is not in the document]
p-offtermatt: do you think getOrElse in general is problematic/unclear? it's just a map get with a default fallback if there is no entry for the key

tests/mbt/model/ccv_model.qnt
```diff
@@ -632,4 +653,271 @@ module ccv_model {
...
+    action KeyAssignment(
+        chain: Chain,
+        validator: Node,
+        consumerAddr: ConsumerAddr
+    ): bool =
+        val result = assignConsumerKey(currentState, chain, validator, consumerAddr)
+        all {
+            hasError(result) == false,
+            currentState' = result.newState,
+            trace' = trace.append(
```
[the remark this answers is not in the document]
p-offtermatt: different=better or worse?

the .with().with().with() syntax and the {...original,a:foo,b:bar} syntax do the same thing.
Not sure which I should prefer, do you have an opinion?

tests/mbt/model/ccv.qnt
```diff
@@ -93,6 +99,30 @@ module ccv_types {
             // a monotonic strictly increasing and positive ID that is used
             // to uniquely identify the Vscs sent to the consumer chains.
             runningVscId: int,
+
+            // For every consumer chain, stores the consumer address assigned by each validator.
+            validatorConsumerAddr: Chain -> (Node -> ConsumerAddr),
```
insumity, line 104: While reading the rest of the code, I had to come back to see again what exactly `validatorConsumerAddr` is. I think it might helpfult to rename to something like `validatorToConsumerAddr` and rename `validatorByConsumerAddr: Chain -> (ConsumerAddr -> Node),` to something like `consumerAddrToValidator`?
Or even (might be a bit too verbose) something like `chainToConsumerAddrToValidator`. I'm not sure if Quint has any naming conventions for this.

insumity (approved):
Great work!

I focused my review on the Quint code.
I concur with Mariu's comment in regards to introducing more comments in general.

Two last comments:
1. I'm not sure if it would simplify the model in not considering `0` powers. If I understood correctly from our discussion, this stems from ABCI which might be too implementation specific. If it doesn't simplify, then it's fine to leave as is.
2. Now `addr` and `key` seem to be used interchangeably in names. Would it make sense to stick with one for consistency?

tests/mbt/model/ccv.qnt
p-offtermatt: I think the only convention is the one we come up with :) I went with your first suggestion

p-offtermatt:
> Great work!
> 
> I focused my review on the Quint code. I concur with Mariu's comment in regards to introducing more comments in general.
> 
> Two last comments:
> 
>     1. I'm not sure if it would simplify the model in not considering `0` powers. If I understood correctly from our discussion, this stems from ABCI which might be too implementation specific. If it doesn't simplify, then it's fine to leave as is.
> 
>     2. Now `addr` and `key` seem to be used interchangeably in names. Would it make sense to stick with one for consistency?

1. Yes, I think this would simplify a lot. Let me do that - it will be some bigger changes, e.g. I can get rid of keyAssignmentReplacements afaict.
2. I think both are sort-of needed: The feature is called key assignment, but in the model I think I don't want to talk about keys, but rather addresses, but maybe this is confusing. Let me think about this a bit, but feel free to chime in if you have an opinion @mpoke @sainoe

tests/mbt/model/ccv_model.qnt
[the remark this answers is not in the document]
p-offtermatt: I am undecided. It's not consistent at the moment, and imo not worth refactoring, I want to get more data on what people like more (but I think it probably doesn't matter much)

tests/mbt/model/ccv.qnt
```diff
@@ -192,6 +227,87 @@ module ccv_types {
...
+        consumer: Chain,
+        valSet: ValidatorSet
+    ): ValidatorSet = {
+        // map each validator in the valSet to a set of
+        // new updates. necessary to allow one validator to generate multiple updates,
+        // which in turn is needed when someone changes keys
+        // (set old key to 0, new key to current power)
+        pure val newValSet = valSet.keys().map(
+            (node) => 
+            pure val power = valSet.get(node)
```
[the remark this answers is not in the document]
p-offtermatt: rewrote this :) there wasn't, now there is no more confusion at all

tests/mbt/model/ccv_model.qnt
```diff
@@ -632,4 +654,270 @@ module ccv_model {
...
+            // ...the validator set under key assignment is in a sent packet...
+            val providerState = currentState.providerState
+            consumer => providerState.sentVscPacketsToConsumer.get(consumer).toSet().exists(
+                packet => 
+                    packet.validatorSet == 
+                    applyKeyAssignmentToValSet(providerState, consumer, providerValSetInCurBlock)
+            )
+            // or the consumer was just started, which we detect by the consumer having a timestamp of 0
+            // and the consumer having the validator set that was just sent in the block
+            or
```
mpoke, line 729: > currentState.consumerStates.get(consumer).chainState.currentValidatorSet == providerValSetInCurBlock

Shouldn't this also have the key assignments applied?

mpoke (approved):
Great work @p-offtermatt.

sainoe (approved):

tests/mbt/model/ccv_model.qnt
mpoke: General feedback on the invariants: Some of them are very hard to read as they are written in a very concise form. Maybe it's just a matter of getting used to quint though.

p-offtermatt: I'll add some more comments to the invariants, specifically.
It's why I also try to split them up sometimes, but some are just not easily decomposed. I'll see if I can split them more

p-offtermatt: Also seems I was wrong here - the precondition is fine, rereading the property, it only talks about updates to validator powers, not to changes in key assignment.
The problem is that the or statement never mattered, since it's already an invariant with just the first disjunct, so the solution was to remove the second disjunct :)

Issue #1571, opened 2024-01-18
Write an ADR for Partial Set Security


insumity:
## Problem

Write an [ADR](https://github.com/cosmos/interchain-security/tree/main/docs/docs/adrs) for the upcoming Partial Set Security implementation.

## Closing criteria

The ADR should contain all relevant information on how we intend to go about implementing Partial Set Security, its consequences, etc. We can close this issue when the ADR has been reviewed and has been accepted.

PR #1584, opened 2024-01-22
docs: add ADR for Partial Set Security

insumity:
## Description

Closes: #1571

Add ADR for Partial Set Security.

---

### Author Checklist

I have...

- [x] provided a link to the relevant issue or specification
- [ ] reviewed "Files changed" and left comments if necessary 
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

I have...


docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+        ...
+```
+
+Note that we use `0` to remove validators that are not opted in as described [here](https://docs.cometbft.com/v0.37/spec/abci/abci++_methods#endblock) and as is done for the [assignment of consumer keys](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/key_assignment.go#L525).
+
+### How do we distribute rewards?
+Currently, rewards are distributed as follows: The consumer [periodically sends rewards](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/consumer/keeper/distribution.go#L148) on the provider `ConsumerRewardsPool` address. The provider then [transfers those rewards to the fee collector address](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/distribution.go#L77) and those transferred rewards are distributed to validators and delegators.
+
+In PSS, we distribute rewards only to validators that actually validated the consumer chain. To do this we generate a slice with bogus `VoteInfo` that contains the opted in (or Top N) validators and call [AllocateTokens](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/distribution/keeper/allocation.go#L14).
```
jtremback, line 231: Why not use AllocateTokensToValidator https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/distribution/keeper/allocation.go#L68?

jtremback: Doesn't seem like AllocateTokens will work because it is pulling out of the distribution at the top level instead of taking tokens as an input

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+    Header header = 5;
+}
+```
+
+The message includes the `vote` that was signed by the validator. Validators that vote on a fraud vote can look at the `BlockID` contained in a `vote` and the `appHash`, `lastResultHash`, etc. in the `header` (that led to this `BlockID`) to conclude whether the to-be-voted validator has performed incorrect execution or not.
+
+#### Double signing
+We do not change the way slashing for double signing and light client attacks functions. If a validator misbehaves on a consumer, then we slash that validator on the provider.
+
+#### Downtime
```
jtremback, line 273: We will need to check that a validator is actually opted in to a consumer chain when receiving a downtime packet about them

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+```
+message MsgOptIn {
+// the chain id of the consumer chain to opt in to
+string chain_id = 1;
+// the provider address of the validator on Cosmos Hub
+string provider_addr = 2;
+}
+```
+Note that in a Top N consumer chain, the top N% Cosmos Hub validators are forced by default to validate the consumer chain. Nevertheless, validators in the bottom 100 - N% can opt in to validate as well.
+Additionally, note that due to undelegations, etc. a validator that is in the top N% validators might fall down to the 100 - N% in which case this validator is not forced to validate anymore. Therefore, a validator that might move out of the top N% might want to send a `MsgOptIn` to keep validating even if this happens.
```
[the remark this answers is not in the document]
p-offtermatt: I agree with Jehans point. I think the more natural thing is to have people explicitly send a MsgOptOut if they were once top-N and don't want to validate anymore (they have to turn off their node anyways, since they were running it before, so I think we can also expect them to send an opt-out message when they shut it down). I think this is good because it (very mildly) incentivizes people to just keep validating(= more security for the chain)

p-offtermatt (approved):
Some first comments. I got to parts that we clearly discussed and decided on differently yesterday, so I'll keep off it and let you revise before I continue.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: This could be done as well but my concern is that it will not be obvious to a validator that it has to opt out.
For example, a validator `V` belongs to the Top N but votes `No` during a `ConsumerAdditionProposal` that introduces a Top N chain. Now, assume that `V` drops below the top N% validators, `V` would be forced to send a `MsgOptOut` even though it never opted in which seems peculiar. This case in reverse already happens today, where you might have to validate even though you do not want to, but I don't see why we should introduce another variant of this.
Another (extreme) example could be the following: A validator does not want to secure a consumer chain `C` before `C` starts but the validator belongs to the Top N so decides to re-delegate half of the tokens to another validator (under its control) so that both the old and the new validator are under Top N. This validator would need to quickly send a `MsgOptOut` when the consumer chain starts so not to get jailed for downtime.

> [...] incentivizes people to just keep validating(= more security for the chain)

This was one argument in favor of the delayed opted out.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+
+Additionally, a validator that opts in can then use `MsgAssignConsumerKey` to change the consumer key on the consumer chain as is currently done in Replicated Security.
+
+#### State & Query
+We store the list of validators that have opted in under the key:
+```
+ChainIdAndConsAddrKey(OptedInBytePrefix, chainID)
+```
+
+And for simplicity for each validator we store a boolean on whether this validator has opted in on this chain or not.
```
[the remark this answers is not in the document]
insumity: You mean, we could check whether a validator has opted in or not by traversing the list of opted in validators and hence only need one record?
This can be done but I wonder if there might be efficiency implications later on when we do things like:
```
for _, val := range validators {
    if hasOptedIn(val) {
    
    }
}
```
If `hasOptedIn` has to go through a list, then the time complexity of the above snippet is quadratic in comparison to having `hasOptedIn` utilize an underlying map.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+            })
+       } else {
+       if k.IsValidatorOptedIn(ctx, chainId, val.PubKey) {
+            updates = append(updates, abci.ValidatorUpdate{
+                PubKey: val.PubKey,
+                Power:  val.Power,
+            })
+        } else {
+            // remove this validator
+            updates = append(updates, abci.ValidatorUpdate{
```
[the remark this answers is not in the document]
insumity: Nice catch! There ways to keep track of the validator set that the consumer chain is consuming and not send unnecessary validator updates. I'll try to think more on if there's an easy way to do so but won't investigate too much time on this because when we move to [epochs](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-014-epochs.md) this would unlikely be an issue.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Good point. I clarified this here and added a note earlier that we need to keep track of validators that were opted in in the past even though they might not be opted in now.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+
+### How do validators opt out?
+Validators that have opted in on a chain can opt out as well. The opt out is not instantaneous and the validator only opts out after a tunable delay of `X` weeks (e.g., 5 weeks). We introduce the delay in order to inform consumer chains about a certain upcoming validator set change in case they want to use this information to take action (e.g., all validators in an opt in chain decide to opt out).
```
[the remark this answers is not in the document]
insumity: Sounds good. I was thinking we could do something like:
```
if val has opted in for more than X blocks {
     // give rewards to val
```
This way, even if you opt in at block `Y` you will start getting rewards after `Y + X` block. Additionally, if you opt out and then opt in back, you would still have to wait for `X` blocks before getting any rewards.
To do the above, we just have to keep track of the current block height when a validator opts in.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Decision
+In Replicated Security, 95% (see  soft opt-out) of the top Cosmos Hub validators have to secure a consumer chain. In Partial Set Security (PSS) we introduce a parameter `N` that is associated with a consumer chain and require that the top `N%` of the Cosmos Hub validators have to secure this consumer chain. Additionally, we allow validators that do not belong to the top `N%` of the Cosmos Hub validators to dynamically opt in if they want to validate on the consumer chain or not.
+For example, if a consumer chain has `N = 95%`, then it ultimately receives the same security it receives today by Replicated Security. On the other hand, if a consumer chain has `N = 0`%, then no validator is forced to validate the chain but validators can opt in to do so instead.
+In what follows, we call a consumer chain _Top N_ if it has joined as Top N chain with `N > 0` and _Opt In_ chain otherwise.  An Opt In consumer chain is secured only by the validators that have opted in to secure that chain.
```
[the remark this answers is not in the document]
insumity: @mpoke not sure what's your take on that (?)
We could only talk about `Top N` chains as you mentioned in the meeting but it seems to me that making the clear distinction between `Top N` and `Opt In` (i.e., `Top N` with `N = 0`) chains at the beginning simplifies the presentation because  later on we can just say "For Opt In chains we do this and for Top N chains we do that" instead of saying "If N = 0 then we do this and if N > 0 we do something else."

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
p-offtermatt: That explanation makes sense to me, thanks for taking the time to write it out. Maybe could have some of the pros/cons of this exact choice somewhere in the ADR, or maybe it's not worth it

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+
+If a proposal has those arguments wrongly set, it should get rejected in [ValidateBasic](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/types/proposal.go#L86).
+
+Note that we could easily distinguish whether a chain is _Top N_ or _Opt In_ by only having the single field `top_N` and if `top_N == 0`, then we’re in the _Opt In_ case, although for clarity, we introduce both fields.
+
+In a future version of PSS, we intend to introduce a way to modify the parameters of a consumer chain, e.g, a chain that is _Opt In_ to become _Top N_, etc.
+
+#### State & Query
+We augment the provider module’s state to keep track of the `is_top_N` and `top_N` values for each consumer chain. They key to store this information would be:
```
sainoe, line 53:
```suggestion
We augment the provider module’s state to keep track of the `is_top_N` and `top_N` values for each consumer chain. The key to store this information would be:
```

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+```
+ChainIdWithLenKey(IsTopNBytePrefix, chainID)
+```
+reusing [`ChainIdWithLenKey`](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/types/keys.go#L418).
+
+Then in the [keeper](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/keeper.go) we introduce methods as follows:
+```
+func (k Keeper) SetIsTopN(ctx sdk.Context, chainID string, isTopN bool) error
```
sainoe, line 62: nit: since this method sets only chainID for `top_N` chains, the `isTopN` argument could be removed.
very-nit: we try to avoid returning errors in setters.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+func (k Keeper) IsTopN(ctx sdk.Context, chainID string) bool
+
+// returns the N if Top N chain, otherwise an error
+func (k Keeper) GetTopN(ctx sdk.Context, chainID string) (uint8, error)
+```
+
+In a future version of PSS, we intend to extend the `gaiad query provider list-consumer-chains` query to also return information on whether a consumer chain is an Opt In or a Top N chain and with what N. This way, block explorers like [Mintscan](https://www.mintscan.io/) would show for a chain “This chain is secured by N% of Cosmos Hub.”
```
sainoe, line 69:
```suggestion
In a future version of PSS, we intend to extend the `interchain-security-pd query provider list-consumer-chains` query to also return information on whether a consumer chain is an Opt In or a Top N chain and with what N. This way, block explorers like [Mintscan](https://www.mintscan.io/) would show for a chain “This chain is secured by N% of Cosmos Hub.”
```

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
sainoe: Since consumer chains can't update their validator set, validators leaving the `TopN` are forced to send a `MsgOptIn` to keep validating the chain.
Maybe we should communicate to small validators the importance of manually opting in, even if they are momentarily in the TopN.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+            continue
+        }   
+        ...
+```
+
+### How do we send the partial validator sets to the consumer chains?
+To send the validator set to a consumer chain we first have to generate the validator updates with a function similar to this one:
+
+```
+func (k Keeper) GetPartialSetValidators(ctx types.Context, chainId string valUpdates []abci.ValidatorUpdate) (newUpdates []abci.ValidatorUpdate)
```
sainoe, line 160:
```suggestion
func (k Keeper) GetPartialSetValidators(ctx types.Context, chainId string, valUpdates []abci.ValidatorUpdate) (newUpdates []abci.ValidatorUpdate)
```

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+#### Fraud votes
+For fraud votes, we add a new type `FraudVoteProposal` similar to [EquivocationProposal](https://github.com/cosmos/interchain-security/pull/703). The time to detect incorrect execution evidence is 7 days (the same time as that of detecting a light client attack on a client with a `trustingPeriod` of 14 days); We have 7 days to detect an incorrect execution and then 14 days for the proposal to be accepted.
+
+##### When are fraud votes eligible?
+Because fraud votes allow the slashing of a validator through a governance proposal, we are cautious on when a fraud vote is eligible and try to restrict their eligibility as much as possible to avoid an adversary proposing a bogus fraud vote.
+As described earlier, because in Top N chains we consider that `N > 50%` we only consider validators that have opted in an Opt In chain. Fraud votes are also **only** intended to be used for **malicious** behaviour, for example, in a scenario similar to what [happened to Neutron](https://blog.neutron.org/neutron-halt-post-mortem-927dbe4540c8), validators should vote against slashing those validators.
+
+#### Message
+```
+message FraudVoteProposal {
```
sainoe, line 268: Is there a reason why this proposal is just for one validator fraud vote?

sainoe (approved):

docs/docs/adrs/adr-015-partial-set-security.md
insumity: You mean in contrast to being able to slash multiple validators for the same fraud vote?
This could be done as well, similarly to [EquivocationProposal](https://github.com/cosmos/interchain-security/pull/703). Nevertheless, having multiple validators to be slashed on a proposal could lead to the following confusing scenario: A proposal is proposed with validators `V1, V2, V3` but a validator `V` only considers that `V1, V2` performed incorrect execution but not `V3`. What should `V` vote for ? Yes or No?
It seems better to only consider one validator per proposal to avoid such edge cases. Of course, in this case we might have multiple proposals but there's i) no cost involved in this because the proposal deposit would be returned at the end, and ii) if someone has created one proposal already it would be straightforward to create another one with a different validator address.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+In PSS, we distribute rewards only to validators that actually validate the consumer chain. To do this, we have a pool associated with each consumer chain and consumers IBC transfer the rewards to this pool.
+Then, because we do not use the canonical fee pool and hence rewards are not automatically distributed, we will create a new and modified version of [AllocateTokens](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/distribution/keeper/allocation.go#L14) as follows:
+
+```
+func (k Keeper) AllocateTokensFromPool(ctx sdk.Context, poolAddress string, totalPreviousPower int64, bondedVotes []abci.VoteInfo) {
+	rewardsCollectedInt := k.bankKeeper.GetAllBalances(ctx, poolAddress)
+	// rest in the same that `AllocateTokens` operates
+	...
+```
```
mpoke, line 224: It is not clear to me how this works. How is this different than `AllocateTokens`?

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+    ...
+    totalPower := 0
+    var []abci.VoteInfo votes
+    for _, val := range (validators that opted in or Top N) {
+        if val has opted in for more than X blocks {
+            votes = append(votes, []abci.VoteInfo{{Validator: val, SignedLastBlock: true}})
+        }
+    }
+    
+    k.AllocateTokens(ctx, pool associated with this chain, totalPower, votes)
```
mpoke, line 239: Why `totalPower = 0`? Also, shouldn't this be `AllocateTokensFromPool`?

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+In [TransferRewardsToFeeCollector](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/distribution.go#L61) we then do:
+
+```
+for _, chain := range consumerChains {
+    ...
+    totalPower := 0
+    var []abci.VoteInfo votes
+    for _, val := range (validators that opted in or Top N) {
```
mpoke, line 233: How are we getting the validators in Top N? I was thinking that maybe it's just easier to store the consumer validator set for every consumer chain. When sending VSCPackets, we just update these valsets. Yeah, it's more storage, but it shouldn't be that much (unless we have 1000s of consumer chains).

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+    sdk.NewCoins(balance),
+)
+```
+
+Note that we would only distribute rewards to validators that are opted in but are opted in for some time (e.g., 10000 blocks) to avoid cases where validators opt in and out in short intervals just in order to receive rewards.
+
+### Misbehaviour
+
+#### Fraud votes
+For fraud votes, we add a new type `FraudVoteProposal` similar to [EquivocationProposal](https://github.com/cosmos/interchain-security/pull/703). The time to detect incorrect execution evidence is 7 days (the same time as that of detecting a light client attack on a client with a `trustingPeriod` of 14 days); We have 7 days to detect an incorrect execution and then 14 days for the proposal to be accepted.
```
mpoke, line 260: This numbers depend on the UnbondigPeriods on both provider and consumers.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+##### When are fraud votes eligible?
+Because fraud votes allow the slashing of a validator through a governance proposal, we are cautious on when a fraud vote is eligible and try to restrict their eligibility as much as possible to avoid an adversary proposing a bogus fraud vote.
+As described earlier, because in Top N chains we consider that `N > 50%` we only consider validators that have opted in an Opt In chain. Fraud votes are also **only** intended to be used for **malicious** behaviour, for example, in a scenario similar to what [happened to Neutron](https://blog.neutron.org/neutron-halt-post-mortem-927dbe4540c8), validators should vote against slashing those validators.
```
mpoke, line 264:
```suggestion
As described earlier, because Top N have `N > 50%`, we only enable fraud votes for Opt In chains. 
Fraud votes are also **only** intended to be used for **malicious** behaviour, for example, in a scenario similar to what [happened to Neutron](https://blog.neutron.org/neutron-halt-post-mortem-927dbe4540c8), validators should vote against slashing those validators.
```

mpoke: What do you mean by "Fraud votes are also **only** intended to be used for **malicious** behaviour"? Also, probably not a good idea to single out Neutron in the ADR.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+    
+    // validator's vote contains BlockId on what was signed
+    Vote vote = 4;
+    
+    // header for which the hash led to the blockId in the vote
+    Header header = 5;
+}
+```
+
+The message includes the `vote` that was signed by the validator. Validators that vote on a fraud vote can look at the `BlockID` contained in a `vote` and the `appHash`, `lastResultHash`, etc. in the `header` (that led to this `BlockID`) to conclude whether the to-be-voted validator has performed incorrect execution or not.
```
mpoke, line 283: I would suggest to leave Fraud Votes our of this ADR. THere are some open questions that need to be addressed: 

- How are validators supposed to know from the header whether the execution is correct or not? 
- How do we deal with the tradeoff between encouraging submitter of Fraud Votes and discouraging spamming?

docs/docs/adrs/adr-015-partial-set-security.md
mpoke: I would also disable fraud votes for Opt In chains with at least 50% of the provider voting power opted in.

mpoke (commented):
Great work @insumity. See my comments below.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+func (k Keeper) AllocateTokensFromPool(ctx sdk.Context, poolAddress string, totalPreviousPower int64, bondedVotes []abci.VoteInfo) {
+	rewardsCollectedInt := k.bankKeeper.GetAllBalances(ctx, poolAddress)
+	// rest in the same that `AllocateTokens` operates
+	...
+```
+
+In [TransferRewardsToFeeCollector](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/distribution.go#L61) we then do:
+
+```
+for _, chain := range consumerChains {
```
jtremback, line 229: Need to specify the mechanism for tracking which rewards are for which consumer chain. Probably separate fee pool accounts is easiest. https://github.com/cosmos/interchain-security/issues/1588

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+func (k Keeper) IsValidatorOptedIn(ctx sdk.Context, chainID string, val Validator) bool
+```
+
+We introduce a query to retrieve the validators that are opted in. The query would operate as follows.
+```
+gaiad query provider optedInValidators $chainId
+```
+
+#### When do validators opt in?
```
jtremback, line 112: To implement top-n, we should just have an automatic process every block or epoch or whatever, which automatically opts in validators in the top-n, just as if they had sent the message themselves.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+during the `ConsumerAdditionProposal` that introduced a consumer chain. This simplifies validators operations because
+they do not have to send an additional message to opt in.
+
+Because the `Tally` method [deletes the votes](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/tally.go#L71) after reading them, we cannot check the votes of the validators after the votes have been tallied. To circumvent this, we introduce
+a hook for [`AfterProposalVote`](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/vote.go#L35) and keep track
+of all the votes cast by a validator. If a validator votes more than once, we only consider the latest vote.
+Finally, we only consider a validator has opted in if it casts a 100% `Yes` vote in case of a [weighted vote](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-037-gov-split-vote.md).
+
+
+### How do validators opt out?
```
jtremback, line 123: To implement top-n, we should have an if condition here which checks if they are in the top-n and rejects their opt out message if they are.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
jtremback: We've decided that it should be possible to implement top-n completely in the opt in and opt out handlers. See above.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+As a solution, we present _Partial Set Security_ (PSS). As the name suggests, PSS allows for only a subset of the validator set of the Cosmos Hub to secure a consumer chain. In what follows we propose the exact steps we need to take to implement PSS. This is a first iteration of Partial Set Security, and therefore we present the most minimal solution that achieves PSS possible.
+
+## Decision
+In Replicated Security, 95% (see  soft opt-out) of the top Cosmos Hub validators have to secure a consumer chain. In Partial Set Security (PSS) we introduce a parameter `N` that is associated with a consumer chain and require that the top `N%` of the Cosmos Hub validators have to secure this consumer chain. Additionally, we allow validators that do not belong to the top `N%` of the Cosmos Hub validators to dynamically opt in if they want to validate on the consumer chain or not.
+For example, if a consumer chain has `N = 95%`, then it ultimately receives the same security it receives today by Replicated Security. On the other hand, if a consumer chain has `N = 0`%, then no validator is forced to validate the chain but validators can opt in to do so instead.
```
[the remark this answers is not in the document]
jtremback: @mpoke that is incorrect

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+Consumer chains join Partial Set Security the same way chains now join Replicated Security (e.g., see [Stride](https://www.mintscan.io/cosmos/proposals/799)). We extend [`ConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v3.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L27) with 2 optional fields:
+`bool is_top_N`: flag that captures whether the consumer chain joins under the Top N case.
+`int8 top_N`: value that captures the percentage of the Top N% validators that should validate this consumer chain. This is only set if `is_top_N` is `true`.
+
+Note that `top_N` resides between `[50, 95]`.
+Assuming that at most `1/3` of validators could be malicious, by having 50% as the minimum value for `top_N` we guarantee that we cannot have a successful incorrect-execution attack on a Top N consumer chain. This is because, a Top N consumer chain with `N >= 50%` would have at least `1/3` correct validators that would be able to censor any incorrect-execution attack.
```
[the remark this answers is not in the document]
jtremback: IMO this is an interesting argument, but we should probably just leave it out

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+Additionally, by having a `N >= 50%` and hence `N > 33%` we provide the ability for validators to `Veto` a `ConsumerAdditionProposal` if they do not desire to validate a consumer chain.
+
+`top_N` can be up to 95% to capture the current case of Replicated Security where we allow the bottom 5% of validators to opt out. There is no reason for `top_N` to be higher than 95%. Smaller chains that belong in the bottom 5% validators can choose to opt in if they want to validate.
+
+So, the fields can take the following values:
```
jtremback, line 39: This might change, not a huge deal

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+// returns the N if Top N chain, otherwise an error
+func (k Keeper) GetTopN(ctx sdk.Context, chainID string) (uint8, error)
+```
+
+In a future version of PSS, we intend to extend the `interchain-security-pd query provider list-consumer-chains` query to also return information on whether a consumer chain is an Opt In or a Top N chain and with what N. This way, block explorers like [Mintscan](https://www.mintscan.io/) would show for a chain “This chain is secured by N% of Cosmos Hub.”
+
+### How do validators opt in?
+A validator can opt in by sending a new type of message that we introduce in [tx.proto](https://github.com/cosmos/interchain-security/blob/v3.3.0/proto/interchain_security/ccv/provider/v1/tx.proto#L1).
+```
+message MsgOptIn {
```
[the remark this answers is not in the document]
jtremback: I would keep the messages separate for now, and add extra conveniences like combined messages once the system is built.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+```
+func (k Keeper) GetPartialSetValidators(ctx types.Context, chainId string, valUpdates []abci.ValidatorUpdate) (newUpdates []abci.ValidatorUpdate)
+
+    powerSum := sdk.ZeroDec()
+    totalPower := k.ComputeTotalPower(ctx)
+    topNThreshold := k.GetTopN(ctx, chainId) / 100
+    for _, val := range valUpdates {
+        powerSum = powerSum.Add(sdk.NewDecFromInt(sdk.NewInt(val.Power)))
+        // if powerSum / totalPower > topNThreshold
+        if k.IsTopN(ctx, chainId) && powerSum.Quo(totalPower).GT(topNThreshold) {
```
jtremback, line 168: As discussed, we don't need to do anything with top n in here, since we do everything just by automatically opting validators in as discussed above

docs/docs/adrs/adr-015-partial-set-security.md
jtremback: As discussed above, we are now going to implement top-n by just having a process that opts validators in automatically if they are in the top-n, so nothing else really needs to know about top-n, including this code.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+In [TransferRewardsToFeeCollector](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/distribution.go#L61) we then do:
+
+```
+for _, chain := range consumerChains {
+    ...
+    totalPower := 0
+    var []abci.VoteInfo votes
+    for _, val := range (validators that opted in or Top N) {
+        if val has opted in for more than X blocks {
```
jtremback, line 234: This probably needs to be specced in more detail

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
mpoke: The argument is still there. I just added a new line to make it easier to review and add suggestions.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
mpoke: I agree to not complicate now, but we should take these UX decisions before releasing PSS as afterwards it will require more comms to notify users that we changed the API

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+For example, if a consumer chain has `N = 95%`, then it ultimately receives the same security it receives today by Replicated Security. On the other hand, if a consumer chain has `N = 0`%, then no validator is forced to validate the chain but validators can opt in to do so instead.
+In what follows, we call a consumer chain _Top N_ if it has joined as Top N chain with `N > 0` and _Opt In_ chain otherwise.  An Opt In consumer chain is secured only by the validators that have opted in to secure that chain.
+
+We intend to implement PSS using a feature branch off [v3.3.0 interchain security](https://github.com/cosmos/interchain-security/tree/v3.3.0).
+
+### How do consumer chains join?
+As a [simplification](https://forum.cosmos.network/t/pss-permissionless-vs-premissioned-lite-opt-in-consumer-chains/12984/17), a consumer chain can only join Partial Set Security through a governance proposal and not in a permissionless way.
```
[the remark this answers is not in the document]
insumity: Link pointing to a specific forum post was intentional because the specific post explained the chain id squatting issue. Rephrased the sentence to make this clear.
Also, used everywhere the PSS abbreviation.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+Consumer chains join Partial Set Security the same way chains now join Replicated Security (e.g., see [Stride](https://www.mintscan.io/cosmos/proposals/799)). We extend [`ConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v3.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L27) with 2 optional fields:
+`bool is_top_N`: flag that captures whether the consumer chain joins under the Top N case.
+`int8 top_N`: value that captures the percentage of the Top N% validators that should validate this consumer chain. This is only set if `is_top_N` is `true`.
```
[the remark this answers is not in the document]
insumity: Made it a `string` and remove the (unecessary) `boolean` field.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+
+
+If a proposal has those arguments wrongly set, it should get rejected in [ValidateBasic](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/types/proposal.go#L86).
+
+Note that we could easily distinguish whether a chain is _Top N_ or _Opt In_ by only having the single field `top_N` and if `top_N == 0`, then we’re in the _Opt In_ case, although for clarity, we introduce both fields.
+
+In a future version of PSS, we intend to introduce a way to modify the parameters of a consumer chain, e.g, a chain that is _Opt In_ to become _Top N_, etc.
+
+#### State & Query
+We augment the provider module’s state to keep track of the `is_top_N` and `top_N` values for each consumer chain. The key to store this information would be:
```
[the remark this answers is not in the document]
insumity: We do not need both. Thought for simplicity to have both but thought about it again and removed `is_top_N`. See my previous [comment](https://github.com/cosmos/interchain-security/pull/1584/files#r1469677466).

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+A validator can opt in by sending a new type of message that we introduce in [tx.proto](https://github.com/cosmos/interchain-security/blob/v3.3.0/proto/interchain_security/ccv/provider/v1/tx.proto#L1).
+```
+message MsgOptIn {
+    // the chain id of the consumer chain to opt in to
+    string chain_id = 1;
+    // the provider address of the validator on Cosmos Hub
+    string provider_addr = 2;
+}
+```
+Note that in a Top N consumer chain, the top N% Cosmos Hub validators are forced by default to validate the consumer chain. Nevertheless, validators in the bottom 100 - N% can opt in to validate as well.
```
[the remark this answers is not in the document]
insumity: I rephrased to:
> Note that in a Top N consumer chain, the top `N%` provider validators have to validate the consumer chain.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: After discussing more about this we reached to the conclusion that top N% validators are automatically opted in the case of a Top N consumer chain. A validator that belongs to the bottom `(100-N)%` would also automatically opt in if it enters (e.g., through delegations) the top N% validators. For a validator to opt out, the validator has to explicitly opt out by sending a `MsgOptOut` message.

> Depending on the reward mechanism, automatically opting a validator our may take away some of its rewards.
With the above approach we wouldn't have this issue.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+```
+ChainIdWithLenKey(OptedInByteKey, chainID)
+```
+
+And for simplicity for each validator we store a boolean on whether this validator has opted in on this chain or not.
+```
+ChainIdAndConsAddrKey(OptedInByteKey, chainID, addr)
+```
+
+We augment the state of the provider to keep track for each chain all the validators that have opted in and if a specific validator has opted in.
```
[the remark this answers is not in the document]
insumity: True. Removed.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: Intend to revamp the whole section here. But in regards to your question:
> Is it fine for Comet to receive a val update with power 0 for a validator that is not part of the set?

It does not seem to be fine. From what I see, in CometBFT, when we "delete"(i.e., set power to 0) more validators than the actual validators of the chain we get a [panic](https://github.com/cometbft/cometbft/blob/v0.38.5/types/validator_set.go#L550). Something like this could trigger, if in a consumer chain with only 5 opted in validators, we send 175 deleted ones.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Just briefly mention fraud votes but hint to a future ADR for them.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this because we will not talk about fraud votes in this ADR anymore.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this because we do not talk about fraud votes anymore in this ADR.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,300 @@
...
+    // the provider address of the validator
+    string provider_addr = 2;
+    // (optional) the consensus public key to use on the consumer
+    optional string consumer_key = 3;
+}
+```
+Note that in a Top N consumer chain, the top `N%` provider validators have to validate the consumer chain.
+Nevertheless, validators in the bottom `(100 - N)%` can opt in to validate as well.
+Provider validators that belong to the top `N%` validators are immediately opted in to validate a Top N consumer chain.
+This means that if a validator `V` belongs to the top `N%` validators but later falls (e.g., due to undelegations) to the bottom `(100 - N)%`, `V` would still be considered opted in and has to validate unless `V` sends a `MsgOptOut` message (see below).
```
p-offtermatt, line 91: So you would have to opt out again every time you get in the top N, right?
i.e.
val V is in top N
falls below top N
opts out
gets back in top N
falls below again
has to opt out again

not really a problem, because this is already annoying and validators that skirt in and out of top N probably also don't really want to turn on/off their nodes all the time, so they'd probably just stay opted in until they are for sure not in the top N anymore.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: Resolving this since we removed the partial-set computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the exact reward distribution computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: Added consumer key in the message.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+### How do we send the partial validator sets to the consumer chains?
+To send the validator set to a consumer chain we first have to generate the validator updates with a function similar to this one:
+
+```
+func (k Keeper) GetPartialSetValidators(ctx types.Context, chainId string, valUpdates []abci.ValidatorUpdate) (newUpdates []abci.ValidatorUpdate)
+
+    powerSum := sdk.ZeroDec()
+    totalPower := k.ComputeTotalPower(ctx)
```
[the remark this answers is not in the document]
insumity: Resolving this since we removed the partial-set computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the exact reward distribution computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the exact reward distribution computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the exact reward distribution computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Added a sentence on this.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the partial-set computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: Resolving this since we removed the partial-set computation from the ADR for now.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,224 @@
...
+    string providerAddr = 2;
+}
+```
+Validators can only opt out after a consumer chain has started and hence the above message returns an error if the chain with `chainID` is not running.
+Additionally, a validator that belongs to the top `N%` validators cannot opt out from a Top N chain and hence a `MsgOptOut` would error in such a case.
+
+#### State & Query
+We also update the state of the opted-in validators when a validator has opted out by removing the opted-out validator.
+
+Note that we only punish validators for downtime that were  opted in on a consumer chain.
```
mpoke, line 154:
```suggestion
Note that only opted-in validators can be punished for downtime on a consumer chain.
```

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,224 @@
...
+For this, we keep historical info of all the validators that have opted in during the last `X` (to be defined) blocks.
+This way we can jail validators for downtime knowing that indeed the validators have opted in at some point in the past. 
+Otherwise, we can think of a scenario where a validator `V` is down for a period of time, but before `V` gets punished for downtime, validator `V` opts out, and then we do not know whether `V` should be punished or not.
```
mpoke, line 157: We could use the opted in state `optedInBytePrefix | len(chainID) | chainID | addr` and add the blockheight as a value.

mpoke (approved):
Great work @insumity. Just some nits.

docs/docs/adrs/adr-015-partial-set-security.md
```diff
@@ -0,0 +1,224 @@
...
+As a simplification and to avoid [chain id squatting](https://forum.cosmos.network/t/pss-permissionless-vs-premissioned-lite-opt-in-consumer-chains/12984/17), a consumer chain can only join PSS through a governance proposal and not in a permissionless way.
+
+Consumer chains join PSS the same way chains now join Replicated Security, namely through a `ConsumerAdditionProposal` proposal.
+We extend [`ConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.0.0/proto/interchain_security/ccv/provider/v1/provider.proto#L27) with one optional field:
+
+`uint32 top_N`: Corresponds to the percentage of validators that join under the Top N case.
+For example, `53` corresponds to a Top 53% chain, meaning that the top `53%` provider validators have to validate the proposed consumer chain.
+`top_N`  can be `0` or include any value in `[50, 95]`. A chain can join with `top_N == 0` as an Opt In, or with `top_N ∈ [50, 95]` as a Top N chain.
+
+In case of a Top N chain, we restrict the possible values of `top_N` from `(0, 100]` to `[50, 95]`.
```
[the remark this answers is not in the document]
insumity: It seems then that we do not really need `SoftOptOutThreshold` and it would make sense to just have `top_N` be anything between `[50, 100]`. If a consumer chain is added with `top_N = 95` then it's like `SoftOptOutThreshold` was set to 5, if a consumer chain is set with `top_N = 90`, then it's like `SoftOptOutThreshold` was set to 10.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: We intend to store the blockheight as a value in any case. See above:
> For each validator, we store on whether the validator has opted in and at which block height the validator opted in under the key:
```
optedInBytePrefix | len(chainID) | chainID | addr
```
I think we can introduce one more `boolean` flag on whether the validator is currently opted in or not and have something similar to `addr -> (blockHeight, isOptedIn)` and this way if someone opts out we just set the `boolean` flag to `false`.  But we would still keep the information that the validator was opted in at some earlier point in time and at which height.

insumity: Updated the ADR to capture the above. Resolving this.

docs/docs/adrs/adr-015-partial-set-security.md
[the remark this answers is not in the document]
insumity: Updated the ADR to capture the above. Resolving this.

docs/docs/adrs/adr-015-partial-set-security.md
insumity: cc-ing @jtremback in regards to the later communication part

Issue #1574, opened 2024-01-19
Spike: Prototype Partial Set Security in Quint


p-offtermatt:
## Problem

We want to get familiar with doing spike implementations in Quint, in particular we want to assess whether it is even 
viable. Therefore, we should model partial set security in Quint first, and pay attention to whether this is useful during later implementation stages.

## Closing criteria

Partial set security has been added to the Quint model of CCV.

## Problem details

These are the steps to spiking partial set security in the CCV Quint model:
* Add data structures that are necessary for partial set security to ProviderState/ConsumerState
* Implement Partial Set Security API. Some incomplete suggestions:
   * Add new consumer chain as top-n
   * Add new consumer chain permissionlessly
   * Adjust top-n percentage
   * Validator opts in to a chain (either with inclusive top-n or into a permissionless consumer chain)
   * Validator opts out of a chain
   * ...more?
* Adjust logic in existing model. Some incomplete suggestions:
   * Different VSCPackets should be sent to different chains, depending on opted-in validators
   *  When blocks end and the validator set changes, who is top-n forced to what chain needs to change

TODO: this might become an epic and need to be split, but for now I think it's probably not big enough for an epic

PR #1614, opened 2024-01-31
test: MBT: Add partial set security to model

p-offtermatt:
## Description

Closes: #1574 

Adds partial set security to the Quint model.

This PR only contains logic, tests, and sanity checks, but no invariants yet.
The invariants are becoming more subtle and harder to define, so I want to see whether to prioritise MBT or defining the invariants first.

Highlights:
* ccv_pss contains the logic for calculating the top N set. Notice it can be much simpler because I don't need synthetic updates, I just send the whole set
* in ccv.qnt, some things changed to accomodate sending the partial validator sets

---

### Author Checklist

I have...

- [x] Confirmed all CI checks have passed (test failure for MBT is expected)

### Reviewers Checklist

I have...


tests/mbt/model/libraries/extraSpells.qnt
```diff
@@ -139,6 +139,71 @@ module extraSpells {
...
+      // If __x was not inserted, it was GT than every other element, so it goes at the end
+      __result.acc.append(__x)
+    else
+      __result.acc
+  }
+
+  run sortedListInsertTest = all {
+    assert(List().sortedListInsert(3, intCompare) == List(3)),
+    assert(List(1,2,4).sortedListInsert(3, intCompare) == List(1,2,3,4)),
+    assert(List(4,1,2).sortedListInsert(3, intCompare) == List(3,4,1,2)),
```
insumity, line 194: This seems a peculiar test because it does not satisfy the precondition of `sortedListInsert`, namely that `List(4, 1, 2)` is sorted. Might make sense to replace with a test that has a non-empty list and we append something at the its tail.

tests/mbt/model/libraries/extraSpells.qnt
```diff
@@ -139,6 +139,71 @@ module extraSpells {
...
+  }
+
+  //// Returns a list of all elements of a set.
+  //// The ordering will be arbitrary.
+  ////
+  //// - @param __set a set
+  //// - @param __cmp an ordering over the elements of the set
```
insumity, line 201: nit: different comment for `__cmp` above:
```
/// - @param __cmp an operator defining an `Ordering` of the elemnts of type `a`
```

tests/mbt/model/libraries/extraSpells.qnt
```diff
@@ -139,6 +139,71 @@ module extraSpells {
...
+    if (__a < __b)
+      { LT }
+    else if (__a > __b)
+      { GT }
+    else
+      { EQ }
+  }
+
+  /// Assuming `__l` is sorted according to `__cmp`, returns a list with the element `__x`
+  /// insterted in order.
```
insumity, line 162:
```suggestion
  /// inserted in order.
```

tests/mbt/model/libraries/extraSpells.qnt
```diff
@@ -139,6 +139,71 @@ module extraSpells {
         __set.fold(List(), (__l, __e) => __l.append(__e))
     }
 
+  /// The type of orderings between comparable things
```
insumity, line 142: Why the `///` here and later instead of `//`?

p-offtermatt: Good question, this is copy-pasted from the Quint "standard library" (there is no actual standard library yet, so copy-pasting what you need is atm the canonical way of using it) and their style differs sometimes

tests/mbt/model/ccv_pss.qnt
```diff
@@ -0,0 +1,157 @@
...
+// to secure a consumer chain.
+// Not all logic related to PSS is inside this module, as some logic is
+// too tightly coupled with the core CCV logic,
+// which is instead found in ccv.qnt
+module ccv_pss {
+    import ccv_types.* from "./ccv"
+    import extraSpells.* from "./libraries/extraSpells"
+    import ccv_utils.* from "./ccv_utils"
+
+    // Given a base validator set, an N for a top N chain, and a set of validators that have opted in to the chain,
```
insumity, line 13: The way I read this, I would expect that an `N` parameter to this function which is not the case.

tests/mbt/model/ccv_pss.qnt
```diff
@@ -0,0 +1,157 @@
...
+    // returns the validator set that should be sent to the chain.
+    // Assumes that the value for N is valid.
+    pure def GetPSSValidatorSet(providerState: ProviderState, origValSet: ValidatorSet, consumer: Chain): ValidatorSet = {
```
insumity, line 16: Why do we need to take the `origValSet` as a parameter here? Isn't this something we can extract from the `providerState`?

tests/mbt/model/ccv_pss.qnt
```diff
@@ -0,0 +1,157 @@
...
+        GetPSSValidatorSet_helper(origValSet, optedInVals)
+    }
+
+    pure def GetPSSValidatorSet_helper(origValSet: ValidatorSet, optedInVals: Set[Node]): ValidatorSet = {
+        origValSet.mapFilter(v => optedInVals.contains(v))
+    }
+
+    // Given a validator set and N, returns the top N% of validators by power.
+    // Note that in the edge case of multiple validators having the same power,
+    // all validators with the lowest voting power that is forced to validate are included.
```
insumity, line 27:
```suggestion
    // all validators with the lowest voting power that are forced to validate are included.
```

tests/mbt/model/ccv_pss.qnt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    // Runs the PSS logic for a single consumer.
+    // Should only be run for running chains.
+    pure def endBlockPSS_helper(providerState: ProviderState, consumer: Chain): ProviderState = {
+        val proviValSet = providerState.chainState.currentValidatorSet
+        val topNVals = GetTopNVals(proviValSet, providerState.topNByConsumer.get(consumer))
+        val prevOptedInVals = providerState.optedInVals.getOrElse(consumer, Set())
+        // opt in all the top N validators, i.e. union the top N vals with the previous opted in vals
+        val newOptedInVals = providerState.optedInVals.put(consumer, prevOptedInVals.union(topNVals))
```
insumity, line 151: I guess because we have sets here, we're guaranteed that a validator that `IsTopN` and has opted in only appears once at the `optedInVals`, right?

tests/mbt/model/ccv_pss.qnt
```diff
@@ -0,0 +1,157 @@
...
+
+        topNValSet.contains(validator)
+    }
+
+    // Returns true if the given validator has opted in to the given consumer chain,
+    pure def IsOptedIn(currentState: ProtocolState, validator: Node, consumer: Chain): bool = {
+        currentState.providerState.optedInVals.getOrElse(consumer, Set()).contains(validator)
+    }
+
+    // Opts a validator out. Safe to call before the consumer chain even runs.
```
insumity, line 110: > Safe to call before the consumer chain even runs.

Regarding the implementation of this (although we do not have the code yet), I was thinking that we should not allow validators to opt out before the chain has started running. This is because, we assume validator that voted `Yes` on a `ConsumerAdditionProposal` are opted in by default, so it would be weird to give them the right to opt out before the chain starts. What do you think?

tests/mbt/model/ccv_pss_model.qnt
```diff
@@ -0,0 +1,113 @@
...
+            ),
+            params' = params,
+        }
+    }
+
+    // Different sets of possible values for the topN parameter.
+    val allFullConsumers: Set[int] = Set(100)
+    val allOptIn: Set[int] = Set(0)
+    // only choose a few values for top N here to not make the "edge cases" of 0 and 100 too unlikely
+    val randomTopN: Set[int] = Set(0, 50, 70, 80, 90, 100)
```
insumity, line 63: nit: The value chosen from the set is randomly chosen but not the set itself. It might make sense to just name this something akin to `possibleTopNValues`.

tests/mbt/model/ccv_pss_test.qnt
```diff
@@ -0,0 +1,34 @@
...
+// which is instead found in ccv.qnt
+module ccv_pss_test {
+    import ccv_types.* from "./ccv"
+    import extraSpells.* from "./libraries/extraSpells"
+    import ccv_utils.* from "./ccv_utils"
+    import ccv_pss.* from "./ccv_pss"
+
+    run TopNTest = 
+        val valSet =
+            Map("a" -> 5, "a1" -> 5, "a2" -> 5, "b" -> 10, "b1" -> 10, "b2" -> 10, "c" -> 15, "c1" -> 15, "e" -> 25)
```
insumity, line 16: nit: Why `e` is the last one and not `d`?

tests/mbt/model/ccv_pss_test.qnt
```diff
@@ -0,0 +1,34 @@
...
+            assert(GetTopNVals(valSet, 25) == Set("e")),
+            // if one validator with a power is included, all validators with that power need to be included
+            assert(GetTopNVals(valSet, 26) == Set("e", "c1", "c")),
+            assert(GetTopNVals(valSet, 45) == Set("e", "c1", "c")),
+            assert(GetTopNVals(valSet, 55) == Set("e", "c1", "c")),
+            assert(GetTopNVals(valSet, 56) == Set("e", "c1", "c", "b2", "b1", "b")),
+            assert(GetTopNVals(valSet, 85) == Set("e", "c1", "c", "b2", "b1", "b")),
+            assert(GetTopNVals(valSet, 86) == Set("e", "c1", "c", "b2", "b1", "b", "a2", "a1", "a")),
+            assert(GetTopNVals(valSet, 95) == Set("e", "c1", "c", "b2", "b1", "b", "a2", "a1", "a")),
+            assert(GetTopNVals(valSet, 100) == Set("e", "c1", "c", "b2", "b1", "b", "a2", "a1", "a")),
```
insumity, line 32: nit: Why not `GetTopNVals(valSet, 100) == valSet`?

tests/mbt/model/libraries/extraSpells.qnt
insumity: I see. I guess in this case we could adhere to our own standard with just `//`.

insumity (approved):
Great work! Thanks @p-offtermatt.

Would it make sense to merge this under the `feat/partial-set-security` branch?  I'm suggesting this because it might take quite some time until we merge `feat/partial-set-security` back to `main` so it might complicate your upcoming MBT work.

tests/mbt/model/ccv_pss.qnt
p-offtermatt: Yup, exactly!

tests/mbt/model/ccv_pss.qnt
p-offtermatt: Good point. My thought was to use this, at the time the consumer starts, to determine who voted yes on the proposal (and is thus opted in), but maybe I am cutting too many corners here and it is not as clear.
But essentially I think this is an implementation detail and maybe abstracting away like I did is ok, but I would love to know if you disagree

tests/mbt/model/ccv_pss_model.qnt
p-offtermatt: Great point

tests/mbt/model/ccv_pss_test.qnt
p-offtermatt: d was rationalized away at some point and I didn't change this :D
yeah, sorting the map the other way around is also a great point

tests/mbt/model/ccv_pss.qnt
insumity: > at the time the consumer starts, to determine who voted yes on the proposal (and is thus opted in) [...]
Thanks for your answer. This makes sense. We can leave as is :)

tests/mbt/model/ccv_pss_test.qnt
p-offtermatt: Sure. need to be valSet.keys(), because valSets are actually maps.. but still more descriptive :)

p-offtermatt:
> Great work! Thanks @p-offtermatt.
> 
> Would it make sense to merge this under the `feat/partial-set-security` branch? I'm suggesting this because it might take quite some time until we merge `feat/partial-set-security` back to `main` so it might complicate your upcoming MBT work.

Great point - let me retarget this.

p-offtermatt:
> > Great work! Thanks @p-offtermatt.
> > Would it make sense to merge this under the `feat/partial-set-security` branch? I'm suggesting this because it might take quite some time until we merge `feat/partial-set-security` back to `main` so it might complicate your upcoming MBT work.
> 
> Great point - let me retarget this.

That gave a lot of modified files, I think I will rather just make a new branch with the correct branch and cherry-pick.

Issue #1586, opened 2024-01-23
Partial Set Security: Add MsgOptIn & MsgOptOut 


insumity:
## Problem

Add the `MsgOptIn` and `MsgOptOut` messages that would be used for Partial Set Security (see [ADR](https://github.com/cosmos/interchain-security/pull/1584)).

## Closing criteria

1. Investigate on whether we can read the votes of proposals to automatically opt in and if so implement this aspect as well (see Problem Details).
2. State should be updated to keep track of all the validators that opted in (and when they opted in).
3. State should be updated to remove validators when they opt out.
4. Provide keeper methods as outlined in the [ADR](https://github.com/cosmos/interchain-security/pull/1584).
5. Thoroughly test. 

## Problem details

It's not clear whether we can use the votes of a `ConsumerAdditionProposal` to automatically opt in a validator. This is because  the `Tally` method [deletes the votes](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/tally.go#L71) after reading them. This also makes sense; There's no reason to keep the votes around after tallying them. Therefore, we wouldn't be able to extract any information on who voted Yes after a proposal has passed through [GetVotes](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/vote.go#L59).

PR #1620, opened 2024-02-05
feat!: introduce MsgOptIn and MsgOptOut

insumity:
## Description

Partially closes (https://github.com/cosmos/interchain-security/issues/1586).
This PR introduces:
* state to keep track of all the validators that opted in (and when they opted in);
* state keep track of validators that are are about to opt in or opt out, namely `toBeOptIn` and `toBeOptOut` respectively;
* the `MsgOptIn` and `MsgOptOut` messages.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/proto/interchain_security/ccv/provider/v1/tx.proto
+++ b/proto/interchain_security/ccv/provider/v1/tx.proto
@@ -15,6 +15,8 @@ service Msg {
   rpc AssignConsumerKey(MsgAssignConsumerKey) returns (MsgAssignConsumerKeyResponse);
   rpc SubmitConsumerMisbehaviour(MsgSubmitConsumerMisbehaviour) returns (MsgSubmitConsumerMisbehaviourResponse);
   rpc SubmitConsumerDoubleVoting(MsgSubmitConsumerDoubleVoting) returns (MsgSubmitConsumerDoubleVotingResponse);
+  rpc OptIn(MsgOptIn) returns (MsgOptInResponse);
+  rpc OptOut(MsgOptOut) returns (MsgOptOutResponse);
 }
 
 message MsgAssignConsumerKey {
@@ -61,3 +63,28 @@ message MsgSubmitConsumerDoubleVoting {
 }
 
 message MsgSubmitConsumerDoubleVotingResponse {}
+
+message MsgOptIn {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  // the chain id of the consumer chain to opt in to
+  string chain_id = 1;
+  // the validator address on the provider
+  string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+  // the consensus public key to use on the consumer in json string format corresponding to proto-any,
+  // for example `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}`
+  string consumerKey = 3;
+}
+
+message MsgOptInResponse {}
+
+message MsgOptOut {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  // the chain id of the consumer chain to opt out from
+  string chain_id = 1;
+  // the validator address on the provider
+  string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+}
+
+message MsgOptOutResponse {}
--- a/proto/interchain_security/ccv/provider/v1/tx.proto
+++ b/proto/interchain_security/ccv/provider/v1/tx.proto
@@ -71,9 +71,10 @@ message MsgOptIn {
   string chain_id = 1;
   // the validator address on the provider
   string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
-  // the consensus public key to use on the consumer in json string format corresponding to proto-any,
+  // (optional) the consensus public key to use on the consumer in json string format corresponding to proto-any,
   // for example `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}`
-  string consumerKey = 3;
+  // we can set `consumer_key = ""` if we do not consider the `consumer_key`
+  string consumer_key = 3;
 }
 
 message MsgOptInResponse {}
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -1184,6 +1184,7 @@ func (k Keeper) IsOptIn(ctx sdk.Context, chainID string) bool {
 	topN, found := k.GetTopN(ctx, chainID)
 	return !found || topN == 0
 }
+
 func (k Keeper) SetOptedIn(
 	ctx sdk.Context,
 	chainID string,
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -1160,7 +1160,7 @@ func (k Keeper) DeleteTopN(
 	store.Delete(types.TopNKey(chainID))
 }
 
-// GetTopN returns (N, true) if chain `chainID` has a top N associated, and (0, false) otherwise. 
+// GetTopN returns (N, true) if chain `chainID` has a top N associated, and (0, false) otherwise.
 func (k Keeper) GetTopN(
 	ctx sdk.Context,
 	chainID string,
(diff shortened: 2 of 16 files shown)


proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -61,3 +63,29 @@ message MsgSubmitConsumerDoubleVoting {
...
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  // the chain id of the consumer chain to opt in to
+  string chain_id = 1;
+  // the validator address on the provider
+  string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+  // (optional) the consensus public key to use on the consumer in json string format corresponding to proto-any,
+  // for example `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}`
+  // we can set `consumer_key = ""` if we do not consider the `consumer_key`
+  string consumer_key = 3;
```
insumity, line 77: Could not set as `optional` field because:
```
Generating Protobuf files
Generating gogo proto code
Warning: plugin "gocosmos" does not support required features.
  Feature "proto3 optional" is required by 1 file(s):
    interchain_security/ccv/provider/v1/tx.proto
Warning: plugin "grpc-gateway" does not support required features.
  Feature "proto3 optional" is required by 1 file(s):
    interchain_security/ccv/provider/v1/tx.proto
```

x/ccv/provider/keeper/keeper.go
```diff
@@ -1184,3 +1184,141 @@ func (k Keeper) IsOptIn(ctx sdk.Context, chainID string) bool {
...
+func (k Keeper) IsToBeOptedOut(
+	ctx sdk.Context,
+	chainID string,
+	providerAddr types.ProviderConsAddress,
+) bool {
+	store := ctx.KVStore(k.storeKey)
+	return store.Get(types.ToBeOptedOutKey(chainID, providerAddr)) != nil
+}
+
+func (k Keeper) GetToBeOptedOut(
```
p-offtermatt, line 1309: Might just be me, but the modifications in this file seem like tripley duplicated code,
the only thing that really changes is the Key/Preifx. Feel free to ignore this comment, but
it might be worth refactoring to make this more concise (probably at the expense of adding a layer of abstraction)

x/ccv/provider/keeper/msg_server.go
```diff
@@ -174,3 +174,62 @@ func (k msgServer) SubmitConsumerDoubleVoting(goCtx context.Context, msg *types.
...
+	if err != nil {
+		return nil, err
+	}
+
+	if msg.ConsumerKey != "" {
+		k.Keeper.HandleOptIn(ctx, msg.ChainId, providerAddr, &msg.ConsumerKey)
+		ctx.EventManager().EmitEvents(sdk.Events{
+			sdk.NewEvent(
+				ccvtypes.EventTypeOptIn,
+				sdk.NewAttribute(types.AttributeProviderValidatorAddress, msg.ProviderAddr),
```
p-offtermatt, line 195: is there a reason to prefer having an opt in event with both fields? could also emit an optIn and an 	AssignConsumerKey event instead

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -0,0 +1,36 @@
...
+	if k.IsToBeOptedOut(ctx, chainID, providerAddr) {
+		// a validator to be opted in cancels out with a validator to be opted out
+		k.DeleteToBeOptedOut(ctx, chainID, providerAddr)
+	} else if !k.IsToBeOptedIn(ctx, chainID, providerAddr) && !k.IsOptedIn(ctx, chainID, providerAddr) {
+		// a validator can only be set for opt in if it is not opted in and not already set for opt in
+		k.SetToBeOptedIn(ctx, chainID, providerAddr)
+	}
+
+	if consumerKey != nil {
+		// TODO: assign consumer key in this case
```
p-offtermatt, line 24: Hm, I would resolve this TODO in this PR to make sure there aren't any complications that need other refactors

p-offtermatt (approved):
Don't think anything looks problematic, some nits. Great work on keeping it modular :)

x/ccv/provider/keeper/msg_server.go
sainoe: Agree. You could also have the `types.AttributeConsumerConsensusPubKey` of the event be an empty `string`, if a consumer key isn't passed.

x/ccv/provider/keeper/partial_set_security.go
sainoe: This func has to return an error if the consumer isn't either in voting or registered.

The `k.IsConsumerProposedOrRegistered(ctx, chainID)` check can be used. Perhaps the key assignment can be handled in msg_server.go instead.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types.ProviderConsAddress) {
```
sainoe, line 28: Same thing here, it should return an error if the consumer doesn't exist as explained above.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1184,3 +1184,141 @@ func (k Keeper) IsOptIn(ctx sdk.Context, chainID string) bool {
 	topN, found := k.GetTopN(ctx, chainID)
 	return !found || topN == 0
```
sainoe, line 1185: Is it possible to have a chain with TopN set to `0`?

x/ccv/provider/keeper/keeper.go
```diff
@@ -1184,3 +1184,141 @@ func (k Keeper) IsOptIn(ctx sdk.Context, chainID string) bool {
...
 }
+func (k Keeper) SetOptedIn(
```
sainoe, line 1187:
```suggestion

func (k Keeper) SetOptedIn(
```

sainoe:
Great work! See my comments.

x/ccv/provider/keeper/keeper.go
p-offtermatt: I think it exactly indicates opt in chains.
Probably, this function should error on !found, and otherwise return topN == 0

x/ccv/provider/keeper/keeper.go
```diff
@@ -1184,3 +1184,141 @@ func (k Keeper) IsOptIn(ctx sdk.Context, chainID string) bool {
...
+	store.Delete(types.OptedInKey(chainID, providerAddr))
+}
+
+func (k Keeper) IsOptedIn(
+	ctx sdk.Context,
+	chainID string,
+	providerAddr types.ProviderConsAddress,
+) bool {
+	store := ctx.KVStore(k.storeKey)
+	return store.Get(types.OptedInKey(chainID, providerAddr)) != nil
```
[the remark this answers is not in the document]
insumity: The block height would always be smaller or equal to the current block height. It would be equal to the current block height when we first opt in that validator, for instance at the end of an epoch.

x/ccv/provider/keeper/keeper.go
insumity: That's true. At least the `ToBeOptedIn` and `ToBeOptedOut` are pretty much the same but with a different key. Would think if there is a nice way to refactor this by just providing the key in a future iteration.

x/ccv/provider/keeper/msg_server.go
insumity: > could also emit an optIn and an AssignConsumerKey event instead
Emitting an `AssignConsumerKey` event as well might be problematic because a relayer might consider that a `MsgAssignConsumerKey` message was sent independently which is not the case. 

Followed @sainoe approach on passing an empty `string`.

x/ccv/provider/keeper/keeper.go
insumity: > Probably, this function should error on !found, and otherwise return topN == 0
Might augment this function to return an `error` later on.

x/ccv/provider/keeper/key_assignment.go
```diff
@@ -15,6 +16,53 @@ import (
 	ccvtypes "github.com/cosmos/interchain-security/v4/x/ccv/types"
 )
 
+// ParseConsumerKey parses the ED25519 PubKey`consumerKey` from a JSON string
+// and constructs its corresponding `tmprotocrypto.PublicKey`
+func (k Keeper) ParseConsumerKey(consumerKey string) (tmprotocrypto.PublicKey, error) {
```
insumity, line 21: Took this method code _as is_ from `msg_server.go` (see below). I chose not to reuse the `AssignConsumerKey` method in `partial_set_security.go` when opting in a validator with a consumer key because this would entail:
* creating a `MsgAssign` message inside `HandleOptIn` which seems counterintuitive in the sense that a message creates another message internally to do something;
*  `AssignConsumerKey` would emit emit an assignment event and potentially this could confuse relayers becuse for one `MsgOptIn` message, relayers get events as if another message was executed as well.

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -31,6 +37,53 @@ func TestHandleOptIn(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// assert that the consumeKey was assigned to `providerAddr` validator on chain with id `chainID`
+	actualConsumerPubKey, found := providerKeeper.GetValidatorConsumerPubKey(ctx, "chainID", providerAddr)
+	require.True(t, found)
+	require.Equal(t, expectedConsumerPubKey, actualConsumerPubKey)
+
+	// assert that the `consumerAddr` to `providerAddr` association was set as well
+	consumerAddr, _ := ccvtypes.TMCryptoPublicKeyToConsAddr(actualConsumerPubKey)
+	actualProviderConsAddr, found := providerKeeper.GetValidatorByConsumerAddr(ctx, "chainID", types.NewConsumerConsAddress(consumerAddr))
```
p-offtermatt, line 83: we should definitely test that it is possible to assign a key the "normal way" after doing it like this. not sure if here is the right place to do it, or if it is easier with higher-level tests, e.g. e2e

insumity: Conceptually, if we exclude opting in, a `MsgOptIn` (with a consumer key) followed by a `MsgAssign` message is the same as two `MsgAssign`s the one after the other. If we can call twice or more `MsgAssign` today, I don't see why `MsgOptIn` would cause issues with this. 
In any case, we could probably test this in an e2e test.

p-offtermatt:
Changes look good to me, thank you! Good to merge from my side

PR #1629, opened 2024-02-08
feat!: automatically opt in validators that vote Yes on consumer addition proposals

insumity:
## Description

(together with https://github.com/cosmos/interchain-security/pull/1620) closes: #1586

Overwrites the `AfterProposalVote` hook so that we can see whether a validator votes `Yes` on a `ConsumerAdditionProposal`. Validators that vote `Yes` are automatically opted in.

Note that we had to overwrite `AfterProposalVote` because the Tally method [deletes the votes](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/tally.go#L71) after reading them. This also makes sense; There's no reason to keep the votes around after tallying them. Therefore, we wouldn't be able to extract any information on who voted Yes after a proposal has passed through [GetVotes](https://github.com/cosmos/cosmos-sdk/blob/v0.47.7/x/gov/keeper/vote.go#L59).

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/ccv/provider/keeper/hooks.go
+++ b/x/ccv/provider/keeper/hooks.go
@@ -217,6 +217,9 @@ func (h Hooks) AfterProposalVote(ctx sdk.Context, proposalID uint64, voterAddr s
 			return
 		}
 		if !weight.Equal(math.LegacyNewDec(1)) {
+			h.k.Logger(ctx).Error("single vote does not have a weight of 1",
+				"vote", vote,
+			)
 			return
 		}
 
--- a/x/ccv/provider/keeper/hooks.go
+++ b/x/ccv/provider/keeper/hooks.go
@@ -1,8 +1,8 @@
 package keeper
 
 import (
+	"cosmossdk.io/math"
 	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkgov "github.com/cosmos/cosmos-sdk/x/gov/types"
 	v1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
@@ -176,6 +176,40 @@ func (h Hooks) AfterProposalDeposit(ctx sdk.Context, proposalID uint64, deposito
 }
 
 func (h Hooks) AfterProposalVote(ctx sdk.Context, proposalID uint64, voterAddr sdk.AccAddress) {
+	validator, found := h.k.stakingKeeper.GetValidator(ctx, voterAddr.Bytes())
+	if !found {
+		return
+	}
+
+	consAddr, err := validator.GetConsAddr()
+	if err != nil {
+		return
+	}
+
+	chainID := h.k.GetProposedConsumerChain(ctx, proposalID)
+
+	vote, found := h.k.govKeeper.GetVote(ctx, proposalID, voterAddr)
+	if !found {
+		return
+	}
+
+	if len(vote.Options) == 1 && vote.Options[0].Option == v1.VoteOption_VOTE_OPTION_YES {
+		// only consider votes that vote YES with 100% weight
+		weight, err := sdk.NewDecFromStr(vote.Options[0].Weight)
+		if err != nil {
+			return
+		}
+		if !weight.Equal(math.LegacyNewDec(1)) {
+			return
+		}
+
+		// in case validator is already to-be-opted in, we just overwrite
+		h.k.SetToBeOptedIn(ctx, chainID, providertypes.NewProviderConsAddress(consAddr))
+	} else {
+		// if vote is not a YES vote with 100% weight, we delete the validator from to-be-opted in
+		// if the validator is not to-be-opted in, the `DeleteToBeOptedIn` is a no-op
+		h.k.DeleteToBeOptedIn(ctx, chainID, providertypes.NewProviderConsAddress(consAddr))
+	}
 }
 
 func (h Hooks) AfterProposalFailedMinDeposit(ctx sdk.Context, proposalID uint64) {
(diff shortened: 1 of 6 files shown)


x/ccv/provider/keeper/hooks.go
```diff
@@ -175,7 +175,46 @@ func (h Hooks) AfterProposalVotingPeriodEnded(ctx sdk.Context, proposalID uint64
...
+// validator votes multiple times, only the last vote would be considered on whether the validator is opted in or not.
 func (h Hooks) AfterProposalVote(ctx sdk.Context, proposalID uint64, voterAddr sdk.AccAddress) {
+	validator, found := h.k.stakingKeeper.GetValidator(ctx, voterAddr.Bytes())
+	if !found {
+		return
+	}
+
+	consAddr, err := validator.GetConsAddr()
+	if err != nil {
+		return
```
insumity, line 188: Note: These are cases where we could also `panic` instead of simply `return`ing. We could `panic` because those are exceptional cases that should never trigger (e.g., a validator with no consensus address). However, I choose to simply `return` because a potential buggy `panic` call would potentially imply that noone could `MsgVote` on such a proposal.

x/ccv/provider/keeper/hooks.go
```diff
@@ -175,7 +175,61 @@ func (h Hooks) AfterProposalVotingPeriodEnded(ctx sdk.Context, proposalID uint64
...
+			return
+		}
+		if !weight.Equal(math.LegacyNewDec(1)) {
+			h.k.Logger(ctx).Error("single vote does not have a weight of 1",
+				"vote", vote,
+			)
+			return
+		}
+
+		// in case validator is already to-be-opted in, we just overwrite
```
p-offtermatt, line 226: isn't this a noop in case the val is already to be opted in, i.e. not overwriting anything?

x/ccv/provider/keeper/hooks_test.go
```diff
@@ -223,3 +229,53 @@ func TestGetConsumerAdditionLegacyPropFromProp(t *testing.T) {
...
+			v1.Vote{ProposalId: 1, Voter: "voter", Options: options}, true,
+		),
+	)
+
+	// set the validator to-be-opted in first to assert that a NO vote removes the validator from to-be-opted in
+	k.SetToBeOptedIn(ctx, "chainID", providerAddr)
+	require.True(t, k.IsToBeOptedIn(ctx, "chainID", providerAddr))
+	k.Hooks().AfterProposalVote(ctx, 1, sdk.AccAddress{})
+	require.False(t, k.IsToBeOptedIn(ctx, "chainID", providerAddr))
+}
```
p-offtermatt, line 281: Could we add a test with a weighted vote, just to safeguard against potential future regressions?

x/ccv/provider/keeper/hooks.go
```diff
@@ -175,7 +175,61 @@ func (h Hooks) AfterProposalVotingPeriodEnded(ctx sdk.Context, proposalID uint64
...
+	vote, found := h.k.govKeeper.GetVote(ctx, proposalID, voterAddr)
+	if !found {
+		h.k.Logger(ctx).Error("could not find vote for validator",
+			"validator acc addr", voterAddr,
+			"proposalID", proposalID,
+		)
+		return
+	}
+
+	if len(vote.Options) == 1 && vote.Options[0].Option == v1.VoteOption_VOTE_OPTION_YES {
```
p-offtermatt, line 209: Need to be careful here, appearantly it *might* be that votes are created that use the vote.Option field instead of the vote.Options field, see 

```
// Deprecated: Prefer to use `options` instead. This field is set in queries
  // if and only if `len(options) == 1` and that option has weight 1. In all
  // other cases, this field will default to VOTE_OPTION_UNSPECIFIED.
  VoteOption option = 3 [deprecated = true];
```

from https://docs.cosmos.network/v0.47/build/modules/gov#vote

maybe also check for this just in case, it's deprecated but not removed for a reason I guess

x/ccv/provider/keeper/hooks_test.go
p-offtermatt: Also, in case you think https://github.com/cosmos/interchain-security/pull/1629/files#r1482790758 should be taken into account, we also should add test cases for it

p-offtermatt (commented):
Thanks, great work. The small PRs are so pleasant to review!

In this case, I'm concerned about the structure of votes, we might need to take the old, deprecated field into account, see https://github.com/cosmos/interchain-security/pull/1629/files#r1482790758

x/ccv/provider/keeper/hooks.go
insumity: Thanks for pointing this out.
The Governance keeper exposes:
```
GetVote(ctx sdk.Context, proposalID uint64, voterAddr sdk.AccAddress) (vote v1.Vote, found bool)
```
that I use to retrieve a `vote`. There's no `vote.Option` in `v1.Vote`. 
`vote.Option` appears in legacy `v1beta1.MsgVote` that can be used in [`Vote`](https://github.com/cosmos/cosmos-sdk/blob/v0.47.0/x/gov/keeper/msg_server.go#L244). But `Vote` [internally](https://github.com/cosmos/cosmos-sdk/blob/v0.47.0/x/gov/keeper/msg_server.go#L120) would build the `Options` field. So, when we extract the vote, it just has `Options`.

x/ccv/provider/keeper/hooks.go
insumity: Yes, this is a no-op. I rewrote comment so it's similar to the one below:
```
// if the validator is not to-be-opted in, the `DeleteToBeOptedIn` is a no-op
```

x/ccv/provider/keeper/hooks.go
p-offtermatt: Thanks for double-checking!

p-offtermatt (approved):
Thanks for addressing the comments, LGTM

x/ccv/provider/keeper/hooks_test.go
insumity: Added one more test case and made the "No" vote test tabular.

Issue #1588, opened 2024-01-23
Spike: Only opted-in validators should get consumer rewards


insumity:
## Problem

Currently, all validators share the rewards sent over from consumer chains. While moving to Partial Set Security, **only** opted in (and Top N) validators should get consumer rewards.

What happens today is that all rewards are sent to the same pool and the rewards are automatically distributed to validators and delegators through the [distribution module](https://github.com/cosmos/cosmos-sdk/blob/main/x/distribution/keeper/). We cannot reuse the same pool here because the way distribution works, **all** validators would get rewards this way.

## Closing criteria

This is a spike issue. No implementation is necessary. We want to investigate how we can distribute rewards only to opted in (and Top N) validators. Investigation on this can include:
- Should we create multiple pools (one for each consumer chain)? 
- Can we replicate the code of [`AllocateTokens`](https://github.com/cosmos/cosmos-sdk/blob/main/x/distribution/keeper/allocation.go#L17) and use it [here](https://github.com/cosmos/interchain-security/blob/v3.3.0/x/ccv/provider/keeper/distribution.go#L77) instead of sending all the rewards to the same pool?

jtremback:
Seems like [AllocateTokensToValidator](https://github.com/cosmos/cosmos-sdk/blob/main/x/distribution/keeper/allocation.go#L95) is the function to use- are there any issues with it?

insumity:
> Seems like [AllocateTokensToValidator](https://github.com/cosmos/cosmos-sdk/blob/main/x/distribution/keeper/allocation.go#L95) is the function to use- are there any issues with it?

@jtremback the final solution is not yet clear to us. But [AllocateTokensToValidator](https://github.com/cosmos/cosmos-sdk/blob/main/x/distribution/keeper/allocation.go#L95) has a `tokens sdk.DecCoins` parameter so we need a way to estimate the exact `tokens` per validator to distribute the rewards. This is why we suggest to use `AllocateTokens` instead that also takes into account the community tax.

jtremback:
Notes from discussion:
- We need to keep the denom whitelist to avoid DOS
- Perhaps we can add a feature to the permissioned-lite consumer chain addition proposal to add denoms automatically

Issue #1610, opened 2024-01-31
Partial Set Security: Distribute rewards only to opted-in validators


insumity:
## Problem

This issue depends on the "[Spike: Only opted-in validators should get consumer rewards](https://github.com/cosmos/interchain-security/issues/1588)" issue. After we have an idea for a solution, this issue would track the actual implementation where we distribute rewards only to opted in validators (based on the block height the validators opted in).

PR #1632, opened 2024-02-09
feat!: add PSS reward distribution spike

sainoe:
This PR introduces the PSS reward distribution. It brings the necessary logic to allocate token rewards to validators for each consumer separately, according to their opted-in validators. A time condition imposing them to be opted in for a minimum amount of time, e.g. 1000 blocks, will be added subsequently.

This solution doesn't add any new module account to receive IBC token rewards from consumer chains and uses an IBC middleware to track the source chain of IBC transfer packets.  Note that it may cause some issues with the IBC Packet Forwarding Middleware. 

Finally, the PR's changes aren't strictly tested as they require significant changes on the testing suite setup. 

Closes: #1588, #1610


app/provider/app.go
```diff
@@ -470,12 +471,15 @@ func New(
 		app.BankKeeper,
 		scopedTransferKeeper,
 	)
-	transferModule := transfer.NewAppModule(app.TransferKeeper)
-	ibcmodule := transfer.NewIBCModule(app.TransferKeeper)
+
+	// Add a IBC middleware callback to track the consumer rewards
```
p-offtermatt, line 475:
```suggestion
	// Add an IBC middleware callback to track the consumer rewards
```

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -300,3 +302,12 @@ message ConsumerAddrsToPrune {
   uint64 vsc_id = 2;
   AddressList consumer_addrs = 3;
 }
+
+// ConsumerRewardsAllocation is used to serialize the allocation of consumer chain rewards
```
p-offtermatt, line 306: nit: could you expand this comment a bit? I don't understand it as written. it serializes the rewards, given as coin amounts, sent by consumer chains?

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,179 @@
...
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v7/modules/core/05-port/types"
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
+)
+
+var _ porttypes.Middleware = &IBCMiddleware{}
+
+// IBCMiddleware implements the ICS26 callbacks for the fee middleware given the
+// fee keeper and the underlying application.
```
p-offtermatt, line 17: Unless I'm missing something, keeper.Keeper here is the provider keeper, right?
Comment seems a bit misleading to me, maybe clarify this

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,179 @@
...
+	chanCap *capabilitytypes.Capability,
+	counterparty channeltypes.Counterparty,
+	version string,
+) (string, error) {
+	// call underlying app's OnChanOpenInit callback with the appVersion
+	return im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version)
+}
+
+// OnChanOpenTry implements the IBCMiddleware interface
+// If the channel is not fee enabled the underlying application version will be returned
```
p-offtermatt, line 47: I don't get this comment - the function logic seems to just pass the call through.
the comment makes it seem like much more should happen here

x/ccv/provider/types/keys.go
```diff
@@ -161,6 +161,7 @@ const (
 	ToBeOptedOutBytePrefix
 
 	// NOTE: DO NOT ADD NEW BYTE PREFIXES HERE WITHOUT ADDING THEM TO getAllKeyPrefixes() IN keys_test.go
+	ConsumerRewardPoolByteBytePrefix
```
p-offtermatt, line 164: nit: Move this before the comment for consistency, and add a comment as explanation like for the other definitions

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,165 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+// AllocateTokensToConsumerValidators allocates the consumer rewards pool to validator
+// according to their voting power
+func (k Keeper) AllocateTokensToConsumerValidators(
+	ctx sdk.Context,
+	chainID string,
+	totalPreviousPower int64,
+	bondedVotes []abci.VoteInfo,
+	tokens sdk.DecCoins,
+) sdk.DecCoins {
+	optedInVals := map[string]struct{}{}
```
p-offtermatt, line 136: Could we just get rid of the whole "checking for opted-in" logic here? I know it's not the final form, but what's there right now is weird to me.
1) only opted in validators could even vote, so I think the whole check whether vals are opted in right now seems unnecessary
2) it probably shouldn't matter whether a val is opted in *right now* - e.g. this behaves weirdly when a validator opts out, but then rewards for a block come in where the val was still validating.

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,165 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
 	return consumerRewardDenoms
 }
 
-// TransferRewardsToFeeCollector transfers all consumer rewards to the fee collector address
-func (k Keeper) TransferRewardsToFeeCollector(ctx sdk.Context) {
-	// 1. Get the denom whitelist from the store
-	denoms := k.GetAllConsumerRewardDenoms(ctx)
+// AllocateTokens performs reward and fee distribution to all validators based
+// on the Partial Set Security distribution specification.
```
p-offtermatt, line 72: maybe comment on what happens to the remaining funds that were not allocated to a validator
(from reading the code, I think they are just sent to the community pool)

p-offtermatt (approved):

p-offtermatt (commented):
Thanks for the work!
I left a couple of comments, mostly asking for clarifications to some comments.
Apart from that, we should just be careful to make sure the placeholder token allocation logic for validators is revamped - I think what's there right now is probably ok as a placeholder

p-offtermatt (approved):
Le

proto/interchain_security/ccv/provider/v1/provider.proto
sainoe: Thanks for pointing this out. That's correct. It's used to track how many tokens were allocated to a consumer chain to the consumer rewards pool (module account).

x/ccv/provider/keeper/distribution.go
sainoe: Works for me! Will add a TODO instead.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+// OnRecvPacket executes the IBC transfer and records the coins received
+// from a consumer chain by updating its consumer rewards states
+func (im IBCMiddleware) OnRecvPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	relayer sdk.AccAddress,
+) exported.Acknowledgement {
+	// TODO: ensure it's a consumer chain!
+	chainID, err := im.keeper.GetPacketSourceChain(ctx, packet)
```
jtremback, line 109: As we discussed, this will make it impossible for chains to send e.g. native Noble USDC as rewards, because this requires unwinding the denom by sending it through Noble because then it will be coming from Noble. Might be ok for the first version anyway.

An alternative would be to use the memo or any other field that the sending chain can write to to identify the source, which would be durable when passed through Noble's PFM

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return channeltypes.NewErrorAcknowledgement(err)
+	}
+
+	consuRewardsPool := im.keeper.GetConsumerRewardsPool(ctx)
+
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
```
jtremback, line 116: Add comment about how you are executing the transfer and then looking at the change below

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,159 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+		//
+		// Ref: https://github.com/cosmos/cosmos-sdk/issues/2525#issuecomment-430838701
+		powerFraction := math.LegacyNewDec(vote.Validator.Power).QuoTruncate(math.LegacyNewDec(totalPreviousPower))
+		tokensFraction := tokens.MulDecTruncate(powerFraction)
+
+		k.distributionKeeper.AllocateTokensToValidator(
 			ctx,
-			k.accountKeeper.GetModuleAccount(ctx, types.ConsumerRewardsPool).GetAddress(),
-			denom,
+			k.stakingKeeper.ValidatorByConsAddr(ctx, consAddr),
```
jtremback, line 151: Consider tweaking the validator's commission here if it is possible. This would involve storing a mapping of validator/consumer chain -> commission, so that validators could have custom commissions per consumer chain. This would be an excellent feature to add.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+	if ack.Success() {
+		// Calculate coin received
+		coinsReceived := im.keeper.GetConsumerRewardsPool(ctx).Sub(consuRewardsPool...)
```
jtremback, line 119: Look into whether a million fake denoms going through this n^2 .Sub method could DOS a relayer

jtremback: Should we be consulting the whitelist here?

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// Add coins to consumer reward
+		alloc := im.keeper.GetConsumerRewardsAllocation(ctx, chainID)
+		alloc.Rewards = alloc.Rewards.Add(sdk.NewDecCoinsFromCoins(coinsReceived...)...)
```
jtremback, line 123: Or checking the whitelist here?

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,159 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+}
+
+// TransferConsumerRewardsToDistributionModule transfers the collected rewards of the given consumer chain
+// from the consumer rewards pool module account to a the distribution module
+func (k Keeper) TransferConsumerRewardsToDistributionModule(
+	ctx sdk.Context,
+	chainID string,
+) (sdk.Coins, error) {
+	// Get coins of the consumer reward pool
+	pool := k.GetConsumerRewardsAllocation(ctx, chainID)
```
jtremback, line 167: You could have a million fake denoms not in the whitelist at this point

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,159 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+	return coinsToSend, nil
+}
+
+// consumer reward pools getter and setter
+
+// GetConsumerRewardsAllocation the onsumer rewards allocation for the given chain ID
+func (k Keeper) GetConsumerRewardsAllocation(ctx sdk.Context, chainID string) (pool types.ConsumerRewardsAllocation) {
+	store := ctx.KVStore(k.storeKey)
+	b := store.Get(types.ConsumerRewardsAllocationKey(chainID))
+	k.cdc.MustUnmarshal(b, &pool)
```
jtremback, line 208: Million fake denoms clogging up your unmarshal method

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,159 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
-				ctx,
-				types.ConsumerRewardsPool,
-				k.feeCollectorName,
-				sdk.NewCoins(balance),
-			)
-			if err != nil {
-				k.Logger(ctx).Error("cannot sent consumer rewards to fee collector:", "reward", balance.String())
-			}
+	// Truncate coin rewards
+	rewards, _ := pool.Rewards.TruncateDecimal()
```
jtremback, line 170: Iterating over a million fake denoms here

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,159 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+
+	// Extract the whitelisted denoms
+	coinsToSend := sdk.Coins{}
+	denoms := k.GetAllConsumerRewardDenoms(ctx)
+	for _, denom := range denoms {
+		if found, coin := rewards.Find(denom); found {
```
jtremback, line 176: Million fake denoms being iterated by .Find

jtremback (commented):
LGTM except for a few things:

- This is currently DOSable. Filter against the whitelist at the earliest point, or use a chainId/Denom composite key to avoid handling arrays with a million denoms in them

- Look into the possibility of implementing the per-consumer commission because this would be really awesome to have!

testutil/integration/__debug_bin2941279796
mpoke, line 1: Remove this file

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,149 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+	totalReward := sdk.DecCoins{}
+	for _, vote := range bondedVotes {
+		// TODO: should check if validator IsOptIn or continue here
+		consAddr := sdk.ConsAddress(vote.Validator.Address)
+
+		// TODO: Consider micro-slashing for missing votes.
+		//
+		// Ref: https://github.com/cosmos/cosmos-sdk/issues/2525#issuecomment-430838701
+		powerFraction := math.LegacyNewDec(vote.Validator.Power).QuoTruncate(math.LegacyNewDec(totalPreviousPower))
+		tokensFraction := tokens.MulDecTruncate(powerFraction)
```
mpoke, line 147: The amount of rewards should be computed based on the amount of power on the consumer chain. Here, the `totalPreviousPower` is for the provider.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+		// check if the packet sender is a consumer chain
+		chainID, err := im.keeper.IdentifyConsumerChainIDFromIBCPacket(ctx, packet)
+		if err != nil {
+			return ack
+		}
+
+		// At this point the transfer was successful,
+		// so the IBC packet can be safely deserialized
+		var data ibctransfertypes.FungibleTokenPacketData
+		_ = types.ModuleCdc.UnmarshalJSON(packet.GetData(), &data)
```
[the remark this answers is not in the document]
sainoe: In `ack.Success()` if condition we know that the IBC transfer succeeded so the packet data can't be deserialized without checking errors.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+
+// OnChanOpenAck implements the IBCMiddleware interface
+func (im IBCMiddleware) OnChanOpenAck(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	counterpartyChannelID string,
+	counterpartyVersion string,
+) error {
+
```
insumity, line 71: nit: remove new line

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+	return im.app.OnChanOpenConfirm(ctx, portID, channelID)
+}
+
+// OnChanCloseInit implements the IBCMiddleware interface
+func (im IBCMiddleware) OnChanCloseInit(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+) error {
+	return im.app.OnChanCloseInit(ctx, portID, channelID)
```
insumity, line 92: nit: not needed but consistency-wise, this method and ` OnChanOpenTry` do not have a comment before `return`

x/ccv/provider/keeper/keeper.go
```diff
@@ -1324,3 +1324,25 @@ func (k Keeper) GetToBeOptedOut(
...
+	}
+	connectionID := channel.ConnectionHops[0]
+	_, tmClient, err := k.getUnderlyingClient(ctx, connectionID)
+	if err != nil {
+		return "", err
+	}
+
+	chainID := tmClient.ChainId
+	if _, ok := k.GetChainToChannel(ctx, chainID); !ok {
+		return "", errorsmod.Wrapf(channeltypes.ErrTooManyConnectionHops, "no CCV channel found for chain with ID: %s", chainID)
```
insumity, line 1344: Probably the error here should not be `ErrTooManyConnectionHops`.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+		_ = types.ModuleCdc.UnmarshalJSON(packet.GetData(), &data)
+
+		// check if the recipient is the consumer reward's pool address
+		receiver, _ := sdk.AccAddressFromBech32(data.Receiver)
+		if receiver.String() != im.keeper.GetConsumerRewardsPoolAddressStr(ctx) {
+			return ack
+		}
+
+		coinDenom := getProviderDenom(data.Denom, packet)
+		coinAmt, _ := math.NewIntFromString(data.Amount)
```
insumity, line 135: In all of the above where we do not check on the returned `err`or, would it make sense to still have a condition like:
```
if err != nil {
    log an error
}
```
?

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+
+		// Iterate over the whitelisted consumer denoms
+		for _, denom := range im.keeper.GetAllConsumerRewardDenoms(ctx) {
```
insumity, line 138: Could we use `ConsumerRewardDenomExists` here instead of looping through all denoms?

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+func (im IBCMiddleware) SendPacket(
+	sdk.Context,
+	*capabilitytypes.Capability,
+	string,
+	string,
+	clienttypes.Height,
+	uint64,
+	[]byte,
+) (uint64, error) {
+	return 0, nil
```
insumity, line 190: Why not have `return im.SendPacket(...)` here instead of `return 0, nil`?

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+	return nil
+}
+
+// GetAppVersion returns the application version of the underlying application
+func (im IBCMiddleware) GetAppVersion(ctx sdk.Context, portID, channelID string) (string, bool) {
+	return "", false
+}
+
+// getProviderDenom returns the updated given denom according to the given IBC packet
+// It follows the same logic than the OnRecvPacket method of the IBC transfer module
```
insumity, line 209:
```suggestion
// It follows the same logic as the OnRecvPacket method of the IBC transfer module
```

x/ccv/provider/keeper/keeper.go
```diff
@@ -1324,3 +1324,25 @@ func (k Keeper) GetToBeOptedOut(
...
+		return "", errorsmod.Wrap(channeltypes.ErrTooManyConnectionHops, "must have direct connection to consumer chain")
+	}
+	connectionID := channel.ConnectionHops[0]
+	_, tmClient, err := k.getUnderlyingClient(ctx, connectionID)
+	if err != nil {
+		return "", err
+	}
+
+	chainID := tmClient.ChainId
+	if _, ok := k.GetChainToChannel(ctx, chainID); !ok {
```
insumity, line 1343: It might make sense to have another check here that the channel returned by `k.GetChainToChannel` is the same as the channel that was used to send the packet?
It seems to me that with the current approach someone could send an IBC packet from a non-consumer chain that happens to have the same `chainID` as a consumer chain and this would still return a `chainID`.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+// see https://github.com/cosmos/ibc-go/blob/v7.3.2/modules/apps/transfer/keeper/relay.go#L162
+func getProviderDenom(denom string, packet channeltypes.Packet) (providerDenom string) {
+	// If the the prefix denom corresponds to the provider transfer channel info,
```
insumity, line 212:
```suggestion
	// If the prefix denom corresponds to the provider transfer channel info,
```

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,149 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+		return sdk.Coins{}, err
+	}
+
+	k.SetConsumerRewardsAllocation(ctx, chainID, allocation)
+	return rewardsToSend, nil
+}
+
+// consumer reward pools getter and setter
+
+// GetConsumerRewardsAllocation the onsumer rewards allocation for the given chain ID
```
insumity, line 194:
```suggestion
// GetConsumerRewardsAllocation returns the consumer rewards allocation for the given chain ID
```

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,149 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+	totalPreviousPower int64,
+	bondedVotes []abci.VoteInfo,
+	tokens sdk.DecCoins,
+) sdk.DecCoins {
+	totalReward := sdk.DecCoins{}
+	for _, vote := range bondedVotes {
+		// TODO: should check if validator IsOptIn or continue here
+		consAddr := sdk.ConsAddress(vote.Validator.Address)
+
+		// TODO: Consider micro-slashing for missing votes.
```
insumity, line 143: remove this TODO ... I guess it's not something we intend to do.

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,149 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+	// return if there is no coins in the consumer rewards pool
+	if k.GetConsumerRewardsPool(ctx).IsZero() {
+		return
+	}
+
+	// Iterate over all registered consumer chains
+	for _, consumer := range k.GetAllConsumerChains(ctx) {
+
+		// transfer the consumer rewards to the distribution module account
+		// note that the rewards transferred are only consumer whitelisted denoms
```
insumity, line 83: Are they "whitelisted" because of the whitelisting that took part in IBC's middleware `OnRecvPacket`?

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,149 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+		rewardsCollected, err := k.TransferConsumerRewardsToDistributionModule(ctx, consumer.ChainId)
```
insumity, line 84: I feel I'm missing something here: Why do we send the rewards to the distribution module in the first place? Aren't we already distributing the rewards to validators & delegators through `AllocateTokensToValidator`?

x/ccv/provider/ibc_middleware.go
sainoe: Great catch thanks!

x/ccv/provider/ibc_middleware.go
sainoe: `return im.SendPacket(...) ` creates an infinite recursive call...
AFAIU in IBC transfer app module doesn't implement `SendPacket` and sends packet using [ics4wrapper](https://github.com/cosmos/ibc-go/blob/v7.3.2/modules/apps/transfer/keeper/relay.go#L133) instead.

x/ccv/provider/keeper/distribution.go
sainoe: That's correct.

x/ccv/provider/keeper/distribution.go
sainoe: `AllocateTokensToValidator` function wraps the distribution module's `AllocateTokensToValidator` function that allocates the rewards from the distribution module account to the validators.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1324,3 +1324,25 @@ func (k Keeper) GetToBeOptedOut(
...
+	}
+	if len(channel.ConnectionHops) != 1 {
+		return "", errorsmod.Wrap(channeltypes.ErrTooManyConnectionHops, "must have direct connection to consumer chain")
+	}
+	connectionID := channel.ConnectionHops[0]
+	_, tmClient, err := k.getUnderlyingClient(ctx, connectionID)
+	if err != nil {
+		return "", err
+	}
+
```
sainoe, line 1341: Comment that it's a CCV channel

tests/integration/distribution.go
```diff
@@ -103,13 +112,29 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
...
+
+	// transfer rewards from consumer to provider
+	relayAllCommittedPackets(s, s.consumerChain, s.transferPath, transfertypes.PortID, s.transferPath.EndpointA.ChannelID, 1)
+
+	// check that the consumer rewards allocation are empty since relayAllCommittedPackets call BeginBlock
+	rewardsAlloc := s.providerApp.GetProviderKeeper().GetConsumerRewardsAllocation(s.providerCtx(), s.consumerChain.ChainID)
+	s.Require().True(rewardsAlloc.Rewards.Empty())
 
-	// Check that the reward pool has no more coins because they were transferred to the fee pool
+	// Check that the reward pool still have the first coins transferred that were never allocated
```
insumity, line 130:
```suggestion
	// Check that the reward pool still has the first coins transferred that were never allocated
```

tests/integration/distribution.go
```diff
@@ -103,13 +112,29 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
...
+	s.Require().NoError(err)
+
+	// pass two blocks
+	s.consumerChain.NextBlock()
+	s.consumerChain.NextBlock()
+
+	// transfer rewards from consumer to provider
+	relayAllCommittedPackets(s, s.consumerChain, s.transferPath, transfertypes.PortID, s.transferPath.EndpointA.ChannelID, 1)
+
+	// check that the consumer rewards allocation are empty since relayAllCommittedPackets call BeginBlock
```
insumity, line 126: Not sure I fully understood this comment. It seems to me that the consumer rewards allocation should be empty because we did not have a `OnRecvPacket` yet. Why does `BeginBlock` come into play?

tests/integration/distribution.go
```diff
@@ -103,13 +112,29 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
...
+	rewardsAlloc := s.providerApp.GetProviderKeeper().GetConsumerRewardsAllocation(s.providerCtx(), s.consumerChain.ChainID)
+	s.Require().True(rewardsAlloc.Rewards.Empty())
 
-	// Check that the reward pool has no more coins because they were transferred to the fee pool
+	// Check that the reward pool still have the first coins transferred that were never allocated
 	rewardCoins = providerBankKeeper.GetAllBalances(s.providerCtx(), rewardPool)
-	s.Require().Equal(0, len(rewardCoins))
+	s.Require().True(rewardCoins[ibcCoinIndex].Amount.Equal(providerExpectedRewards[0].Amount))
```
insumity, line 132: Couldn't we use `s.Require.Equal(providerExpectedRewards[0].Amount, ...` here?

tests/integration/distribution.go
```diff
@@ -454,6 +479,398 @@ func (s *CCVTestSuite) TestSendRewardsToProvider() {
...
+				data.Receiver = "cosmos149lw9fktlqfed3zt8ah48r5czmsug5s7kw77u9" // random acct address
+				packet.Data = data.GetBytes()
+			},
+			false,
+			false,
+		},
+		{
+			"IBC Transfer coin denom isn't registered",
+			func(ctx sdk.Context, keeper *providerkeeper.Keeper, bankKeeper icstestingutils.TestBankKeeper) {
+				fmt.Println("middleware", getIBCDenom(packet.DestinationPort, packet.DestinationChannel))
```
insumity, line 526: nit:
```suggestion
```

tests/integration/distribution.go
```diff
@@ -454,6 +479,398 @@ func (s *CCVTestSuite) TestSendRewardsToProvider() {
...
+					s.providerCtx(),
+					getIBCDenom(packet.DestinationPort, packet.DestinationChannel),
+				)
+			},
+			false,
+			false,
+		},
+		{
+			"successful token transfer to empty pool",
+			func(ctx sdk.Context, keeper *providerkeeper.Keeper, bankKeeper icstestingutils.TestBankKeeper) {
```
insumity, line 537: The `setup` of this test is the same as the one from the test above `"IBC Transfer coin denom isn't registered",` so it's hard for me to understand while `rewardsAllocated` is `true` in this test but `false` in the other one.

tests/integration/distribution.go
```diff
@@ -454,6 +479,398 @@ func (s *CCVTestSuite) TestSendRewardsToProvider() {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		s.Run(tc.name, func() {
+			s.SetupTest()
+			s.SetupCCVChannel(s.path)
+			s.SetupTransferChannel()
+
+			providerkeeper := s.providerApp.GetProviderKeeper()
```
insumity, line 581:
```suggestion
			providerKeeper := s.providerApp.GetProviderKeeper()
			bankKeeper := s.providerApp.GetTestBankKeeper()
```

tests/integration/distribution.go
```diff
@@ -454,6 +479,398 @@ func (s *CCVTestSuite) TestSendRewardsToProvider() {
...
+				s.Require().True(ack.Success())
+				// verify that the consumer rewards pool received the IBC coins
+				s.Require().True(rewardsTransferred.IsEqual(sdk.Coins{expRewards}))
+
+				if tc.rewardsAllocated {
+					// check the data receiver address is set to the consumer rewards pool address
+					s.Require().Equal(data.GetReceiver(), providerkeeper.GetConsumerRewardsPoolAddressStr(s.providerCtx()))
+
+					// verify that consumer rewards allocation is updated
+					s.Require().True(rewardsAllocated.IsEqual(sdk.NewDecCoinsFromCoins(expRewards)))
```
insumity, line 658: nit: can we use `Equal` here ... similarly below

tests/integration/distribution.go
sainoe: `relayAllCommittedPackets` relays the IBC packets and hence executes successively the IBC transfer middleware and BeginBlock. Therefore, the provider module's `AllocateTokens` logic is executed and the consumer allocations are emptied.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+	// deserialized without checking errors.
+	if ack.Success() {
+		// extract the coin info received from the packet data
+		var data ibctransfertypes.FungibleTokenPacketData
+		_ = types.ModuleCdc.UnmarshalJSON(packet.GetData(), &data)
+
+		// check if the recipient is the consumer reward's pool address
+		receiver, _ := sdk.AccAddressFromBech32(data.Receiver)
+		if receiver.String() != im.keeper.GetConsumerRewardsPoolAddressStr(ctx) {
+			return ack
```
jtremback, line 135: I'm curious about these `return ack`s- how do they work in the context of middleware?

I'm assuming that when you do this is just passes control flow out of the middleware, so if the packet is going somewhere else it can?

sainoe: These `ack`s are returned directly to [the IBC core RecvPacket handler](https://github.com/cosmos/ibc-go/blob/v7.3.1/modules/core/keeper/msg_server.go#L453), since in our context, we've only one middleware wrapping the transfer module.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+
+		coinAmt, _ := math.NewIntFromString(data.Amount)
+		coinDenom := GetProviderDenom(data.Denom, packet)
+
+		// verify that the coin's denom is a whitelisted consumer denom,
+		// and if so, adds it to the consumer chain rewards allocation
+		if im.keeper.ConsumerRewardDenomExists(ctx, coinDenom) {
```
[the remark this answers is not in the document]
sainoe: Yep, that's correct.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,240 @@
...
+func (im IBCMiddleware) SendPacket(
+	sdk.Context,
+	*capabilitytypes.Capability,
+	string,
+	string,
+	clienttypes.Height,
+	uint64,
+	[]byte,
+) (uint64, error) {
+	return 0, nil
```
mpoke, line 191: Panic instead. This middleware doesn't have an ICS4 Wrapper. This means that none of the `SendPacket`, `WriteAcknowledgement`, and `GetAppVersion` methods are expected to go through the middleware. If that happens, something is clearly wrong (probably with the wiring) and we should panic.

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,240 @@
...
+	// executes the IBC transfer OnRecv logic
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
+
+	// execute the middleware logic only if the sender is a consumer chain
+	consumerID, err := im.keeper.IdentifyConsumerChainIDFromIBCPacket(ctx, packet)
+	if err != nil {
+		return ack
+	}
+
+	// Note that inside the above if condition statement,
```
mpoke, line 123:
```suggestion
	// Note that inside the below if condition statement,
```

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +70,201 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+		if rewardsCollected.IsZero() {
+			continue
+		}
+
+		rewardsCollectedDec := sdk.NewDecCoinsFromCoins(rewardsCollected...)
+
+		// temporary workaround to keep CanWithdrawInvariant happy
+		// general discussions here: https://github.com/cosmos/cosmos-sdk/issues/2906#issuecomment-441867634
+		feePool := k.distributionKeeper.GetFeePool(ctx)
+		if totalPreviousPower == 0 {
```
mpoke, line 104: This `totalPreviousPower` should be computed using `ComputeConsumerTotalVotingPower`.

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +70,201 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
+	return k.bankKeeper.GetAllBalances(
+		ctx,
+		k.accountKeeper.GetModuleAccount(ctx, types.ConsumerRewardsPool).GetAddress(),
+	)
+}
+
+// ComputeConsumerTotalVotingPower returns the total voting power for a given consumer chain
+// by summing its opted-in validators votes
+func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string, votes []abci.VoteInfo) int64 {
```
mpoke, line 232: This method doesn't need the votes but rather the consumer valset. For now leave it like this, but it needs to be updated once we have the consumer valsets in store.

mpoke (approved):
LGTM. Check my comments below.

Issue #1631, opened 2024-02-08
MBT: Add driver logic for PSS


p-offtermatt:
## Problem

Prerequisite: https://github.com/cosmos/interchain-security/issues/1574

We need to adjust the driver for handling traces generated by the PSS Quint model.
This needs some adjustments in the core driver logic.

## Closing criteria

MBT tests pass on the PSS feature branch.

PR #1636, opened 2024-02-09
test: Add driver for PSS

p-offtermatt:
## Description

Closes: #1631 

* Handles new action types OptIn and OptOut
* Handles the new topN field when adding consumer chains

Some things I decided against doing:
* I think comparing the sets of opted in validators between model and system is probably unnecessary, because those will become the validator sets on the consumer chain anyways, so if there is an issue with who is opted in, this will be detected when the validator sets are eventually different.
* Starting the chain with an initial validator set different from the replicated security valset will come later, since I will want to reuse the function to compute the partial set from the keeper

---

### Author Checklist

I have...

- [x] Confirmed all CI checks have passed (mbt test is expected to fail, since PSS behaviour is not yet completely implemented)

### Reviewers Checklist

I have...

### DIFF
--- a/tests/mbt/driver/core.go
+++ b/tests/mbt/driver/core.go
@@ -20,6 +20,7 @@ import (
 	abcitypes "github.com/cometbft/cometbft/abci/types"
 	cmttypes "github.com/cometbft/cometbft/types"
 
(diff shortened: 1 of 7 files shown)


sainoe (approved):
LGTM.
Approving without the full context since the changes are small and only touch testing.

Issue #1606, opened 2024-01-31
Partial Set Security: Construct and send the partial sets to the consumers


insumity:
## Problem

Construct the partial sets that are stored in the `VSCPacket`s that are queued and send to the consumer chain. If the epoch ADR has been implemented, we should combine it with [epochs](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-014-epochs.md).

## Closing criteria

1. Make sure the partial set is also initialized correctly.
2. Thoroughly test

p-offtermatt:
Hey, I think this is done and should be closed/moved to done in the board, right?

PR #1642, opened 2024-02-13
feat!: compute partial set

insumity:
## Description

Closes: https://github.com/cosmos/interchain-security/issues/1606

This PR introduces the logic needed to compute the `ValidatorUpdate`s in `ComputeValidatorUpdates`. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,336 @@
...
+			PubKey: pubKey,
+			Power:  0,
+		}
+
+	}
+
+	var out []abci.ValidatorUpdate
+	for _, update := range m {
+		out = append(out, update)
+	}
```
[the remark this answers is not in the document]
insumity: We're sorting afterwards so we've determinism.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -139,19 +139,30 @@ func (k msgServer) SubmitConsumerDoubleVoting(goCtx context.Context, msg *types.
 func (k msgServer) OptIn(goCtx context.Context, msg *types.MsgOptIn) (*types.MsgOptInResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	valAddress, err := sdk.ConsAddressFromBech32(msg.ProviderAddr)
+	providerValidatorAddr, err := sdk.ValAddressFromBech32(msg.ProviderAddr)
```
insumity, line 142: Fixed this. Similarly to `AssignConsumerKey` we consider that `ProviderAddr` in the message is the operator address and use this operator address to get the consensus address that is used later on.
We made a similar change in `OptOut`.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1190,14 +1190,17 @@ func (k Keeper) SetOptedIn(
 	chainID string,
 	providerAddr types.ProviderConsAddress,
 	blockHeight uint64,
+	power uint64,
```
insumity, line 1193: Introduce `Power` in the opted-in state so we can see if we should send a `ValidatorUpdate` down to CometBFT based on whether the power changed since last time it was sent. Only makes sense in an epoch-based setting because in a non-epoch-based setting, we could use the validator updates from the `staking` module.

x/ccv/provider/keeper/proposal.go
```diff
@@ -255,6 +255,12 @@ func (k Keeper) MakeConsumerGenesis(
 		return false
 	})
 
+	// currentValidators := k.GetOptedIn(ctx, chainID)
```
insumity, line 258: Leaving those commented for now to show how `ComputeValidatorUpdates`, etc. are going to be used.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,171 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+}
+
+// ResetCurrentValidators resets the opted-in validators with the newest set that was computed by
+// `ComputePartialSetValidatorUpdates` and hence this method should only be called after
+// `ComputePartialSetValidatorUpdates` has completed. Method also clears all the `ToBeOptedIn` and `ToBeOptedOut` states.
+func (k Keeper) ResetCurrentValidators(ctx sdk.Context, chainID string, nextValidators []OptedInValidator) {
+	k.DeleteAllOptedIn(ctx, chainID)
+	for _, val := range nextValidators {
+		k.SetOptedIn(ctx, chainID, val.ProviderAddr, val.BlockHeight, val.Power)
+	}
```
mpoke, line 240: Seems very inneficient to delete all opten in entries and set the new ones. If in an epoch there will be no changes in the voting powers, this will be a very expensive way to leave the state unchanged.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1190,14 +1190,17 @@ func (k Keeper) SetOptedIn(
...
 
-	// validator is considered opted in
 	blockHeightBytes := make([]byte, 8)
 	binary.BigEndian.PutUint64(blockHeightBytes, blockHeight)
 
-	store.Set(types.OptedInKey(chainID, providerAddr), blockHeightBytes)
+	powerBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(powerBytes, power)
+
+	store.Set(types.OptedInKey(chainID, providerAddr), append(blockHeightBytes, powerBytes...))
```
mpoke, line 1203: Store `OptedInValidator` as a value (define it in the proto files).

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,171 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+// getValAddressAndPublicKey is a helper function that returns the `ValAddress` and the public key of
+// the corresponding validator
+func (k Keeper) getValAddressAndPublicKey(ctx sdk.Context, addr types.ProviderConsAddress,
+) (sdk.ValAddress, tmprotocrypto.PublicKey, error) {
+	validator, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, addr.ToSdkConsAddr())
+	if !found {
+		return sdk.ValAddress{}, tmprotocrypto.PublicKey{}, stakingtypes.ErrNoValidatorFound
+	}
+
+	consAddr, err := validator.GetConsAddr()
```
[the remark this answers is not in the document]
insumity: Removed `getValAddressAndPublicKey`.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,171 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
 
 	return nil
 }
+
+// getValAddressAndPublicKey is a helper function that returns the `ValAddress` and the public key of
+// the corresponding validator
+func (k Keeper) getValAddressAndPublicKey(ctx sdk.Context, addr types.ProviderConsAddress,
```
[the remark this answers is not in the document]
insumity: Good point. Removed `getValAddressAndPublicKey`.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,171 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+			continue
+		}
+
+		// if the voting power did not change since the last time the validator opted in, do not create an update
+		if val.Power == uint64(k.stakingKeeper.GetLastValidatorPower(ctx, validator.GetOperator())) {
+			continue
+		}
+
+		// if `val` has unbonded, its `GetLastValidatorPower` power returns 0
+		m[pubKey.String()] = abci.ValidatorUpdate{
```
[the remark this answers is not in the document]
insumity: I could use `ProviderAddr` but I need the `pubKey` in any case to build the `ValidatorUpdate`, so it's not that I get the `pubKey` and only use it as a key for the map `m`.
Do you recommend using `ProviderAddr` for a specific reason?

x/ccv/provider/keeper/partial_set_security.go
insumity: Is my understanding correct that at this stage, nothing is written to disk? Only if the transactions succeed, at the end all the changes would get written. If that's the case, why's this expensive? Wouldn't this correspond to a simple write in memory?
I agree that most likely the validator set won't change between epochs but probably it's unlikely that powers would remain the same between epochs for all validators, although this depends on how long an epoch is. Even in this case, I would need to loop through all currently opted in validators and compare with the `nextValidators` to see if a change took place.

x/ccv/provider/keeper/keeper.go
insumity: Is the recommendation here to define it in proto to simplify our life later on when we implement queries that return opted-in validators?

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +74,171 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+		}
+		return out[i].PubKey.String() > out[j].PubKey.String()
+	})
+
+	return out
+}
+
+// ComputeNextValidators computes the next validator set that is responsible for validating on a consumer chain.
+// The returned opted-in validators correspond to the next `currentValidators`.
+func (k Keeper) ComputeNextValidators(ctx sdk.Context,
```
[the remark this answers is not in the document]
insumity: Yes, that's true. I believe that splitting this into two distinct methods makes it easier to read because the `ValidateUpdate`s do not have a 1-to-1 correspodence with the next `OptedInValidator`s.
But I don't have strong feelings on this and we can combine them as well.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +72,152 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
 
 	return nil
 }
+
+// ComputeValidatorUpdates computes the validator updates needed to be sent to the consumer chain to capture
+// the newly opted-in and opted-out validators, as well as validators that unbonded.
+func (k Keeper) ComputeValidatorUpdates(ctx sdk.Context,
```
p-offtermatt, line 78: I feel like I am missing context so my judgement may not be accurate, but it feels like one of ComputeValidatorUpdates and ComputeNextValidators is unnecessary?
Wouldn't NextValidators + a diffing method like Jehan suggested a bit ago do the trick?

x/ccv/provider/keeper/keeper.go
```diff
@@ -1209,6 +1212,19 @@ func (k Keeper) DeleteOptedIn(
 	store.Delete(types.OptedInKey(chainID, providerAddr))
 }
 
+func (k Keeper) DeleteAllOptedIn(
```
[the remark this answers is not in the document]
insumity: Added them although they feel slightly redundant.

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: might be premature optimization to worry about this? especially once we have epochs this happens rarely and it's at most ~200 entries being deleted and written per consumer

p-offtermatt (commented):
Added a few comments. Having both ComputeValidatorUpdates and ComputeNextValidators feels weird to me - maybe we could reconsider the idea of taking diffs between new and old validator sets here

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +72,152 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+		if !found {
+			continue
+		}
+
+		pubKey, err := validator.TmConsPublicKey()
+		if err != nil {
+			continue
+		}
+
+		// if the voting power did not change since the last time the validator opted in, do not create an update
```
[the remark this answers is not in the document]
insumity: The comment should probably be rephrased to:
```
// if the voting power is the same as the one the validator had when it opted in, do not create an update
```
because the voting power could change multiple times during an epoch, but at the end of an epoch, it could be that the validator has again the same voting power it had when it opted in.

> Also, to better understand, with epochs, GetLastValidatorPower will need to be replaced with something like GetValidatorPowerAtLastEpoch, right?

Not sure I understand what you mean here. `GetLastValidatorPower` returns the most recent validator power and `ComputeValidatorUpdates` would be called at the end of every epoch.
Think of the following example. A validator `V` opts in at the end of an epoch `e` with power `p`. If at a later  epoch `e + X` (`X >= 1`), validator `V` has still the same power `p`, then we should not send a `ValidatorUpdate` for `V` and this is what this checks here.

x/ccv/provider/keeper/partial_set_security.go
insumity: See my previous comment [here](https://github.com/cosmos/interchain-security/pull/1642#discussion_r1492109174).
More specific to the potentiality of a diff method, I guess you could first call `ComputeNextValidators` and then call `diff(currentValidators, nextValidators)` but this `diff` method would be equally complex `ComputeValidatorUpdates` and would have a different method signature compared to `ComputeNextValidators`. So, I don't see much the advantage to this.
Does that sound reasonable?

x/ccv/provider/keeper/partial_set_security.go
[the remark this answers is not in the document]
p-offtermatt: Right. GetLastValidatorPower will return the validator power in the last block. But what we need to compare against is the validator power at the last epoch. If epochLength==1, (as is the case right now) these coincide, but if epochLength != 1 this is not the case anymore.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -67,3 +72,152 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+		currentValPower := k.stakingKeeper.GetLastValidatorPower(ctx, validator.GetOperator())
+		if val.Power == uint64(currentValPower) {
+			continue
+		}
+
+		// if `val` has unbonded, its `GetLastValidatorPower` power returns 0
```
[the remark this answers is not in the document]
p-offtermatt: Yeah, that would be clearer I think

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: Yeah, I agree. I think I would still find the diff method more elegant, but it probably comes down to personal preference, and at the end of the day both are ok.

x/ccv/provider/keeper/proposal.go
```diff
@@ -279,7 +285,14 @@ func (k Keeper) MakeConsumerGenesis(
 	}
 
 	// Apply key assignments to the initial valset.
-	initialUpdatesWithConsumerKeys := k.MustApplyKeyAssignmentToValUpdates(ctx, chainID, initialUpdates)
+	initialUpdatesWithConsumerKeys := k.MustApplyKeyAssignmentToValUpdates(ctx, chainID, initialUpdates,
+		func(address types.ProviderConsAddress) bool {
```
sainoe, line 289: Why this check can't be done in `MustApplyKeyAssignmentToValUpdates` directly?

Issue #1559, opened 2024-01-11
Support backward compatiblity checks in e2e testing


bermuell:
## Problem

Currently it's not possible to check if current provider/consumer versions are compatible with previous versions within the development test environment such as e2e tests. 
These checks are performed in testnet environment which is late in the development process.

Support for backward compatibility checks in e2e test is needed to ensure current implementation does not break backwards compatiblity, e.g. by testing against last major release.

## Closing criteria

- e2e compatibility tests are defined checking sanity of main features across major versions (at least against last major version)
- identification of transformation rule needed for a given e2e test run based on configured provider/consumer versions
- CI job running defined compatibility tests of current implementation against last major version
- address remaining TODOs in e2e tests where applicable

## Problem details

With #1133 e2e tests can be initialized with different consumer/provider versions.
To be able to test backwards compatiblity in the e2e test environment:
- a dedicated test case for compatiblity checks needs to be defined
- required consumer genesis transformation needs to be identified based on consumer/provider versions

PR #1656, opened 2024-02-22
test: support backward compatiblity checks in e2e testing

bermuell:
## Description

Closes: #1559 

Add backward compatibility tests to e2e testing

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/e2e/actions.go
```diff
@@ -546,17 +549,122 @@ func (tr *TestConfig) getConsumerGenesis(providerChain, consumerChain ChainID, t
...
+	}
+
+	if !semver.IsValid(consumerVersion) || !semver.IsValid(providerVersion) {
+		fmt.Println("unable to identify the need for genesis transformation: invalid sem-version", consumerVersion, providerVersion)
+		return false
+	}
+
+	breakages := []string{"v3.0.0", "v3.3.0", "v4.0.0"}
+	for _, breakage := range breakages {
+		if semver.Compare(consumerVersion, breakage) < 0 && semver.Compare(providerVersion, breakage) >= 0 {
```
p-offtermatt, line 585: I'd prefer to put both conditions into the same conditional to reduce code duplication in the inner part

tests/e2e/actions.go
```diff
@@ -546,17 +549,122 @@ func (tr *TestConfig) getConsumerGenesis(providerChain, consumerChain ChainID, t
...
+	if cfg.providerVersion == "latest" {
+		providerVersion = "v4.0.0"
+	}
+
+	if !semver.IsValid(consumerVersion) || !semver.IsValid(providerVersion) {
+		fmt.Println("unable to identify the need for genesis transformation: invalid sem-version", consumerVersion, providerVersion)
+		return false
+	}
+
+	breakages := []string{"v3.0.0", "v3.3.0", "v4.0.0"}
```
p-offtermatt, line 583: nice way of doing this breakage check :)

tests/e2e/actions.go
```diff
@@ -546,17 +549,122 @@ func (tr *TestConfig) getConsumerGenesis(providerChain, consumerChain ChainID, t
...
+	breakage_retry_delay := "88499b7c650ea0fb2c448af2b182ad5fee94d795"
+	// is consumer before v2.x
+	// -- if provider is v3.3.0 (T)
+	transformParams := map[string][]string{
+		"v2.x":   {breakage_prehash},
+		"v3.3.x": {breakage_splitgenesisMain, breakage_splitgenesisV33x},
+		"v4.x":   {breakage_retry_delay},
+	}
+
+	targetVersion := "v4.x"
```
p-offtermatt, line 629: I don't quite get why this is hardcoded instead of a param

tests/e2e/config.go
```diff
@@ -263,9 +398,109 @@ func SlashThrottleTestConfig() TestConfig {
 	return tr
 }
 
+// CompatibilityTestConfig returns a test configuarion for a given version of a consumer and provider
+// The version must match
```
p-offtermatt, line 402: I don't get what `match` means in this context

tests/e2e/config.go
```diff
@@ -630,3 +858,335 @@ func (s *TestConfig) validateStringLiterals() {
...
+// getValidatorConfigFromVersion returns validator configuration based on provider/consumer version used.
+func getValidatorConfigFromVersion(providerVersion, consumerVersion string) map[ValidatorID]ValidatorConfig {
+	var validatorCfg map[ValidatorID]ValidatorConfig
+	switch providerVersion {
+	case "v2.0.0", "v2.1.0", "v2.2.0", "v2.3.0", "v2.4.0",
+		"v3.0.0", "v3.1.0", "v3.2.0", "v3.3.0", "v3.3.2-lsm":
+		fmt.Println("Using old validator configs: ", providerVersion)
+
+		validatorCfg = map[ValidatorID]ValidatorConfig{
+			ValidatorID("alice"): {
```
p-offtermatt, line 871: I wonder if there is some common kernel that we should try to pull out here to reduce duplication, but maybe it's not possible or worth it - just from looking at a diffchecker, a few things seem to be common among versions, but maybe not enough to justify effort to unify it

tests/e2e/test_runner.go
```diff
@@ -10,27 +23,68 @@ type TestRunner struct {
...
+	Duration  time.Duration
+	Result    string
+}
+
+func (result *TestResult) Started() {
+	result.StartTime = time.Now()
+	result.Status = TEST_STATUS_STARTED
+}
+
+func (res *TestResult) Failed() {
```
p-offtermatt, line 41: Maybe we want to check whether the result was already set and reject if it's set twice? not sure if it matters, probably overkill though

p-offtermatt (commented):
Thanks for the work! Mostly looks good, left a couple of comments. I think in particular we need a bit more documentation on this.

I think a lot of this means extra work when we do new releases, and this will for sure be an extra maintenance burden - I'd suggest thinking of ways we can make this e.g. fail visibly when we make a new release and something hasn't been updated properly on the compatibility tests, e.g. somewhere the new release is not set as the latest version

mpoke:
>  I am slightly concerned this introduces a lot of changes when we release a new version, but I don't really have the experience to say. 

@p-offtermatt This PR doesn't modify any production code, thus it doesn't affect the release process.

MSalopek:
> LGTM, thanks for addressing the comments.
> 
> I think this would benefit from having someone who is more involved in the release process take a look. I am slightly concerned this introduces a lot of changes when we release a new version, but I don't really have the experience to say. @MSalopek @mpoke I think one of you should probably have a look

It does not affect releasing. There could be further changes for later versions and we may need to revisit the compatility matrix (some versions will not be complatible because of test initialization, differenc CLI comands used etc.)

p-offtermatt:
> > LGTM, thanks for addressing the comments.
> > I think this would benefit from having someone who is more involved in the release process take a look. I am slightly concerned this introduces a lot of changes when we release a new version, but I don't really have the experience to say. @MSalopek @mpoke I think one of you should probably have a look
> 
> It does not affect releasing. There could be further changes for later versions and we may need to revisit the compatility matrix (some versions will not be complatible because of test initialization, differenc CLI comands used etc.)

Yeah, the points you made are what I was wondering about, not the release process itself, probably didn't express that clearly.

Issue #1669, opened 2024-02-29
MBT: Add epochs


p-offtermatt:
This is an issue that tracks the progress on adding epochs to MBT:
https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-014-epochs.md

Several main tasks should be done:

* Adjust the Quint model to include the feature
* Add tests in Quint to test the feature
* Make sure that the generated traces use the feature in non-trivial ways (or add new generation methods, i.e. invariants, that do)
* Expand the driver to understand/utilize the new feature

PR #1676, opened 2024-03-04
test: Add epochs to MBT

p-offtermatt:
## Description

Closes: #1669 

Adds epochs to MBT.

Because the MBT driver produces a ton of extra blocks (due to the use of ibctesting, which mainly produces extra blocks when updating clients) I get around ~15 extra blocks per block in the model (but this is not a constant, depends on how many consumers are running, there are more extra blocks for starting consumers, etc)

So I solve epochs in a slightly hacky way:
I set the blocksPerEpoch parameter to a very large value, which will not fill up during normal execution.
Then, when an epoch in the model ends, I detect this and produce many blocks to actually end the epoch in the system.
There is a "safety breaker", i.e. I check that the system does not enter a new epoch unless the model does, too, and give a warning to increase blocksPerEpoch if this is violated.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/mbt/driver/mbt_test.go
```diff
@@ -523,6 +558,9 @@ func ComparePacketQueues(
 	consumer string,
 	timeOffset time.Time,
 ) {
+	if consumer == "consumer1" {
```
insumity, line 561: Why is this needed here?

tests/mbt/driver/mbt_test.go
```diff
@@ -200,8 +208,16 @@ func RunItfTrace(t *testing.T, path string) {
 	t.Log("Reading the trace...")
 
 	for index, state := range trace.States {
+		t.Log("Height modulo epoch length:", driver.providerChain().CurrentHeader.Height%blocksPerEpoch)
+		t.Log("Model height:", ProviderHeight(state.VarValues["currentState"].Value.(itf.MapExprType))%modelBlocksPerEpoch)
```
insumity, line 212: nit: similar to the line above maybe say `"Model height modulo model epoch length"`

tests/mbt/driver/mbt_test.go
```diff
@@ -200,8 +208,16 @@ func RunItfTrace(t *testing.T, path string) {
...
 		t.Logf("Reading state %v of trace %v", index, path)
 
+		// store the height of the provider state before each step.
+		// The height should only pass an epoch when there passes an epoch in the model, too,
```
insumity, line 216:
```suggestion
		// The height should only pass an epoch when it passes an epoch in the model, too,
```

tests/mbt/driver/mbt_test.go
```diff
@@ -239,22 +255,33 @@ func RunItfTrace(t *testing.T, path string) {
...
+
+				// produce blocks until the next block ends the epoch
+				for i := heightInEpoch; i < blocksPerEpoch; i++ {
+					driver.endAndBeginBlock("provider", 1*time.Nanosecond)
+				}
+			}
+
 			// we need at least 2 blocks, because for a packet sent at height H, the receiving chain
 			// needs a header of height H+1 to accept the packet
 			// so, we do `blocksPerEpoch` time advancements with a very small increment,
```
insumity, line 276: Remove the `// so, we do ... ` comment.

tests/mbt/driver/setup.go
```diff
@@ -397,21 +397,15 @@ func (s *Driver) ConfigureNewPath(consumerChain, providerChain *ibctesting.TestC
 	// their channel, and are ready for anything to happen.
 	s.consumerKeeper(consumerChainId).SetProviderChannel(s.ctx(consumerChainId), consumerEndPoint.ChannelID)
 
-	// Commit a block on both chains, giving us two committed headers from
-	// the same time and height. This is the starting point for all our
-	// data driven testing.
+	// Commit a block on the consumer to get a header
```
insumity, line 400: Why do we have this change here from `provider` to `consumer`?

tests/mbt/model/ccv.qnt
```diff
@@ -86,11 +89,8 @@ module ccv_types {
             // Stores VscPackets which have been sent but where the provider has *not received a response yet*.
             sentVscPacketsToConsumer: Chain -> List[VscPacket],
 
-            // stores whether, in this block, the validator set has changed.
-            // this is needed because the validator set might be considered to have changed, even though
-            // it is still technically identical at our level of abstraction, e.g. a validator power change on the provider
-            // might leave the validator set the same because a delegation and undelegation cancel each other out.
-            providerValidatorSetChangedInThisBlock: bool,
+            // stores for which consumer chains, in this epoch, the validator set is considered to have changed and we thus need to send a vscPacket to the consumer chains.
```
insumity, line 92: nit:
```suggestion
            // stores for which consumer chains, in this epoch, the validator set is considered to have changed and we thus need to send a VscPacket to the consumer chains.
```

tests/mbt/model/ccv_model.qnt
```diff
@@ -517,9 +541,10 @@ module ccv_model {
                     assert(currentState.providerState.chainState.votingPowerHistory == List(InitialValidatorSet.put("node1", 150), InitialValidatorSet)),
                     // change voting power on provider again
                     VotingPowerChange("node1", 50).then(
-                        // end another block
-                        EndAndBeginBlockForProvider(1 * Second, Set(), Set())
-                    ).then(
+                        // end the epoch
+                        EndProviderEpoch(3 * Second, Set(), Set())
```
insumity, line 545: Why do we have 3 seconds here instead of 1 as is done in `EndProviderEpoch` below?

tests/mbt/driver/mbt_test.go
p-offtermatt: Oops, a legacy left over from debugging. Thanks for catching!

tests/mbt/driver/setup.go
p-offtermatt: I removed an extra block production for the provider here because I thought I could make it so model and system heights would line up, if I managed to remove this (but it turns out there are waaay more places where blocks are produced than I thought), so the comment needed to change, too

tests/mbt/model/ccv_model.qnt
p-offtermatt: "Historical reasons", i.e. a relic from trying things while debugging :)
good catch, adjusted this!

tests/mbt/driver/setup.go
p-offtermatt: Turns out I was able to remove the extra block for the consumer, too, so now the comment is completely gone :)

insumity (approved):

tests/mbt/model/ccv_model.qnt
```diff
@@ -319,16 +340,19 @@ module ccv_model {
...
+
+    val CurrentBlockEndsEpoch = currentState.providerState.chainState.currentBlockHeight % epochLength == 0
+
     // Any update in the power of a validator on the provider
     // MUST be present in a ValidatorSetChangePacket that is sent to all registered consumer chains
     val ValUpdatePrecondition = trace[trace.length()-1].kind == "EndAndBeginBlockForProvider"
     val ValidatorUpdatesArePropagatedInv = 
-        // when the provider has just entered a validator set into a block...
-        ValUpdatePrecondition and currentState.providerState.providerValidatorSetChangedInThisBlock
+        // when the an epoch ends and the provider has just entered a validator set into a block...
```
sainoe, line 350:
```suggestion
        // when an epoch ends and the provider has just entered a validator set into a block...
```

tests/mbt/model/ccv_model.qnt
```diff
@@ -235,6 +239,23 @@ module ccv_model {
...
+    // The rest of the blocks will not start or stop any consumers, and will advance time by 1 second each.
+    run EndProviderEpoch(
+        timeAdvancement: Time,
+        consumersToStart: Set[Chain],
+        consumersToStop: Set[Chain]
+    ): bool =
+        epochLength.reps(
+            i => 
+            if (i == 0)
+            EndAndBeginBlockForProvider(timeAdvancement-((epochLength-1)*Second), consumersToStart, consumersToStop)
```
mpoke, line 254: What if `timeAdvancement-((epochLength-1)*Second)` is negative?

tests/mbt/model/ccv_model.qnt
```diff
@@ -235,6 +239,23 @@ module ccv_model {
         stepCommon
     }
 
+    // Runs epochLength many blocks on the provider.
+    // The first block will start all consumers in consumersToStart and stop all consumers in consumersToStop,
+    // and advance time by timeAdvancement - ((epochLength-1) * Seconds).
+    // The rest of the blocks will not start or stop any consumers, and will advance time by 1 second each.
+    run EndProviderEpoch(
```
mpoke, line 246: This is only used for tests, right? Could you rename it accordingly or make it clear in the comment.

mpoke (approved):
LGTM. Great work @p-offtermatt

sainoe (approved):
LGTM!

PR #1701, opened 2024-03-12
test: update confusing e2e democracy setup

MSalopek:
## Description

Democracy and democracy reward e2e tests were wired in a confusing way. This PR removes the confusion.

Before the PR, `democracy` tests actually ran `democracy-reward` tests and vice versa. It was difficult to work with the tests since there are configuration nuances that are easy to overlook with regards to reward distribution tests.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


p-offtermatt:
Suggestion: Since the two steps steps_democracy and steps_reward_denom only differ in exactly the true/false statements, could we make these steps a single function that just takes a boolean as an input to reduce code duplication?
Also, could we rename them to more evocative names? The only difference seems to be whether the denom is enabled or not. Could we call one democracy_missing_denom and the other democracy_ok_denom?

MSalopek:
> Suggestion: Since the two steps steps_democracy and steps_reward_denom only differ in exactly the true/false statements, could we make these steps a single function that just takes a boolean as an input to reduce code duplication? Also, could we rename them to more evocative names? The only difference seems to be whether the denom is enabled or not. Could we call one democracy_missing_denom and the other democracy_ok_denom?

We can, I'm doing this to unblock some of the v50 work

p-offtermatt (approved):
LGTM!

tests/e2e/steps_democracy.go
```diff
@@ -181,12 +181,13 @@ func stepsDemocracy(consumerName string) []Step {
 			},
 			State: State{
 				ChainID("provi"): ChainState{
-					// Check that tokens are minted and sent to provider chain and distributed to validators and their delegators on provider chain
+					// Check that ARE NOT minted and sent to provider chain and distributed to validators and their delegators on provider chain
```
bermuell, line 184:
```suggestion
					// Check that tokens ARE NOT minted and sent to provider chain and distributed to validators and their delegators on provider chain
```

PR #1702, opened 2024-03-13
Insumity/add opt in set

insumity:
Please go to the `Preview` tab and select the appropriate sub-template:

* [Production code](?expand=1&template=production.md) - for types `fix`, `feat`, and `refactor`.
* [Docs](?expand=1&template=docs.md) - for documentation changes.
* [Others](?expand=1&template=others.md) - for changes that do not affect production code.


x/ccv/provider/keeper/proposal.go
```diff
@@ -289,7 +289,23 @@ func (k Keeper) MakeConsumerGenesis(
...
+
+	considerOnlyOptIn := func(validator stakingtypes.Validator) bool {
+		consAddr, err := validator.GetConsAddr()
+		if err != nil {
+			return false
+		}
+		return k.IsOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr))
+	}
+
+	if topN, found := k.GetTopN(ctx, chainID); found {
```
sainoe, line 302: The `TopN` value for the consumer isn't set at this poin since SetTopN() is called [after](https://github.com/cosmos/interchain-security/blob/03af237c800211642e81759b5f2e23bdcbad1113/x/ccv/provider/keeper/proposal.go) `CreateConsumerClientInCachedCtx`. 
So `prop.Top_N` should be used here.

x/ccv/provider/keeper/hooks.go
```diff
@@ -225,12 +225,13 @@ func (h Hooks) AfterProposalVote(ctx sdk.Context, proposalID uint64, voterAddr s
 			return
 		}
 
-		// in the validator is already to-be-opted in, the `SetToBeOptedIn` is a no-op
-		h.k.SetToBeOptedIn(ctx, chainID, providertypes.NewProviderConsAddress(consAddr))
+		// in the validator is already opted in, the `SetOptedIn` is a no-op
```
p-offtermatt, line 228:
```suggestion
		// if the validator is already opted in, then `SetOptedIn` is a no-op
```

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -7,22 +7,19 @@ import (
...
 	}
 
-	if k.IsToBeOptedOut(ctx, chainID, providerAddr) {
-		// a validator to be opted in cancels out with a validator to be opted out
-		k.DeleteToBeOptedOut(ctx, chainID, providerAddr)
-	} else if !k.IsToBeOptedIn(ctx, chainID, providerAddr) && !k.IsOptedIn(ctx, chainID, providerAddr) {
-		// a validator can only be set for opt in if it is not opted in and not already set for opt in
-		k.SetToBeOptedIn(ctx, chainID, providerAddr)
-	}
+	k.SetOptedIn(ctx, chainID, providerAddr)
```
p-offtermatt, line 22: just to make sure: if this function returns an error, this SetOptedIn is rolled back, right? Otherwise, it could happen that someone tries to opt in, key assignment fails, but this still opts them in

testutil/ibc_testing/generic_setup.go
```diff
@@ -140,6 +140,11 @@ func AddConsumer[Tp testutil.ProviderApp, Tc testutil.ConsumerApp](
 	prop := testkeeper.GetTestConsumerAdditionProp()
 	prop.ChainId = chainID
 	prop.Top_N = consumerTopNParams[index] // isn't used in CreateConsumerClient
+
+	// set the consumer TopN here since the test suite setup only used the consumer addition prop
+	// to create the consumer genesis, see BeginBlockInit in /x/ccv/provider/keeper/proposal.go.
+	providerKeeper.SetTopN(providerChain.GetContext(), chainID, prop.Top_N)
```
sainoe, line 146: Moving this line makes the consumer test setup different from the consumer addition proposal logic - see my comment above.

x/ccv/provider/keeper/hooks.go
sainoe, line 228:
```suggestion
		// If the validator is already opted in, the `SetOptedIn` is a no-op
```

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -139,3 +123,94 @@ func TestHandleSetConsumerCommissionRate(t *testing.T) {
 	require.Equal(t, sdk.OneDec(), cr)
 	require.True(t, found)
 }
+
+func TestOptInTopNValidators(t *testing.T) {
```
p-offtermatt, line 127: Could we add a test that has multiple validators with the smallest power that is opted in? just to check that edge case

x/ccv/provider/keeper/relay.go
```diff
@@ -222,7 +222,23 @@ func (k Keeper) QueueVSCPackets(ctx sdk.Context) {
 
 	for _, chain := range k.GetAllConsumerChains(ctx) {
 		currentValidators := k.GetConsumerValSet(ctx, chain.ChainId)
-		nextValidators := k.ComputeNextEpochConsumerValSet(ctx, chain.ChainId, bondedValidators)
+		var nextValidators []providertypes.ConsumerValidator
+
+		considerOnlyOptIn := func(validator stakingtypes.Validator) bool {
```
p-offtermatt, line 227: code duplicated from proposal.go, could we just extract a function here and put it into partial_set_security.go?

x/ccv/provider/keeper/relay_test.go
```diff
@@ -659,41 +659,66 @@ func TestEndBlockVSU(t *testing.T) {
...
+				}
+			}
+		},
+	).AnyTimes()
 	mocks.MockStakingKeeper.EXPECT().GetLastValidators(gomock.Any()).Return(lastValidators).AnyTimes()
 	mocks.MockStakingKeeper.EXPECT().GetLastValidatorPower(gomock.Any(), gomock.Any()).Return(int64(2)).AnyTimes()
 
+	for _, val := range lastValidators {
+		consAddr, _ := val.GetConsAddr()
+		// FIXME: probably `consAddr` is not needed
```
p-offtermatt, line 696: can we resolve this?

x/ccv/provider/keeper/validator_set_update.go
```diff
@@ -174,11 +128,53 @@ func DiffValidators(
...
+	}, nil
+}
+
+// ComputeNextEpochConsumerValSet returns the next validator set that is responsible for validating consumer
+// chain `chainID`. The next validator set corresponds to the `bondedValidator`s that `shouldConsider` evaluates to `true`.
+func (k Keeper) ComputeNextEpochConsumerValSet(
+	ctx sdk.Context,
+	chainID string,
+	bondedValidators []stakingtypes.Validator,
+	shouldConsider func(validator stakingtypes.Validator) bool,
```
p-offtermatt, line 161: Is ShouldConsider the thing we would use for what Jehan calls power shaping?
Because I think a predicate (val -> bool) is probably not expressive enough.
In reality, we want to be able to put functions
[]ConsumerValidator -> []ConsumerValidator

The predicate is not enough to express e.g. power caps

p-offtermatt (commented):
Left some comments. My main concern is that the predicate seems too restrictive to be used for power shaping.

x/ccv/provider/keeper/partial_set_security.go
insumity: As far as I know, the execution is performed on cached context and state would only be committed if all the messages of a transactions are successful. See [here](https://github.com/cosmos/cosmos-sdk/blob/main/baseapp/baseapp.go#L814):
```
// [...]. All state transitions occur through
// a cached Context depending on the mode provided. State only gets persisted
// if all messages get executed successfully and the execution mode is DeliverTx.
```

x/ccv/provider/keeper/partial_set_security_test.go
insumity: Introduced one more validator with power `1` so we have two validators with power `1`. Also added a case where we have `minPowerToOptIn == 0`.

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: Ok, yeah that matches my understanding

x/ccv/provider/keeper/relay.go
insumity: Good point. I created `ShouldConsiderOnlyOptIn` (and added a test for it) and now do:
```
nextValidators := k.ComputeNextEpochConsumerValSet(ctx, chain.ChainId, bondedValidators,
	func(validator stakingtypes.Validator) bool {
		return k.ShouldConsiderOnlyOptIn(ctx, chain.ChainId, validator)
	})
```

x/ccv/provider/keeper/validator_set_update.go
insumity: We discussed offline: We agree that this can't be used for power shaping. But it is something we can do add/refactor later when we actually implement power shaping, etc.

x/ccv/provider/keeper/proposal.go
insumity: Good catch! Fixed.

x/ccv/provider/keeper/relay.go
```diff
@@ -222,7 +222,17 @@ func (k Keeper) QueueVSCPackets(ctx sdk.Context) {
 
 	for _, chain := range k.GetAllConsumerChains(ctx) {
 		currentValidators := k.GetConsumerValSet(ctx, chain.ChainId)
-		nextValidators := k.ComputeNextEpochConsumerValSet(ctx, chain.ChainId, bondedValidators)
+
+		if topN, found := k.GetTopN(ctx, chain.ChainId); found {
```
p-offtermatt, line 226: Can we de-duplicate this more? Lines 226 to 235 here are still the same as lines 293 to 301 in proposal.go (but it's already much better like this)

p-offtermatt: Yeah, you are right, let's keep this in mind to maybe consolidate more in the future, but leave as is for now

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -53,13 +51,91 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+			if err != nil {
+				k.Logger(ctx).Error("could not retrieve validators consensus address",
+					"validator", val,
+					"error", err)
+				continue
+			}
+
+			// if validator already exists it gets overwritten
+			k.SetOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr))
+		} else {
```
sainoe, line 92: No need to enclose this comment in a `else` statement here.

x/ccv/provider/keeper/validator_set_update.go
```diff
@@ -174,11 +128,53 @@ func DiffValidators(
 	return updates
```
sainoe, line 128: It seems to me that this `updates` slice is created by iterating on `map`s which are not deterministic.
Shouldn't we sort them, as we do in these [tests](https://github.com/cosmos/interchain-security/blob/330c085024997377d6d01e2c3159f9562a0d913a/x/ccv/provider/keeper/validator_set_update_test.go#L183-L191)?

insumity: The `updates` are created by executing two `for` loops:
1. `for _, currentVal := range currentValidators {`
2. `for _, nextVal := range nextValidators {`
where `currentValidators` and `nextValidators` are slices. The used maps are used internally in the `for` loops. So I do not believe we would have a non-determinism problem here.

sainoe: You're right. Misread it!

p-offtermatt (approved):

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -53,13 +51,89 @@ func (k Keeper) HandleOptOut(ctx sdk.Context, chainID string, providerAddr types
...
+
+			// if validator already exists it gets overwritten
+			k.SetOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr))
+		} // else validators that do not belong to the top N validators but were opted in, remain opted in
+	}
+}
+
+// ComputeMinPowerToOptIn returns the minimum power needed for a validator (from the bonded validators)
+// to belong to the `topN` validators
+func (k Keeper) ComputeMinPowerToOptIn(ctx sdk.Context, bondedValidators []stakingtypes.Validator, topN uint32) int64 {
```
p-offtermatt, line 98: Hey,

I think there is a bug here.
If topN == 0, topNThreshold will be the maximal power.

However, this is invoked e.g. here https://github.com/cosmos/interchain-security/blob/30b99aa8eb193a08bef96d181fb21c13ddcc9ca5/x/ccv/provider/keeper/partial_set_security.go#L65 to check whether someone can opt out.
This disallows the val with maximal power from opting out, for example.

I think it should just return something to signify infinity if topN == 0

insumity: Great catch. I fixed the condition on `HandleOptOut` to:
```
if topN, found := k.GetTopN(ctx, chainID); found && topN > 0 {
```
to guarantee that `ComputeMinPowerToOptIn` does not get `topN == 0`.
But, also changed `ComputeMinPowerToOptIn` to return `math.MaxInt64` when `topN == 0` to avoid potential pitfalls.

Issue #1609, opened 2024-01-31
Partial Set Security: Introduce queries


insumity:
## Problem

Introduce (or modify) queries as outlined in [ADR-015](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-015-partial-set-security.md). This means, at least the following queries:

-  modify `interchain-security-pd query provider list-consumer-chains` to return information on whether a consumer chain is an Opt In or a Top N chain and with what N;
- `interchain-security-pd query provider optedInValidators $chainID`  to retrieve the validators that are opted in and hence the validators that need to validate the consumer chain; 
- `interchain-security-pd query provider hasToValidate $providerAddr` that given a validator's address returns all the chains this validator has to validate. (FYI, a note mentioned by @sainoe was that: i) we could create and store a reverse mapping on the Hub from `validator` to `chains` but this would be error prone (e.g., more code), and ii) we could just loop through all the chains and find the validator which would slow down those operations).

## Closing criteria

Implement and thoroughly test.

PR #1733, opened 2024-03-26
feat: Add queries for PSS and consumer commission rate

sainoe:
## Description

Closes: #1609

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -99,6 +100,33 @@ service Query {
       option (google.api.http).get =
     "/interchain_security/ccv/provider/params";
     }
+
+  // QueryConsumerChainOptedInValidators returns a list of validators consensus address
```
insumity, line 104:
```suggestion
  // QueryConsumerChainOptedInValidators returns a list of validators consensus addresses
```

x/ccv/provider/keeper/keeper.go
```diff
@@ -249,9 +249,16 @@ func (k Keeper) GetAllConsumerChains(ctx sdk.Context) (chains []types.Chain) {
...
 
+		// Get the consumer TopN value by checking
+		// if the chain is a TopN chain,
+		// otherwise it's a OptIn chain
+		topN, isTopN := k.GetTopN(ctx, chainID)
+
 		chains = append(chains, types.Chain{
 			ChainId:  chainID,
 			ClientId: clientID,
+			Opt_In:   !isTopN,
```
insumity, line 260: I think this should be:
```suggestion
			Opt_In:   !isTopN || topN == 0,
```
`GetTopN` would return a `found` on whether `TopN` is set, but `TopN` could be set to `0`.

insumity: Similarly, below in `keeper_test.go` you probably won't need to check `if topN > 0 {` in

```
if topN > 0 {
    pk.SetTopN(ctx, chainID, topN)
```

insumity (approved):
LGTM

Just a minor comment: I wonder if we need to extend `QueryConsumerChainsValidatorHasToValidate` so it returns the opted-in validators, besides the consumer validators. The way the query is phrased, returning only the consumer validators is fine, because those are the ones that _have to validate_.
But, if a validator opts-in and then checks this query, they might expect to see themselves there as well (e.g., "validators does not have to validate yet but has to validate starting at block height `X`").

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -127,6 +155,10 @@ message QueryConsumerChainStopProposalsResponse {
 message Chain {
   string chain_id = 1;
   string client_id = 2;
+  // If the `chainID` is an Opt-In chain, i.e., no validator is forced to validate chain `chainID`
+  bool opt_In = 3;
```
p-offtermatt, line 159: Why do we need this? (it's more clear for users is a perfectly fine reason if you think it applies)

 Can't we keep the TopN == 0 means it's an opt-in chain, and just have the TopN field?

p-offtermatt (approved):

p-offtermatt (approved):
LGTM, one nit.

I think it would be cool but not mandatory to have also what Karolos mentioned, probably as another query, something that gets all chains the validator will have to validate when the next epoch starts if nothing else changes until then.
it should account for the validator being newly in the topN , opting in voluntarily, or opting out.

and then the current query can stay as is and just say which chains the validator has to validate right at this moment

x/ccv/provider/keeper/keeper.go
sainoe: Make sense, thanks. Updated.

proto/interchain_security/ccv/provider/v1/query.proto
sainoe: We don't need it in the logic. The intention was to make it clearer for the user. But after some thought, it can actually be removed.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -266,11 +266,13 @@ func (k Keeper) QueryConsumerChainsValidatorHasToValidate(goCtx context.Context,
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	// get all the consumer chains for which the validator is opted-in
+	// get all the consumer chains for which the validator is already or will be
+	// opted-in starting from the next epoch
```
p-offtermatt, line 270: This comment doesn't match the code below, right?

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -266,11 +266,13 @@ func (k Keeper) QueryConsumerChainsValidatorHasToValidate(goCtx context.Context,
...
 	consumersToValidate := []string{}
 	for _, consumer := range k.GetAllConsumerChains(ctx) {
 		chainID := consumer.ChainId
-		if k.IsConsumerValidator(ctx, chainID, types.NewProviderConsAddress(consAddr)) || k.IsOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr)) {
+
+		if k.IsOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr)) {
```
p-offtermatt, line 275: Note that top N validators are only opted in right when the epoch begins, so this doesn't tell you who will get opted into the chain because of being in the top N.

Of course we cannot know for sure, but I would change this to just compute the MinPowerToOptIn for the chain, and then compute the validators that would be opted in, if the epoch would end without further power changes

x/ccv/provider/keeper/grpc_query.go
sainoe: I think it does. When `k.IsOptedIn` returns `true`, it means that the validator is either already on the consumer valset or is opted-in and will be added to the set at the start of the next epoch.

x/ccv/provider/keeper/grpc_query.go
sainoe: Good point. Will add that. Thanks!

x/ccv/provider/keeper/grpc_query.go
p-offtermatt: Hm, I guess you are right. I'm just a bit confused what should happen when someone has opted out, but is still currently in power on the consumer chain

insumity (approved):

x/ccv/provider/client/cli/query.go
```diff
@@ -409,3 +412,123 @@ $ %s query provider params
...
+}
+
+// Command to query the consumer chains list a given validator has to validate
+func CmdConsumerChainsValidatorHasToValidate() *cobra.Command {
+	bech32PrefixConsAddr := sdk.GetConfig().GetBech32ConsensusAddrPrefix()
+	cmd := &cobra.Command{
+		Use:   "has-to-validate [provider-validator-address]",
+		Short: "Query the consumer chains list a given validator has to validate",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Query the consumer chains list a given validator has to validate.
```
p-offtermatt, line 457: Could we make this a bit more explicit?

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -249,7 +249,8 @@ func (k Keeper) QueryConsumerChainOptedInValidators(goCtx context.Context, req *
 	}, nil
 }
 
-// QueryConsumerChainsValidatorHasToValidate returns all consumer chains a given validator has to validate
+// QueryConsumerChainsValidatorHasToValidate returns all consumer chains that the given validator has to validate now
+// or in the next epoch if nothing change.
```
insumity, line 253:
```suggestion
// or in the next epoch if nothing changes.
```

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -266,15 +267,31 @@ func (k Keeper) QueryConsumerChainsValidatorHasToValidate(goCtx context.Context,
...
+			if topN, found := k.GetTopN(ctx, chainID); found && topN > 0 {
+				val, found := k.stakingKeeper.GetValidatorByConsAddr(ctx, consAddr)
+				if !found {
+					return nil, status.Error(codes.InvalidArgument, "invalid provider address")
+				}
+				power := k.stakingKeeper.GetLastValidatorPower(ctx, val.GetOperator())
+				minPowerToOptIn := k.ComputeMinPowerToOptIn(ctx, chainID, k.stakingKeeper.GetLastValidators(ctx), topN)
+
+				// Check that the validator's voting power is smaller
+				// than the minimum to be automatically opt-in
```
insumity, line 287:
```suggestion
				// than the minimum to be automatically opted in
```

insumity (approved):

PR #1741, opened 2024-03-28
test: Quint model - Use `indices` instead of `range` to avoid problem with Apalache

bugarela:
## Description

This is a simple fix for the model, so it works with Apalache (`quint verify`). Apalache doesn't support non-constant values in `range`, but in this case, we can simply use `indices` to achieve the same thing and avoid the issue. The behavior on simulation should stay the same, and this unblocks usage with Apalache.

---

### Author Checklist

I have...

- [X] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [X] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
- [-] Provided a link to the relevant issue or specification
  - There are no issues related to this. I've been fixing some other issues on the quint repo related to this, and bumped into this problem myself.
  - I need authorization for running them first

### Reviewers Checklist

I have...

### DIFF
--- a/tests/mbt/model/libraries/extraSpells.qnt
+++ b/tests/mbt/model/libraries/extraSpells.qnt
@@ -257,7 +257,7 @@ module extraSpells {
     /// Returns:
     /// - The index of the first element that satisfies the predicate function, or -1 if no such element exists.
     pure def firstIndex(__list: List[a], __f: a => bool): int = {
-        range(0, __list.length()).foldl(
+        __list.indices().fold(
             -1,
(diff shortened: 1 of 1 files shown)


bugarela:
I can't request reviews, but will ping @p-offtermatt directly here :smile:

p-offtermatt (approved):
Nice, thanks Gabriela! LGTM!

PR #1762, opened 2024-04-05
fix!: introduce v4 provider migration code

MSalopek:
Epochs code changes the paramaters store, so the new parameter needs to be added to the store via a migration.

Had v4.0.0 and v4.1.0 stayed the same consensus version, provider chains using v4.0.0 would panic when migrating to v4.1.0 because of the missing migration.

To mitigate this issue provider consensus version got bumped to `4` and the migration code was refactored to introduce the `v4` migration.

### DIFF
--- a/x/ccv/provider/migrations/migrator.go
+++ b/x/ccv/provider/migrations/migrator.go
@@ -6,6 +6,7 @@ import (
 
 	providerkeeper "github.com/cosmos/interchain-security/v4/x/ccv/provider/keeper"
 	v3 "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/v3"
+	v4 "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/v4"
 )
 
 // Migrator is a struct for handling in-place store migrations.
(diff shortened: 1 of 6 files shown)


x/ccv/provider/migrations/migrator.go
```diff
@@ -29,6 +30,12 @@ func (m Migrator) Migrate1to2(ctx sdktypes.Context) error {
 
 // Migrate2to3 migrates x/ccvprovider state from consensus version 2 to 3.
 func (m Migrator) Migrate2to3(ctx sdktypes.Context) error {
-	v3.MigrateParams(ctx, m.paramSpace)
```
insumity, line 32: Why is this removed?

sainoe (approved):

mpoke (approved):

x/ccv/provider/migrations/migrator.go
MSalopek: Because it was moved to v4 (migration used for ICS @ v4.1).

`Migrate2to3` is for v4.0.0.
`Migrate3to4` is for v4.1.0.

PR #1773, opened 2024-04-08
fix!: register provider v3 to v4 migration handler

MSalopek:
This code registers a migration handler that was missing in https://github.com/cosmos/interchain-security/pull/1762.

This was discovered during upgrade testing.


insumity (approved):

MSalopek:
Link check fails on purpose, the release `v4.1.0-rc2` does not exist yet

Issue #853, opened 2023-04-14
EPIC: Partial Set Security


mpoke:
# Problem

This is a meta issue to track the work on opt-in security. With opt in security, validators will be able to decide which consumer chains they would like to validate on, and will only receive rewards for those that they are active on.

EDIT: Renamed to Partial Set Security. See this [doc](https://informalsystems.notion.site/Partial-Set-Security-398ca9a1453740068be5c7964a4059bb) for a high level design. 

# Closing criteria

Partial Set Security is implemented.

# Task list

```[tasklist]
### Must have
- [x] Analysis of whether "governance fraud proofs" are Ok
- [ ] https://github.com/cosmos/interchain-security/issues/788
- [ ] https://github.com/cosmos/interchain-security/issues/1571
- [ ] https://github.com/cosmos/interchain-security/issues/1587
- [ ] https://github.com/cosmos/interchain-security/issues/1586
- [ ] https://github.com/cosmos/interchain-security/issues/1574
- [ ] https://github.com/cosmos/interchain-security/issues/1606
- [ ] https://github.com/cosmos/interchain-security/issues/1607
- [ ] https://github.com/cosmos/interchain-security/issues/1588
- [ ] https://github.com/cosmos/interchain-security/issues/1610
- [ ] https://github.com/cosmos/interchain-security/issues/1611
- [ ] https://github.com/cosmos/interchain-security/issues/1609
- [ ] https://github.com/cosmos/interchain-security/issues/1608
- [ ] https://github.com/cosmos/interchain-security/issues/1630
- [ ] https://github.com/cosmos/interchain-security/issues/1633
- [ ] https://github.com/cosmos/interchain-security/issues/1658
- [ ] Migrate Neutron and Stride to Top N with N = 95 or N = 100? When we build `gaia` v17
- [ ] https://github.com/cosmos/interchain-security/issues/1707
- [ ] https://github.com/cosmos/interchain-security/issues/1697
- [ ] https://github.com/cosmos/interchain-security/issues/1781
- [ ] https://github.com/cosmos/interchain-security/issues/1772
- [ ] https://github.com/cosmos/interchain-security/issues/1785
- [ ] https://github.com/cosmos/interchain-security/issues/1245
- [ ] https://github.com/cosmos/interchain-security/issues/1926
```

```[tasklist]
### Nice to have
- [ ] https://github.com/cosmos/interchain-security/issues/1634
- [ ] https://github.com/cosmos/interchain-security/issues/1635
[...]

PR #1809, opened 2024-04-23
feat!: Introduce Partial Set Security

p-offtermatt:
## Description

Related: #853 

This PR adds the logic for partial set security.
It can logically be broken down into three features:
1) Handling opt-in and outs, and sending a filtered validator set according to who is opted in
2) Adding the top-N parameter for consumer chains, which automatically opts in validators in the top N% of the validator set
3) Allowing validators to set a separate commission rate for each consumer chain

Some more minor changes:
- Changes the behaviour when a validator tries to assign a key that is already assigned by them from error to noop

Highlights:
- Adds 3 new message types on the provider:
   * MsgOptIn
   * MsgOptOut
   * MsgSetConsumerCommissionRate
- Adds 3 new queries:
   * QueryConsumerChainOptedInValidators
   * QueryConsumerChainsValidatorHasToValidate
   * QueryValidatorConsumerCommissionRate

The actual main logic change for PSS is in relay.go, partial_set_security.go, and validator_set_update.go.

Things to figure out for migrations:
* Versioning for Migrations (upgrade handler only upgrades with integer versions, but v5 is sdk v50; what do we do?)
* What to migrate for outstanding consumer addition proposals (do we even need to migrate; is it ok if they get a top N of 0 by default?)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/provider/types/proposal.go
```diff
@@ -135,6 +137,12 @@ func (cccp *ConsumerAdditionProposal) ValidateBasic() error {
 		return errorsmod.Wrap(ErrInvalidConsumerAdditionProposal, "unbonding period cannot be zero")
 	}
 
+	// Top N corresponds to the top N% of validators that have to validate the consumer chain and can only be 0 (for an
+	// Opt In chain) or in the range [50, 100] (for a Top N chain).
+	if cccp.Top_N != 0 && cccp.Top_N < 50 || cccp.Top_N > 100 {
```
p-offtermatt, line 142: Should we allow more generous values for Top_N here?

x/ccv/provider/module.go
```diff
@@ -144,6 +144,8 @@
 	am.keeper.BeginBlockCCR(ctx)
 	// Check for replenishing slash meter before any slash packets are processed for this block
 	am.keeper.BeginBlockCIS(ctx)
+	// BeginBlock logic need for the  Reward Distribution sub-protocol
+	am.keeper.BeginBlockRD(ctx, req)
```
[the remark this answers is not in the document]
p-offtermatt: @sainoe could you check if this is problematic?

mpoke:
I added changelog entries for PSS, see https://github.com/cosmos/interchain-security/pull/1809/commits/e0e6524f1d91def5a5fa14918e0f100bba8c8f82

x/ccv/provider/types/proposal.go
insumity: Rationale behind 50 being a lower bound is described [here](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-015-partial-set-security.md):
> In case of a Top N chain, we restrict the possible values of top_N from (0, 100] to [50, 100]. By having top_N >= 50 we can guarantee that we cannot have a successful attack, assuming that at most 1/3 of provider validators can be malicious. This is because, a Top N chain with N >= 50% would have at least 1/3 honest validators, which is sufficient to stop attacks. Additionally, by having N >= 50% (and hence N > (VetoThreshold = 33.4%)) we enable the top N validators to Veto any ConsumerAdditionProposal for consumer chains they do not want to validate.

Additionally, with `top_N >= 50` we also do **not** need to have fraud votes for _Top N_ chains. This is because, if we assume < 1/3 of validators can be malicious on the Hub, then a consumer with 50% of the Hub's power would have >= 1/3 of honest validators. In order to perform incorrect execution we need >2/3 of malicious validators and hence with `top_N >= 50` we cannot have incorrect execution and therefore do not need fraud votes.
Not having fraud votes for _Top N_ chains is also nice because validators in a _Top N_ chain might have been forced to validate and hence we want to reduce the potentiality of them getting slashed for something they were forced to do.

x/ccv/types/events.go
```diff
@@ -12,6 +12,8 @@ const (
 	EventTypeRemoveConsumerRewardDenom  = "remove_consumer_reward_denom"
 	EventTypeSubmitConsumerMisbehaviour = "submit_consumer_misbehaviour"
 	EventTypeSubmitConsumerDoubleVoting = "submit_consumer_double_voting"
+	EventTypeOptIn                      = "opt_in"
+	EventTypeOptOut                     = "opt_out"
```
mpoke, line 16: Move to provider events

x/ccv/provider/types/codec.go
```diff
@@ -46,6 +46,14 @@ func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 		(*sdk.Msg)(nil),
 		&MsgSubmitConsumerDoubleVoting{},
 	)
+	registry.RegisterImplementations(
+		(*sdk.Msg)(nil),
+		&MsgOptIn{},
+	)
+	registry.RegisterImplementations(
+		(*sdk.Msg)(nil),
+		&MsgOptOut{},
```
mpoke, line 55: Shouldn't we register also MsgSetConsumerCommissionRate?

x/ccv/provider/types/consumer.go
```diff
@@ -25,3 +27,10 @@ func NewConsumerStates(
 		SlashDowntimeAck:     slashDowntimeAck,
 	}
 }
+
+// zero consumer rewards allocation
+func InitialConsumerRewardsAllocation() ConsumerRewardsAllocation {
```
mpoke, line 32: Never used.

x/ccv/provider/types/proposal.go
mpoke: nit: add parentheses for clarity

x/ccv/provider/types/msg.go
```diff
@@ -203,3 +209,163 @@ func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
...
+}
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgOptIn) ValidateBasic() error {
+	if strings.TrimSpace(msg.ChainId) == "" {
+		return errorsmod.Wrapf(ErrInvalidConsumerChainID, "chainId cannot be blank")
+	}
+	// It is possible to assign keys for consumer chains that are not yet approved.
+	// This can only be done by a signing validator, but it is still sensible
+	// to limit the chainID size to prevent abuse.
```
mpoke, line 249:
```suggestion
	// It is possible to opt in to validate on consumer chains that are not yet approved.
	// This can only be done by a signing validator, but it is still sensible
	// to limit the chainID size to prevent abuse.
```

x/ccv/provider/types/msg.go
```diff
@@ -203,3 +209,163 @@ func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
...
+func (msg MsgOptOut) ValidateBasic() error {
+	if strings.TrimSpace(msg.ChainId) == "" {
+		return errorsmod.Wrapf(ErrInvalidConsumerChainID, "chainId cannot be blank")
+	}
+	// It is possible to assign keys for consumer chains that are not yet approved.
+	// This can only be done by a signing validator, but it is still sensible
+	// to limit the chainID size to prevent abuse.
+	if 128 < len(msg.ChainId) {
+		return errorsmod.Wrapf(ErrInvalidConsumerChainID, "chainId cannot exceed 128 length")
+	}
```
mpoke, line 309: Not needed as we return error when handling the message if the chain ID is not found. There is not risk to spam the store. 
```suggestion
```

x/ccv/provider/types/msg.go
```diff
@@ -203,3 +209,163 @@ func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
...
+}
+
+func (msg MsgSetConsumerCommissionRate) ValidateBasic() error {
+	if strings.TrimSpace(msg.ChainId) == "" {
+		return errorsmod.Wrapf(ErrInvalidConsumerChainID, "chainId cannot be blank")
+	}
+
+	if 128 < len(msg.ChainId) {
+		return errorsmod.Wrapf(ErrInvalidConsumerChainID, "chainId cannot exceed 128 length")
+	}
```
mpoke, line 346: Not needed as we call `IsConsumerProposedOrRegistered(ctx, chainID)` when handling this message. 
```suggestion
```

x/ccv/provider/types/msg.go
mpoke: Actually also not needed as we call IsConsumerProposedOrRegistered when handling MsgOptIn messages. We should remove this also from MsgAssignConsumerKey.

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,225 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+// on the condition that the given consumer chain exists.
+func (k Keeper) HandleSetConsumerCommissionRate(ctx sdk.Context, chainID string, providerAddr types.ProviderConsAddress, commissionRate sdk.Dec) error {
+	// check that the consumer chain exists
+	if !k.IsConsumerProposedOrRegistered(ctx, chainID) {
+		return errorsmod.Wrapf(
+			types.ErrUnknownConsumerChainId,
+			"unknown consumer chain, with id: %s", chainID)
+	}
+	// set per-consumer chain commission rate for the validator address
+	k.SetConsumerCommissionRate(
```
mpoke, line 284: Do we want to have a minimum consumer commission rate, like we have on the Hub?

app/provider/app.go
```diff
@@ -470,12 +471,15 @@ func New(
 		app.BankKeeper,
 		scopedTransferKeeper,
 	)
-	transferModule := transfer.NewAppModule(app.TransferKeeper)
-	ibcmodule := transfer.NewIBCModule(app.TransferKeeper)
+
+	// Add an IBC middleware callback to track the consumer rewards
+	var transferStack porttypes.IBCModule
+	transferStack = transfer.NewIBCModule(app.TransferKeeper)
+	transferStack = provider.NewIBCMiddleware(transferStack, app.ProviderKeeper)
```
mpoke, line 478: Need to be properly tested on Gaia as the transfer stack is much larger. @sainoe we should have upgrading instructions on how to integrate this and if there are any constraints on the order in the stack.

mpoke (commented):
Great work team. Here are some comments after a partial review.

mpoke (dismissed):
Asking for changes mainly due to this https://github.com/cosmos/interchain-security/pull/1809#discussion_r1577712376

x/ccv/provider/ibc_middleware.go
```diff
@@ -0,0 +1,242 @@
...
+func (im IBCMiddleware) OnRecvPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	relayer sdk.AccAddress,
+) exported.Acknowledgement {
+	// executes the IBC transfer OnRecv logic
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
+
+	// execute the middleware logic only if the sender is a consumer chain
+	consumerID, err := im.keeper.IdentifyConsumerChainIDFromIBCPacket(ctx, packet)
```
mpoke, line 120: Move this call inside the `if ack.Success()` block.

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,225 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
+	return k.bankKeeper.GetAllBalances(
+		ctx,
+		k.accountKeeper.GetModuleAccount(ctx, types.ConsumerRewardsPool).GetAddress(),
+	)
+}
+
+// ComputeConsumerTotalVotingPower returns the validator set total voting power
+// for the given consumer chain
+func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string) (totalPower int64) {
```
mpoke, line 241: Move function to `validator_set_update.go`

x/ccv/provider/migrations/migrator.go
```diff
@@ -7,6 +7,7 @@ import (
 	providerkeeper "github.com/cosmos/interchain-security/v4/x/ccv/provider/keeper"
 	v3 "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/v3"
 	v4 "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/v4"
+	vPSS "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/vPSS"
```
mpoke, line 10: This should be v5 and the consensus version in module.go should be bump to 5.

mpoke (commented):
Finished the review without migration. See my comments below.

x/ccv/provider/keeper/distribution.go
p-offtermatt: Seems like a good idea; should probably just use the same value rather than doing smth more complicated like having a different minimal rate per consumer?

x/ccv/provider/keeper/distribution.go
```diff
@@ -57,32 +68,225 @@ func (k Keeper) GetAllConsumerRewardDenoms(ctx sdk.Context) (consumerRewardDenom
...
+	}
+
+	k.SetConsumerRewardsAllocation(ctx, chainID, allocation)
+	return rewardsToSend, nil
+}
+
+// consumer reward pools getter and setter
+
+// GetConsumerRewardsAllocation returns the consumer rewards allocation for the given chain ID
+func (k Keeper) GetConsumerRewardsAllocation(ctx sdk.Context, chainID string) (pool types.ConsumerRewardsAllocation) {
```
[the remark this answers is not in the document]
p-offtermatt: Added a simple UT just for the setter/getter
(fc09cf3e)
@sainoe can you confirm that is enough?

x/ccv/provider/keeper/distribution.go
p-offtermatt: Why? It's called in this file, but not in validator_set_update, and it also has nothing to do with the actual valset updates

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -225,6 +225,124 @@ func (k Keeper) QueryParams(c context.Context, _ *types.QueryParamsRequest) (*ty
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// get all the consumer chains for which the validator is either already
+	// opted-in, currently a consumer validator or if its voting power is within the TopN validators
+	consumersToValidate := []string{}
+	for _, consumer := range k.GetAllConsumerChains(ctx) {
+		chainID := consumer.ChainId
+		provAddr := types.NewProviderConsAddress(consAddr)
+		if !k.IsOptedIn(ctx, chainID, provAddr) && !k.IsConsumerValidator(ctx, chainID, provAddr) {
+			// check that the validator voting power isn't in the TopN
```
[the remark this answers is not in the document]
p-offtermatt: Yes there is - in an epoch where the validator has moved into the top N, but before the epoch ends

x/ccv/provider/keeper/hooks.go
```diff
@@ -35,8 +36,64 @@ func (k *Keeper) Hooks() Hooks {
...
+	consAddr, err := validator.GetConsAddr()
+	if err != nil {
+		ctx.Logger().Error(err.Error())
+		return nil
+	}
+
+	// get all consumers where the validator is in the validator set
 	for _, chain := range h.k.GetAllConsumerChains(ctx) {
-		consumerChainIDS = append(consumerChainIDS, chain.ChainId)
+		if h.k.IsConsumerValidator(ctx, chain.ChainId, types.NewProviderConsAddress(consAddr)) {
```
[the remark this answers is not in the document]
p-offtermatt: Don't think so. I looked into adding an integration test, but I think an e2e test might be easier.

app/provider/app.go
sainoe: Agree. I will create an issue to make sure we test the transfer stack in Gaia.

x/ccv/provider/keeper/distribution.go
p-offtermatt: I will try to do this one

x/ccv/provider/keeper/keeper.go
```diff
@@ -1136,3 +1143,163 @@ func (k Keeper) GetAllRegisteredAndProposedChainIDs(ctx sdk.Context) []string {
...
+func (k Keeper) SetConsumerCommissionRate(
+	ctx sdk.Context,
+	chainID string,
+	providerAddr types.ProviderConsAddress,
+	commissionRate sdk.Dec,
+) {
+	store := ctx.KVStore(k.storeKey)
+	bz, err := commissionRate.Marshal()
+	if err != nil {
+		panic(fmt.Errorf("consumer commission rate marshalling failed: %s", err))
```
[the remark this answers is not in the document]
sainoe: Agree with returning an error. I will update it.
>provided can be marshaled
I think the marshalling works, since it's now UTed, and also tested during the devnet. Is it correct @p-offtermatt ?

x/ccv/provider/keeper/proposal.go
```diff
@@ -189,6 +189,12 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, closeChan boo
 		k.DeleteVscSendTimestampsForConsumer(ctx, chainID)
 	}
 
+	// delete consumer commission rate
```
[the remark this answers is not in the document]
sainoe: Good catch. We've to make sure the rewards have been distributed to validators before deleting `ConsumerRewardsAllocation`.
I will create an issue for it since it's not trivial.

x/ccv/provider/keeper/relay.go
```diff
@@ -391,6 +401,15 @@ func (k Keeper) HandleSlashPacket(ctx sdk.Context, chainID string, data ccv.Slas
 		"infractionType", data.Infraction,
 	)
 
+	// Check that the validator belongs to the consumer chain valset
+	if !k.IsConsumerValidator(ctx, chainID, providerConsAddr) {
```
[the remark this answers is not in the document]
sainoe: Good catch!

x/ccv/provider/migrations/migrator.go
sainoe: Isn't it cleaner to do in Gaia since ICS V5 already exists?

x/ccv/provider/types/codec.go
sainoe: Good catch.

x/ccv/provider/keeper/distribution.go
p-offtermatt: Done, added also some extra validation on the rate which we didn't have before:
https://github.com/cosmos/interchain-security/pull/1834/files

x/ccv/provider/module.go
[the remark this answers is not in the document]
sainoe: Even though I don't see how a consumer commission unmarshalling can panic when the marshalling succeeds, let's handle this gracefully and log it.

x/ccv/provider/keeper/grpc_query.go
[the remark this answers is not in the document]
p-offtermatt: Yeah, I will resolve this. not sure the other PR should be targetting main now; we can merge it after this main pss PR, though

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -0,0 +1,148 @@
...
+
+			// if validator already exists it gets overwritten
+			k.SetOptedIn(ctx, chainID, types.NewProviderConsAddress(consAddr))
+		} // else validators that do not belong to the top N validators but were opted in, remain opted in
+	}
+}
+
+// ComputeMinPowerToOptIn returns the minimum power needed for a validator (from the bonded validators)
+// to belong to the `topN` validators. `chainID` is only used for logging purposes.
+func (k Keeper) ComputeMinPowerToOptIn(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator, topN uint32) int64 {
```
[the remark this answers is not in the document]
p-offtermatt: Good idea. I'll work on that

MSalopek (commented):
Adding some comments. I'm still reviewing the core logic.

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -13,6 +13,8 @@ import "ibc/lightclients/tendermint/v1/tendermint.proto";
 import "tendermint/crypto/keys.proto";
 import "cosmos/evidence/v1beta1/evidence.proto";
 import "cosmos/base/v1beta1/coin.proto";
+import "amino/amino.proto";
```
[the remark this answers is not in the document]
insumity: It is used in `ConsumerRewardsAllocation` with `(amino.dont_omitempty)   = true`.

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -61,3 +64,49 @@ message MsgSubmitConsumerDoubleVoting {
...
+message MsgOptIn {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+  // the chain id of the consumer chain to opt in to
+  string chain_id = 1;
+  // the validator address on the provider
+  string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+  // (optional) the consensus public key to use on the consumer in json string format corresponding to proto-any,
+  // for example `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}`
+  // we can set `consumer_key = ""` if we do not consider the `consumer_key`
```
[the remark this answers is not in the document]
insumity: Is trying to communicate on how to use the `consumer_key` field. This comment is on par with the [MsgAssignConsumerKey](https://github.com/cosmos/interchain-security/blob/6e2618b15ff39de6e7ecf983d1fb85d2d80143d1/proto/interchain_security/ccv/provider/v1/tx.proto#L30) comment.

Just slightly rephrased the comment. Let me know if you have any concerns.

x/ccv/consumer/keeper/distribution_test.go
```diff
@@ -90,3 +91,39 @@ func TestAllowedRewardDenoms(t *testing.T) {
 	require.Equal(t, allowedDenoms[0], "ustake")
 	require.True(t, strings.HasPrefix(allowedDenoms[1], "ibc/"))
 }
+
+func TestSetConsumerRewardsAllocation(t *testing.T) {
```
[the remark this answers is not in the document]
p-offtermatt: Yeah, my bad!

x/ccv/provider/keeper/keeper.go
```diff
@@ -1136,3 +1143,163 @@ func (k Keeper) GetAllRegisteredAndProposedChainIDs(ctx sdk.Context) []string {
...
+	return !found || topN == 0
+}
+
+func (k Keeper) SetOptedIn(
+	ctx sdk.Context,
+	chainID string,
+	providerConsAddress types.ProviderConsAddress,
+) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.OptedInKey(chainID, providerConsAddress), []byte{})
```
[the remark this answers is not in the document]
insumity: We can add this in a future iteration where we could use the height to see if a validator should get rewards or not. I do not believe it would add value at this stage to simply store the height without using it.

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -100,6 +101,33 @@ service Query {
...
+  rpc QueryConsumerChainsValidatorHasToValidate(
+    QueryConsumerChainsValidatorHasToValidateRequest)
+    returns (QueryConsumerChainsValidatorHasToValidateResponse) {
+      option (google.api.http).get =
+    "/interchain_security/ccv/provider/consumer_chains_per_validator";
+    }
+
+   // QueryValidatorConsumerCommissionRate returns the commission rate a given
+   // validator charges on a given consumer chain
+  rpc QueryValidatorConsumerCommissionRate(
```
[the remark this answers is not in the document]
p-offtermatt: Was added

x/ccv/provider/keeper/partial_set_security.go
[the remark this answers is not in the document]
p-offtermatt: There is a PR for it here: https://github.com/cosmos/interchain-security/pull/1840
But in retrospect I think it's not necessary for now, and I'm not confident making the change now, since it touches on pretty core logic.
It's something that could be added later if we deem it necessary; it needs a migration either way.

x/ccv/provider/keeper/keeper.go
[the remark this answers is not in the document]
p-offtermatt: Dante tested it manually, yeah

x/ccv/provider/types/msg.go
p-offtermatt: We decided to leave this because the sanity check doesn't hurt

x/ccv/provider/types/msg.go
p-offtermatt: See above

x/ccv/provider/types/msg.go
p-offtermatt: See above

x/ccv/provider/keeper/keeper.go
```diff
@@ -1136,3 +1143,163 @@ func (k Keeper) GetAllRegisteredAndProposedChainIDs(ctx sdk.Context) []string {
...
+) (sdk.Dec, bool) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.ConsumerCommissionRateKey(chainID, providerAddr))
+	if bz == nil {
+		return sdk.ZeroDec(), false
+	}
+
+	cr := sdk.Dec{}
+	if err := cr.Unmarshal(bz); err != nil {
+		k.Logger(ctx).Error("consumer commission rate unmarshalling failed: %s", err)
```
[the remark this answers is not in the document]
p-offtermatt: Error logging seems like a reasonable failsafe and we want to avoid the panic during EndBlock

x/ccv/types/expected_keepers.go
```diff
@@ -53,6 +57,7 @@ type StakingKeeper interface {
 	GetUnbondingDelegationsFromValidator(ctx sdk.Context, valAddr sdk.ValAddress) (ubds []stakingtypes.UnbondingDelegation)
 	GetRedelegationsFromSrcValidator(ctx sdk.Context, valAddr sdk.ValAddress) (reds []stakingtypes.Redelegation)
 	GetUnbondingType(ctx sdk.Context, id uint64) (unbondingType stakingtypes.UnbondingType, found bool)
+	BlockValidatorUpdates(ctx sdk.Context) []abci.ValidatorUpdate
```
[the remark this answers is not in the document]
p-offtermatt: Removed

insumity (approved):

sainoe (approved):

x/ccv/provider/migrations/migrator.go
MSalopek: Clarification for this conversation:

The comment says to change
`vPSS "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/vPSS"` 
to
`v5 "github.com/cosmos/interchain-security/v4/x/ccv/provider/migrations/v5"`.

> Note: consensus version 5 is used for the provider on [ICS@v5.x](mailto:ICS@v5.x).

You can use 5 here regardless. We'll figure out a clean way to handle the provider upgrade on ICS@v5.

This comment says that consensus version 5 is already used for [another release (ICS v5.0.0)](https://github.com/cosmos/interchain-security/releases/tag/v5.0.0-rc0).

No changes should be made to the go.mod file

PR #1814, opened 2024-04-23
feat: add consumer params upgrade message

MSalopek:
This PR adds support for `MsgUpdateParams` on the consumer.

In order to use this feauture the consumers must assign an `authority` account during build time.

Feature is available for consumers using the `x/ccv/democracy` module via `x/gov` or with a custom module using `x/ccv/consumer`.


p-offtermatt (approved):
LGTM

PR #1817, opened 2024-04-24
chore: add consumer and provider migrations for ICS v5

MSalopek:
Closes: #1818 

This PR refactors migrations on `release/v5.x.`

Consumer consensus version: `3`
Provider consensus version: `5`

Versions were bumped because module parameters got migrated to their respective modules instead of relying on `x/params` and legacy param subspaces.

This is done in accordance to cosmos-sdk migration v50 upgrade instructions.


x/ccv/consumer/module.go
```diff
@@ -127,6 +127,9 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(consumertypes.ModuleName, 1, m.Migrate1to2); err != nil {
```
mpoke, line 127: Remove as v3.0.0 and v3.1.0 have consensus version 1 and they are both deprecated. The consumer module on >= v3.2.0 has consensus version 2.

x/ccv/provider/module.go
```diff
@@ -130,7 +130,7 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 		panic(fmt.Sprintf("failed to register migrator for %s: %s -- from 3 -> 4", providertypes.ModuleName, err))
 	}
 	if err := cfg.RegisterMigration(providertypes.ModuleName, 4, migrator.Migrate4to5); err != nil {
-		panic(fmt.Sprintf("failed to register migrator for %s: %s -- from 4 -> 4", providertypes.ModuleName, err))
+		panic(fmt.Sprintf("failed to register migrator for %s: %s -- from 4 -> 5", providertypes.ModuleName, err))
```
mpoke, line 133: nit: for consistency, also update the panic above when registering Migrate2to3

mpoke (approved):

x/ccv/consumer/module.go
MSalopek: Let's leave the migrations as they are written now.

Removing the migrations would make it difficult to figure out what was happening and when. Having all migrations in front of you makes it somewhat easier.

This is from experience we got during provider v2 -> v3 migration which was a mess and some previously registered migrations were just deleted. Due to the deletion previous code versions needed to be cross-checked to make sure we're not dropping something.

x/ccv/provider/module.go
MSalopek: done

Issue #1045, opened 2023-06-21
refactor PutUnbondingOnHold handling


MSalopek:
## Problem

cosmos-sdk v0.47 introduced a feature called `CancelUnbonding` which enables users to cancel unbonding operations.

To make this operation feasible, `UnbondingDelegationEntries` that were created on the same height are now merged into a single entry (previously there would be multiple entries).

This created a problem for `AfterUnbondingInitiated` hook users where the hook could be called with an `UnbondingId` that does not exist (because it was merged into another unbonding entry).

Since the changes are state-machine breaking, the feature will be available in `cosmos-sdk > 0.47`. To accomodate for the change in behaviour `provider` code had to be modified to handle cases where `PutUnbondingOnHold` operation could fail because the unbonding ID does not actually exist in state.

This change should be reverted when ICS migrates to `cosmos-sdk v0.47`.

Context:
Previous handling of `PutUnbondingOnHold` would treat missing `unbondingId` as corrupted state. Since missing `unbondingId`s are to be expected, we cannot panic the chain any more since that is a valid case and not a state corruption.

## Closing criteria

Refactor `PutUnbondingOnHold` calls to reflect changes in `cosmos-sdk >0.47`.

The behavior must match the one in [ICS@v1.2.0-multiden](https://github.com/cosmos/interchain-security/blob/v1.2.0-multiden/x/ccv/provider/keeper/hooks.go#L53).

## Problem details

Additional problem context:
* https://github.com/cosmos/cosmos-sdk/pull/16043/files
* https://github.com/cosmos/cosmos-sdk/issues/16042

Original PR & Issue introducing the change in cosmos-sdk:
* https://github.com/cosmos/cosmos-sdk/pull/12967
* https://github.com/cosmos/cosmos-sdk/issues/12931

This is the PR that made the change that needs to be reverted:
* https://github.com/cosmos/interchain-security/pull/1120

PR #1819, opened 2024-04-24
fix!: revert PutUnbondingOnHold to desired behaviour

MSalopek:
Closes: #1045.

This reverts the `AfterUnbondingInitiated` behavior to the version used on [ICS@v1.2.0-multiden](https://github.com/cosmos/interchain-security/blob/v1.2.0-multiden/x/ccv/provider/keeper/hooks.go#L53).

See issue for more context.
____

TLDR;
cosmos-sdk v0.50 has required state breaking changes that could not be backported to cosmos-sdk v0.47.
This PR takes note of that and enables an additional correctness check on provider.


mpoke (approved):
LGTM. Is there a way to test this?

MSalopek:
> LGTM. Is there a way to test this?

Practically - you need to create inconsistent staking state. I'll try adding an integration test.

PR #1829, opened 2024-04-29
chore: add release/v5.x bots targets

MSalopek:
Closes: XXXXX


p-offtermatt (approved):

p-offtermatt:
just to make sure, we don't need to in turn remove any old release from the bots at this point, right?

MSalopek:
> just to make sure, we don't need to in turn remove any old release from the bots at this point, right?

Maybe later, it's not the point of this PR

PR #1834, opened 2024-04-30
fix: Validate consumer commission rate against minimal rate

p-offtermatt:
Validate the rate against the minimum when setting the consumer commission rate.
Note this means that the minimum for all consumers is simply the one on the provider.

I messed up the template, but it's a small PR anyways.


sainoe (approved):
LGTM. thanks!

insumity (approved):
LGTM
However, it's not clear to me why we need to enforce the provider's commission rate on consumer chains?

p-offtermatt:
I guess optimally, we would have a minimum commission per consumer chain. This seems like the second best thing, while being much less work

PR #1840, opened 2024-04-30
feat!: store the minimal power to be in the top N on EndBlock, instead of computing on-the-fly

p-offtermatt:
## Description

Closes: #XXXX

Compute the minimal power in the top N on EndBlock, when we compute it anyways for sending VSCPackets.

This should, at worst, delay the ability to opt out by an epoch
(intuitively, when someone is allowed to opt out not because their own power decreases, but because others powers increase, they would still compare against the old, stored MinPowerInTopN and not be allowed to opt out.
This would be solved after the epoch ends and the MinPowerInTopN is updated, however.)

NOTE: Needs a migration; the MInPowerToOptIn needs to be initialized during the migration

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/ccv/provider/keeper/keeper.go
+++ b/x/ccv/provider/keeper/keeper.go
@@ -1313,7 +1313,7 @@ func (k Keeper) SetMinimumPowerInTopN(
 	buf := make([]byte, 8)
 	binary.BigEndian.PutUint64(buf, uint64(power))
 
-	store.Set(types.TopNKey(chainID), buf)
+	store.Set(types.MinimumPowerInTopNKey(chainID), buf)
 }
 
 // GetMinimumPowerInTopN returns the minimum power required for a validator to be in the top N
@@ -1323,7 +1323,7 @@ func (k Keeper) GetMinimumPowerInTopN(
 	chainID string,
 ) (int64, bool) {
 	store := ctx.KVStore(k.storeKey)
-	buf := store.Get(types.TopNKey(chainID))
+	buf := store.Get(types.MinimumPowerInTopNKey(chainID))
 	if buf == nil {
 		return 0, false
 	}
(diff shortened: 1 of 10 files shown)


x/ccv/provider/keeper/keeper.go
```diff
@@ -1300,3 +1300,32 @@ func (k Keeper) DeleteConsumerCommissionRate(
 	store := ctx.KVStore(k.storeKey)
 	store.Delete(types.ConsumerCommissionRateKey(chainID, providerAddr))
 }
+
+// SetMinimumPowerInTopN sets the minimum power required for a validator to be in the top N
+// for a given consumer chain.
+func (k Keeper) SetMinimumPowerInTopN(
+	ctx sdk.Context,
+	chainID string,
```
insumity, line 1308: Why do we need the `chainID`?
Only reason that `ComputeMinPowerInTopN` has a `chainID` argument is for logging purposes.

x/ccv/provider/types/keys.go
```diff
@@ -166,6 +166,10 @@ const (
 	// per validator per consumer chain
 	ConsumerCommissionRatePrefix
 
+	// ConsumerCommissionRateDenomPrefix is the byte prefix for storing the
```
insumity, line 169: Rename `ConsumerCommissionRateDenomPrefix` to `MinimumPowerInTopNBytePrefix`.

insumity:
Thanks for the change.

I'm not in favor of adding this to PSS. I do not see the benefits (i.e., slightly faster query-ing and opting-out) trumping the slight additional complexity that this PR introduces (i.e., additional state and potentially opting-out delays).

x/ccv/provider/keeper/keeper.go
p-offtermatt: Because different chains have different top N, so naturally have different min power to be forced to validate

x/ccv/provider/types/keys.go
p-offtermatt: oops :D

p-offtermatt:
Replaced by #1952

PR #1879, opened 2024-05-14
docs: Add draft ADR for validators outside of the active set

p-offtermatt:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

- [ ] provided a link to the relevant issue or specification
- [x] reviewed "Files changed" and left comments if necessary 
- [x] confirmed all CI checks have passed

### Reviewers Checklist

I have...

### DIFF
--- a/docs/docs/adrs/adr-017-allowing-inactive-validators.md
+++ b/docs/docs/adrs/adr-017-allowing-inactive-validators.md
@@ -1,0 +1,67 @@
+---
+sidebar_position: 18
+title: ADR Template
+---
+# ADR [17]: [Allowing validators outside the active set to validate on consumer chains]
+
+## Changelog
+* [date]: [changelog]
+
+## Status
+
+Proposed
+
+## Context
+
+Currently, only validators in the active set on the provider can validate on consumer chains.
+This limits the number of validators that can validate on consumer chains. Validators outside of the active set might be willing
+to validate on consumer chains, but we might not want to make the provider validator set larger, e.g. to not put more strain on the consensus engine.
+This runs the risk of leaving consumer chains with too few validators.
+
+The purpose of this ADR is to allow validators that are *not* part of the consensus process on the provider chain (because they are inactive)
+to validate on consumer chains.
+
+## Decision
+
+The proposed solution is to:
+* Increase the validator set size in the staking module to a very large number (e.g. 500)
+* Introduce an additional parameter in the provider module, called MaxProviderConsensusValidators
+* When the validator updates are passed to the consensus engine of the provider during EndBlock, the provider module will simply only pass the first MaxProviderConsensusValidators validators by voting power to the consensus engine
+
+To facilitate this, the provider module will need to:
+* keep the last validator set sent to the consensus engine
+* load the current validator set from the staking module
+* diff the two sets, and send the diff to the consensus engine
+
+Extra considerations:
+* Migration: In the migration, the last consensus validator set would just be sent to the last active validator set from the view of the staking module. Existing consumer chains need to be migrated to have a validator set size cap (otherwise, they could end up with a huge validator set including all the staking-but-not-consensus-active validators from the provider chain)
+* Slashing: Validators that are not part of the active set on the provider chain can still be slashed on the consumer chain, but they *should not* be slashed for downtime on the provider chain. Will those validators accrue missed blocks? If yes, we probably need to make changes in the slashing module to not continuously slash them for downtime on the provider
+* Rewards: Validators that are not part of the active set on the provider chain can still receive rewards on the consumer chain, but they *should not* receive rewards on the provider chain.
+* Where else might the staking module validators be used? We need to carefully assess whether we need to change these references and direct them to the "actual active set" of the provider chain instead, or whether they can still go to the staking module
+
+Comms:
+* This change needs to be communicated, in particular to frontends. They need to be aware that there is a difference between "active validator on the provider chain", and "part of the consensus validator set on the provider chain".
+
+## Consequences
+
+### Positive
+
+* Validators outside of the active set can validate on consumer chains without having an impact on the consensus engine of the provider chain
+* We can do this change without further forking from the standard Cosmos SDK modules
+
+### Negative
+
+* We need to be very careful in making sure all existing references to the validator set are updated to refer to the "actual active set" of the provider chain if necessary
+* In the worst case, this might mean we need many changes to existing Cosmos SDK modules (and this would negate one of the positives of this solution)
+
+### Neutral
+
+## Alternative considerations
+
+We could instead adapt the *staking module* with a similar change.
+This might be better if it turns out that the staking module active set is used in many other places.
+
+## References
+
+* [adr-016-securityaggregation.md] has similar concerns where the staking validator set will differ from the consensus validator set
+* 


p-offtermatt:
@coderabbitai pause
please don't review this PR

jtremback (commented):
LGTM!

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,66 @@
+---
+sidebar_position: 18
+title: ADR Template
+---
+# ADR [17]: [Allowing validators outside the active set to validate on consumer chains]
```
insumity, line 5: nit: remove `[]` square brackets from ADR number and title.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Changelog
```
insumity, line 7: Maybe add a date for the first "Initial Draft"?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,68 @@
...
+* Introduce an additional parameter in the provider module, called MaxProviderConsensusValidators
+* When the validator updates are passed to the consensus engine of the provider during EndBlock, the provider module will simply only pass the first MaxProviderConsensusValidators validators by voting power to the consensus engine
+
+To facilitate this, the provider module will need to:
+* keep the last validator set sent to the consensus engine
+* load the current validator set from the staking module
+* diff the two sets, and send the diff to the consensus engine
+
+Extra considerations:
+* Migration: In the migration, the last consensus validator set would just be sent to the last active validator set from the view of the staking module. Existing consumer chains need to be migrated to have a validator set size cap (otherwise, they could end up with a huge validator set including all the staking-but-not-consensus-active validators from the provider chain)
```
insumity, line 38: Could we use ["unbonded"](https://docs.cosmos.network/v0.46/modules/staking/01_state.html) instead of "staking-but-not-consensus-active"?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,68 @@
...
+* Slashing: Validators that are not part of the active set on the provider chain can still be slashed on the consumer chain, but they *should not* be slashed for downtime on the provider chain. Will those validators accrue missed blocks? If yes, we probably need to make changes in the slashing module to not continuously slash them for downtime on the provider
```
insumity, line 39: What does it mean to  "still be slashed on the consumer chain"? Is provider chain meant here?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,68 @@
...
+### Neutral
+
+## Alternative considerations
+
+We could instead adapt the *staking module* with a similar change.
+This might be better if it turns out that the staking module active set is used in many other places.
+
+## References
+
+* [adr-016-securityaggregation.md] has similar concerns where the staking validator set will differ from the consensus validator set
```
insumity, line 67: nit: The `[adr-016-securityaggregation.md]` does not point anywhere.

insumity:
I believe I understand the general idea: Increase the `max_validators` of the staking module to 500 and then work on top of that to do filtering, etc. so that the provider only still "sees" 180 validators but the remaining 320 (= 500 - 180) can still slashed, etc. 

However I was thinking the following: Unbonded validators can opt in or assign a consumer key to a consumer chain. So, we could send the unbonded validators' public keys down to the consumer chains. For those validators to actually secure a consumer chain we need:
1.  for them to be slashable;
2.  to get rewards. 

For 1. couldn't we just slash those unbonded validators on our own (without utilizing directly the `Slash` method) but using something like [`RemoveValidatorTokens`](https://github.com/cosmos/cosmos-sdk/blob/9a18e855c3a095d0140fabb9fb32e69b3ccbc8a9/x/staking/keeper/validator.go#L165). We already do part of slashing on our [own](https://github.com/cosmos/interchain-security/blob/2b5b6cfba035e942bd35fa32c2da8d8b26a6b487/x/ccv/provider/keeper/consumer_equivocation.go#L418) so this could be adapted. Similarly for 2. we could adapt the reward distribution to give rewards to non-bonded validators.

A final note on the approach: Do we want to have a way for a validator to be able to secure a consumer chain _without being forced_ to secure the Hub? If not, then it might make sense to simply increase the validator-set size on the Hub. After having a discussion with @MSalopek, it seems that increasing the validator-set size would lead to an increase in block times and this makes sense. However, it's not clear to me how big that increase would be. If it is in the range of a few hundred milliseconds then just increasing the validator-set might be a viable approach (?)

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: should read "slashed for downtime on a consumer chain"

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: Thanks, will fix!

cordt-sei (commented):
I was just discussing this with someone the other day, really great idea. At worst you are improving decentralization, at best you are rewarding actual contributors that oltherwise get overlooked.

faddat:
this is extremely cool

p-offtermatt:
> I believe I understand the general idea: Increase the `max_validators` of the staking module to 500 and then work on top of that to do filtering, etc. so that the provider only still "sees" 180 validators but the remaining 320 (= 500 - 180) can still slashed, etc.
> 
> However I was thinking the following: Unbonded validators can opt in or assign a consumer key to a consumer chain. So, we could send the unbonded validators' public keys down to the consumer chains. For those validators to actually secure a consumer chain we need:
> 
>     1. for them to be slashable;
> 
>     2. to get rewards.
> 
> 
> For 1. couldn't we just slash those unbonded validators on our own (without utilizing directly the `Slash` method) but using something like [`RemoveValidatorTokens`](https://github.com/cosmos/cosmos-sdk/blob/9a18e855c3a095d0140fabb9fb32e69b3ccbc8a9/x/staking/keeper/validator.go#L165)? We already do part of slashing on our [own](https://github.com/cosmos/interchain-security/blob/2b5b6cfba035e942bd35fa32c2da8d8b26a6b487/x/ccv/provider/keeper/consumer_equivocation.go#L418) so this could be adapted. Similarly for 2. we could adapt the reward distribution to give rewards to non-bonded validators.


We discussed offline, but just to summarize:

* I listed your alternative solution in the ADR. It's a good one, too, I will try that implementation instead of the current prototype to see if its significantly easier.

> A final note on the approach: Do we want to have a way for a validator to be able to secure a consumer chain _without being forced_ to secure the Hub? If not, then it might make sense to simply increase the validator-set size on the Hub. After having a discussion with @MSalopek, it seems that increasing the validator-set size would lead to an increase in block times and this makes sense. However, it's not clear to me how big that increase would be. If it is in the range of a few hundred milliseconds, then just increasing the validator-set might be a viable approach (?)

I think we want a solution where the potential validator set for consumer chains could be thousands of validators.
We want to remove the inherent link between "consensus validators of the hub" and "potential validators for consumer chains" so that e.g. hub governance can just increase the set of potential validators without even needing to consider the impact on consensus.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
mpoke, line 1: Wait for https://github.com/cosmos/interchain-security/pull/1910 to be merged and update intro.md

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Status
+
+Proposed
+
+## Context
+
+Currently, only validators in the active set on the provider can validate on consumer chains.
+This limits the number of validators that can validate on consumer chains. Validators outside of the active set might be willing
```
mpoke, line 18:
```suggestion
Currently, only validators in the active set on the provider can validate on consumer chains, which limits the number of validators that can participate in Interchain Security (ICS). 
Validators outside of the active set might be willing
```

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Decision
+
+The proposed solution is to:
+* Increase the validator set size in the staking module to a very large number (e.g. 500)
+* Introduce an additional parameter in the provider module, called MaxProviderConsensusValidators
+* When the validator updates are passed to the consensus engine of the provider during EndBlock, the provider module will simply only pass the first MaxProviderConsensusValidators validators by voting power to the consensus engine
+
+To facilitate this, the provider module will need to:
+* keep the last validator set sent to the consensus engine
```
mpoke, line 33: The staking module already keeps that for the 500 validators. You just need to select the first MaxProviderConsensusValidators.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+* load the current validator set from the staking module
+* diff the two sets, and send the diff to the consensus engine
+
+Extra considerations:
+* Migration: In the migration, the last consensus validator set would just be sent to the last active validator set from the view of the staking module. Existing consumer chains need to be migrated to have a validator set size cap (otherwise, they could end up with a huge validator set including all the staking-but-not-consensus-active validators from the provider chain)
```
mpoke, line 38: This is unclear. What the last consensus validator set? Also, do you "would just be SET to the last active validator set"?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+* Slashing: Validators that are not part of the active set on the provider chain can still be slashed for downtime on a consumer chain, but they *should not* be slashed for downtime on the provider chain. Will those validators accrue missed blocks? If yes, we probably need to make changes in the slashing module to not continuously slash them for downtime on the provider
```
mpoke, line 39: > Will those validators accrue missed blocks?

Cf. the [code](https://github.com/cosmos/cosmos-sdk/blob/c059f66ff4264685969a5d16fa55041ed90fb0cd/x/slashing/abci.go#L22) in the slashing module, only the active valset can be slashed for downtime.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+
+At first glance, does not look problematic. When slashing, this gets the last commit info, which afaict is set by CometBFT according to what validator set was given to it.
+See https://github.com/cosmos/cosmos-sdk/blob/2f89b04b1430f244bcbaa3a1a68ef6d700e04dbe/x/slashing/abci.go#L22 for who accrues missed blocks.
```
mpoke, line 42: This shouldn't not be part of the ADR. The ADR should contain the new design and not our journey to get to the new design :)

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+
+* Rewards: Validators that are not part of the active set on the provider chain can still receive rewards on the consumer chain, but they *should not* receive rewards on the provider chain.
+* Where else might the staking module validators be used? We need to carefully assess whether we need to change these references and direct them to the "actual active set" of the provider chain instead, or whether they can still go to the staking module
+
+Comms:
+* This change needs to be communicated, in particular to frontends. They need to be aware that there is a difference between "active validator on the provider chain", and "part of the consensus validator set on the provider chain".
```
mpoke, line 48: Also not part of an ADR but rather of a deployment plan.

mpoke (commented):
Great work @p-offtermatt. See my comments below.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
+---
+sidebar_position: 18
+title: ADR Template
```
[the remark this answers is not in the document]
p-offtermatt: oops :D

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: This is talking about the *last* set we sent over, since we will need to diff that with the new set

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,85 @@
...
+* Slashing: Validators that are not part of the active set on the provider chain can still be slashed for downtime on a consumer chain, but they *should not* be slashed for downtime on the provider chain. Will those validators accrue missed blocks? If yes, we probably need to make changes in the slashing module to not continuously slash them for downtime on the provider
+
+At first glance, does not look problematic. When slashing, this gets the last commit info, which afaict is set by CometBFT according to what validator set was given to it.
+See https://github.com/cosmos/cosmos-sdk/blob/2f89b04b1430f244bcbaa3a1a68ef6d700e04dbe/x/slashing/abci.go#L22 for who accrues missed blocks.
+
+* Rewards: Validators that are not part of the active set on the provider chain can still receive rewards on the consumer chain, but they *should not* receive rewards on the provider chain.
+* Where else might the staking module validators be used? We need to carefully assess whether we need to change these references and direct them to the "actual active set" of the provider chain instead, or whether they can still go to the staking module
```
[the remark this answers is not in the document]
p-offtermatt: That's what they are supposed to be, yeah :)

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: Makes sense! Typo is also correct

p-offtermatt:
I made some adjustments and I will reopen. We are now clearer on the direction we need to go with this.
Also added some nice diagrams :)

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,111 @@
...
+
+In consequence, the provider chain can keep a reasonably-sized validator set, while giving consumer chains a much larger pool of potential validators.
+
+![inactivevals_after.png](../../figures/inactivevals_after.png)
+
+
+Some additional considerations:
+
+* Migration: In the migration, the last consensus validator set will be set to the last active validator set from the view of the staking module. Existing consumer chains are migrated to have a validator set size cap (otherwise, they could end up with a huge validator set including all the staking-but-not-consensus-active validators from the provider chain)
+* Slashing: Validators that are not part of the active set on the provider chain can still be slashed for downtime on a consumer chain, but they *are not* slashed for downtime on the provider chain.
```
insumity, line 54: What does it mean "slashed for downtime on a consumer chain"? Slashing should only take place on the provider.

insumity, line 54: nit: could slightly rephrase. When I first read "can still be slashed for downtime on a consumer chain" I thought the slashing took place on the consumer chain.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,111 @@
...
+To mitigate risks from validators with little stake, we introduce a minimum stake requirement for validators to be able to validate on consumer chains, which can be set by each consumer chain independently, with a default value set by the provider chain.
+
+Additional risk mitigations are to increase the active set size slowly, and to monitor the effects on the network closely. For the first iteration, we propose to increase the active set size to 200 validators (while keeping the consensus validators to 180), thus letting the 20 validators with the most stake outside of the active set validate on consumer chains.
+
+## Consequences
+
+### Positive
+
+* Validators outside of the active set can validate on consumer chains without having an impact on the consensus engine of the provider chain
+* Consumer chains can have a much larger validator set than the provider chain if they prefer this e.g. for decentralization reasons
```
insumity, line 70: You could add as positive the fact that a consumer chain can run its own validator node to make sure the chain does not halt if _all_ the other nodes opt out (the one you mention in the upcoming signaling proposal).

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,111 @@
...
+With the restricted size of the active set today, it’s clear that the set is at least minimally competitive and it is not trivial to spin up multiple nodes as a validator.
+
+When we make the “potential validator set” much larger, we should assume that it becomes much less competitive to be part of that set, and thus trivial for single entities to control many of those validators.
+
+#### Reputational damage is not a deterrent
+
+For validators in the active set, we typically assume that if they would misbehave, they pay a large reputational cost. This represents delegators deciding to switch validators (potentially even on chains other than the one the misbehaviour happened on), and loss of credibility in the ecosystem. With the much larger active set, it seems prudent to assume that reputational damage is not a deterrent for many validators. They might only have minimal amounts of delegated stake and control most of it themselves, so they might not be deterred from performing actions that would usually bring reputational damage.
+
+#### Additional negative consequences
+* Validators outside of the active set become bonded, even if they are not validating on any consumer chains
```
insumity, line 89: What is it negative that validators are bonded?
More general question on this: Are there any side-effects by having more validators bonded? E.g., other `cosmos-sdk` modules that would operate differently when having more bonded validators?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,111 @@
...
+
+We could instead adapt the *staking module* with a similar change.
+This might be better if it turns out that the staking module active set is used in many other places.
+
+### Allowing unbonding validators to validate
+
+Instead of increasing the active set size, we could allow validators that are unbonded (but still exist on the provider) to validate consumer chains.
+For this, we would need to:
+* Modify the VSC updates to consider the set of all validators, even unbonded ones, instead of just active ones
+* Adjust our downtime jailing/equivocation slashing logic to work correctly with unbonded validators
```
insumity, line 107: @mpoke pointed out that no matter what we do we won't be able to slash redelegations for unbonded validators, unless we keep track of redelegations on the provider. Naturally, such a solution would be substantially more complex than the one you propose. It might make sense to add a sentence here on this and on why we did not follow this alternative approach.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: It's negative because it means that those validators need to go through unbonding before they get their tokens back (but it appears this is essentially already how it works, so maybe I should take that out!

Yeah, for gov proposals, those validators now would get to vote. (I mean, the validators would already now get to vote; but they would get to vote with delegated stake). I think this is probably the way it should be, too, because if those validators have delegators, we should assume they represent their delegators exactly like active set validators represent their delegators

p-offtermatt: Although I do still think this is a negative - it's ok if a validator is bonded who is validating on at least some consumer chain, but having a validator that is bonded who is not validating anything is a bit strange to me

p-offtermatt: I'll put it as neutral for now

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+Some additional considerations:
+
+* Migration: In the migration, the last consensus validator set will be set to the last active validator set from the view of the staking module. Existing consumer chains are migrated to have a validator set size cap (otherwise, they could end up with a huge validator set including all the staking-but-not-consensus-active validators from the provider chain)
+* Slashing: Validators that are not part of the active set on the provider chain can still be jailed for downtime on a consumer chain (via an Interchain Security SlashPacket sent to the provider, who will then jail the validator), but they *are not* slashed for downtime on the provider chain.
+This is achieved without any additional changes to the slashing module, because the slashing module checks for downtime by looking at the consensus participants reported by CometBFT, and thus with the proposed solution, validators that are not part of the consensus validators on the provider chain are not considered for downtime slashing (see https://github.com/cosmos/cosmos-sdk/blob/b68da64d8e38caa0438dbf3b065d864d91e03c4b/x/slashing/abci.go#L22).
```
mpoke, line 52: Update links to use the SKD version.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+* Rewards: Validators that are not part of the active set on the provider chain can still receive rewards on the consumer chain, but they *do not* receive rewards from the provider chain. This change is
+achieved without further changes to staking or reward distributions, because similar to downtime, rewards are based on the consensus validator set (see https://github.com/cosmos/cosmos-sdk/blob/b68da64d8e38caa0438dbf3b065d864d91e03c4b/x/distribution/abci.go#L28)
+
+## Risk Mitigations
+
+To mitigate risks from validators with little stake, we introduce a minimum stake requirement for validators to be able to validate on consumer chains, which can be set by each consumer chain independently, with a default value set by the provider chain.
```
mpoke, line 58:
```suggestion
To mitigate risks from validators with small amount of stake, we introduce a minimum stake requirement for validators to be able to validate on consumer chains, which can be set by each consumer chain independently, with a default value set by the provider chain.
```

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+
+Additional risk mitigations are to increase the active set size slowly, and to monitor the effects on the network closely. For the first iteration, we propose to increase the active set size to 200 validators (while keeping the consensus validators to 180), thus letting the 20 validators with the most stake outside of the active set validate on consumer chains.
```
mpoke, line 60:
```suggestion
Additional risk mitigations are to increase the active set size slowly, and to monitor the effects on the network closely. For the first iteration, we propose to increase the active set size to 200 validators (while keeping the consensus validators to 180), thus letting the 20 validators with the most amount of stake outside of the active set validate on consumer chains.
```

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+This is achieved without any additional changes to the slashing module, because the slashing module checks for downtime by looking at the consensus participants reported by CometBFT, and thus with the proposed solution, validators that are not part of the consensus validators on the provider chain are not considered for downtime slashing (see https://github.com/cosmos/cosmos-sdk/blob/b68da64d8e38caa0438dbf3b065d864d91e03c4b/x/slashing/abci.go#L22).
+* Rewards: Validators that are not part of the active set on the provider chain can still receive rewards on the consumer chain, but they *do not* receive rewards from the provider chain. This change is
+achieved without further changes to staking or reward distributions, because similar to downtime, rewards are based on the consensus validator set (see https://github.com/cosmos/cosmos-sdk/blob/b68da64d8e38caa0438dbf3b065d864d91e03c4b/x/distribution/abci.go#L28)
+
+## Risk Mitigations
```
mpoke, line 56: I would also add an option for consumers to not allow validators outside of the provider's active set.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+
+b) do *not* take the updates for CometBFT directly from the bonded validators in the staking module, by wrapping the staking modules `EndBlocker` with a dummy EndBlocker that doesn't return any validator updates. Instead, we adjust the provider module to return validator updates on its EndBlocker. These validator updates are obtained by *filtering* the bonded validators to send only the first `MaxProviderConsensusValidators` (sorted by largest amount of stake first) many validators to CometBFT
+
+c) use the enlarged list of bonded validators from the staking module as basis for the validator set that the provider module sends to consumer chains (again after applying power shaping and filtering out validatiors that are not opted in).
+
+In consequence, the provider chain can keep a reasonably-sized consensus validator set, while giving consumer chains a much larger pool of potential validators.
+
+![inactivevals_after.png](../../figures/inactivevals_after.png)
+
+
```
mpoke, line 47: What extra queries will be added? Also, are there any changes in the state?

mpoke (commented):
Great work @p-offtermatt. See my comments below.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
jtremback: Yea this seems like a good idea

docs/docs/adrs/adr-017-allowing-inactive-validators.md
jtremback: disagree with this change

docs/docs/adrs/adr-017-allowing-inactive-validators.md
jtremback: disagree with this change

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -0,0 +1,110 @@
...
+### Positive
+
+* Validators outside of the active set can validate on consumer chains without having an impact on the consensus engine of the provider chain
+* Consumer chains can have a much larger validator set than the provider chain if they prefer this e.g. for decentralization reasons
+* Consumer chain teams can, with much less cost than today, start up their own consumer chain node to keep the chain running (in a centralized manner) even if no hub validators have opted in to validate on the chain. This is useful to stop the chain from ending up with an empty validator set and becoming recoverable only with a hardfork
+
+### Negative
+
+Allowing validators from the inactive set brings with it some additional risks.
+In general, consumer chains will now face some of the problems also faced by standalone chains. It’s reasonable to assume that the validator set on the hub has a minimal amount of operational quality due to being battle tested and decentralized, and consumer chains with validators from outside the hub active set cannot rely on this as much anymore.
```
jtremback, line 73:
```suggestion
In general, consumer chains will now face some of the problems also faced by standalone chains. It’s reasonable to assume that the validator set on the hub has a minimum amount of operational quality due to being battle tested and decentralized, and consumer chains with validators from outside the hub active set cannot rely on this as much anymore.
```

This kind of means the opposite lol

jtremback:
LGTM!!

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: I think changes to the state are useful to list. I am not convinced queries are useful to list.
Are queries really part of architecture decisions that we should document? It feels like the wrong level of detail to me, but lmk if you disagree, I can add it

mpoke: I agree. Unless the queries are essential to the design, they shouldn't be in the ADR but in the docs.

p-offtermatt:
Note: Markdown link failures unrelated, seem to have smt to do with the Gaia doc update

insumity (approved):

Issue #1926, opened 2024-05-29
Partial Set Security: Apply audit suggestions


insumity:
## Problem
The audit team looked into the PSS code and provided us with a [security report](https://github.com/user-attachments/files/15897852/Hub.2024.Q2_.Partial.Set.Security_Report.2.pdf).
.

## Closing criteria
Consider the suggestions presented in the security report and apply them to PSS.

PR #1929, opened 2024-05-30
feat!: only distribute rewards to validators that have been validating a consumer chain for some time

insumity:
## Description

Closes:  (partially) https://github.com/cosmos/interchain-security/issues/1926 and https://github.com/informalsystems/interchain-security/issues/27 

**Problem**
A validator might opt in to a consumer chain simply to retrieve rewards without validating the consumer chain for __too long__.

**Solution**
We introduce a `height` field in `ConsumerValidator`s. This `height` corresponds to the block height of the provider chain when the consumer validator first became a consumer validator. The set `height` of a `ConsumerValidator` remains stable as long as the validator remains a consumer validator. The `height` of a `ConsumerValidator` only resets if a validator stops being a consumer validator and later becomes again a consumer validator.

We introduce a new param `NumberOfEpochsToStartReceivingRewards` and a validator can only start receiving rewards after being a consumer validator for `NumberOfEpochsToStartReceivingRewards` epochs or after `NumberOfEpochsToStartReceivingRewards * BlocksPerEpoch` blocks.

Note that at the beginning of a provider chain (e.g., when the block height of the provider is still small), no validator could have been a consumer validator for more than `NumberOfEpochsToStartReceivingRewards` epochs and hence in this case we distribute rewards to validators, even though they might have not been validating the consumer chain for more than a single epoch.

Finally, note that in a consumer chain that just starts, no validator would receive rewards during the first `NumberOfEpochsToStartReceivingRewards` epochs. During this time, any rewards sent to the provider from the consumer chain through IBC transfers will be allocated to the community pool.

**Testing**
Two main things are being tested:
- the `height` of a consumer validator is not being reset if the validator remains a consumer validator (unit test `TestQueueVSCPacketsDoesNotResetConsumerValidatorsHeights`);
- that validators only receive rewards if they have been consumer validators for some time (integration test `TestAllocateTokensToConsumerValidatorsWithDifferentValidatorHeights`).

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Implemented a new reward distribution mechanism where validators receive rewards only after validating for a specific number of blocks.
	- Introduced a new parameter, `NumberOfEpochsToStartReceivingRewards`, to configure reward eligibility.

- **Documentation**
	- Updated documentation to include a warning about validator rewards eligibility and revised instructions for withdrawing rewards.

- **Tests**
	- Added tests to validate the new reward eligibility criteria and ensure accurate token allocation based on validator heights.

- **Migration**
	- Implemented functions for migrating provider chain parameters and ensuring compatibility with the new reward distribution system.


tests/integration/distribution.go
```diff
@@ -934,7 +934,7 @@ func (s *CCVTestSuite) prepareRewardDist() {
 	s.coordinator.CommitNBlocks(s.consumerChain, uint64(blocksToGo))
 }
 
-func (s *CCVTestSuite) TestAllocateTokensToValidator() {
+func (s *CCVTestSuite) TestAllocateTokensToConsumerValidators() {
```
insumity, line 937: This test was practically testing `AllocateTokensToConsumerValidators` so renamed it.

tests/integration/distribution.go
```diff
@@ -1060,6 +1060,114 @@ func (s *CCVTestSuite) TestAllocateTokensToValidator() {
...
+	// update the consumer validators
+	consuVals := providerKeeper.GetConsumerValSet(ctx, chainID)
+	// first 2 validators were consumer validators since block height 1 and hence get rewards
+	consuVals[0].Height = 1
+	consuVals[1].Height = 1
+	// last 2 validators were consumer validators since block height 2 and hence do not get rewards because they
+	// have not been consumer validators for `GetNumberOfEpochsToStartReceivingRewards * GetBlocksPerEpoch` blocks
+	consuVals[2].Height = 2
+	consuVals[3].Height = 2
+	providerKeeper.SetConsumerValSet(ctx, chainID, consuVals)
```
sainoe, line 1093: Is it necessary to do that?

tests/integration/distribution.go
```diff
@@ -1060,6 +1060,114 @@ func (s *CCVTestSuite) TestAllocateTokensToValidator() {
...
+
+	chainID := s.consumerChain.ChainID
+
+	tokens := sdk.DecCoins{sdk.NewDecCoinFromDec(sdk.DefaultBondDenom, math.LegacyNewDecFromIntWithPrec(math.NewInt(999), 2))}
+	rate := sdk.OneDec()
+	expAllocated := sdk.DecCoins{sdk.NewDecCoinFromDec(sdk.DefaultBondDenom, math.LegacyNewDecFromIntWithPrec(math.NewInt(999), 2))}
+
+	ctx, _ := s.providerCtx().CacheContext()
+	// If the provider chain has not yet reached `GetNumberOfEpochsToStartReceivingRewards * GetBlocksPerEpoch` block height,
+	// then all validators receive rewards. In this test, we want to check whether validators receive rewards or not based on how
```
sainoe, line 1080: This comment is slightly confusing. What does it imply if the provider chain hasn't reached `GetNumberOfEpochsToStartReceivingRewards * GetBlocksPerEpoch`? Why should the validators get rewards based on that?

x/ccv/provider/keeper/distribution.go
```diff
@@ -240,6 +247,13 @@ func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
 func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string) (totalPower int64) {
 	// sum the opted-in validators set voting powers
 	for _, v := range k.GetConsumerValSet(ctx, chainID) {
+
+		// only consider the voting power of a validator that would receive rewards (i.e., validator has been validating for a number of blocks)
```
sainoe, line 251: nit: since this is used in multiple places, it could be reformatted in a function like `IsEligibleToConsumerRewards(valHeight int)`.

x/ccv/provider/keeper/relay_test.go
```diff
@@ -87,6 +87,55 @@ func TestQueueVSCPackets(t *testing.T) {
...
+
+	// the height of consumer validator A should not be modified because A was already a consumer validator
+	cv, _ := providerKeeper.GetConsumerValidator(ctx, "chainID", providertypes.NewProviderConsAddress(valAConsAddr))
+	require.Equal(t, consumerValidatorA.Height, cv.Height, "the consumer validator's height was erroneously modified")
+
+	// the height of consumer validator B is set to be the same as the one of the current chain height because this
+	// consumer validator becomes a consumer validator for the first time (i.e., was not a consumer validator in the previous epoch)
+	cv, _ = providerKeeper.GetConsumerValidator(ctx, "chainID", providertypes.NewProviderConsAddress(valBConsAddr))
+	require.Equal(t, chainHeight, cv.Height, "the consumer validator's height was not correctly set")
+}
```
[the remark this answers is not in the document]
sainoe: This comment seems accurate.

x/ccv/provider/keeper/validator_set_update.go
```diff
@@ -71,6 +71,23 @@
...
+	store := ctx.KVStore(k.storeKey)
+	marshalledConsumerValidator := store.Get(types.ConsumerValidatorKey(chainID, providerAddr.ToSdkConsAddr()))
+
+	if marshalledConsumerValidator == nil {
+		return types.ConsumerValidator{}, false
+	}
+
+	var validator types.ConsumerValidator
+	if err := validator.Unmarshal(marshalledConsumerValidator); err != nil {
+		panic(fmt.Errorf("failed to unmarshal ConsumerValidator: %w", err))
```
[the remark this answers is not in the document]
sainoe: Maybe it's safer to log an error like "fail to get consumer validator info" and then return `false`. In this scenario, the consumer validator would just have its height reset, right?

tests/integration/distribution.go
insumity: No because `consVals[i] > 1` but I think this makes it clearer when reading the code.

tests/integration/distribution.go
insumity: I have a comment in this PR's description that I slightly updated about this:
> Note that at the beginning of a provider chain (e.g., when the block height of the provider is still small), no validator could have been a consumer validator for more than `NumberOfEpochsToStartReceivingRewards` epochs and hence in this case we distribute rewards to validators, even though they might have not been validating the consumer chain for more than a single epoch.

I introduced the `IsEligibleForConsumerRewards` as you suggested and added that comment there as well to make it clear. If this did not clarify things, let me know and I'll update this.

x/ccv/provider/keeper/validator_set_update.go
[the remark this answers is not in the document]
insumity: Yes, that is correct, the height would reset. This implies that the validator would never get rewards but this seems better than halting the provider chain. Fixed it.

p-offtermatt (commented):
If I get it correctly, the UX for chains with a validator set size cap is pretty bad.
Small validators that get pushed in/out of the set might not be in consumer validators long enough to get any rewards, even though they might actually be validating on the consumer chain and stay opted in all the time, right?

A large validator could even cause this maliciously, opting in/out so that they a) never stay opted in long enough to get jailed for downtime b) validators around the border get pushed in/out, but never stay in the set long enough to obtain any rewards

The bad thing is that the large validator doesn't even have to run a consumer node for this, they just opt out before they would be punished for non-signing, so it doesn't actually cost them anything to do this attack

x/ccv/provider/types/params.go
```diff
@@ -41,18 +41,30 @@ const (
 	// an epoch corresponds to 1 hour (6 * 600 = 3600 seconds).
 	// forcing int64 as the Params KeyTable expects an int64 and not int.
 	DefaultBlocksPerEpoch = int64(600)
+
+	// DefaultNumberOfEpochsToStartReceivingRewards defines the default minimum number of epochs required by a validator to validate
+	// during so that the validator can start receiving rewards. This would mean that a validator has to be a consumer validator for at least
+	// `DefaultNumberOfEpochsToStartReceivingRewards * DefaultBlocksPerEpoch` on a consumer chain to stat receiving rewards from the chain.
```
mpoke, line 47:
```suggestion
	// `NumberOfEpochsToStartReceivingRewards * BlocksPerEpoch` on a consumer chain to stat receiving rewards from the chain.
```

x/ccv/provider/module.go
mpoke, line 1: Update the consensus version

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -329,6 +332,11 @@ message ConsumerValidator {
   int64 power = 2;
   // public key the validator uses on the consumer chain during this epoch
   tendermint.crypto.PublicKey consumer_public_key = 3;
+  // height the validator had when it FIRST became a consumer validator
+  // If a validator becomes a consumer validator at height `H` and is continuously a consumer validator for all the upcoming
+  // epochs, then the height of the validator SHOULD remain `H`. This height only resets to a different height if a validator
+  // stops being a consumer validator during an epoch and later becomes again a consumer validator.
+  int64 height = 4;
```
mpoke, line 339: Doesn't this mean that we need to migrate all the consumer validator state?

x/ccv/provider/keeper/distribution.go
```diff
@@ -127,6 +127,23 @@ func (k Keeper) AllocateTokens(ctx sdk.Context) {
...
+// IsEligibleForConsumerRewards returns `true` if the validator with `consumerValidatorHeight` has been a consumer
+// validator for a long period of time and hence is eligible to receive rewards, and false otherwise
+func (k Keeper) IsEligibleForConsumerRewards(ctx sdk.Context, consumerValidatorHeight int64) bool {
+	numberOfBlocksToStartReceivingRewards := k.GetNumberOfEpochsToStartReceivingRewards(ctx) * k.GetBlocksPerEpoch(ctx)
+
+	// Note that at the beginning of the chain's life (e.g., when the block height of the chain is still small), no validator
+	// could have been a consumer validator for more than `NumberOfEpochsToStartReceivingRewards` epochs and hence in this
+	// case all validators are eligible for rewards, even though they might have not been a consumer validator on the
+	// consumer chain for more than a single epoch.
+	isChainStillNew := ctx.BlockHeight() < numberOfBlocksToStartReceivingRewards
```
mpoke, line 139: This check is for provider with height lower than 24*600. This is clearly an edge case that is not relevant for the Hub. I think we should remove it. Even in case the provider does start from height 0, the worst case scenario is that there are no rewards for ~24h.

x/ccv/provider/keeper/distribution.go
```diff
@@ -240,6 +261,12 @@ func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
 func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string) (totalPower int64) {
 	// sum the opted-in validators set voting powers
 	for _, v := range k.GetConsumerValSet(ctx, chainID) {
+
+		// only consider the voting power of a validator that would receive rewards (i.e., validator has been validating for a number of blocks)
+		if !k.IsEligibleForConsumerRewards(ctx, v.Height) {
+			continue
+		}
```
mpoke, line 268: Does this mean that a validator not being eligible results in the other validators getting more rewards?

mpoke (changes requested):
Blocking merging this before we address [this comment](https://github.com/cosmos/interchain-security/pull/1929#pullrequestreview-2093002194)

x/ccv/provider/migrations/v6/migrations.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+
+	providertypes "github.com/cosmos/interchain-security/v4/x/ccv/provider/types"
+)
+
+// MigrateParams adds missing provider chain params to the param store.
+func MigrateParams(ctx sdk.Context, paramsSubspace paramtypes.Subspace) {
```
MSalopek, line 11: [This PR](https://github.com/cosmos/interchain-security/pull/1952/files#diff-634767bb375db45fd33c40d36b03f596cf760a84b91b78cf5a51661b0a835af3) already bumped consensus to version 6.

You can stay on consensus version 6 and change the migratinon code so it runs both migrator functions.

insumity: Yeah, I haven't rebased.

proto/interchain_security/ccv/provider/v1/provider.proto
insumity: I do not think a migration is _necessary_.  The `height` would be set to `0`. So, validators that were consumer validators during the upgrade would be getting rewards without necessarily having validated the consumer chain for more than `NumberOfEpochsToStartReceivingRewards`. But this does not seem that problematic to me because validators that do not validate would stop getting rewards.

The other option, as you suggest, would be to indeed migrate the state with something like this:
```
for _, chainID := range providerKeeper.GetAllRegisteredConsumerChainIDs(ctx) {
  for _, val := range providerKeeper.GetConsumerValSet(ctx, chainID) {
	  val.Height = X ???
	  providerKeeper.SetConsumerValidator(ctx, chainID, val)
  }
}
```
If `X = ctx.BlockHeight()`, then validators that have been validating for more than `NumberOfEpochsToStartReceivingRewards` might not receive any rewards at the beginning. If `X - NumberOfEpochsToStartReceivingRewards * BlocksPerEpoch`, then we might give rewards to validators that have not been validating for more than `NumberOfEpochsToStartReceivingRewards`.

So, it seems to me that a migration is unnecessary. Let me know if you have any concerns.

x/ccv/provider/keeper/distribution.go
insumity: I removed the check.

x/ccv/provider/keeper/distribution.go
insumity: Yes, this could potentially happen. For instance, consider you have 2 validators `A` and `B` with powers 100 and 400 respectively, so the `totalPower` would have been 500. If both validators are eligible, then `A` would get `100/500` of the tokens, while `B` would get `400 / 500` of the tokens. However, if validator `B` is not eligible for rewards, this would mean that `totalPower` is only `100` and hence `A` gets all the tokens.

x/ccv/provider/module.go
insumity: Good catch. Rebased and now we have the latest consensus version from Philip's [PR](https://github.com/cosmos/interchain-security/pull/1952/files#diff-1b18b7319a6e8b981868c7896ae7f8ad8a621a186c387e18312fdf7da4cb5723).

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -360,6 +363,11 @@ message ConsumerValidator {
...
 }
```
mpoke, line 371: nit: rename to something more intuitive, e.g., init_height, join_height...

x/ccv/provider/keeper/distribution.go
```diff
@@ -240,6 +254,12 @@ func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
 func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string) (totalPower int64) {
 	// sum the consumer validators set voting powers
 	for _, v := range k.GetConsumerValSet(ctx, chainID) {
+
+		// only consider the voting power of a validator that would receive rewards (i.e., validator has been validating for a number of blocks)
+		if !k.IsEligibleForConsumerRewards(ctx, v.Height) {
```
mpoke, line 259: What happens here when a new consumer chain is added? Are all the validators ineligible? If so, is it safe for ComputeConsumerTotalVotingPower to return zero? Also, what happens with the rewads for the first warmup period?

mpoke (approved):
Great work @insumity. I approved the main logic. Left some comments though. See below.

x/ccv/provider/migrations/v6/migrations.go
```diff
@@ -31,3 +33,21 @@ func MigrateMinPowerInTopN(ctx sdk.Context, providerKeeper providerkeeper.Keeper
...
+	if !paramsSubspace.HasKeyTable() {
+		paramsSubspace.WithKeyTable(providertypes.ParamKeyTable())
+	}
+	paramsSubspace.Set(ctx, providertypes.KeyNumberOfEpochsToStartReceivingRewards, providertypes.DefaultNumberOfEpochsToStartReceivingRewards)
+}
+
+// MigrateSomething adds missing provider chain params to the param store.
+func MigrateSomething(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
```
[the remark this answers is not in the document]
insumity: Removed.

proto/interchain_security/ccv/provider/v1/provider.proto
insumity: Good point. `JoinHeight` is a much better name than simply `Height`. I renamed it.

x/ccv/provider/keeper/distribution.go
insumity: See above in the PR description:
>Finally, note that in a consumer chain that just starts, no validator would receive rewards during the first NumberOfEpochsToStartReceivingRewards epochs. During this time, any rewards sent to the provider from the consumer chain through IBC transfers will be allocated to the community pool.

If no validator is eligible, then rewards would [go to the community pool](https://github.com/cosmos/interchain-security/blob/864b3acd5a26842049ec23e6f3bcbe0d4ab78519/x/ccv/provider/keeper/distribution.go#L104). The same applies for the first warmup period. No validator would get rewards and the rewards would end up in the community pool.

PR #1942, opened 2024-06-06
test: Add an e2e test that reproduces the chain halt

p-offtermatt:
## Description

Closes: #XXXX

The test case uses the default config, except it sets the MaxValidators to 2.

Run the test locally by running 
```
go run ./tests/e2e/... --tc active-set-changes
```

These are the steps to reproduce the chain halt:
* Start the provider chain and a consumer
* Jail an active set validator. This will a) take that validator outside of the active set and b) put an inactive validator into the active set. This step will cause a chain halt.

The chain halt will make the test time out. You can look at the provider nodes logs by running 
```
docker exec interchain-security-instance cat provi/validatoralice/logs
```
and they should include this:
```
CONSENSUS FAILURE!!! err="more validators than maxValidators found"
```

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/interchain-security/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Provided a link to the relevant issue or specification
- [ ] Reviewed "Files changed" and left comments if necessary 
- [ ] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] Confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Confirmed all author checklist items have been addressed
- [ ] Confirmed that this PR does not change production code 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new test configuration for scenarios with a small number of maximum validators.
  - Added a new E2E test for active set changes to ensure protocol stability.

- **Chores**
  - Added a nightly E2E test job for active set changes in the CI workflow.

PR #1945, opened 2024-06-06
test: Add integration test reproducing the LastValidators exceeding MaxValidators bug

sainoe:
## Summary by CodeRabbit

- **Tests**
  - Added a new test function `TestTooManyLastValidators` to simulate scenarios causing a panic due to inconsistent state in the staking module.
  - Added methods for interacting with the staking keeper, including jailing/un-jailing validators and updating validator set parameters.

### DIFF
--- a/tests/integration/unbonding.go
+++ b/tests/integration/unbonding.go
@@ -488,7 +488,7 @@ func (s *CCVTestSuite) TestTooManyLastValidators() {
 	// verify that the number of bonded validators is decreased by one
 	s.Require().Equal(len(lastVals)-1, len(sk.GetLastValidators(s.providerCtx())))
 
-	// update maximum validator to the equal number of bonded validators
+	// update maximum validator to equal the number of bonded validators
 	p.MaxValidators = uint32(len(sk.GetLastValidators(s.providerCtx())))
 	sk.SetParams(s.providerCtx(), p)
 
--- a/tests/integration/unbonding.go
+++ b/tests/integration/unbonding.go
@@ -1,6 +1,7 @@
 package integration
 
 import (
+	"fmt"
 	"time"
 
 	"cosmossdk.io/math"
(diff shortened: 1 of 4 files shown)


tests/integration/unbonding.go
```diff
@@ -461,3 +461,47 @@ func (s *CCVTestSuite) TestRedelegationProviderFirst() {
...
+
+	// update maximum validator to the equal number of bonded validators
+	p.MaxValidators = uint32(len(sk.GetLastValidators(s.providerCtx())))
+	sk.SetParams(s.providerCtx(), p)
+
+	// pass one block to apply validator set changes
+	s.providerChain.NextBlock()
+
+	// unjail validator
+	sk.Unjail(s.providerCtx(), consAddr)
```
insumity, line 499: So, I guess validators `vals` are sorted in descending order by power and that's why `vals[0]` when unjailed enters again the active set? Might be worth to write a comment.

tests/integration/unbonding.go
```diff
@@ -461,3 +461,47 @@ func (s *CCVTestSuite) TestRedelegationProviderFirst() {
...
 	checkCCVUnbondingOp(s, s.providerCtx(), s.consumerChain.ChainID, valsetUpdateID, false)
 }
+
+func (s *CCVTestSuite) TestTooManyLastValidators() {
+	sk := s.providerApp.GetTestStakingKeeper()
+
+	// get current staking params
+	p := sk.GetParams(s.providerCtx())
+
+	// get validators, which are all active atm
```
insumity, line 471: nit (being pedantic): 
```suggestion
	// get validators, which are all active at the moment
```
:)

tests/integration/unbonding.go
```diff
@@ -461,3 +461,47 @@ func (s *CCVTestSuite) TestRedelegationProviderFirst() {
 	// Check that ccv unbonding op has been deleted
 	checkCCVUnbondingOp(s, s.providerCtx(), s.consumerChain.ChainID, valsetUpdateID, false)
 }
+
+func (s *CCVTestSuite) TestTooManyLastValidators() {
```
insumity, line 465: Might make sense to have consistent names with this [E2E test](https://github.com/cosmos/interchain-security/pull/1942/files). Either name is fine by me. (cc @p-offtermatt )

tests/integration/unbonding.go
```diff
@@ -461,3 +461,47 @@ func (s *CCVTestSuite) TestRedelegationProviderFirst() {
...
+	// pass one block to apply validator set changes
+	s.providerChain.NextBlock()
+
+	// unjail validator
+	sk.Unjail(s.providerCtx(), consAddr)
+
+	// pass another block to update the validator set
+	// which causes a panic due to a GetLastValidator call in
+	// ApplyAndReturnValidatorSetUpdates where the staking module has a inconsistent state
+	s.Require().Panics(s.providerChain.NextBlock)
```
insumity, line 504: I guess we would have to modify those assertions to `NotPanics` after the fix is merged on `main`?

insumity (approved):
LGTM!

tests/integration/unbonding.go
p-offtermatt: I think it is probably ok to have different names, since we don't have common naming conventions across e2e and integration tests right now, anyways

p-offtermatt, line 465: Would recommend a docstring explaining the test

tests/integration/unbonding.go
```diff
@@ -461,3 +461,47 @@ func (s *CCVTestSuite) TestRedelegationProviderFirst() {
...
+	lastVals := sk.GetLastValidators(s.providerCtx())
+
+	// pass one block to apply the validator set changes
+	// (calls ApplyAndReturnValidatorSetUpdates in the the staking module EndBlock)
+	s.providerChain.NextBlock()
+
+	// verify that the number of bonded validators is decreased by one
+	s.Require().Equal(len(lastVals)-1, len(sk.GetLastValidators(s.providerCtx())))
+
+	// update maximum validator to the equal number of bonded validators
```
p-offtermatt, line 491:
```suggestion
	// update maximum validator to equal the number of bonded validators
```

p-offtermatt (approved):
LGTM!
Just got to remember to change Panics to NotPanics (but hopefully the test will just start failing when we have the fix and test in the same branch)

PR #1982, opened 2024-06-21
fix!: update PSS reward distribution to work with SDK v0.50.0 changes

sainoe:
SDK v0.50.x changes the methods for interacting with the community pool of the distribution module. The current PSS reward distribution data flow is as follows:

1. A consumer sends rewards to the provider module.
2. The provider module sends rewards to the distribution module.
3. The distribution module sends a fraction of the rewards to the validators.
4. The remaining rewards are allocated to the community pool (tokens stay in the distribution module account).

In SDK v0.50.x, step 4 is no longer possible, and the community pool must be updated using the [FundCommunityPool](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/x/distribution/keeper/keeper.go#L206) method, which combines sending funds to the distribution module account (step 1) and allocating them to the community pool (step 4).

Therefore, the PSS reward distribution data flow must be updated:

1. A consumer sends rewards to the provider module.
2. The provider module sends a fraction of the rewards to the distribution module.
3. The distribution module sends a fraction of the rewards to the validators.
4. The provider module sends the remaining rewards to the community pool using  `FundCommunityPool`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the accuracy of reward allocation between consumers, validators, and the community pool.
  - Ensured proper handling of remaining reward decimals for precise distribution.

- **Tests**
  - Enhanced reward distribution test scenarios to verify pool balances during transfers and allocations.

### DIFF
--- a/tests/integration/distribution.go
+++ b/tests/integration/distribution.go
@@ -744,7 +744,7 @@ func (s *CCVTestSuite) TestAllocateTokens() {
 
 	// At this point the distribution module account
 	// only holds the community pool's tokens
-	// since there zero validators outstanding rewards
+	// since there are no validators with outstanding rewards
 	lastCommPool := getDistrAcctBalFn(s.providerCtx())
 
 	// execute BeginBlock to trigger the token allocation
--- a/tests/integration/distribution.go
+++ b/tests/integration/distribution.go
@@ -720,14 +720,14 @@ func (s *CCVTestSuite) TestAllocateTokens() {
 	)
 
 	// Allocate rewards evenly between consumers
-	rewardsPerConsumer := totalRewards.QuoInt(math.NewInt(int64(len(s.consumerBundles))))
+	rewardsPerChain := totalRewards.QuoInt(math.NewInt(int64(len(s.consumerBundles))))
 	for chainID := range s.consumerBundles {
 		// update consumer allocation
 		providerKeeper.SetConsumerRewardsAllocation(
 			s.providerCtx(),
 			chainID,
 			providertypes.ConsumerRewardsAllocation{
-				Rewards: sdk.NewDecCoinsFromCoins(rewardsPerConsumer...),
+				Rewards: sdk.NewDecCoinsFromCoins(rewardsPerChain...),
 			},
 		)
 	}
@@ -754,26 +754,25 @@ func (s *CCVTestSuite) TestAllocateTokens() {
 	consNum := len(s.consumerBundles)
 
 	// compute the expected validators token allocation by subtracting the community tax
-	rewardsPerConsumerDec := sdk.NewDecCoinsFromCoins(rewardsPerConsumer...)
+	rewardsPerChainDec := sdk.NewDecCoinsFromCoins(rewardsPerChain...)
 	communityTax, err := distributionKeeper.GetCommunityTax(s.providerCtx())
 	s.Require().NoError(err)
 
-	rewardsPerChain, _ := rewardsPerConsumerDec.
+	rewardsPerChainTrunc, _ := rewardsPerChainDec.
 		MulDecTruncate(math.LegacyOneDec().Sub(communityTax)).TruncateDecimal()
-	validatorsExpRewardsPerChain := sdk.NewDecCoinsFromCoins(rewardsPerChain...).QuoDec(math.LegacyNewDec(int64(valNum)))
+	validatorsExpRewardsPerChain := sdk.NewDecCoinsFromCoins(rewardsPerChainTrunc...).QuoDec(math.LegacyNewDec(int64(valNum)))
 	// multiply by the number of consumers
 	validatorsExpRewards := validatorsExpRewardsPerChain.MulDec(math.LegacyNewDec(int64(consNum)))
 
 	// verify the validator tokens allocation
 	// note that the validators have the same voting power to keep things simple
-	lastValOutRewards := map[string]sdk.DecCoins{}
 	for _, val := range s.providerChain.Vals.Validators {
 		valRewards, err := distributionKeeper.GetValidatorOutstandingRewards(s.providerCtx(), sdk.ValAddress(val.Address))
 		s.Require().NoError(err)
 
 		s.Require().Equal(
 			valRewards.Rewards,
-			lastValOutRewards[sdk.ValAddress(val.Address).String()].Add(validatorsExpRewards...),
+			validatorsExpRewards,
 		)
 	}
 
--- a/tests/integration/distribution.go
+++ b/tests/integration/distribution.go
@@ -733,19 +733,18 @@ func (s *CCVTestSuite) TestAllocateTokens() {
 	}
 
 	// Iterate over the validators and
-	// store their outstanding rewards
-	lastValOutRewards := map[string]sdk.DecCoins{}
 	totalValsRewards := sdk.DecCoins{}
 	for _, val := range s.providerChain.Vals.Validators {
 		valRewards, err := distributionKeeper.GetValidatorOutstandingRewards(s.providerCtx(), sdk.ValAddress(val.Address))
 		s.Require().NoError(err)
-		lastValOutRewards[sdk.ValAddress(val.Address).String()] = valRewards.Rewards
 		totalValsRewards = totalValsRewards.Add(valRewards.Rewards...)
 	}
 
 	s.Require().True(totalValsRewards.IsZero())
 
-	// store community pool balance
+	// At this point the distribution module account
+	// only holds the community pool's tokens
+	// since there zero validators outstanding rewards
 	lastCommPool := getDistrAcctBalFn(s.providerCtx())
 
 	// execute BeginBlock to trigger the token allocation
@@ -763,10 +762,11 @@ func (s *CCVTestSuite) TestAllocateTokens() {
 		MulDecTruncate(math.LegacyOneDec().Sub(communityTax)).TruncateDecimal()
 	validatorsExpRewardsPerChain := sdk.NewDecCoinsFromCoins(rewardsPerChain...).QuoDec(math.LegacyNewDec(int64(valNum)))
 	// multiply by the number of consumers
-	validatorsExpRewards := validatorsExpRewardsPerchain.MulDec(math.LegacyNewDec(int64(consNum)))
+	validatorsExpRewards := validatorsExpRewardsPerChain.MulDec(math.LegacyNewDec(int64(consNum)))
 
 	// verify the validator tokens allocation
 	// note that the validators have the same voting power to keep things simple
+	lastValOutRewards := map[string]sdk.DecCoins{}
 	for _, val := range s.providerChain.Vals.Validators {
 		valRewards, err := distributionKeeper.GetValidatorOutstandingRewards(s.providerCtx(), sdk.ValAddress(val.Address))
 		s.Require().NoError(err)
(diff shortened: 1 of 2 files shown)


tests/integration/distribution.go
```diff
@@ -147,15 +148,20 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
...
 	)
 
-	// Check that the consumer rewards allocation are empty since relayAllCommittedPackets calls BeginBlockRD,
-	// which in turns calls AllocateTokens.
+	// Consumer allocations are distributed between the validators and the community pool.
+	// The decimals resulting from the distribution are expected to remain in the consumer allocations.
 	rewardsAlloc := providerKeeper.GetConsumerRewardsAllocation(s.providerCtx(), s.consumerChain.ChainID)
-	s.Require().Empty(rewardsAlloc.Rewards)
+	remainingAlloc := rewardsAlloc.Rewards.AmountOf(rewardsIBCdenom)
+	s.Require().True(remainingAlloc.LTE(math.LegacyOneDec()))
```
insumity, line 155: Why `LTE` and not `LT` (i.e., less than)? It seems that if `remainingAlloc` was `1` it should have been allocated?

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
-			// check remaining consumer rewards allocation
-			expCoinTransferred, expRemaining := tc.rewardsAlloc.TruncateDecimal()
-			s.Require().Equal(expCoinTransferred, coinsTransferred)
-
-			s.Require().Equal(
-				expRemaining,
-				providerKeeper.GetConsumerRewardsAllocation(ctx, chainID).Rewards,
-			)
+	// Iterate over the validators and
+	// store their current voting power and outstanding rewards
```
insumity, line 736:
```suggestion
	// store their outstanding rewards
```
Voting power is not stored anymore (?)

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+		lastValOutRewards[sdk.ValAddress(val.Address).String()] = valRewards.Rewards
+		totalValsRewards = totalValsRewards.Add(valRewards.Rewards...)
+	}
 
-			// check updated consuemer rewards pool balance
-			newPool := bankKeeper.GetAllBalances(
-				ctx,
-				distributionKeeper.GetDistributionAccount(ctx).GetAddress(),
-			)
+	s.Require().True(totalValsRewards.IsZero())
```
insumity, line 746: If `totalValsRewards` is zero, wouldn't this imply that all `valRewards.Rewards` in the above `for` loop were zero as well? Hence, `lastValOutRewards` is zero as well for all `s.providerChain.Vals.Validators` validators. If yes, why do we even need to store this data.

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
 
-			s.Require().Equal(newPool.Sub(oldPool...), coinsTransferred)
-		})
+	// store community pool balance
```
insumity, line 748: Doesn't `getDistrAcctBalFn` get the distribution module's balance? Is this the same as that of the community pool?

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+	consNum := len(s.consumerBundles)
+
+	// compute the expected validators token allocation by subtracting the community tax
+	rewardsPerConsumerDec := sdk.NewDecCoinsFromCoins(rewardsPerConsumer...)
+	communityTax, err := distributionKeeper.GetCommunityTax(s.providerCtx())
+	s.Require().NoError(err)
+
+	rewardsPerChain, _ := rewardsPerConsumerDec.
+		MulDecTruncate(math.LegacyOneDec().Sub(communityTax)).TruncateDecimal()
+	validatorsExpRewardsPerchain := sdk.NewDecCoinsFromCoins(rewardsPerChain...).QuoDec(math.LegacyNewDec(int64(valNum)))
```
insumity, line 764:
```suggestion
	validatorsExpRewardsPerChain := sdk.NewDecCoinsFromCoins(rewardsPerChain...).QuoDec(math.LegacyNewDec(int64(valNum)))
```

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+		valRewards, err := distributionKeeper.GetValidatorOutstandingRewards(s.providerCtx(), sdk.ValAddress(val.Address))
+		s.Require().NoError(err)
+
+		s.Require().Equal(
+			valRewards.Rewards,
+			lastValOutRewards[sdk.ValAddress(val.Address).String()].Add(validatorsExpRewards...),
+		)
 	}
+
+	// check that the total expected rewards is transferred to the distribution module account
```
insumity, line 780:
```suggestion
	// check that the total expected rewards are transferred to the distribution module account
```

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+	allocRemPerConsu := providerKeeper.GetConsumerRewardsAllocation(s.providerCtx(), s.consumerChain.ChainID).Rewards
```
insumity, line 781: The suffix `PerConsu` corresponds to **per consu**mer chain, right? If yes, then maybe name `PerChain` to be consistent with previously-named variables.

tests/integration/distribution.go
```diff
@@ -690,197 +696,92 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+	totalRewardsDistributed := sdk.NewDecCoinsFromCoins(totalRewards...).Sub(allocRemPerConsu.MulDec(math.LegacyNewDec(int64(consNum))))
+
+	s.Require().Equal(lastCommPool.Add(totalRewardsDistributed...), getDistrAcctBalFn(s.providerCtx()))
```
insumity, line 784: I do not fully understand this assertion. Since we had a `BeginBlockRd`, it seems to me that tokens would have been transferred (?) from the distribution module to the corresponding validators, etc. So, how come the distribution module now contains what the distribution model had plus the `totalRewardsDistributed`? It seems to me we should have a `Sub` instead of an `Add` here (?)

tests/integration/distribution.go
sainoe: It's due to `SendCoinsFromModuleToModule` that only accepts integer amounts of coins. Therefore when the total reward is split in two, the truncation remainders of the [validator allocation](https://github.com/cosmos/interchain-security/blob/f031dafc5b889899d8f5ee8e862d0b235bd8ed81/x/ccv/provider/keeper/distribution.go#L130) and the [community pool transfer](https://github.com/cosmos/interchain-security/blob/f031dafc5b889899d8f5ee8e862d0b235bd8ed81/x/ccv/provider/keeper/distribution.go#L149) can sum to one. See [here](https://github.com/cosmos/interchain-security/blob/f031dafc5b889899d8f5ee8e862d0b235bd8ed81/x/ccv/provider/keeper/distribution.go#L161-L162) how the consumer rewards value gets updated.

tests/integration/distribution.go
sainoe: Good catch!

tests/integration/distribution.go
sainoe: Yes, in this setup the distribution module's account balance only holds the tokens that belong to the community pool. Note that the community pool is only a state of the distribution module, not a separate account.

x/ccv/provider/keeper/distribution.go
```diff
@@ -121,19 +117,36 @@ func (k Keeper) AllocateTokens(ctx sdk.Context) {
 			)
 			continue
 		}
+
+		// compute rewards for validators
+		consumerRewards := alloc.Rewards
 		voteMultiplier := math.LegacyOneDec().Sub(communityTax)
-		feeMultiplier := rewardsCollectedDec.MulDecTruncate(voteMultiplier)
+		feeMultiplier := consumerRewards.MulDecTruncate(voteMultiplier)
```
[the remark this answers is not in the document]
sainoe: Good point. Update to `validatorRewards` since it defines the fraction of rewards allocated to validators.

tests/integration/distribution.go
sainoe: After `AllocateTokens` is executed, all rewards are transferred to the distribution module account and the validator's outstanding rewards and the community pool balance are updated. It's only when validators call `WithdrawValidatorCommission` that their token rewards are being transferred to their account.

x/ccv/provider/keeper/distribution.go
```diff
@@ -121,19 +117,36 @@ func (k Keeper) AllocateTokens(ctx sdk.Context) {
...
 		voteMultiplier := math.LegacyOneDec().Sub(communityTax)
-		feeMultiplier := rewardsCollectedDec.MulDecTruncate(voteMultiplier)
+		validatorRewards := consumerRewards.MulDecTruncate(voteMultiplier)
+
+		// compute remaining rewards for the community pool
+		remaining := consumerRewards.Sub(validatorRewards)
+
+		// transfer validators rewards to distribution module account
+		validatorRewardsTrunc, validatorRewardsChange := validatorRewards.TruncateDecimal()
+		err = k.bankKeeper.SendCoinsFromModuleToModule(ctx, types.ConsumerRewardsPool, distrtypes.ModuleName, validatorRewardsTrunc)
```
insumity, line 131: nit: s/`validatorRewards`/`validatorsRewards` to be consistent with comment above `// transfer validators rewards ...`

x/ccv/provider/keeper/distribution.go
```diff
@@ -121,19 +117,36 @@ func (k Keeper) AllocateTokens(ctx sdk.Context) {
...
 			consumerChainID,
-			feeMultiplier,
+			sdk.NewDecCoinsFromCoins(validatorRewardsTrunc...),
 		)
-		remaining = remaining.Sub(feeAllocated)
 
-		// allocate community funding
-		remainingCoins, _ := remaining.TruncateDecimal()
+		// allocate remaining rewards to the community pool
+		remainingCoins, remainingChanges := remaining.TruncateDecimal()
```
insumity, line 149: Should this be called `remainingRewards` and `remainingChange` to be consistent with naming above.

tests/integration/distribution.go
```diff
@@ -690,197 +696,97 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
-
-			// check remaining consumer rewards allocation
-			expCoinTransferred, expRemaining := tc.rewardsAlloc.TruncateDecimal()
-			s.Require().Equal(expCoinTransferred, coinsTransferred)
-
-			s.Require().Equal(
-				expRemaining,
-				providerKeeper.GetConsumerRewardsAllocation(ctx, chainID).Rewards,
-			)
+	// Iterate over the validators and
```
insumity, line 735:
```suggestion
	// iterate over the validators and verify that no validator has outstanding rewards
```

tests/integration/distribution.go
```diff
@@ -690,197 +696,97 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+	lastCommPool := getDistrAcctBalFn(s.providerCtx())
+
+	// execute BeginBlock to trigger the token allocation
+	providerKeeper.BeginBlockRD(s.providerCtx())
+
+	valNum := len(s.providerChain.Vals.Validators)
+	consNum := len(s.consumerBundles)
+
+	// compute the expected validators token allocation by subtracting the community tax
+	rewardsPerConsumerDec := sdk.NewDecCoinsFromCoins(rewardsPerConsumer...)
```
insumity, line 757: `rewardsPerChainDec` (?) would consistent with `rewardsPerChain` below.

tests/integration/distribution.go
```diff
@@ -690,197 +696,97 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
+	// verify the validator tokens allocation
+	// note that the validators have the same voting power to keep things simple
+	lastValOutRewards := map[string]sdk.DecCoins{}
+	for _, val := range s.providerChain.Vals.Validators {
+		valRewards, err := distributionKeeper.GetValidatorOutstandingRewards(s.providerCtx(), sdk.ValAddress(val.Address))
+		s.Require().NoError(err)
+
+		s.Require().Equal(
+			valRewards.Rewards,
+			lastValOutRewards[sdk.ValAddress(val.Address).String()].Add(validatorsExpRewards...),
```
insumity, line 776: Isn't `lastValOutRewards[sdk.ValAddress(val.Address).String()]` zero?

tests/integration/distribution.go
insumity: I see. Thank you very much!

insumity (approved):
Great work!
LGTM

tests/integration/distribution.go
```diff
@@ -690,197 +696,97 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
...
-				ctx,
-				distributionKeeper.GetDistributionAccount(ctx).GetAddress(),
-			)
+	s.Require().True(totalValsRewards.IsZero())
 
-			s.Require().Equal(newPool.Sub(oldPool...), coinsTransferred)
-		})
+	// At this point the distribution module account
+	// only holds the community pool's tokens
+	// since there zero validators outstanding rewards
```
insumity, line 747:
```suggestion
	// since there are no validators with outstanding rewards
```

PR #1983, opened 2024-06-21
chore!: only distribute rewards to validators that have been validating a consumer chain for some backport #1929

sainoe:
~[should be merged after #1982]~

backport [1929](https://github.com/cosmos/interchain-security/pull/1929).

## Summary by CodeRabbit

- **Tests**
  - Updated reward distribution logic in test cases to accommodate new block height criteria and context adjustments.
  - Added tests for allocating tokens to consumer validators with varying block heights.

- **New Features**
  - Introduced eligibility criteria for consumer rewards based on validator duration.

- **Refactor**
  - Enhanced parameter handling and validations in migration test functions.


tests/integration/distribution.go
```diff
@@ -137,7 +137,16 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
 	}
 	consuValsRewards := consumerValsOutstandingRewardsFunc(s.providerCtx())
 
-	// Transfer rewards from consumer to provider
+	// increase the block height so validators are eligible for consumer rewards (see `IsEligibleForConsumerRewards`)
```
bermuell, line 140: from where is this change comming from?
did not see this change in the original PR 1929.
can you pls cross check this?

x/ccv/provider/keeper/distribution.go
```diff
@@ -76,42 +76,42 @@ func (k Keeper) AllocateTokens(ctx sdk.Context) {
 
 	// Iterate over all registered consumer chains
 	for _, consumerChainID := range k.GetAllRegisteredConsumerChainIDs(ctx) {
-		// transfer the consumer rewards to the distribution module account
```
bermuell, line 79: also this change here I can't see in original PR?

x/ccv/provider/migrations/v6/migration_test.go
```diff
@@ -7,51 +7,21 @@ import (
...
-
-	err := MigrateLegacyParams(ctx, k, inMemParams.ParamsSubspace)
-	require.NoError(t, err)
-
-	// check that "new" params are available after migration and equal to defaults
-	migratedParams := k.GetParams(ctx)
-	require.NotEmpty(t, migratedParams)
-	require.Equal(t, defaultParams, migratedParams)
-	require.NotEqual(t, emptyParams, migratedParams)
+	// after migration, number of epochs epoch param should exist and be equal to default
```
insumity, line 22:
```suggestion
	// after migration, number of epochs to start receiving rewards param should exist and be equal to default
```

x/ccv/provider/migrations/v7/legacy_params.go
```diff
@@ -1,4 +1,4 @@
-package v6
+package v7
```
insumity, line 1: Do we want to keep the old `legacy_params.go` file under `v6` as well in case someone wants to use it to go from 5 to 6?

x/ccv/provider/migrations/v7/migrations_test.go
```diff
@@ -0,0 +1,57 @@
...
+	require.Equal(t, defaultParams, legacyParams)
+
+	err := MigrateLegacyParams(ctx, k, inMemParams.ParamsSubspace)
+	require.NoError(t, err)
+
+	// check that "new" params are available after migration and equal to defaults
+	migratedParams := k.GetParams(ctx)
+	require.NotEmpty(t, migratedParams)
+	require.Equal(t, defaultParams, migratedParams)
+	require.NotEqual(t, emptyParams, migratedParams)
```
insumity, line 56: nit: Since we already check that `migratedParams` are not empty and that `emptyParams` are empty, I guess this assertion is not necessarily needed, albeit it doesn't hurt.

insumity (approved):
LGTM

x/ccv/provider/migrations/v7/legacy_params.go
insumity: Talked offline. Resolving this because it was based on my misunderstanding.

PR #2001, opened 2024-07-01
docs: introduce ADR for permissionless ICS

insumity:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

- [ ] provided a link to the relevant issue or specification
- [ ] reviewed "Files changed" and left comments if necessary 
- [x] confirmed all CI checks have passed

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Introduced "Permissionless Interchain Security" (ICS) through ADR 18, enabling Opt In consumer chains to join ICS via transactions without governance proposals.
	- Added a visual representation of consumer chain proposals and transactions, enhancing understanding of operational flows.
	- Introduced new messages for managing consumer chains: `MsgRegisterConsumerChain`, `MsgInitializeConsumerChain`, `MsgUpdateConsumerChain`, and `MsgStopConsumerChain`.

- **Documentation**
	- Detailed documentation on "Permissionless Interchain Security" (ICS) and its related processes, including phases and state management.


docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,266 @@
...
+| `ConsumerRewardsAllocationBytePrefix`    |Stores the ICS rewards per chain                                                  |**YES**                   |Every IBC transfer packet that sends rewards to the provider              |
+| `ConsumerCommissionRatePrefix`           |Comission rate per chain per validator                                            |**YES**                   |Every `MsgSetConsumerCommissionRate` message                              |
+| `MinimumPowerInTopNBytePrefix`           |Stores the minimum power needed to opt in for a chain                             |**YES**                   |Every epoch                                                               |
+| `ConsumerAddrsToPruneV2BytePrefix`       |Stores consumer addresses to be pruned (as part of `VSCMaturedPacket`s deprecation)|**YES**                   |Every `MsgAssignConsumerKey` or `MsgOptIn` and later during actual pruning|
+
+Everything stored under a key associated with a `chainID` needs to be migrated to new state under `consumerID`. 
+Because we have to migrate in any case, we can also clean up a number of those keys by building a `ConsumerChainRecord`.
```
insumity, line 113: I'm still not a huge fan of `ConsumerChainRecord`. It seems to slightly complicate things (e.g., someone adding new state now would have to think on whether this should go in `ConsumerChainRecord` or not) without any extra benefits because it does not contain everything currently stored under `chainID`-related keys.

I wonder if it would make sense to have a Go `struct` that is a `ConsumerChainRecord` and provides methods to do everything you want to do for a chain (e.g., `chainRecord.SetConsumerValidator(validator)`, `chainRecord.GetConsumerValidators()`, etc.). Those methods under the hood would be using the `proto`-generated `structs` and this way we won't have to do any migration besides that of Neutron and Stride.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+  // allowlist of provider consensus addresses of validators that are the ONLY ones that can validate the consumer chain
+  repeated string allowlist;
+  // denylist of provider consensus addresses of validators that are the CANNOT validate the consumer chain
+  repeated string denylist;
+}
+```
+
+We store the `ConsumerChainRecord`s using the `consumerID` as their key. The `owner_address` of a consumer chain corresponds
+to the address that issued the `MsgLaunchConsumerChain` and the `owner_address` can be updated by issuing a `MsgUpdateConsumerChain` (see later).
+Note that a `ConsumerChainRecord` is only created when a consumer chain moves from the prelaunched to the _launched_ phase.
```
bermuell, line 149: what is the reason to store it in advance as a proposal?
are there any side-effects doing so, i.e. query on gov proposals will show up these entries as well?

p-offtermatt (commented):
Do you have the source for the diagram? if yes, could you add it? (excalidraw allows downloading/uploading, and this would make it possible for us to modify the diagram if anything changes at some point, without needing to recreate)

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+keep track of the consumer chain's state for some period, so that we are able to punish validators for misbehaviour.
+
+Note that everything described so far and everything that follows applies to standalone chains as well.
+
+### From `chainID` to `consumerID`
+A hindrance in moving to Permissionless ICS is chain-id squatting. In a permissionless setting, anyone could issue a transaction
+to launch a consumer chain with a `chainID` that might already be used by some other consumer chain. This is a problem
+because in the current design the majority of stored state for a consumer chain is indexed using the `chainID` as the key (e.g.,
+see [key used to store client ids](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/keys.go#L233)).
+To tackle this problem, in Permissionless ICS, we introduce the `consumerID` that defines a consumer chain and is simply
```
p-offtermatt, line 58: Could you briefly expand on the format of the consumer id? naively, depending on what "combination" means, there might be problems with the chain id "bleeding" into the sequence number, e.g.

neutron11 = chain_id ("neutron1") + sequence ("1")
but also
neutron11 = chain_id ("neutron") + sequence ("11)

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+a combination of a `chainID` and an increasing sequence number, thus we can support multiple consumer chains with the same `chainID`.
+Nevertheless, as a result of using `consumerID`, we have to migrate a substantial chunk of state to re-index it using `consumerID` as the key.
+
+#### State
+To move from `consumerID` to a `chainID`, we need to revamp the consumer chains' stored state in ICS. Currently, in
```
p-offtermatt, line 63: reversed?
```suggestion
To move from using the `chainID` to the new `consumerID`, we need to revamp the consumer chains' stored state in ICS. Currently, in
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+  string owner_address;
+  // client id to the consumer chain
+  string client_id;
+  // channel id
+  string channel_id;
+  // the chain-id of the consumer chain
+  string chain_id;
+  // consumer genesis
+  interchain_security.ccv.v1.ConsumerGenesisState consumer_genesis = 5
+  // min height needed for an equivocation to be actionable
```
p-offtermatt, line 134: at first, I thought this would continuously update.
In reality, I think this is just the first height the chain is ICS secured for? Could the comment be a bit clearer on this?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+through a governance proposal). Note, that an alternative format for this message could have been to include a single `ConsumerAdditionProposal` field in it,
+but this confuses the semantics of `MsgLaunchConsumerChain` and that of a governance proposal. The response of this message
+(i.e., `MsgLaunchConsumerChainResponse`) contains a single `string` that is the `consumerID`. The `consumerID`s corresponds
+to `len(chainID) | chainID | counter` and this way we can have multiple different `consumerID`s that correspond to a chain
+with the same `chainID`. By doing this, we tackle the main issue behind permissionless consumer chains that is chain-id squatting.
+Note that `consumerID` could just be a `uint64` but we choose to include `chainID` in it as well so that it is easier to see
+what the consumer chain just by looking at the `consumerID`. This means that the `chainID` of a chain cannot be changed
+after launching it (because the `chainID` is part of the `consumerID` key).
+
+We intend to set a fixed cost of a `MsgLaunchConsumerChain` to avoid getting spammed with bogus consumer launch chains.
```
p-offtermatt, line 190:
```suggestion
We intend to set a fixed cost of a `MsgLaunchConsumerChain` to avoid getting spammed with bogus consumer chains.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,266 @@
...
+
+To execute a `MsgLaunchConsumerChain`, we first create a `ConsumerAdditionProposal` under the hoods, with the `top_N` set to 0, and call
+[`HandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L30)
+simulating a proposal that was already voted on and accepted. Note that we need to migrate `ConsumerAdditionProposal`s to use
+the `consumerID` instead using [both](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L382)
```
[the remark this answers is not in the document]
p-offtermatt: bot is right


```suggestion
the `consumerID` instead of using [both](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L382)
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+`chainID` and the `spawnTime` as keys. The [usual validity conditions](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/proposal.go#L114)
+hold for the fields of `MsgLaunchConsumerChain`. Note however, that we intend to have a `spawnTime` upper limit as well.
+For example, if you launch a consumer chain in Permissionless ICS, the `spawnTime` should not be more
+than two months ahead in the future, to avoid having consumer chains lingering for too long before they get added. To do this,
+we introduce a `maxSpawnTime` limit in [`ConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L29).
```
p-offtermatt, line 200: How is the `maxSpawnTime` set? For proposals, they can choose it themselves, and for launches via TX, it is set by current time + some parameter?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+  ...  (all fields contained in ConsumerAdditionProposal except top_N)
+  
+  uint32 validators_power_cap;
+  uint32 validator_set_cap;
+  repeated string allowlist;
+  repeated string denylist;
+}
+```
+
+The `owner_address` is provided as well and hence a validator can change the owner of a consumer chain as well. Because of this,
```
p-offtermatt, line 231: Why `a validator`? This has nothing to do with validators, and it should just say "the current owner account"?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+the `MsgUpdateConsumerChain` needs to contain two signatures: one for the old and one for the new owner address.
```
p-offtermatt, line 232: the new owner needs to provide a /signature/? why? can it just be permissionless?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+
+The `initial_height`, `spawnTime`, etc. exist in `MsgUpdateConsumerChain` but are only be applicable if the chain
```
p-offtermatt, line 234:
```suggestion
The `initial_height`, `spawnTime`, etc. exist in `MsgUpdateConsumerChain` but can only be modified if the chain
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+is still in the prelaunched phase. This can be achieved by looking if there is an ongoing `ConsumerAdditionProposal` for this `consumerID`,
+and then changing the fields of ths proposal. The other fields such as `allowlist`, etc. can be changed at any point before or after a chain has started.
```
p-offtermatt, line 236:
```suggestion
and then changing the fields of this proposal. The other fields such as `allowlist`, etc. can be changed at any point before or after a chain has started.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Stop a Consumer Chain
+With the `MsgStopConsumerChain` we can stop any Opt In chain at any moment. Note that all relevant state for this consumer chain
+remains on the provider's state before getting removed for the provider's unbonding period. This is to enable
```
p-offtermatt, line 240: "getting removed for the provider's unbonding period" does not parse.
maybe 
```suggestion
remains on the provider's state before getting removed after one unbonding period (of the provider). This is to enable
```

p-offtermatt (commented):
Great work. I left some comments

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: This is described later on where we write:
> The `consumerID`s corresponds to `len(chainID) | chainID | counter` and this way we can have multiple different `consumerID`s [...]

which is the common way of creating those keys. So, in the above example we would have two different keys: `8 | neutron1 | 1` and `7 | neutron | 11`.

Added the fact that the `consumerID` contains the `len(chainID`) in this part of the ADR as well.

p-offtermatt:
@coderabbitai pause
I thought I turned the bot off on markdown files :(

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+an Opt In chain can choose to change parameters of the to-be-launched chains, such as `spawnTime`, etc. by issuing a specific transaction.
+This is not the case for Top N chains, where a `ConsumerModificationProposal` can only be issued after a consumer
+chain [has started](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L150).
+
+When the [`spawnTime`](https://github.com/cosmos/interchain-security/blob/v4.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L55)
+passes and [at least one validator has opted in](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L455)
+the chain can launch and moves to the _launched_ phase. While in launched phase, a Top N consumer chain can choose to modify
+its parameters through a `ConsumerModificationProposal` and an Opt In chain can change its parameters by issuing a transaction.
+
+Lastly, a Top N chain can choose to exit ICS by issuing a `ConsumerRemovalProposal` and an Opt In chain can issue a transaction to stop the chain.
```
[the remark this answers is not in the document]
insumity: In the current version of ICS, we cannot issue `ConsumerRemovalProposal` in the prelaunch phase. I could see use cases (e.g., consumer chain decides not to join ICS after all) for being able to remove a chain while it's in its prelaunch phase, but it does not seem critical.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+be either _Top N_ or _Opt In_. If a chain is an Opt In chain, then no validator is required to validate this chain unless they choose to.
+Because of this, we can launch an Opt In consumer chain without going through a governance proposal.
+
+This ADR presents _Permissionless_ ICS, a way in which an [_Opt In_](adr-015-partial-set-security.md) consumer chain can join
+ICS without needing a governance proposal but by simply issuing a transaction.
+
+## Decision
+
+### The Phases of a Consumer Chain
+In Permissionless ICS, launching an Opt In chain is **only** possible through a transaction and not through a `ConsumerAdditionProposal`.
```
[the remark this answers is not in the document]
insumity: That's a good question. My view is that everything _Opt In_-chain related should be done through transactions and everything _Top N_-chain related should be done through governance proposals. This creates a nice distinction between those two.

I do not see a reason why the community would want to remove an Opt In chain because noone is forced to validate it. In case a consumer chain attempts to act maliciously, or so, we probably should have a coordinated upgrade to kick it out as soon as possible and not through a gov proposal

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+its parameters through a `ConsumerModificationProposal` and an Opt In chain can change its parameters by issuing a transaction.
+
+Lastly, a Top N chain can choose to exit ICS by issuing a `ConsumerRemovalProposal` and an Opt In chain can issue a transaction to stop the chain.
+After some period of time (e.g., provider's unbonding period), all state related to the stopped consumer chain can be removed. We
+keep track of the consumer chain's state for some period, so that we are able to punish validators for misbehaviour.
+
+Note that everything described so far and everything that follows applies to standalone chains as well.
+
+### From `chainID` to `consumerID`
+A hindrance in moving to Permissionless ICS is chain-id squatting. In a permissionless setting, anyone could issue a transaction
```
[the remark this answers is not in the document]
insumity: I'm not familiar of any non-cumbersome alternatives. Some ideas are mentioned [here](https://forum.cosmos.network/t/pss-permissionless-vs-premissioned-lite-opt-in-consumer-chains/12984/17) before we thought of the simpler `consumerID` solution. I did not think it made sense to describe here potentially cumbersome solutions.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: The reason is to simplify implementation. `ConsumerAdditionProposal` already exists and contains everything we need. As a result, we would have less code changes.
> are there any side-effects doing so, i.e. query on gov proposals will show up these entries as well?
I do not see this being the case because we manually create `ConsumerAdditionProposal` and hall `HandleConsumerAdditionProposal`. `HandleConsumerAdditionProposal` is in gnerally called after a proposal has already passed. In any case, I'll take another look on this.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Wanted to write a consumer chain instead of a validator.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Not sure I fully understand your question. I think it would be good to have a verification that the new account owner is indeed the one that issued this message. This way, we can avoid cases where the `owner_address` is set incorrectly to an address noone actually owns that would prevent any future modifications.
The second signature is not necessary but seems helpful to avoid cases like the above.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Yes, that is correct. Rephrased.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+
+### The Phases of a Consumer Chain
+In Permissionless ICS, launching an Opt In chain is **only** possible through a transaction and not through a `ConsumerAdditionProposal`.
+Nevertheless, Permissionless ICS does not eliminate the `ConsumerAdditionProposal` governance proposal, as proposals are still necessary
+for Top N chains. Because of this, this ADR outlines a solution that attempts to preserve as much of the governance proposal code
+as possible. In what follows, what we describe applies to both governance-proposed consumer (i.e., Top N) chains and transaction-based (i.e., Opt In) chains.
+
+A consumer chain can reside in three phases: i) _prelaunched_, ii) _launched_, and iii) _stopped_ phase as seen
+in the diagram below:
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
```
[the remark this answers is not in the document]
insumity: Thanks. That's a good point. I modified the diagram and also included the `.excalidraw` file.

docs/docs/adrs/adr-018-permissionless-ics.md
p-offtermatt: Perfect, thanks :)

docs/docs/adrs/adr-018-permissionless-ics.md
p-offtermatt: Ok, makes sense, I am just a bit concerned that it will be very hard/tedious to do this in practice. imagine you want to give control over a consumer chain from one DAO to another DAO, the process now becomes "get dao 1 to sign a message, give that message to dao 2, get dao 2 to sign the message, post the message to chain".
It's possible, just an extra step that might be annoying. but you are right, it helps to avoid typos where a wrong account gets ownership

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+In Permissionless ICS, launching an Opt In chain is **only** possible through a transaction and not through a `ConsumerAdditionProposal`.
+Nevertheless, Permissionless ICS does not eliminate the `ConsumerAdditionProposal` governance proposal, as proposals are still necessary
+for Top N chains. Because of this, this ADR outlines a solution that attempts to preserve as much of the governance proposal code
+as possible.
+Additionally, to make the distinction between governance-proposed versus transaction-launched chains clearer, in Permissionless ICS,
+we can only add, modify, or remove Top N chains with governance proposals (i.e., `ConsumerAdditionProposal`, `ConsumerModificationProposal`,
+and `ConsumerRemovalProposal`) and we can only add, modify, or remove Opt In chains with transactions.
+In what follows, we describe changes that apply to both governance-proposed (i.e., Top N) and transaction-based (i.e., Opt In) consumer chains.
+
+A consumer chain can reside in three phases: i) _prelaunched_, ii) _launched_, and iii) _stopped_ phase as seen
```
mpoke, line 36: Need to align with the FE platform regarding the name of these phases. cc @MSalopek

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+in the diagram below:
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
+
+
+When a Top N chain is first proposed through a `ConsumerAdditionProposal` or an Opt In chain is added through a transaction,
+the consumer chain resides in the _prelaunched_ phase. At this state, validators can choose to opt in on the consumer chain. Additionally,
+an Opt In chain can choose to change parameters of the to-be-launched chains, such as `spawnTime`, etc. by issuing a specific transaction.
```
mpoke, line 43:
```suggestion
an Opt In chain can choose to change the consumer chain parameters, such as `spawnTime`, etc. by issuing a specific transaction.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+because validators were required to validate a consumer chain. However, after the introduction of PSS, a consumer chain can
+be either _Top N_ or _Opt In_. If a chain is an Opt In chain, then no validator is required to validate this chain unless they choose to.
+Because of this, we can launch an Opt In consumer chain without going through a governance proposal.
+
+This ADR presents _Permissionless_ ICS, a way in which an [_Opt In_](adr-015-partial-set-security.md) consumer chain can join
+ICS without needing a governance proposal but by simply issuing a transaction.
+
+## Decision
+
+### The Phases of a Consumer Chain
```
mpoke, line 26: Add a few lines about ownership of consumer chain for Opt In vs Top N. It would add clarity and help explaining who can modify chain params.

docs/docs/adrs/adr-018-permissionless-ics.md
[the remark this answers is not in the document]
mpoke: I think that's more relevant once we remove the init timeout. However, given that we move to consumers being identify by a consumer ID instead of a chain ID, this might not be such an issue. It's more of a garbage collection thing.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+When the [`spawnTime`](https://github.com/cosmos/interchain-security/blob/v4.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L55)
+passes and [at least one validator has opted in](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L455)
+the chain can launch and moves to the _launched_ phase. While in launched phase, a Top N consumer chain can choose to modify
+its parameters through a `ConsumerModificationProposal` and an Opt In chain can change its parameters by issuing a transaction.
+
+Lastly, a Top N chain can choose to exit ICS by issuing a `ConsumerRemovalProposal` and an Opt In chain can issue a transaction to stop the chain.
+After some period of time (e.g., provider's unbonding period), all state related to the stopped consumer chain can be removed. We
+keep track of the consumer chain's state for some period, so that we are able to punish validators for misbehaviour.
+
+Note that everything described so far and everything that follows applies to standalone chains as well.
```
mpoke, line 56:
```suggestion
Note that everything described so far and everything that follows applies to consumer chains that transitions from standalone chains as well.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+chain [has started](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L150).
+
+When the [`spawnTime`](https://github.com/cosmos/interchain-security/blob/v4.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L55)
+passes and [at least one validator has opted in](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L455)
+the chain can launch and moves to the _launched_ phase. While in launched phase, a Top N consumer chain can choose to modify
+its parameters through a `ConsumerModificationProposal` and an Opt In chain can change its parameters by issuing a transaction.
+
+Lastly, a Top N chain can choose to exit ICS by issuing a `ConsumerRemovalProposal` and an Opt In chain can issue a transaction to stop the chain.
+After some period of time (e.g., provider's unbonding period), all state related to the stopped consumer chain can be removed. We
+keep track of the consumer chain's state for some period, so that we are able to punish validators for misbehaviour.
```
mpoke, line 54:
```suggestion
keep track of the consumer chain's state for some period, so that we are able to punish validators for misbehaviours that occurred before the consumer chain stopped.
```

mpoke (commented):
Partial review.

docs/docs/adrs/adr-018-permissionless-ics.md
jtremback: @p-offtermatt we should be preventing this globally by how we do our prefixes and stuff. I haven't been in the code for so long that I don't remember how we do it, but if we aren't any more i would be worried

jtremback (approved):
LGTM

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+through a governance proposal). Note, that an alternative format for this message could have been to include a single `ConsumerAdditionProposal` field in it,
+but this confuses the semantics of `MsgLaunchConsumerChain` and that of a governance proposal. The response of this message
+(i.e., `MsgLaunchConsumerChainResponse`) contains a single `string` that is the `consumerID`. The `consumerID`s corresponds
+to `len(chainID) | chainID | counter` and this way we can have multiple different `consumerID`s that correspond to a chain
+with the same `chainID`. By doing this, we tackle the main issue behind permissionless consumer chains, that is chain-id squatting.
+Note that `consumerID` could just be a `uint64` but we choose to include `chainID` in it as well so that it is easier to see
+what the consumer chain just by looking at the `consumerID`. This means that the `chainID` of a chain cannot be changed
+after launching it (because the `chainID` is part of the `consumerID` key). To easily generate `consumerID`s we store
+for each `len(chainID) | chainID` the latest `counter` value used for this `chainID`.
```
bermuell, line 211: is there a need to protect the provider `chainID` , e.g. prevent consumer chains to use this ID as `chainID`?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+We set a to-be-decided fixed cost of a `MsgLaunchConsumerChain` to avoid getting spammed with bogus consumer chains.
+
+To execute a `MsgLaunchConsumerChain`, we first create a `ConsumerAdditionProposal` under the hoods, with the `top_N` set to 0, and call
+[`HandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L30)
+simulating a proposal that was already voted on and accepted. Note that we need to migrate `ConsumerAdditionProposal`s to use
+the `consumerID` instead of using [both](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L382)
+`chainID` and the `spawnTime` as keys. The [usual validity conditions](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/proposal.go#L114)
+hold for the fields of `MsgLaunchConsumerChain`. Note however, that we intend to have a `spawnTime` upper limit as well.
+For example, if you launch a consumer chain in Permissionless ICS, the `spawnTime` should not be more
+than two months ahead in the future, to avoid having consumer chains lingering for too long before they get added.
```
mpoke, line 222: Why? What's the problem with that?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+message MsgUpdateConsumerChain {
+  // for which chain to change the state ...
+  string consumer_id;
+
+  // if the user intends to change the owner_address
+  // the one issuing this message should be able to sign this message as well?
+  string owner_address;
+  
+  // the chain-id of the new consumer chain
+  string chain_id;
```
mpoke, line 249: Above you mentioned that you cannot modify the chain_id.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+This way, we can respond to queries that ask for all the consumer chain's parameters. For example, retrieving the
+`spawn_time` of consumer chain with a given `consumerID`.
+
+#### Modify a Consumer Chain
+
+We introduce the `MsgUpdateConsumerChain` message so that the owner of a consumer chain can change its parameters
+(e.g., `spawn_time`, PSS-related parameters, etc.) This message can only be executed by the owner of a consumer chain (see `owner_address` in the `ConsumerChainRecord`).
+
+```protobuf
+message MsgUpdateConsumerChain {
```
mpoke, line 240: What about the other fields from ConsumerChainRecord, e.g., client_id?

docs/docs/adrs/adr-018-permissionless-ics.md
mpoke: Yeah, I think we can drop this for now. Just have the current owner sign the message.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+is still in the prelaunched phase. This can be achieved by looking if there is an ongoing `ConsumerAdditionProposal` for this `consumerID`,
+and then changing the fields of this proposal. The other fields such as `allowlist`, etc. can be changed at any point before or after a chain has started.
+
+#### Stop a Consumer Chain
+With the `MsgStopConsumerChain` we can stop any Opt In chain at any moment. Note that all relevant state for this consumer chain
+remains on the provider's state before getting removed after one unbonding period (of the provider). This is to enable
+potential slashing for any infraction that might have been caused until now. After the unbonding period, the `ConsumerChainRecord`
+associated with this chain is removed. Note however that we never recycle used `consumerID`s. Naturally, this message
+can only be issued by the owner of the consumer chain. Also, any remaining IBC rewards that were to be sent to the provider chain
+are lost.
```
mpoke, line 275: > Also, any remaining IBC rewards that were to be sent to the provider chain are lost.

How are IBC rewards lost? If these rewards were sent by the consumer, can be distributed. Otherwise, the IBC rewards will be unescrowed on the consumer.

mpoke (commented):
Completed the review. See my comments below.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Why would this complicate things? Note that `ConsumerAdditionProposal` contains substantial more data (e.g., `spawnTime`) than what the `ConsumerChainRecord` ends up having.

I think it would make sense to rename `ConsumerAdditionProposal` to `PrelaunchConsumerChainRecord` and reuse it for gov proposals and the `MsgLaunchConsumerChain` message that would probably need to be renamed to `MsgPrelaunchConsumerChain`.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: We could just use a counter but seems easier (e.g., for debugging purposes) to also have the `chainID` there.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: (Talked offline) On the Hub, it's fine if a light-client tracks a chain with `chainID` that is `cosmoshub-4`. The same can apply here.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+| `ValidatorsByConsumerAddrBytePrefix`     |Stores consumer to provider validator address                                     |**YES**                   | Every `MsgAssignConsumerKey` or `MsgOptIn`                                                               |
+| `KeyAssignmentReplacementsBytePrefix`    |Deprecated                                                                        |-                         |                                                                                                          |
+| `ConsumerAddrsToPruneBytePrefix`         |Deprecated together with `VSCMaturedPacket`s                                      |-                         |                                                                                                          |
+| `SlashLogBytePrefix`                     |Not used                                                                          |-                         |                                                                                                          |
```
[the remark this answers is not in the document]
insumity: It is used but not for anything that matters, so we could as well consider that it's not used.
`SlashLogBytePrefix` is used to construct the `SlashLogKey` that is used in `SetSlashLog` and `GetSlashLog` methods. `SetSlashLog` is used [here](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/relay.go#L334) while `GetSlashLog` is **only** used in an integration test, so the fact that we set something but we have no getters to read it makes this unusable in my view. Additionally, the whole slashing for equivocation evidence chunk where `SetSlashLog` is called is not really used. Equivocation evidence is to be submitted through `MsgSubmitConsumerDoubleVoting` messages.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: There's no problem. With a maximum `spawnTime` we can enable garbage collection. Otherwise we might have consumer chains that stay there for ever.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: True, not needed. I'll remove.

insumity: True, good point. I'll remove.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Those would fall under `...  (all fields contained in ConsumerAdditionProposal except top_N)`.
I'll update to make it clearer.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: If we stop a consumer chain and delete everything in relation to it then this [call](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/distribution.go#L79) won't iterate through the stopped consumer chain and no tokens would be allocated to the consumer's validators. However, the IBC reward packet could have successfully been sent over to the provider chain (see [here](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/ibc_middleware.go#L125)). This seems to be the case today as well.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+  // the owner of this consumer chain
+  string owner_address;
+  // client id to the consumer chain
+  string client_id;
+  // channel id
+  string channel_id;
+  // the chain-id of the consumer chain
+  string chain_id;
+  // consumer genesis
+  interchain_security.ccv.v1.ConsumerGenesisState consumer_genesis = 5
```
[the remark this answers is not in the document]
insumity: Yes, this one is not needed here.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+  // First height of the consumer chain that the chain is secured for. Any infraction on the consumer chain is eligible
+  // if the infraction is performed on a height >= equivocation_evidence_min_height. Unless manually set, for a completely new chain,
+  // this would be 0, while for a standalone chain it would the height of when the chain converted to a consumer chain.
+  uint64 equivocation_evidence_min_height;
```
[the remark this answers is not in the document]
insumity: Yes, the owner of a consumer chain should **not** be able to modify the `equivocation_evidence_min_height` parameter. This parameter does not exist in `MsgUpdateConsumerChain`.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+
+#### Launch a Consumer Chain
+To prepare a consumer chain for launch, we issue a `MsgLaunchConsumerChain` message that is as follows:
+
+```protobuf
+message MsgLaunchConsumerChain {
+  // the title of this launch 
+  string title;
+  // the description of this consumer chain
+  string description;
```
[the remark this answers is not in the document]
insumity: Yes. We intend to have the checks that already exist for `ConsumerAdditionProposal` as mentioned:
> The [usual validity conditions](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/proposal.go#L114) hold for the fields of `MsgLaunchConsumerChain`.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+for each `len(chainID) | chainID` the latest `counter` value used for this `chainID`.
+
+We set a to-be-decided fixed cost of a `MsgLaunchConsumerChain` to avoid getting spammed with bogus consumer chains.
+
+To execute a `MsgLaunchConsumerChain`, we first create a `ConsumerAdditionProposal` under the hoods, with the `top_N` set to 0, and call
+[`HandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L30)
+simulating a proposal that was already voted on and accepted. Note that we need to migrate `ConsumerAdditionProposal`s to use
+the `consumerID` instead of using [both](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L382)
+`chainID` and the `spawnTime` as keys. The [usual validity conditions](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/proposal.go#L114)
+hold for the fields of `MsgLaunchConsumerChain`. Note however, that we intend to have a `spawnTime` upper limit as well.
```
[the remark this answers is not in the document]
insumity: You mean we need additional checks on the fields?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+`MsgLaunchConsumerChain` contains everything that is contained in [`ConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/proto/interchain_security/ccv/provider/v1/provider.proto#L29)
+except the `top_N` field because we can only launch Opt In consumer chains with this message (Top N chains still need to go
+through a governance proposal). Note, that an alternative format for this message could have been to include a single `ConsumerAdditionProposal` field in it,
+but this confuses the semantics of `MsgLaunchConsumerChain` and that of a governance proposal. The response of this message
+(i.e., `MsgLaunchConsumerChainResponse`) contains a single `string` that is the `consumerID`. The `consumerID`s corresponds
+to `len(chainID) | chainID | counter` and this way we can have multiple different `consumerID`s that correspond to a chain
+with the same `chainID`. By doing this, we tackle the main issue behind permissionless consumer chains, that is chain-id squatting.
+Note that `consumerID` could just be a `uint64` but we choose to include `chainID` in it as well so that it is easier to see
```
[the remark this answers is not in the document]
insumity: I'm okay with having this simply as a `counter`. However, having an `uint64` would need substantially more refactoring because the currently used `chainId` is a `string`.  Also, having the `counter` stored as a `string` would simplify our life if later on we intend to introduce more data to the `consumerId`. Finally, note that `clientId` is similarly a `string`, although it could simply be a `uint64`.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: @mpoke Updated the ADR to just use a `counter`.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Removed the maximum `spawnTime`. We iterate chains to be launched by ascending order based on their `spawnTime` so we do not have to iterate through all of them in any case.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+what the consumer chain just by looking at the `consumerID`. This means that the `chainID` of a chain cannot be changed
+after launching it (because the `chainID` is part of the `consumerID` key). To easily generate `consumerID`s we store
+for each `len(chainID) | chainID` the latest `counter` value used for this `chainID`.
+
+We set a to-be-decided fixed cost of a `MsgLaunchConsumerChain` to avoid getting spammed with bogus consumer chains.
+
+To execute a `MsgLaunchConsumerChain`, we first create a `ConsumerAdditionProposal` under the hoods, with the `top_N` set to 0, and call
+[`HandleConsumerAdditionProposal`](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L30)
```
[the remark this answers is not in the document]
insumity: Renamed to `PrelaunchConsumerChainRecord`.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Actually, removed the maximum `spawnTime` altogether. Did not see a benefit in it because when we traverse consumer chains to launch, we do so in ascending order of `spawnTime` and hence slowing down a [`BeginBlockInit`](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/keeper/proposal.go#L413).

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+Nevertheless, ICS does not eliminate the `MsgConsumerAddition` governance proposal, as proposals are still necessary
+for Top N chains. Because of this, this ADR outlines a solution that attempts to preserve as much of the governance proposal code
+as possible.
+Additionally, to make the distinction between governance-proposed versus transaction-launched chains clearer, in Permissionless ICS,
+we can only add, modify, or remove Top N chains with governance proposals (i.e., `MsgConsumerAddition`, `MsgConsumerModification`,
+and `MsgConsumerRemoval`) and we can only add, modify, or remove Opt In chains with transactions.
+Note however that a Top N chain can transform to an Opt In Chain through a `MsgConsumerModification` that sets `top_N == 0` but not vice versa.
+In what follows, what we describe applies mostly for transaction-based (i.e., Opt In) consumer chains, unless stated otherwise.
+
+A consumer chain can reside in four phases: i) waiting, ii) _prelaunched_, iii) _launched_, and iv) _stopped_ phase as seen
```
MSalopek, line 37: Seems there are extra phases.

I'm seeing these phases:
1. registered
* `consumerId` is assigned,`chain-id` is registered
* only for opt-in
* top_n can skip -> submitting a prop both registers and initializes the chain

2. initialized (pre-launch)
* it's signified by `spawn_time` being populated
* * all top_N start here - they must provide a `spawn_time` anyway
* opt-in chains transition to this stage by executing `MsgInitializeChain`
* spawn_time is present, hash fields are populated and unbonding perios is known

3. launched
* `spawn_time` passed
* IBC clients created
* opted-in validators list has been created and is available through `providerd q provider consumer-genesis`
* chain is live - some validators are running it

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+In the prelaunched phase, an Opt In chain can choose to change the consumer chain parameters, such as `spawnTime`, etc. by issuing anew `MsgInitiateConsumerChain`.
+This is not the case for Top N chains, where a `MsgConsumerModification` can only be issued after a consumer
+chain [has started](https://github.com/cosmos/interchain-security/blob/v5.1.0/x/ccv/provider/keeper/legacy_proposal.go#L89).
+
+When the [`spawnTime`](https://github.com/cosmos/interchain-security/blob/v5.1.0/proto/interchain_security/ccv/provider/v1/provider.proto#L57)
+passes and [at least one validator has opted in](https://github.com/cosmos/interchain-security/blob/v5.1.0/x/ccv/provider/keeper/proposal.go#L430)
+the chain can launch and moves to the _launched_ phase. While in launched phase, a Top N consumer chain can choose to modify
+its parameters through a `MsgConsumerModification` and an Opt In chain can change its parameters by issuing the `MsgUpdateConsumerChain` transaction.
+
+Lastly, a Top N chain can choose to  ICS by issuing a `MsgConsumerRemoval` and an Opt In chain can issue a transaction to stop the chain.
```
MSalopek, line 55: Governance cannot remove an `Opt-in` chain?

insumity:
@coderabbitai resolve

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Resolving this as outdated. `ConsumerChainRecord` or any substantial refactoring on keys is orthogonal to permissionless. I do not believe it should be part of this ADR.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+| `VSCMaturedHandledThisBlockBytePrefix`   |Deprecated together with `VSCMaturedPacket`s                                      |-                         |                                                                                                          |
+| `EquivocationEvidenceMinHeightBytePrefix` |Stores min height for a consumer chain                                            |**YES**                   | Only once (during set up)                                                                                |
+| `ProposedConsumerChainByteKey`           |Stores `proposalID` for a specific chain                                          |**YES**                   |                                                                                                          |
+| `ConsumerValidatorBytePrefix`            |Stores consumer validators for a specific chain                                   |**YES**                   | Potentially at every epoch                                                                               |
```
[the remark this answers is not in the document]
insumity: I totally agree but this is is not related to this ADR. I created this [issue](https://github.com/cosmos/interchain-security/issues/2087) so we do not forget.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: @MSalopek is aware of this ADR and intend to look on it, so resolving this comment.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,304 @@
...
+to check on whether a validator can be slashed.
+
+
+#### State
+As a result of using `consumerID`, we have to migrate a substantial chunk of state to re-index it using `consumerID` as the key.
+To move from using the `chainID` to the new `consumerID`, we need to revamp the consumer chains' stored state in ICS. Currently, in
+ICS we have state that is indexed by a multitude of [keys](https://github.com/cosmos/interchain-security/blob/v4.3.0/x/ccv/provider/types/keys.go#L40).
+In the table below, we see which ones are associated with a `chainID` and how often state under those keys gets updated.
+
+| Key                                     |Description                                                                       |Associated with `chainID`?| How often are `chainID`-associated keys updated?                                                         |
```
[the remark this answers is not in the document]
insumity: I agree. As any work on revamping the provider's state by using a new `ConsumerChainRecord` or something alike is orthogonal to this ADR, I'm resolving this.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Yes, a gov proposal cannot remove an Opt In chain. I would like to keep the distinction between those as clear as possible. Their only connection is when a Top N chain becomes an Opt In chain, otherwise gov proposals are only for Top N chains and transactions are only used for Opt In chains.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: > Seems there are extra phases.

It seems we still have the same number of phases if you include the _stopped_ phase. Naturally, after a chain has stopped and before having its data fully GCed, it resides in a different phase, what we call in this ADR, the stopped phase.

> top_n can skip -> submitting a prop both registers and initializes the chain

A Top N chain only gets initialized if its gets voted upon, so I think it makes sense to consider it registered before the corresponding proposal passes.

docs/docs/adrs/adr-018-permissionless-ics.md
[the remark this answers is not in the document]
mpoke: Ah yeah, this is legacy code from before we introduced cryptographic equivocation. @insumity could you please open an issue to remove this?

docs/docs/adrs/adr-018-permissionless-ics.md
mpoke: @insumity any thoughts on this?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,270 @@
...
+}
+```
+
+We store the `ConsumerChainRecord`s using the `consumerID` as their key. The `owner_address` of a consumer chain corresponds
+to the address that issued the `MsgLaunchConsumerChain` and the `owner_address` can be updated by issuing a `MsgUpdateConsumerChain` (see later).
+Note that a `ConsumerChainRecord` is only created when a consumer chain moves from the prelaunched to the _launched_ phase.
+The information needed to set up a `ConsumerChainRecord` are stored in the corresponding `ConsumerAdditionProposal` (see `MsgLaunchConsumerChain` below). 
+
+### New Messages
+In what follows, we describe the new messages that Permissionless ICS introduces and on how those can be used.
```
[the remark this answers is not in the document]
mpoke: nit: this still needs to be addressed

docs/docs/adrs/adr-018-permissionless-ics.md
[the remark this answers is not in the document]
mpoke: The usual validity conditions are not enough as the description field is not limited in length.

docs/docs/adrs/adr-018-permissionless-ics.md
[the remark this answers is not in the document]
mpoke: Not sure that I understand why is this more refactor. We need to migrate the state anyway to change the key. Does it matter how the new keys look like or what type we use? The keys are marshaled to []byte anyway.

docs/docs/adrs/adr-018-permissionless-ics.md
mpoke: So we should call AllocateTokens before removing a consumer chain, right?

docs/docs/adrs/adr-018-permissionless-ics.md
mpoke: Can we have these names for the phases:

- registered (there is a mapping consumerID -> chainID)
- ready-to-launch (there is also a spawn time)
- launched or active or running (there is also a clientID and a validator set)
- stopped

insumity:
@coderabbitai resolve

docs/docs/adrs/adr-018-permissionless-ics.md
mpoke: I agree with Matija. Rename the phases to registered, initialized, launched, stopped. Also, make consistent throughout the doc.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+
+A consumer chain can reside in four phases: i) waiting, ii) _prelaunched_, iii) _launched_, and iv) _stopped_ phase as seen
+in the diagram below:
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
+
+
+When a Top N chain is first proposed through a `MsgConsumerAddition` proposal or an Opt In chain is registered (more on this later) using the `MsgRegisterConsumerChain` transaction,
+the consumer chain resides in the _waiting_ phase. A consumer chain in the waiting phase might not launch (i.e., the `MsgConsumerAddition` proposal does not pass or the registered Opt In chain is not set to launch) and hence is "waiting."
+At this phase, as well as in the prelaunched and launched phases, validators can choose to opt in on the consumer chain.
+If the `MsgConsumerAddition` of a Top N chain passes or a registered Opt In chain is set to launch with the `MsgInitiateConsumerChain` transaction, then the chain moves to the _prelaunched_ phase.
```
mpoke, line 45:
```suggestion
If the `MsgConsumerAddition` of a Top N chain passes or a registered Opt In chain is set to launch with the `MsgInitializeConsumerChain` transaction, then the chain moves to the _prelaunched_ phase.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+we can only add, modify, or remove Top N chains with governance proposals (i.e., `MsgConsumerAddition`, `MsgConsumerModification`,
+and `MsgConsumerRemoval`) and we can only add, modify, or remove Opt In chains with transactions.
+Note however that a Top N chain can transform to an Opt In Chain through a `MsgConsumerModification` that sets `top_N == 0` but not vice versa.
+In what follows, what we describe applies mostly for transaction-based (i.e., Opt In) consumer chains, unless stated otherwise.
+
+A consumer chain can reside in four phases: i) waiting, ii) _prelaunched_, iii) _launched_, and iv) _stopped_ phase as seen
+in the diagram below:
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
+
+
```
mpoke, line 41: To improve clarity, separate the paragraphs, i.e., one paragraph per phase. 

For example, 
> **Registered phase.** When a Top N chain is first proposed through ... 

> **Initialized phase.** If the `MsgConsumerAddition` of a Top N chain passes or ...

> **Launched phase.** When the `spawnTime` passes ... 

> **Stopped phase.** Lastly, a Top N chain can choose to exit ICS by issuing ...

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+Note that everything described so far and everything that follows applies to consumer chains that transition from standalone chains as well.
+
+### From `chainId` to `consumerId`
+A hindrance in moving to Permissionless ICS is [chain-id squatting](https://forum.cosmos.network/t/pss-permissionless-vs-premissioned-lite-opt-in-consumer-chains/12984/17).
+In a permissionless setting, anyone could issue a transaction to launch a consumer chain with a `chainId` that might already be used by some other consumer chain. This is a problem
+because in the current design the majority of stored state for a consumer chain is indexed using the `chainId` as the key (e.g.,
+see [key used to store client ids](https://github.com/cosmos/interchain-security/blob/v5.1.0/x/ccv/provider/types/keys.go#L245)).
+To tackle this problem, in Permissionless ICS, we introduce the `consumerId` that defines a consumer chain and is simply
+an increasing counter (i.e., `counter`), thus we can support multiple consumer chains with the same `chainId`.
```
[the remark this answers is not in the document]
mpoke: The "i.e., `counter`" is not necessary. 
```suggestion
an increasing counter, thus we can support multiple consumer chains with the same `chainId`.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+Another way to understand this is with an analogy between consumer chains and IBC clients: Imagine having multiple IBC clients
+that each point to different consumer chains, but all share the exact same `chainId`. It is then up to the user to select the
+appropriate client (i.e., `clientID`) based on the actual chain they want to communicate with. Similarly, there can be multiple
+consumer chains with the exact same `chainId`, and it is the responsibility of the validator to choose the one they wish
```
mpoke, line 71:
```suggestion
consumer chains with the exact same `chainId`, and it is the responsibility of the validators to choose the one they wish
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+to interact with by providing the right `consumerId`.
+
+Although a `counter`, we store it as a `string` and not as an integer (e.g., `uint64`) for three reasons:
+i) avoid extensive refactorings when moving from `chainId` that is of type `string` to `consumerId`,
+ii) simplify migrations because we do not need to keep different methods at the same time (e.g., `ChainToClientKey(string)`
+and `ChainToClientKeyV2(uint64)` to do something like `client := Get(ChainToClientKey("currentConsumerChain))` and
+`Set(ChainToClientKeyV2(1234), client)`, and iii) allow for easy ways to incorporate more data to a `consumerId` if needed in the future.
```
mpoke, line 78: Not sure that this is needed. Also, I'm not sure I agree with this. Why does it have to be a string. What's the problem with having `ChainToClientKey(uint64)`? Why would you need the string version? In the migration code you can just use the definition of the old `ChainToClientKey(string)` (we do this in a lot of migrations).

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+
+
+Note that with Permissionless ICS, all interactions on a consumer chain have to use the `consumerId` instead of the `chainId`.
+For example, if a validator opts in on a chain using `MsgOptIn`, the validator has to provide the `consumerId`. To also
+provide the `consumerId` for Top N consumers chains, we store a mapping between `proposalID` to `consumerId`. This storing
+takes place in the [`AfterProposalSubmission`](https://github.com/cosmos/cosmos-sdk/blob/v0.50.8/x/gov/types/hooks.go#L19) hook.
+Note that in some cases, we still have to use the actual `chainId`, when, for example, we [verify equivocation evidence](https://github.com/cosmos/interchain-security/blob/v5.1.0/x/ccv/provider/keeper/consumer_equivocation.go#L143).
+Specifically, for the equivocation evidence, we update the `MsgSubmitConsumerMisbehaviour` and `MsgSubmitConsumerDoubleVoting` messages to include the `consumerId`,
+and change [Hermes](https://github.com/informalsystems/hermes) to include `consumerId` in those constructed messages as well.
```
mpoke, line 87: This is inconsistent.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+Hermes can find out the `consumerId` by querying the provider's `clientId` for some consumer chain (i.e., `query ccvconsumer provider-info`)
+and then asking the provider chain for the `consumerId` that corresponds to this `clientId`. To do this, we need to store
+the `clientId` to `consumerId` association on the provider and introduce a query to retrieve the `clientId`
+given the `consumerId`.
+
+#### State
+As a result of using `consumerId`, we have to migrate a substantial chunk of state to re-index it using `consumerId` as the key.
+To move from using the `chainId` to the new `consumerId`, we need to revamp the consumer chains' stored state in ICS. Currently, in
```
mpoke, line 95: These two sentences say the same thing.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+  // the owner of this consumer chain
+  string owner_address;
+}
+```
+
+This response of this message contains a single `string`, that is the `consumerId` for this registered consumer chain and initiates
+a consumer chain in its waiting phase. With the returned `consumerId`, validators can already opt in on the consumer
+chain to show their potential interest on the chain. Additionally, a front-end ICS launchpad can also present
+this chain. Additionally, this allows consumer chains to show that they are interested in joining ICS even though,
+they might not yet know the specific ICS parameters (see `MsgInitiateConsumerChain`) they would like to use.
```
mpoke, line 170:
```suggestion
they might not yet know the specific ICS parameters they would like to use (see `MsgInitiateConsumerChain`). 
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,342 @@
...
+
+This response of this message contains a single `string`, that is the `consumerId` for this registered consumer chain and initiates
+a consumer chain in its waiting phase. With the returned `consumerId`, validators can already opt in on the consumer
+chain to show their potential interest on the chain. Additionally, a front-end ICS launchpad can also present
+this chain. Additionally, this allows consumer chains to show that they are interested in joining ICS even though,
+they might not yet know the specific ICS parameters (see `MsgInitiateConsumerChain`) they would like to use. 
+
+This message contains the `owner_address` that corresponds to the address that would be able to initiate or later update this consumer chain.
+We store the owner address of each Opt In consumer chain by creating an association between `consumerId`s and `owner_address`es.
+Top N chains do not have an `onwer_address` because they can only be modified through governance proposals.
```
mpoke, line 174: Isn't that the governance account?

mpoke (commented):
Partial re-review.

insumity:
@coderabbitai resolve

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Yes, but there's no benefit for `owner_address` to contain something when we have Top N chains.

insumity:
@coderabbitai resolve

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+in the diagram below:
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
+
+
+**Registered phase.** When a Top N chain is first proposed through a `MsgConsumerAddition` proposal or an Opt In chain is registered (more on this later) using the `MsgRegisterConsumerChain` transaction,
+the consumer chain resides in the _registered_ phase. A consumer chain in the registered phase might not launch (i.e., the `MsgConsumerAddition` proposal does not pass or the registered Opt In chain is not set to launch).
+At this phase, as well as in the initialized and launched phases, validators can choose to opt in on the consumer chain.
+
+**Initialized phase.** If the `MsgConsumerAddition` of a Top N chain passes or a registered Opt In chain is set to launch with the `MsgInitializeConsumerChain` transaction, then the chain moves to the _initialized_ phase.
+In the initialized phase, an Opt In chain can choose to change the consumer chain parameters, such as `spawnTime`, etc. by issuing anew `MsgInitializeConsumerChain`.
```
mpoke, line 47:
```suggestion
In the initialized phase, an Opt In chain can choose to change the consumer chain parameters, such as `spawnTime`, etc. by issuing a new `MsgInitializeConsumerChain`.
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+  string chain_id;
+  // the owner of this consumer chain
+  string owner_address;
+}
+```
+
+This response of this message contains a single `string`, that is the `consumerId` for this registered consumer chain and initializes
+a consumer chain in its registered phase. With the returned `consumerId`, validators can already opt in on the consumer
+chain to show their potential interest on the chain. Additionally, a front-end ICS launchpad can also present
+this chain. Additionally, this allows consumer chains to show that they are interested in joining ICS even though,
```
mpoke, line 146: The front-end platform is orthogonal to this work and should not be part of this ADR. 
```suggestion
chain to show their potential interest on the chain. Additionally, this allows consumer chains to show that they are interested in joining ICS even though,
```

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+  // ---------- ---------- ----------
+  
+  // the title of the chain to-be-launched 
+  string title;
+  // the description of the chain to-be-launched
+  string description;
+  // the chain id of the new consumer chain
+  string chain_id;
+  // consumer id of the to-be-launched chain
+  string consumer_id;
```
mpoke, line 183: Why are these fields part of the initialization record? More important, why are these fields part of the MsgInitializeConsumerChain?

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+  // the owner of this consumer chain
+  string owner_address;
+  
+  PowerShapingParameters power_shaping_parameters;
+}
+```
+
+where `PowerShapingParameters` contains the following:
+```
+message PowerShapingParameters {
```
mpoke, line 255: nit: rename to `PowerShapingConfig`

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+  // consumer id of the to-be-updated consumer chain
+  string consumer_id;
+  // the initialization record that contains more specific parameters for the upcoming chain
+  InitializationRecord record;
+}
+```
+
+where `InitializationRecord` contains the following:
+```protobuf
+message InitializationRecord {
```
mpoke, line 171: rename to `ConsumerRecord`

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,339 @@
...
+}
+```
+
+Note, that even though a consumer chain is initialized with all the arguments in `InitializationRecord`,
+the `MsgUpdateConsumerChain` updates only the `owner_address` and the `power_shaping_parameters`. This is because 
+all the other arguments are either useless (e.g., `spanwTime`) after a chain has started, or can be updated directly
+by the consumer chain params (e.g., `consumer_redistribution_fraction`).
+
+are the arguments that can be changed through the provider and
+exist for a
```
mpoke, line 304: typo?

insumity:
@coderabbitai resolve
@codereabbitai stop
@codereabbitai pause

docs/docs/adrs/adr-018-permissionless-ics.md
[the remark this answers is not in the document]
insumity: Created this [issue](https://github.com/cosmos/interchain-security/issues/2093).

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Yes, that would be an option. Does not seem critical to me because the sending of IBC rewards over is not synchronized with the provider. In any case, created this [issue](https://github.com/cosmos/interchain-security/issues/2094) because this is orthogonal to permissionless work.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: After the offline discussion we had the other day, this has changed substantially so I'm resolving this as outdated.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: After the offline discussion we had the other day, this has changed substantially so I'm resolving this as outdated.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: So added parts referring to the fact that Top N chains are owned by the governance account, while others by the `owner_address` set internally. Feel free to re-open if this did not answer your question.

docs/docs/adrs/adr18_flows_of_launching_a_consumer_chain.png
mpoke, line 1: Not sure I agree with the text in red. I think no messages can be executed after the proposal is submitted as that will change ownership to the gov account.

docs/docs/adrs/adr-018-permissionless-ics.md
```diff
@@ -0,0 +1,347 @@
...
+![Phases of a consumer chain](./adr18_phases_of_a_consumer_chain.png)
+
+**Registered phase.** In the _registered phase_, a consumer chain is assigned a unique identifier that identifies a consumer chain
+that can later be used to interact with the specific consumer chain (e.g., when a validator opts in on a chain).
+After a chain has been registered, it can later be initialized and then launched. Specifically, Permissionless ICS introduces
+a `MsgRegisterConsumer` message that can be used to register **both** Top N and Opt In consumer chains.
+In the registered phase, it is not yet known if the consumer chain would end up being a Top N or an Opt In chain and hence
+the owner of the consumer chain at this phase is the one that signed the `MsgRegisterConsumer`.
+Note that currently, a consumer chain is registered when first proposed through a `MsgConsumerAddition` proposal
+message but with Permissionless ICS, the `MsgConsumerAddition` is deprecated and chains have to issue a `MsgRegisterConsumer` message instead.
```
mpoke, line 54: We need to make sure we migrate any existing `ConsumerAddtionProposal` when we upgrade.

mpoke (approved):
Great work @insumity.

docs/docs/adrs/adr18_flows_of_launching_a_consumer_chain.png
insumity: No message would be able to execute after the proposal is submitted because the ownership would already have changed to the gov account the moment the proposal gets proposed. As a result, any transaction issued afterwards would fail if it is not executed by the gov account.

The reason I had the text in red was to simplify things. Because otherwise we might have some weird race condition where an Opt In chain is about to start but then before it does we try to move it to a Top N chain by issuing a governance proposal. But this is not a big deal. We can simply erase any `ConsumerInitializationRecord`s or `ConsumerUpdateRecord`s if they already exist when we get a Top N proposal. So, let's allow this. I'm removing the text in red.

docs/docs/adrs/adr-018-permissionless-ics.md
insumity: Good point. Added a note in the `Additional Modifications` section.

Issue #1916, opened 2024-05-24
Inactive validators: Allow consumer chains to disable non-provider-vals from validating


p-offtermatt:
## Problem

Some consumer chains may not want to be validated by validators that are outside of the hub active set.
We should give those chains the option to disable the feature, or restrict the potential validators to a smaller set (e.g. using the first 200 validators on the hub instead of all 500 bonded ones)

## Closing criteria

We have implemented a way for consumer chains to *not* be validated by validators that are inactive on the hub.

## Implementation

A possible implementation is to add a new param that allows consumers to filter the *potential* validator set to the first N  validators. Ultimately, the need is that consumer chains want to be able to set a floor on "quality" of validators allowed to validate the chain, which includes the stake/rank they have on the hub.

PR #2035, opened 2024-07-16
feat!: Let consumer chains choose a minimum stake and validator rank

p-offtermatt:
## Description

Closes: #1916 

Read the ADR and changes to the docs first, then the rest is mostly boilerplate.

---

### Author Checklist

I have...

* [x] If this PR is library API breaking, bump the go.mod version string of the repo, and follow through on a new major release

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
  - Introduced functionality for setting minimum stake and maximum validator rank requirements for consumer chains.
  - Added mechanisms for managing validator power distribution, including capping power, creating allowlists and denylists, and defining minimum validator stake.

- **Bug Fixes**
  - Corrected a bug in the computation of the maximum capped power in the interchain security module.

- **Documentation**
  - Updated documentation to clarify the impact of new validator management features on chain security and validator participation.

- **Tests**
  - Added test cases for minimum stake and maximum validator rank parameters to ensure correct system behavior under different conditions.


docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -62,12 +62,16 @@ The following changes to the state are required:
...
+* Introduce two extra per-consumer-chain parameters: `MinStake` and `MaxValidatorRank`. `MinStake` is the minimum amount of stake a validator must have to be considered for validation on the consumer chain. `MaxValidatorRank` is the maximum rank of a validator that can validate on the consumer chain. The provider module will only consider the first `MaxValidatorRank` validators that have at least `MinStake` stake as potential validators for the consumer chain.
 
 ## Risk Mitigations
 
 To mitigate risks from validators with little stake, we introduce a minimum stake requirement for validators to be able to validate on consumer chains, which can be set by each consumer chain independently, with a default value set by the provider chain.
 
-Additionally, we indepdently allow individual consumer chains to disable this feature, which will disallow validators from outside the provider active set from validating on the consumer chain and revert them to the previous behaviour of only considering validators of the provider that are part of the active consensus validator set.
+Additionally, we indepdently allow individual consumer chains to set a maximum rank for validators.
```
p-offtermatt, line 71:
```suggestion
Additionally, we independently allow individual consumer chains to set a maximum rank for validators.
```

.changelog/unreleased/features/provider/2035-min-stake-max-rank.md
```diff
@@ -0,0 +1,3 @@
+- Add minimum stake and maximum validator rank requirements to let consumer chains
+  determine requirements for validators that validate their chain. ([\#2035](https://github.com/cosmos/interchain-
+  security/pull/2035))
```
insumity, line 3: Link probably needs to be in the same line. Here and in the file under `state-breaking/provider` as well.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -62,12 +62,16 @@ The following changes to the state are required:
...
 
 ## Risk Mitigations
 
 To mitigate risks from validators with little stake, we introduce a minimum stake requirement for validators to be able to validate on consumer chains, which can be set by each consumer chain independently, with a default value set by the provider chain.
 
-Additionally, we indepdently allow individual consumer chains to disable this feature, which will disallow validators from outside the provider active set from validating on the consumer chain and revert them to the previous behaviour of only considering validators of the provider that are part of the active consensus validator set.
+Additionally, we independently allow individual consumer chains to set a maximum rank for validators.
+This means that validators above a certain position in the validator set cannot validate on the consumer chain.
+Setting this to be equal to `MaxProviderConsensusValidators` effectively disables this feature,
```
insumity, line 73:
```suggestion
Setting `MaxValidatorRank` to be equal to `MaxProviderConsensusValidators` effectively disables this feature,
```

Can `MaxValidatorRank` be smaller than `MaxProviderConsensusValidators`?

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -106,6 +106,10 @@ message ConsumerAdditionProposal {
   repeated string allowlist = 18;
   // Corresponds to a list of provider consensus addresses of validators that CANNOT validate the consumer chain.
   repeated string denylist = 19;
+  // Corresponds to the minimal amount of (provider chain) stake required to validate on the consumer chain.
+  int64 min_stake = 20;
```
insumity, line 110: Why aren't those `uint...`s?

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -832,32 +833,63 @@ func TestDenylist(t *testing.T) {
 	require.True(t, providerKeeper.IsDenylistEmpty(ctx, chainID))
 }
 
-func TestMinimumPowerInTopN(t *testing.T) {
+// Tests setting, getting and deleting number parameters that are stored per-consumer chain.
+// The tests cover the following parameters:
+// - MinimumPowerInTopN
+// - MinStake
+// - MaxValidatorRank
+func TestKeeperIntConsumerParams(t *testing.T) {
```
insumity, line 841: This test seems inconsistent with the way we test other params (e.g., `TestValidatorSetCap`, `TestValidatorsPowerCap`, etc.) where we have one test per param.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,54 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+		return false
+	}
+
+	// validator has enough stake to validate the chain
+	return validator.GetTokens().GTE(math.NewInt(minStake))
+}
+
 // ComputeNextValidators computes the validators for the upcoming epoch based on the currently `bondedValidators`.
 func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator) []types.ConsensusValidator {
+	// sort the bonded validators by power in descending order
```
insumity, line 330: Comments states we sort by `power`, but sorting by number of tokens here.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,54 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+	sort.Slice(bondedValidators, func(i, j int) bool {
+		iTokens := bondedValidators[i].GetTokens()
+		jTokens := bondedValidators[j].GetTokens()
+		result := iTokens.GT(jTokens)
+		return result
```
insumity, line 335: Why not simply do this in just one line: `return bondedValidators[i].GetTokens().GT(bondedValidators[j].GetTokens())`?

Could `sort.Slice` lead to different orderings if 2 validators have the same number of tokens? For example, if we have validator `A (tokens = 50)` and `B  (tokens = 50)` that we end up with in one run with `A, B` and in another with `B, A`? I do not think this is the case (we already use `sort.Slice` extensively but might be good to verify).

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,54 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+	})
+
+	// take only the first `MaxValidatorRank` many validators; others are not allowed to validate
+	maxRank, found := k.GetMaxValidatorRank(ctx, chainID)
+	if found && maxRank > 0 && int(maxRank) < len(bondedValidators) {
+		tmpValidators := bondedValidators[:maxRank]
+
+		// also include other validators that have the same number of tokens as the last validator in the list
+		for i := maxRank; int(i) < len(bondedValidators); i++ {
```
insumity, line 344: nit:
```suggestion
		for i := int(maxRank); i < len(bondedValidators); i++ {
```

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,54 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+	sort.Slice(bondedValidators, func(i, j int) bool {
+		iTokens := bondedValidators[i].GetTokens()
+		jTokens := bondedValidators[j].GetTokens()
+		result := iTokens.GT(jTokens)
+		return result
+	})
+
+	// take only the first `MaxValidatorRank` many validators; others are not allowed to validate
+	maxRank, found := k.GetMaxValidatorRank(ctx, chainID)
+	if found && maxRank > 0 && int(maxRank) < len(bondedValidators) {
```
insumity, line 340: What happens if `maxRank >= len(bondedValidators)`?
Shouldn't we only get up to `maxRank` somehow the `bondedValidators`?

insumity:
What the relation between minimum stake and the validator power cap? It seems that a validator `V` might not be added due to a minimum stake requirement but after the application of the validator power cap, `V` might have an increase in its power (and hence an increase in `V`'s supposedly tokens). If that's the case, why shouldn't we add `V`?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: Yeah, don't see a reason why not, it just gives more expressive power to consumer chains. Maybe they think that the last 100 validators in the active set shouldn't be able to validate their chain; this allows them to express that.
They can for sure set parameters that will not make much sense, but I don't see a reason to forbid them programatically.

x/ccv/provider/keeper/keeper_test.go
p-offtermatt: This is just a more concise way to test it imo, one table-based test with N cases vs N tests that repeat the same logic

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: Woops, this was for debugging. Thanks for the catch!

x/ccv/provider/keeper/keeper_test.go
p-offtermatt: Actually, I'll move the other params that use integers here, too, unless you see a reason to prefer the separate tests?

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: bondedValidators[:maxRank] just fails if maxRank > len(bondedValidators) 
it's just the way slices work in golang.

> Shouldn't we only get up to maxRank somehow the bondedValidators?
Could you clarify this part? Not sure what you mean

proto/interchain_security/ccv/provider/v1/provider.proto
p-offtermatt: Thanks, changed them to uints.

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: Oh, I reread the text and realize why this might be confusing.

x/ccv/provider/keeper/keeper.go
```diff
@@ -1582,3 +1582,81 @@ func (k Keeper) DeleteMinimumPowerInTopN(
...
+	chainID string,
+	minStake int64,
+) {
+	store := ctx.KVStore(k.storeKey)
+
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(minStake))
+
+	store.Set(types.MinStakeKey(chainID), buf)
+}
```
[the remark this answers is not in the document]
p-offtermatt: The Set, Get and Delete keeper functions should never return errors.

p-offtermatt:
The minimum stake is meant to be a minimum ATOM amount of stake; this is supposed to mitigate attacks by validators with very little stake, so the amount of stake is relevant, not the voting power

x/ccv/provider/keeper/keeper_test.go
insumity: This was more of a nit comment. You can leave as is or change it to include other integer params.I just wanted to mention it.

x/ccv/provider/keeper/partial_set_security.go
insumity: >Shouldn't we only get up to maxRank somehow the `bondedValidators`?

Actually, this was something I misunderstood from my side. For example, `bondedValidators` could be up to 500 validators, right? If so, my comment does not make sense and you can skip.

>bondedValidators[:maxRank] just fails if maxRank > len(bondedValidators)
it's just the way slices work in golang.

Is this possible to occur? E.g., if I start a chain permissionlessly and put `maxRank = 501`, could this lead to a `panic` and hence a chain halt? If yes, I guess we could restrict `maxRank` to be 500 (if that's the maximum number of bonded validators), but then what happens if the bonded validators are later set to 450 while a running consumer chain has `maxRank > 450`?

p-offtermatt: No, I don't think so. This if condition is what makes it so that no matter what your max rank is, you will not try to take more elements from the bondedValidators than are in that list. Does it make sense?

insumity (approved):

Issue #1913, opened 2024-05-24
Allow validators that are not participating in consensus on the provider to validate consumer chains


p-offtermatt:
## Problem

Currently, only validators in the active set on the provider can validate on consumer chains, which limits the number of validators that can participate in Interchain Security (ICS). 
Validators outside of the active set might be willing
to validate on consumer chains, but we might not want to make the provider validator set larger, e.g. to not put more strain on the consensus engine.
This runs the risk of leaving consumer chains with too few validators.

## Closing criteria

We have implemented a way to let validators that are not part of consensus on the provider chain
validate on consumer chains.

## Problem details

See this draft ADR for more context: https://github.com/cosmos/interchain-security/pull/1879/files

PR #2079, opened 2024-07-22
feat!: Allow inactive provider chain validators to validate on consumer chains

p-offtermatt:
## Description

Closes: #1913 

Combines all changes for allowing inactive validators to validate on consumer chains.
Recommended order of reviewing:
* Review the ADR for context
* Review the core logic changes:
   * New parameters/consumer chain configurations in the provider and how they are used
   * Check the bonded validators and bonded active validators functions and where they are used: GetLastActiveBondedValidators vs GetLastBondedValidators need to be used correctly (GetLastBondedValidators if all bonded validators should be included, GetLastActiveBondedValidators if only active validators should be included)
   * Review how provider validator updates are computed in EndBlock (see module.go in x/ccv/provider)
   * Review how validator updates for consumer chains are computed in x/ccv/provider/keeper/partial_set_security.go
   * Review the new app wiring in app/provider/app.go
* Review tests:
   * Simulator is now enabled and checks some properties, see app/provider/sim_test.go
   * e2e tests: there are a lot of e2e tests because the feature is better suited to be tested e2e. integration tests are not very extensive here in comparison

---

### Author Checklist

I have...

* [x] If this PR is library API breaking, bump the go.mod version string of the repo, and follow through on a new major release

### Reviewers Checklist

I have...


x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,62 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+			// if the validator has the same number of tokens as the last validator in the list, include it
+			if bondedValidators[i].GetTokens().Equal(bondedValidators[maxRank-1].GetTokens()) {
+				tmpValidators = append(tmpValidators, bondedValidators[i])
+			} else {
+				// since validators are sorted, we can break if we get to a validator with less tokens
+				break
+			}
+		}
+		bondedValidators = tmpValidators
+	}
```
mpoke, line 362: Do we actually want this max validator rank? It adds a bunch of logic and I'm not sure what's the use case.

x/ccv/provider/keeper/genesis.go
```diff
@@ -103,6 +105,50 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
 
 	k.SetParams(ctx, genState.Params)
 	k.InitializeSlashMeter(ctx)
+
+	// Set the last provider consensus validator set
+	k.SetLastProviderConsensusValSet(ctx, genState.LastProviderConsensusValidators)
```
mpoke, line 110: This action seems redundant given that InitGenesisValUpdates will set the valset to something else.

x/ccv/provider/keeper/relay.go
```diff
@@ -160,6 +169,46 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+
+	return valUpdates, nil
+}
+
+// ProviderValidatorUpdates returns changes in the provider consensus validator set
+// from the last block to the current one.
+// It retrieves the bonded validators from the staking module and creates a `ConsumerValidator` object for each validator.
+// The maximum number of validators is determined by the `maxValidators` parameter.
+// The function returns the difference between the current validator set and the next validator set as a list of `abci.ValidatorUpdate` objects.
+func (k Keeper) ProviderValidatorUpdates(ctx sdk.Context) []abci.ValidatorUpdate {
```
mpoke, line 181: Is this function tested?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -62,15 +62,111 @@ The following changes to the state are required:
...
 
-Additionally, we independently allow individual consumer chains to disable this feature, which will disallow validators from outside the provider active set from validating on the consumer chain and revert them to the previous behaviour of only considering validators of the provider that are part of the active consensus validator set.
+Additionally, we independently allow individual consumer chains to set a maximum rank for validators.
+This means that validators above a certain position in the validator set cannot validate on the consumer chain.
+Setting this to be equal to `MaxProviderConsensusValidators` effectively disables inactive validators from validating on the consumer chain and thus
+disables the main feature described in this ADR.
 
 Additional risk mitigations are to increase the active set size slowly, and to monitor the effects on the network closely. For the first iteration, we propose to increase the active set size to 200 validators (while keeping the consensus validators to 180), thus letting the 20 validators with the most stake outside of the active set validate on consumer chains.
 
+## Testing Scenarios
```
mpoke, line 81: Nice

mpoke (commented):
Great work @p-offtermatt. See my comments below.

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -106,6 +106,12 @@ message ConsumerAdditionProposal {
   repeated string allowlist = 18;
   // Corresponds to a list of provider consensus addresses of validators that CANNOT validate the consumer chain.
   repeated string denylist = 19;
+  // Corresponds to the minimal amount of (provider chain) stake required to validate on the consumer chain.
+  uint64 min_stake = 20;
+  // Corresponds to the maximal rank in the provider chain validator set that a validator can have to validate on the consumer chain.
+  uint32 max_rank = 21;
```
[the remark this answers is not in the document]
p-offtermatt: Example: 
* Set it to 200 if you want only active validators + the first 20 validators with the most stake to validate
* Set it to 100 if you only want the 100 active validators with the most stake to be able to validate
is it confusing?

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -183,6 +183,12 @@ message MsgConsumerAddition {
   repeated string denylist = 17;
   // signer address
   string authority = 18 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // Corresponds to the minimal amount of (provider chain) stake required to validate on the consumer chain.
+  uint64 min_stake = 19;
+  // Corresponds to the maximal rank in the provider chain validator set that a validator can have to validate on the consumer chain.
+  uint32 max_rank = 20;
+  // Corresponds to whether inactive validators are allowed to validate the consumer chain.
+  bool allow_inactive_vals = 21;
```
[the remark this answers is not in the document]
p-offtermatt: Yeah, good point.

docs/docs/features/power-shaping.md
```diff
@@ -28,21 +28,29 @@ This is a soft cap, and the actual power of a validator can exceed this fraction
...
 
 :::warning
 Note that if denylisting is used in a Top N consumer chain, then the chain might not be secured by N% of the total provider's
 power. For example, consider that the top validator `V` on the provider chain has 10% of the voting power, and we have a Top 50% consumer chain,
 then if `V` is denylisted, the consumer chain would only be secured by at least 40% of the provider's power.
 :::
 
+4) **Maximum validator rank**: The consumer chain can specify a maximum position in the validator set that a validator can have on the provider chain to be able to validate the consumer chain. This can be used to ensure that only validators with relatively large amounts of stake can validate the consumer chain. For example, setting this to 20 would mean only the 20 validators with the most voting stake on the provider chain can validate the consumer chain.
```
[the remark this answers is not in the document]
p-offtermatt: Validator with the most stake = rank 1
Last active validator = rank 180

I think this is the more natural way of counting

.changelog/unreleased/state-breaking/provider/1925-apply-audit-suggestions.md
```diff
@@ -0,0 +1,2 @@
+- Apply audit suggestions that include a bug fix in the way we compute the
```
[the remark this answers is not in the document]
p-offtermatt: It got put back when I merged; will remove

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/cosmos/interchain-security/v5/x/ccv/provider/types"
+)
+
+// getValidatorKey constructs the key to access a given validator, stored under a given prefix.
+func (k Keeper) getValidatorKey(prefix []byte, providerAddr types.ProviderConsAddress) []byte {
+	return append(prefix, providerAddr.ToSdkConsAddr()...)
+}
+
+// setValidator stores the given `validator` in the validator set stored under the given prefix.
+func (k Keeper) setValidator(
```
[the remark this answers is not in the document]
p-offtermatt: Good point, changing to an error return

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,106 @@
...
+	ctx sdk.Context,
+	prefix []byte,
+	providerConsAddr types.ProviderConsAddress,
+) {
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(k.getValidatorKey(prefix, providerConsAddr))
+}
+
+// deleteValSet deletes all the stored consumer validators under the given prefix.
+func (k Keeper) deleteValSet(
```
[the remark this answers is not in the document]
p-offtermatt: Not sure what you mean, could you explain whether you think something should change in the function?

x/ccv/provider/simulation/genesis.go
```diff
@@ -0,0 +1,44 @@
...
+	maxProviderConsensusValidators = "max_provider_consensus_validators"
+)
+
+// genMaxProviderConsensusValidators returns randomized maxProviderConsensusValidators
+func genMaxProviderConsensusValidators(r *rand.Rand) int64 {
+	return int64(r.Intn(250) + 1)
+}
+
+// RandomizedGenState generates a random GenesisState for staking
+func RandomizedGenState(simState *module.SimulationState) {
```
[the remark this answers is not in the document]
p-offtermatt: This is not a test, this is the genesis setup for the simulator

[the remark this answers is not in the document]
p-offtermatt: it is supposed to initialize parameters at random. see this in the gov module: https://github.com/cosmos/cosmos-sdk/blob/a32186608aab0bd436049377ddb34f90006fcbf7/x/gov/simulation/genesis.go#L28

x/ccv/provider/migrations/v7/legacy_params.go
```diff
@@ -95,5 +95,7 @@ func GetParamsLegacy(ctx sdk.Context, paramspace ccvtypes.LegacyParamSubspace) t
 		getConsumerRewardDenomRegistrationFee(ctx, paramspace),
 		getBlocksPerEpoch(ctx, paramspace),
 		getNumberOfEpochsToStartReceivingRewards(ctx, paramspace),
+		// this parameter is new so it doesn't need to be migrated, just initialized
+		types.DefaultMaxProviderConsensusValidators,
```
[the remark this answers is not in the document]
p-offtermatt: I added the migration in vX.
I still need to have this line here afaict because the parameters that NewParams takes changed

x/ccv/provider/keeper/relay.go
p-offtermatt: It's hit in tests, yes, in particular many integration tests pass through this.
Do you think a separate unit test is necessary? I think it doesn't provide a ton of value, since DiffValidators is tested, and this function doesn't really have any logic other than that call,
so I'm happy with how it's tested, but I can add a test just for this if your call is that it's important to have it.

p-offtermatt: Added a unit test for the function anyways :)

proto/interchain_security/ccv/provider/v1/provider.proto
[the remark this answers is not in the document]
mpoke: With the example is easier to understand. For me a higher rank sounds like a positive which means that the top validator has highest rank. I know that this might not be true. So to address the potential confusion of users, either change the name to something else or just add the examples to the docstring in the proto file and to the documentation of power shaping.

docs/docs/features/power-shaping.md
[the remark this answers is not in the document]
mpoke: See https://github.com/cosmos/interchain-security/pull/2079#discussion_r1691088985

x/ccv/provider/simulation/genesis.go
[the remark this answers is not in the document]
mpoke: Can we expand it for more params besides `MaxProviderConsensusValidators`. It can be part of another PR.

proto/interchain_security/ccv/provider/v1/provider.proto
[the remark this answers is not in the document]
p-offtermatt: Added

x/ccv/provider/keeper/validator_set_storage.go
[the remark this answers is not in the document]
mpoke: You can use the exact same function to delete any prefix, not just the valset. My suggestion is to add this function as a general utility function. However, this is not a must.

x/ccv/provider/keeper/partial_set_security.go
mpoke: I also think that minStake should be enough. I would like to avoid adding more params just for the sake of it without actually having a real need. It just complicates the code and the UX.

x/ccv/provider/simulation/genesis.go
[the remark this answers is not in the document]
p-offtermatt: Yeah we can, it's a question of what benefit it gives us. We aren't checking anything about the other parameters, and in fact most are completely unused without consumer chains. I want to avoid putting lots of stuff here that just does nothing in relaity

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: I will remove it then. Agreed that this probably makes UX more confusing. We can add it back if someone asks for it

p-offtermatt:
I have removed the maxRank parameter

proto/interchain_security/ccv/provider/v1/tx.proto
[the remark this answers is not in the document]
insumity: Created [issue](https://github.com/cosmos/interchain-security/issues/2103) so we keep track of this.

p-offtermatt:
@mpoke I also added some more simulator functionality to check that when there are no inactive validators, the provider keeper does the same as the staking keeper.
commits summarized for convenience:
https://github.com/cosmos/interchain-security/pull/2079/files/60bf2611e129475696237652ec996317740db304..a73ef5a9

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -106,6 +106,10 @@ message ConsumerAdditionProposal {
   repeated string allowlist = 18;
   // Corresponds to a list of provider consensus addresses of validators that CANNOT validate the consumer chain.
   repeated string denylist = 19;
+  // Corresponds to the minimal amount of (provider chain) stake required to validate on the consumer chain.
```
insumity, line 109: This applies irrespectively of whether the `allow_inactive_vals` flag is set or not, right?

p-offtermatt: Yes!

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -343,24 +343,6 @@ func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedVal
 		}
 	}
 
-	// take only the first `MaxValidatorRank` many validators; others are not allowed to validate
```
mpoke: Is the sorting above still needed?

go.mod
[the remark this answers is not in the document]
p-offtermatt: Thanks!

tests/e2e/actions.go
```diff
@@ -1916,7 +1918,7 @@ func (tr Chain) registerRepresentative(
 				panic(fmt.Sprintf("failed writing ccv consumer file : %v", err))
 			}
 			defer file.Close()
-			err = os.WriteFile(file.Name(), []byte(fileContent), 0600)
+			err = os.WriteFile(file.Name(), []byte(fileContent), 0o600)
```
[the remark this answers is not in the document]
p-offtermatt: No, it happens when running `make format` (which afaict is the source-of-truth for what formatting is correct)

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: The sorting above is for the inactive vals, too, see the comment below
```
// if inactive validators are not allowed, only consider the first `MaxProviderConsensusValidators` validators
// since those are the ones that participate in consensus
```

In general, this is in principle called with a sorted list already, but sorting here seems reasonable as a defensive measure (since not having a sorted list here is pretty bad and breaks PoS)

x/ccv/provider/migrations/vX/migrations.go
```diff
@@ -15,10 +15,11 @@ func InitializeMaxValidatorsForExistingConsumers(ctx sdk.Context, providerKeeper
...
-// for all existing consumer chains.
-func InitializeAllowInactiveVals(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
-	for _, chainID := range providerKeeper.GetAllRegisteredConsumerChainIDs(ctx) {
-		providerKeeper.SetAllowInactiveValidators(ctx, chainID, false)
+// InitializeMaxProviderConsensusParam initializes the MaxProviderConsensusValidators parameter.
+func InitializeMaxProviderConsensusParam(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
+	params := providerKeeper.GetParams(ctx)
+	if params.MaxProviderConsensusValidators == 0 {
+		params.MaxProviderConsensusValidators = 180
+		providerKeeper.SetParams(ctx, params)
```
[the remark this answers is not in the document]
mpoke: Yes.

x/ccv/provider/keeper/genesis.go
```diff
@@ -140,5 +188,6 @@ func (k Keeper) ExportGenesis(ctx sdk.Context) *types.GenesisState {
 		k.GetAllValidatorConsumerPubKeys(ctx, nil),
 		k.GetAllValidatorsByConsumerAddr(ctx, nil),
 		consumerAddrsToPrune,
+		lastProviderConsensusValSet,
```
mpoke, line 191: Why do we export this if we never use it on import?

x/ccv/provider/keeper/genesis.go
```diff
@@ -81,6 +83,47 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState *types.GenesisState) {
...
+// from the staking module's validator set.
+func (k Keeper) InitGenesisValUpdates(ctx sdk.Context) []abci.ValidatorUpdate {
+	// get the staking validator set
+	valSet, err := k.stakingKeeper.GetBondedValidatorsByPower(ctx)
+	if err != nil {
+		panic(fmt.Errorf("retrieving validator set: %w", err))
+	}
+
+	// restrict the set to the first MaxProviderConsensusValidators
+	maxVals := k.GetParams(ctx).MaxProviderConsensusValidators
```
mpoke, line 101:
```suggestion
	maxVals := k.GetMaxProviderConsensusValidators(ctx)
```

x/ccv/provider/migrations/v7/legacy_params.go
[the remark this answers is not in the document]
insumity: Where's `vX`?
Shouldn't we have something that sets the param similar to [this](https://github.com/cosmos/interchain-security/blob/3e7cd06035bb8863778001b7b77ae57a436222fb/x/ccv/provider/migrations/v4/migrations.go#L13) somewhere?

x/ccv/provider/keeper/relay.go
```diff
@@ -144,17 +196,26 @@ func (k Keeper) QueueVSCPackets(ctx sdk.Context) {
...
 		if topN > 0 {
 			// in a Top-N chain, we automatically opt in all validators that belong to the top N
-			minPower, err := k.ComputeMinPowerInTopN(ctx, bondedValidators, topN)
+			// of the active validators
+			activeValidators, err := k.GetLastActiveBondedValidators(ctx)
+			if err != nil {
+				// we just log here and do not panic because panic-ing would halt the provider chain
+				k.Logger(ctx).Error("failed to get active validators", "error", err)
+			}
+			minPower, err := k.ComputeMinPowerInTopN(ctx, activeValidators, topN)
```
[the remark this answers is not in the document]
p-offtermatt: You are right of course. This feels problematic.
I see two ways, we can either assume minPower = 0 and make everyone validate in this case, or we do panic.
I decided to assume 0 power for now, but lmk what you think

x/ccv/provider/types/keys.go
```diff
@@ -301,6 +307,17 @@ func getKeyPrefixes() map[string]byte {
 		// consumer validators addresses that need to be pruned.
 		ConsumerAddrsToPruneV2KeyName: 41,
 
+		// LastProviderConsensusValsPrefix is the byte prefix for storing the last validator set
```
insumity, line 310:
```suggestion
		// LastProviderConsensusValsKeyName is the byte prefix for storing the last validator set
```

insumity, line 310: The comment should start with `LastProviderConsensusValsKey`. 
I'm not sure if the comment should be `LastProviderConsensusValsKeyName` or `LastProviderConsensusValsKey` actually. It seems all key names remove the `Name` suffix from the comment (?)

x/ccv/provider/types/genesis_test.go
```diff
@@ -39,6 +39,7 @@ func TestValidateGenesisState(t *testing.T) {
 				nil,
 				nil,
 				nil,
+				nil,
```
insumity, line 42: Would a test make sense here on invalid `lastProviderConsensusValidators `?
Wouldn't a `nil` provider consensus val set be invalid (?) Could that ever be possible?

x/ccv/provider/keeper/distribution.go
```diff
@@ -191,7 +191,17 @@ func (k Keeper) AllocateTokensToConsumerValidators(
 	}
 
 	// Allocate tokens by iterating over the consumer validators
-	for _, consumerVal := range k.GetConsumerValSet(ctx, chainID) {
+	consumerVals, err := k.GetConsumerValSet(ctx, chainID)
+	if err != nil {
+		k.Logger(ctx).Error(
+			"cannot get consumer validator set while allocating rewards from consumer chain",
```
insumity, line 197:
```suggestion
			"cannot get consumer validator set while allocating rewards for consumer chain",
```

x/ccv/provider/keeper/distribution.go
```diff
@@ -272,7 +282,17 @@ func (k Keeper) GetConsumerRewardsPool(ctx sdk.Context) sdk.Coins {
 // for the given consumer chain
 func (k Keeper) ComputeConsumerTotalVotingPower(ctx sdk.Context, chainID string) (totalPower int64) {
 	// sum the consumer validators set voting powers
-	for _, v := range k.GetConsumerValSet(ctx, chainID) {
+	vals, err := k.GetConsumerValSet(ctx, chainID)
```
insumity, line 285: Why not call this `consumerVals` instead of `vals` as you do above in `AllocateTokensToConsumerValidators`?

x/ccv/provider/keeper/grpc_query_test.go
```diff
@@ -165,6 +168,11 @@ func TestQueryConsumerChainsValidatorHasToValidate(t *testing.T) {
 	// set `providerAddr` as an opted-in validator on "chain3"
 	pk.SetOptedIn(ctx, "chain3", providerAddr)
 
+	// set max provider consensus vals to include all validators
+	params := pk.GetParams(ctx)
+	params.MaxProviderConsensusValidators = 180
```
insumity, line 173: Does this have to be `180` for this test or it has to be smaller?

x/ccv/provider/keeper/keeper.go
```diff
@@ -1158,6 +1162,89 @@ func (k Keeper) DeleteMinimumPowerInTopN(
...
+	if allowed {
+		k.AllowInactiveValidators(ctx, chainID)
+	} else {
+		k.DeleteAllowInactiveValidators(ctx, chainID)
+	}
+}
+
+// AllowInactiveValidators sets the flag to signal that inactive validators are allowed to validate
+// a given consumer chain.
+func (k Keeper) AllowInactiveValidators(
```
insumity, line 1220: Name too similar with `AllowsInactiveValidators` that is a getter. Might make sense to try other names, e.g., `EnableInactiveValidators` for the setting and `AreInactiveValidatorsEnabled`.

insumity: Also the name of `SetAllowInactiveValidators` above is slightly peculiar because you can call it with `false` to disallow inactive validators but the method is called `SetAllow`.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,44 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+	// validator has enough stake to validate the chain
+	return validator.GetBondedTokens().GTE(math.NewIntFromUint64(minStake))
+}
+
 // ComputeNextValidators computes the validators for the upcoming epoch based on the currently `bondedValidators`.
-func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator) []types.ConsumerValidator {
+func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator) []types.ConsensusValidator {
+	// sort the bonded validators by number of staked tokens in descending order
+	sort.Slice(bondedValidators, func(i, j int) bool {
+		return bondedValidators[i].GetTokens().GT(bondedValidators[j].GetTokens())
```
insumity, line 332: I guess since those are bonded validators, `GetTokens()` is the same as `GetBondedTokens()`. Still, I would rather use `GetBondedTokens()` the same way you use it above in `FulfillsMinStake`.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,44 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+		maxProviderConsensusVals := k.GetMaxProviderConsensusValidators(ctx)
+		if len(bondedValidators) > int(maxProviderConsensusVals) {
+			bondedValidators = bondedValidators[:maxProviderConsensusVals]
+		}
+	}
+
 	nextValidators := k.FilterValidators(ctx, chainID, bondedValidators,
 		func(providerAddr types.ProviderConsAddress) bool {
-			return k.CanValidateChain(ctx, chainID, providerAddr)
+			return k.CanValidateChain(ctx, chainID, providerAddr) && k.FulfillsMinStake(ctx, chainID, providerAddr)
```
insumity, line 348: Does this mean you might not have enabled inactive validators but have set a minimum stake and the minimum stake stills gets applied? I thought that minimum stake was only applicable if inactive validators were allowed.

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -683,3 +683,122 @@ func findConsumerValidator(t *testing.T, v types.ConsumerValidator, valsAfter []
...
+	}
+	// set up mocks
+	for index, val := range validators {
+		mocks.MockStakingKeeper.EXPECT().GetValidatorByConsAddr(ctx, consAddrs[index].Address).Return(val, nil).AnyTimes()
+	}
+
+	return validators, consAddrs
+}
+
+// TestMinStake checks that FulfillsMinStake returns true if the validator has more than the min stake
```
insumity, line 712: "more than" implies `>` while I guess we have `>=` for `FullfillsMinStake`.

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -683,3 +683,122 @@ func findConsumerValidator(t *testing.T, v types.ConsumerValidator, valsAfter []
...
+	providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+
+	// create two validators with powers 1 and 2
+	_, consAddrs := createStakingValidatorsAndMocks(ctx, mocks, 1, 2)
+
+	testCases := []struct {
+		name           string
+		minStake       uint64
+		expectedFulfil []bool
```
insumity, line 724:
```suggestion
		expectedFulfill []bool
```

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -683,3 +683,122 @@ func findConsumerValidator(t *testing.T, v types.ConsumerValidator, valsAfter []
...
+				result := providerKeeper.FulfillsMinStake(ctx, "chainID", valAddr)
+				require.Equal(t, tc.expectedFulfil[i], result)
+			}
+		})
+	}
+}
+
+// TestMaxProviderConsensusValidators checks that the number of validators in the next validator set is at most
+// the maxProviderConsensusValidators parameter if the consumer chain does not allow inactive validators to validate.
+func TestIfInactiveValsDisallowedProperty(t *testing.T) {
```
insumity, line 756: Different name to the godoc comment ` TestMaxProviderConsensusValidators `

x/ccv/provider/keeper/partial_set_security_test.go
```diff
@@ -683,3 +683,122 @@ func findConsumerValidator(t *testing.T, v types.ConsumerValidator, valsAfter []
...
+			for i, valAddr := range consAddrs {
+				result := providerKeeper.FulfillsMinStake(ctx, "chainID", valAddr)
+				require.Equal(t, tc.expectedFulfil[i], result)
+			}
+		})
+	}
+}
+
+// TestMaxProviderConsensusValidators checks that the number of validators in the next validator set is at most
+// the maxProviderConsensusValidators parameter if the consumer chain does not allow inactive validators to validate.
```
insumity, line 755: nit: ```suggestion
// the MaxProviderConsensusValidators parameter if the consumer chain does not allow inactive validators to validate.
```

File: x/ccv/provider/keeper/partial_set_security_test.go [hunk 32]
```diff
@@ -683,3 +683,122 @@ func findConsumerValidator(t *testing.T, v types.ConsumerValidator, valsAfter []
...
+		providerKeeper.SetMinStake(ctx, "chainID", minStake)
+		params := providerKeeper.GetParams(ctx)
+		params.MaxProviderConsensusValidators = int64(maxProviderConsensusVals)
+		providerKeeper.SetParams(ctx, params)
+
+		// Compute the next validators
+		nextVals := providerKeeper.ComputeNextValidators(ctx, "chainID", vals)
+
+		// Check that the length of nextVals is at most maxProviderConsensusVals
+		require.LessOrEqual(r, len(nextVals), int(maxProviderConsensusVals), "The length of nextVals should be at most maxProviderConsensusVals")
```
insumity, line 785: I guess those might not be equal if we have less randomly chosen `valPowers` validators than the randomly chosen `maxProviderConsensusVals`?

x/ccv/provider/keeper/partial_set_security_test.go
insumity, line 756: Nice! 😃

x/ccv/provider/keeper/genesis.go
p-offtermatt: Right, it doesn't need to be part of the genesis state actually. Changed

x/ccv/provider/keeper/proposal.go
```diff
@@ -217,6 +219,8 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, closeChan boo
 	k.DeleteValidatorSetCap(ctx, chainID)
 	k.DeleteAllowlist(ctx, chainID)
 	k.DeleteDenylist(ctx, chainID)
+	k.DeleteAllowInactiveValidators(ctx, chainID)
+	k.DeleteMinStake(ctx, chainID)
```
insumity, line 223: nit: move `DeleteMinStake` before `DeleteAllowInactiveValidators` so it's in the same order as in the `HandleConsumerModificationProposal`.

x/ccv/provider/keeper/proposal.go
```diff
@@ -263,20 +267,26 @@ func (k Keeper) MakeConsumerGenesis(
 	}
 
 	if prop.Top_N > 0 {
+		// get the consensus active validators
```
insumity, line 270: Add more to the comment on why we need the active here, similar to your comment on the bonded validators below. I guess the reasoning here is that you do not want to "force" inactive validators to validate (?)

x/ccv/provider/keeper/proposal_test.go
```diff
@@ -917,6 +933,10 @@ func TestBeginBlockInit(t *testing.T) {
 
 	valAddr, _ := sdk.ValAddressFromBech32(validator.GetOperator())
 	mocks.MockStakingKeeper.EXPECT().GetLastValidatorPower(gomock.Any(), valAddr).Return(int64(1), nil).AnyTimes()
+
+	// for each validator, expect a call to GetValidatorByConsAddr with its consensus address
```
insumity, line 937: Why "for each"? This seems to only be doing it for a _single_ validator?
nit: remove new line after comment

x/ccv/provider/keeper/provider_consensus.go
```diff
@@ -0,0 +1,87 @@
...
+// as part of the last provider consensus validator set
+func (k Keeper) SetLastProviderConsensusValidator(
+	ctx sdk.Context,
+	validator types.ConsensusValidator,
+) {
+	k.setValidator(ctx, []byte{types.LastProviderConsensusValsPrefix()}, validator)
+}
+
+// SetLastProviderConsensusValSet resets the stored last validator set sent to the consensus engine on the provider
+// to the provided `nextValidators“.
```
insumity, line 24:
```suggestion
// to the provided `nextValidators`
```

proto/interchain_security/ccv/provider/v1/genesis.proto
mpoke, line 1: Make sure to run `make proto-gen`

x/ccv/provider/keeper/provider_consensus_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	validator2 := types.ConsensusValidator{
+		ProviderConsAddr: []byte("providerConsAddr2"),
+		Power:            3,
+		PublicKey:        &crypto.PublicKey{},
+	}
+
+	nextValidators := []types.ConsensusValidator{validator1, validator2}
+
+	providerKeeper.SetLastProviderConsensusValSet(ctx, nextValidators)
```
insumity, line 99: nit: Check here that the set is not empty.

x/ccv/provider/keeper/relay.go
```diff
@@ -67,6 +76,49 @@ func (k Keeper) EndBlockVSU(ctx sdk.Context) {
...
+	nextValidators := []types.ConsensusValidator{}
+	maxValidators := k.GetMaxProviderConsensusValidators(ctx)
+	// avoid out of range errors by bounding the max validators to the number of bonded validators
+	if maxValidators > int64(len(bondedValidators)) {
+		maxValidators = int64(len(bondedValidators))
+	}
+	for _, val := range bondedValidators[:maxValidators] {
+		nextValidator, err := k.CreateProviderConsensusValidator(ctx, val)
+		if err != nil {
+			k.Logger(ctx).Error("error when creating provider consensus validator", "error", err)
```
insumity, line 110: nit: log the problematic `val` as well

x/ccv/provider/keeper/relay_test.go
```diff
@@ -630,6 +637,74 @@ func TestEndBlockVSU(t *testing.T) {
...
+	providerKeeper, ctx, ctrl, mocks := testkeeper.GetProviderKeeperAndCtx(t, testkeeper.NewInMemKeeperParams(t))
+	defer ctrl.Finish()
+
+	// Mocking bonded validators in the staking keeper
+	validators := []stakingtypes.Validator{
+		createStakingValidator(ctx, mocks, 3, 30, 3),
+		createStakingValidator(ctx, mocks, 2, 20, 2),
+		createStakingValidator(ctx, mocks, 1, 10, 1),
+	}
+	// sort validators by power
```
insumity, line 652: Where does the sorting take place?

x/ccv/provider/keeper/relay.go
```diff
@@ -144,21 +196,35 @@ func (k Keeper) QueueVSCPackets(ctx sdk.Context) {
...
-				k.OptInTopNValidators(ctx, chainID, bondedValidators, minPower)
-			} else {
+			// of the active validators
+			activeValidators, err := k.GetLastActiveBondedValidators(ctx)
+			var minPower int64
+			if err != nil {
 				// we just log here and do not panic because panic-ing would halt the provider chain
-				k.Logger(ctx).Error("failed to compute min power to opt in for chain", "chain", chainID, "error", err)
+				k.Logger(ctx).Error("failed to get active validators", "error", err)
+				// assume that the minPower is 0, since we cannot compute it because the validator set is broken
```
mpoke, line 213: This error handling is not consistent. In some cases we panic (e.g.,  ProviderValidatorUpdates), while in other we just log an error and continue. We need to make this consistent. and add comments on the design decisions. 

In this case, why would `GetLastActiveBondedValidators` return an error? If it's because the state is corrupted, then we should panic. Otherwise, we need to ask if it's safe to continue. Assuming minPower is zero means that all validators need to run TopN chains, which will lead to a bunch of small validators getting jailed for downtime as they don't expect that.

x/ccv/provider/keeper/relay.go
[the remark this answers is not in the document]
mpoke: See https://github.com/cosmos/interchain-security/pull/2079/files#r1701624550

x/ccv/provider/keeper/staking_keeper_interface.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+// IterateBondedValidatorsByPower iterates over the consensus-active validators by power.
+// The same as IterateBondedValidatorsByPower in the StakingKeeper,
+// but only returns the first MaxProviderConsensusValidators validators.
+// This is used to implement the interface of the staking keeper to interface with
```
insumity, line 14: Not sure, but "to interact with" might sound slightly better than "to interface with"

x/ccv/provider/keeper/staking_keeper_interface.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+
+	return totalBondedTokens, nil
+}
+
+// IterateDelegations is the same as IterateDelegations in the StakingKeeper.
+// Necessary to implement the interface of the staking keeper to interface with
+// other modules.
+func (k Keeper) IterateDelegations(ctx context.Context, delegator sdk.AccAddress, fn func(index int64, delegation stakingtypes.DelegationI) (stop bool)) error {
+	return k.stakingKeeper.IterateDelegations(ctx, delegator, fn)
```
insumity, line 50: Could this iterate through delegations of non-consensus-active validators here and below?
Could this be problematic?

x/ccv/provider/keeper/staking_keeper_interface_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+			params := providerKeeper.GetParams(ctx)
+			params.MaxProviderConsensusValidators = tt.maxValidators
+			providerKeeper.SetParams(ctx, params)
+
+			// sort vals by descending number of tokens
+			sort.Slice(
+				vals,
+				func(i, j int) bool {
+					return vals[i].GetTokens().Int64() > vals[j].GetTokens().Int64()
```
insumity, line 96: Other comment on `GetTokens()` versus `GetBondedTokens()`.

x/ccv/provider/keeper/distribution.go
p-offtermatt: the rewards are indeed also *from* a consumer chain, right? as in, they come from the consumer chain

x/ccv/provider/keeper/relay.go
p-offtermatt: Changed this to panic. It's what the sdk does when the underlying method fails (IterateLastValidatorPowers)

x/ccv/provider/keeper/distribution.go
p-offtermatt: because it's clear from the context of being in ComputeConsumerTotalVotingPower, and the function is not so large that it's easy to forget

x/ccv/provider/keeper/grpc_query_test.go
p-offtermatt: Changed to 3 to make it clear this can be small

x/ccv/provider/keeper/keeper.go
p-offtermatt: Renamed both, and also delete

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: Changed, thx

x/ccv/provider/keeper/staking_keeper_interface_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		{
+			name:            "validators equal to max",
+			maxValidators:   3,
+			powers:          []int64{10, 20, 30},
+			expectedPowers:  []int64{30, 20, 10}, // get all validators back
+			mockTotalSupply: 60,
+			// 30 + 20 + 10 = 60
+			expectedTotalBonded: math.NewInt(60),
+			expectedBondedRatio: math.LegacyNewDec(100).Quo(math.LegacyNewDec(100)), // 100% bonded
```
insumity, line 66: Shouldn't this be `60 / 60`? Same result but `mockTotalSupply` is 60.

x/ccv/provider/keeper/staking_keeper_interface_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		{
+			name:            "validators with equal powers",
+			maxValidators:   3,
+			powers:          []int64{10, 10, 10, 20, 20, 30, 30},
+			expectedPowers:  []int64{30, 30, 20}, // get the top 3 validators back
+			mockTotalSupply: 1000,
+			// 30 + 30 + 20 = 80
+			expectedTotalBonded: math.NewInt(80),
+			expectedBondedRatio: math.LegacyNewDec(8).Quo(math.LegacyNewDec(100)), // 8% bonded
```
insumity, line 76: Similarly with above, `80 / 1000` instead?

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,114 @@
...
+// setValidator stores the given `validator` in the validator set stored under the given prefix.
+func (k Keeper) setValidator(
+	ctx sdk.Context,
+	prefix []byte,
+	validator types.ConsensusValidator,
+) error {
+	store := ctx.KVStore(k.storeKey)
+	bz, err := validator.Marshal()
+	if err != nil {
+		return fmt.Errorf("marshalling ConsumerValidator: %w", err)
```
insumity, line 28:
```suggestion
		return fmt.Errorf("marshalling ConsensusValidator: %w", err)
```

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, val := range nextValidators {
+		err := k.setValidator(ctx, prefix, val)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// deleteValidator removes validator with `providerAddr` address from the
```
insumity, line 47:
```suggestion
// deleteValidator removes validator with `providerConsAddr` address from the
```

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,114 @@
...
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		keysToDel = append(keysToDel, iterator.Key())
+	}
+	for _, delKey := range keysToDel {
+		store.Delete(delKey)
+	}
+}
+
+// isValidator returns `true` if the validator with `providerAddr` exists
```
insumity, line 76: Previously you used `providerConsAddr`, while here `providerAddr`. Maybe make those consistent across.

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,114 @@
...
+) (validators []types.ConsensusValidator, err error) {
+	store := ctx.KVStore(k.storeKey)
+	iterator := storetypes.KVStorePrefixIterator(store, prefix)
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		iterator.Value()
+		var validator types.ConsensusValidator
+		if err := validator.Unmarshal(iterator.Value()); err != nil {
+			return validators, fmt.Errorf("unmarshalling ConsumerValidator: %w", err)
```
insumity, line 96:
```suggestion
			return validators, fmt.Errorf("unmarshalling ConsensusValidator: %w", err)
```

x/ccv/provider/keeper/validator_set_storage.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"fmt"
+
+	"cosmossdk.io/math"
+	storetypes "cosmossdk.io/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/interchain-security/v5/x/ccv/provider/types"
+)
+
```
insumity, line 12: I guess all those methods here are indirectly tested through the methods that use them?
Maybe add a comment here on where those are used/tested.

tests/mbt/driver/mbt_test.go
```diff
@@ -520,8 +520,6 @@ func RunItfTrace(t *testing.T, path string) {
 		for _, consumerChainID := range actualRunningConsumerChainIDs {
```
insumity, line 520: We do not use MBT anymore, right? If so, why do we "touch" those files here?

tests/e2e/v4/state.go
```diff
@@ -660,3 +661,20 @@ func (tr Commands) GetConsumerCommissionRate(chain ChainID, validator ValidatorI
 func uintPtr(i uint) *uint {
 	return &i
 }
+
+func (tr Commands) GetInflationRate(
```
insumity, line 665: Why is this needed in this version?

tests/e2e/main.go
```diff
@@ -293,7 +335,9 @@ func getTestCases(selectedPredefinedTests, selectedTestFiles TestSet, providerVe
 			"partial-set-security-validator-set-cap", "partial-set-security-validators-power-cap",
 			"partial-set-security-validators-allowlisted", "partial-set-security-validators-denylisted",
 			"partial-set-security-modification-proposal",
-			"active-set-changes",
+			"active-set-changes", "inactive-vals-topN",
```
insumity, line 338: 1. What about the `bestcase`s shown above?
2. Can you reorder to follow the same order as the ones above (e.g., first `"inactive-provider-validators-on-consumer"` and then `"inactive-vals-topN"`)?

x/ccv/provider/keeper/relay_test.go
p-offtermatt: was removed, thanks for the catch. I just create them sorted now

docs/docs/features/power-shaping.md
```diff
@@ -28,21 +28,27 @@ This is a soft cap, and the actual power of a validator can exceed this fraction
...
+
 All these mechanisms are set by the consumer chain in the `ConsumerAdditionProposal`. They operate *solely on the provider chain*, meaning the consumer chain simply receives the validator set after these rules have been applied and does not have any knowledge about whether they are applied.
 
-Each of these mechanisms is *set during the consumer addition proposal* (see [Onboarding](../consumer-development/onboarding.md#3-submit-a-governance-proposal)), and is currently *immutable* after the chain has been added.
+Each of these mechanisms is *set during the consumer addition proposal* (see [Onboarding](../consumer-development/onboarding.md#3-submit-a-governance-proposal)).
 
 The values can be seen by querying the list of consumer chains:
 ```bash
 interchain-security-pd query provider list-consumer-chains
+
```
insumity, line 50: nit: Why the new lines here?

x/ccv/provider/keeper/staking_keeper_interface.go
p-offtermatt: This is on purpose, afaict the SDK also iterates over non-active delegations

tests/mbt/driver/mbt_test.go
p-offtermatt: To get rid of compilation errors. Maybe it's time to remove the files completely and not have the dead code anymore, but maybe this should be a separate PR

UPGRADING.md
```diff
@@ -1,5 +1,32 @@
...
+
+### Provider
+
+Upgrading a provider from v5.1.x requires state migrations. The following migrators should be added to the upgrade handler of the provider chain:
+
+```go
+// InitializeMaxValidatorsForExistingConsumers initializes the max validators
+// parameter for existing consumers to the MaxProviderConsensusValidators parameter.
+// This is necessary to avoid those consumer chains having an excessive amount of validators.
+func InitializeMaxValidatorsForExistingConsumers(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
```
insumity, line 13: Shouldn't those functions appear in our migration code somewhere 🤔 ?

x/ccv/provider/keeper/validator_set_storage.go
p-offtermatt: Made it consistently providerConsAddr in this file

x/ccv/provider/types/genesis_test.go
p-offtermatt: it was removed because the parameter was unused, anyways.
the set is computed on-the-fly

UPGRADING.md
p-offtermatt: Negative, see here for a discussion I had with Marius about this:
https://github.com/cosmos/interchain-security/pull/2079#discussion_r1694853611

x/ccv/provider/keeper/partial_set_security.go
bermuell: is there a need to ensure ordering of vals with same amount of tokens?

tests/integration/partial_set_security_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	appConsumer "github.com/cosmos/interchain-security/v5/app/consumer"
+	appProvider "github.com/cosmos/interchain-security/v5/app/provider"
+)
+
+// we need a stake multiplier because tokens do not directly correspond to voting power
+// this is needed because 1000000 tokens = 1 voting power, so lower multipliers
+// will be verbose and harder to read because small token numbers
+// won't correspond to at least one voting power
+const stake_multiplier = 1000000
```
insumity, line 21: Should we use camel case for `const`s? So, `stakeMultplier`?

tests/e2e/main.go
p-offtermatt: 1. I think the basecases should not be run by default. They are sanity checks that tbh were mostly for me during testing, and for others to be able to confirm how it would work without the inactive vals feature
2. Sure

tests/integration/partial_set_security_test.go
p-offtermatt: Sure

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -62,15 +62,105 @@ The following changes to the state are required:
 * Store the provider consensus validator set in the provider module state under the `LastProviderConsensusValsPrefix` key. This is the last set of validators that the provider sent to the consensus engine. This is needed to compute the ValUpdates to send to the consensus engine (by diffing the current set with this last sent set).
 * Increase the `MaxValidators` parameter of the staking module to the desired size of the potential validator
 set of consumer chains.
+* Introduce extra per-consumer-chain parameters:
```
insumity, line 65: Line 61 above where it writes:
> Introduce the `MaxProviderConsensusValidators` parameter to the provider module, which is the number of validators that the provider module will send to consumer chains.

Shouldn't it be "number of validators that the provider module sends to the consensus engine of the provider."?

x/ccv/provider/keeper/validator_set_update.go
```diff
@@ -229,5 +196,14 @@ func (k Keeper) FilterValidators(
 // GetLastBondedValidators iterates the last validator powers in the staking module
 // and returns the first MaxValidators many validators with the largest powers.
 func (k Keeper) GetLastBondedValidators(ctx sdk.Context) ([]stakingtypes.Validator, error) {
```
insumity, line 198: In the ADR you write:
> For the first iteration, we propose to increase the active set size to 200 validators (while keeping the consensus validators to 180), thus letting the 20 validators with the most stake outside of the active set validate on consumer chains.

I was wondering. How are we going to set the max validators? Is the idea to go through a gov proposal to increase it on the `staking` module or are we going to do it directly as part of the migration?

x/ccv/provider/keeper/validator_set_update.go
```diff
@@ -229,5 +196,14 @@ func (k Keeper) FilterValidators(
...
-	return ccv.GetLastBondedValidatorsUtil(ctx, k.stakingKeeper, k.Logger(ctx))
+	maxVals, err := k.stakingKeeper.MaxValidators(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return ccv.GetLastBondedValidatorsUtil(ctx, k.stakingKeeper, k.Logger(ctx), maxVals)
+}
+
+func (k Keeper) GetLastActiveBondedValidators(ctx sdk.Context) ([]stakingtypes.Validator, error) {
```
insumity, line 206: Add a godoc comment here. I wonder if other names here would be easier to parse. Why not call this `GetLastProviderConsensusBondedValidators` or something similar to make the distinction clear between this method and `GetLastBondedValidators`?

x/ccv/provider/keeper/relay.go
```diff
@@ -55,7 +58,13 @@ func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet) err
 
 // EndBlockVSU contains the EndBlock logic needed for
 // the Validator Set Update sub-protocol
-func (k Keeper) EndBlockVSU(ctx sdk.Context) {
+func (k Keeper) EndBlockVSU(ctx sdk.Context) ([]abci.ValidatorUpdate, error) {
+	// logic to update the provider consensus validator set.
+	// Important: must be called before the rest of EndBlockVSU, because
+	// we need to know the updated provider validator set
```
insumity, line 64: Not sure I fully understand this comment. In `QueueVSCPacket`s we would need to compute the minimum power in the Top N. However, there we use directly `GetLastActiveBondedValidators` that talks directly to the `staking` keeper and we do not seem to be using the stored provider consensus validator set for this. Why would we need to update the provider consensus validator set to make sure that the minimum power is computed correctly?

app/provider/app.go
```diff
@@ -628,6 +635,8 @@ func New(
 		providertypes.ModuleName,
 	)
 
+	// NOTE: provider module needs to come after the staking module, since
+	// it needs the information the staking module provides to compute validator updates.
```
insumity, line 639: Do we need to have the `no_valupdates_staking` module's `EndBlocker` here as well?

app/provider/app.go
```diff
@@ -654,6 +663,8 @@ func New(
 	// NOTE: Capability module must occur first so that it can initialize any capabilities
 	// so that other modules that want to create or claim capabilities afterwards in InitChain
 	// can do so safely.
+	// NOTE: The provider module must come after genutils and staking, since it relies on the
+	// information about the validators these modules provide to compute validator updates.
 	app.MM.SetOrderInitGenesis(
```
insumity, line 668: Do we need to have the `no_valupdates_*` modules here as well?

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,44 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+	}
+
+	// validator has enough stake to validate the chain
+	return validator.GetBondedTokens().GTE(math.NewIntFromUint64(minStake))
+}
+
 // ComputeNextValidators computes the validators for the upcoming epoch based on the currently `bondedValidators`.
-func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator) []types.ConsumerValidator {
+func (k Keeper) ComputeNextValidators(ctx sdk.Context, chainID string, bondedValidators []stakingtypes.Validator) []types.ConsensusValidator {
+	// sort the bonded validators by number of staked tokens in descending order
```
insumity, line 330: Aren't those already sorted?  The way I understand this, is that when we call `ComputeNextValidators`, we pass `bondedValidators` that stem from `GetLastBondedValidatorsUtil` that should have sorted those already.

x/ccv/provider/keeper/partial_set_security.go
insumity: I agree. But it might raise some questions on what happens when we have `min_stake` set in Top N chains. E.g., validator `V` has to validate because it belongs to the first top 50% of the validators in a Top 50% chain, but the `min_stake` is set in such a way that `V` should not validate.
We already have similar questions in regards to other power-shaping parameters as well, so, I do not think this is a real issue. We can revisit later if needed.

app/provider/app.go
```diff
@@ -352,9 +353,7 @@ func New(
 	// Remove the ConsumerRewardsPool from the group of blocked recipient addresses in bank
 	// this is required for the provider chain to be able to receive tokens from
 	// the consumer chain
-	bankBlockedAddrs := app.ModuleAccountAddrs()
-	delete(bankBlockedAddrs, authtypes.NewModuleAddress(
-		providertypes.ConsumerRewardsPool).String())
+	bankBlockedAddrs := BankBlockedAddrs(app)
```
insumity, line 356: The comment above "// Remove ..." should probably be moved to the function `BankBlockedAddrs`.
Also, I would rename the `BankBlockedAddrs` since it does something more than retrieving the blocked addresses, it also deletes one, e.g., something akin to `BankBlockedAddrsWithoutTheConsumerRewardsPool(app)`?

app/provider/app.go
```diff
@@ -374,15 +373,6 @@ func New(
...
 	)
-	app.MintKeeper = mintkeeper.NewKeeper(
-		appCodec,
-		runtime.NewKVStoreService(keys[minttypes.StoreKey]),
-		app.StakingKeeper,
-		app.AccountKeeper,
-		app.BankKeeper,
-		authtypes.FeeCollectorName,
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
-	)
```
insumity, line 385: `DistrKeeper` should use the staking keeper or the provider one?

app/provider/app.go
p-offtermatt: No, this just sets the order of EndBlockers by stating the names of the modules; this is ok to say staking.
Thanks for double-checking

app/provider/app.go
p-offtermatt: Same as above: no

app/provider/app.go
p-offtermatt: changed to ComputeBankBlockedAddrs
I don't like the extremely long name

docs/docs/adrs/adr-017-allowing-inactive-validators.md
p-offtermatt: Yes, thanks for the catch!

x/ccv/provider/keeper/partial_set_security.go
p-offtermatt: Agreed. I have it for defensive-programming purposes here, since it will be inexpensive if the input is already sorted and means this function has less (implicit) assumptions

p-offtermatt: lmk if you see a reason to change it

x/ccv/provider/keeper/relay.go
p-offtermatt: You are right, I changed that and didn't adjust the comment. Have removed the comment

x/ccv/provider/keeper/validator_set_update.go
p-offtermatt: Good point, they were indeed a bit too close. aldo added a comment

x/ccv/provider/keeper/validator_set_update.go
p-offtermatt: Gov prop imo, but good question. @mpoke wdyt?

app/provider/app.go
p-offtermatt: See https://informalsystems.notion.site/Staking-dependencies-in-Cosmos-SDK-v0-50-5-and-ibc-go-v8-3-2-669889523e1840f68996f69946bf388e?pvs=4
They are all ok using staking

x/ccv/provider/keeper/validator_set_update.go
mpoke: From the [signaling prop for this feature](https://www.mintscan.io/cosmos/proposals/930):
> We plan to mitigate this risk by going in small steps and, for the first iteration, only allowing the first 20 validators outside the Cosmos Hub’s active validator set to validate on consumer chains. We expect that this will still be a relatively competitive set.

So I think we don't need another gov prop. We will have a software upgrade proposal for Gaia v20 detailing all the changes.

mpoke: > coordination with block explorers (e.g., mintscan, etc.) is needed so they do not start presenting validator sets in an erroneous way.

This is true. I'll add it as a TODO.

p-offtermatt: Good point, thanks for clarifying

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -308,11 +308,44 @@ func (k Keeper) CanValidateChain(ctx sdk.Context, chainID string, providerAddr t
...
+// by checking its staked tokens against the minimum stake required to validate the chain.
+func (k Keeper) FulfillsMinStake(ctx sdk.Context, chainID string, providerAddr types.ProviderConsAddress) bool {
+	minStake, found := k.GetMinStake(ctx, chainID)
+	if !found {
+		return true
+	}
+
+	validator, err := k.stakingKeeper.GetValidatorByConsAddr(ctx, providerAddr.Address)
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
p-offtermatt: Thx, done

x/ccv/provider/client/legacy_proposal_handler.go
```diff
@@ -61,7 +61,9 @@ Where proposal.json contains:
     "validators_power_cap": 32,
     "validator_set_cap": 50,
     "allowlist": [],
-    "denylist": ["validatorAConsensusAddress", "validatorBConsensusAddress"]
+    "denylist": ["validatorAConsensusAddress", "validatorBConsensusAddress"],
+	"min_stake": 100000000000,
+	"allow_inactive_vals": false
```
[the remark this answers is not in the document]
p-offtermatt: Thanks, yes, I guess there should be (top N chains should not allow inactive validators)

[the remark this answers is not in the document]
p-offtermatt: Removed my previous comment - there isn't, and there doesn't seem to be a strong reason to prevent it

tests/e2e/steps_inactive_vals.go
```diff
@@ -0,0 +1,975 @@
...
+
+// stepsInactiveValidatorsOnConsumer tests situations where validators that are *not* in the active set on the
+// provider chain validate on the consumer chain.
+// The provider chain is set to have at most *2* validators active in consensus, and there are 3 validators in total.
+// high-level, this test does:
+// - start the provider chain
+// - start a consumer chain
+// - check that non-consensus validators do not get slashed for downtime on the provider; and that they don't get rewards
+// - check that active validators *do* get slashed for downtime on the provider, and don't get rewards while they are down
+// - check that non-consensus validators *do* get jailed for downtime on consumer chains
```
insumity, line 18:
```suggestion
// - check that non-consensus validators *do* get jailed for consumer downtime on the provider
```

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -58,19 +58,109 @@ achieved without further changes to staking or reward distributions, because sim
...
+Checked as part of the e2e test `inactive-provider-validators-on-consumer`.
+
+### Scenario 4: Inactive validators should not get slashed/jailed for downtime on the provider chain
+
+This can be tested by having an inactive validator go offline on the provider chain for long enough to accrue downtime.
+The validator should be neither slashed nor jailed for downtime.
+
+Checked as part of the e2e test `inactive-provider-validators-on-consumer`.
+
+### Scenario 5: Inactive validators *should* get jailed for downtime on the provider chain
```
insumity, line 118: nit: 
```suggestion
### Scenario 5: Inactive validators *should* get jailed for consumer downtime on the provider chain
```

app/provider/app.go
insumity: Oh, yes. Thanks. Is there a problem with using the `provider` keeper and hence the consensus-active validator set even though they're fine using the `staking` keeper? Is there a potentiality where distr, slashing, evidence, ... might need to use non-consensus-active validators?

insumity (approved):
Amazing work Philip! 🎉 

I reviewed everything last week and went again over today to look more closely at the general idea, wiring, E2E tests. Specifically, the reviewing chunks (i.e., "ADR for context", "core logic changes", and "tests" (I only skimmed through the simulator))  you mention in your PR.

mpoke (approved):
Great work @p-offtermatt

mpoke:
What's the purpose for https://github.com/cosmos/interchain-security/pull/2079/commits/c3b36e68c022460789771ad55d6dc901fbca2c5d?

docs/docs/adrs/adr-017-allowing-inactive-validators.md
```diff
@@ -58,19 +58,109 @@ achieved without further changes to staking or reward distributions, because sim
...
+
+Inactive validators should not get rewards from the provider chain.
+
+This can be tested by starting a provider chain with inactive validators and checking the rewards of inactive validators.
+
+Checked as part of the e2e test `inactive-provider-validators-on-consumer`.
+
+### Scenario 3: Inactive validators should get rewards from consumer chains
+
+An inactive validator that is validating on a consumer chain should receive rewards in the consumer chain token.
```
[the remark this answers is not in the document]
p-offtermatt: I will merge for now to get this out of the way, and we can handle this in a separate PR imo

app/provider/app.go
p-offtermatt: It just means our provider keeper would need to implement the interface that these keepers expect from the staking keeper, too, so it just increases the surface of the provider keeper for no benefit

PR #2117, opened 2024-07-31
feat: first iteration on Permissionless ICS

insumity:
## Description

This is a first iteration on the Permissionless ICS. It is by far not ready, e.g., the integration and E2E tests are all failing that we expect would lead to more changes.
However, the main ideas on how Permissinless ICS would look like are already in this PR.

The plan is to have this PR as a first step that we can merge into `feat/permissionless` and afterwards parallelize the fixing of integration and E2E tests together with @bermuell .

**It makes sense to review this PR commit by commit.**

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Introduced new message types for managing consumer operations, including registration and removal.
	- Added functionalities for handling consumer addition and removal proposals.
- **Documentation**
	- Updated documentation for Permissionless Inter-Chain Standard (ICS) to clarify terminology and processes.
- **Bug Fixes**
	- Enhanced validation logic to ensure proper handling of consumer IDs.
- **Refactor**
	- Renamed parameters and functions throughout the codebase to shift focus from chains to consumers for improved clarity. 
	- Removed deprecated functionalities related to consumer modification proposals.


proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -120,8 +120,8 @@ message ConsumerRemovalProposal {
   string title = 1;
   // the description of the proposal
   string description = 2;
-  // the chain-id of the consumer chain to be stopped
-  string chain_id = 3;
+  // the consumer id of the consumer chain to be stopped
+  string consumer_id = 3;
```
mpoke, line 124: Do we still need the `ConsumerRemovalProposal` and `ConsumerModificationProposal`? I thought the plan is to deprecate legacy gov proposals from ICS.

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -189,17 +191,13 @@ message MsgConsumerAddition {
...
 
 
-// MsgConsumerRemoval message contains a governance proposal on the provider chain to
-// remove (and stop) a consumer chain. If it passes, all the consumer chain's
-// state is removed from the provider chain. The outstanding unbonding operation
-// funds are released.
-//
-// Note: this replaces ConsumerRemovalProposal which is deprecated and will be removed soon
-message MsgConsumerRemoval {
+// MsgConsumerRemoval defines the message used to remove (and stop) a consumer chain.
```
mpoke, line 194: nit: make sure to update also the docstrings

x/ccv/provider/types/keys.go
mpoke, line 1: rebase on main to include the removal of VSCMaturedPackets

x/ccv/provider/types/keys.go
```diff
@@ -49,11 +49,11 @@ const (
 
 	SlashMeterReplenishTimeCandidateKeyName = "SlashMeterReplenishTimeCandidateKey"
 
-	ChainToChannelKeyName = "ChainToChannelKey"
+	ChainToChannelKeyName = "ConsumerIdToChannelIdKey"
```
mpoke, line 52: Make the name consistent with the string, i.e., `ConsumerIdToChannelIdKeyName`

x/ccv/provider/types/keys.go
```diff
@@ -81,7 +81,7 @@ const (
 
 	ThrottledPacketDataKeyName = "ThrottledPacketDataKey"
 
-	GlobalSlashEntryKeyName = "GlobalSlashEntryKey"
+	GlobalSlashEntryKeyName = "DeprecatedGlobalSlashEntryKey"
```
mpoke, line 84: ditto

x/ccv/provider/types/keys.go
```diff
@@ -281,6 +296,27 @@ func getKeyPrefixes() map[string]byte {
 		// minimum power required to be in the top N per consumer chain.
 		MinimumPowerInTopNKeyName: 40,
 
+		// ConsumerIdKeyName is the key for storing the consumer id for the next registered consumer chain
+		ConsumerIdKeyName: 41,
```
mpoke, line 300: Once you get changes from main, you'll have a collision here.

x/ccv/provider/types/keys.go
```diff
@@ -281,6 +296,27 @@ func getKeyPrefixes() map[string]byte {
...
+
+		// ConsumerIdToChainIdKeyName is the key for storing the chain id for the given consumer id
+		ConsumerIdToChainIdKeyName: 42,
```
mpoke, line 303: Why do we need this. Isn't the RegistrationRecord containing the chainID?

x/ccv/provider/types/keys.go
```diff
@@ -281,6 +296,27 @@ func getKeyPrefixes() map[string]byte {
...
+		ConsumerIdToRegistrationRecordKeyName: 43,
+
+		// ConsumerIdToInitializationRecordKeyName is the key for storing the initialization record for the given consumer id
+		ConsumerIdToInitializationRecordKeyName: 44,
+
+		// ConsumerIdToOwnerAddressKeyName is the key for storing the owner address for the given consumer id
+		ConsumerIdToOwnerAddressKeyName: 45,
+
+		// ConsumerIdToPhaseKeyName is the key for storing the phase of a consumer chain with the given consumer id
+		ConsumerIdToPhaseKeyName: 46,
```
mpoke, line 315: Do we actually need this?

proto/interchain_security/ccv/provider/v1/provider.proto
insumity: No we do not. Those will be eventually removed. Probably renaming their fields was not worth it, I agree.

x/ccv/provider/types/keys.go
insumity: Good catch! Removed.

x/ccv/provider/types/keys.go
insumity: I believe we do because when a chain stops, we should keep the registration and initialization record. If we do not keep it, I guess we can get rid of this state.
The reason, we need the phase state if we keep the records of a stopped chain is to distinguish between a chain that just registered & initialized, with a chain that registered & initialized & launched & stopped.

It also seems to me that it simplifies the code because now we can see exactly where a chain moves from a phase `a` to another phase `b`.

app/provider/app.go
```diff
@@ -581,7 +580,6 @@ func New(
 					paramsclient.ProposalHandler,
 					ibcproviderclient.ConsumerAdditionProposalHandler,
 					ibcproviderclient.ConsumerRemovalProposalHandler,
-					ibcproviderclient.ConsumerModificationProposalHandler,
```
insumity, line 584: Those will be eventually removed as part of this [work](https://github.com/cosmos/interchain-security/pull/2130).

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,414 @@
...
+
+// DeleteClientIdToConsumerId deletes the client id to consumer id association
+func (k Keeper) DeleteClientIdToConsumerId(ctx sdk.Context, clientId string) {
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.ClientIdToConsumerIdKey(clientId))
+}
+
+// GetInitializedConsumersReadyToLaunch returns the consumer ids of the pending initialized consumer chains
+// that are ready to launch,  i.e., consumer clients to be created.
+func (k Keeper) GetInitializedConsumersReadyToLaunch(ctx sdk.Context) []string {
```
insumity, line 268: @mpoke This replaces the loop we had over the pending consumer additional proposals to see which ones we can execute. It does not contain `spawnTime` in the key and hence the loop is not `spawnTime`-sorted and the loop does not stop early. This loop executes in `BeginBlockInit` and hence might slow things down.
Let me know if you think it's worth it to include `spawnTime` in the key under some new state.

Similarly in  `GetLaunchedConsumersReadyToStop` below we do not loop by `stopTime`.

p-offtermatt:
Is there a reason for the consumer id to be a string, and not an integer? Since ultimately, it's just a uint64, why isn't it used as one everywhere, but instead treated as a string in many places?

x/ccv/provider/keeper/proposal.go
```diff
@@ -360,76 +320,21 @@ func (k Keeper) GetPendingConsumerAdditionProp(ctx sdk.Context, spawnTime time.T
...
 
-// BeginBlockInit iterates over the pending consumer addition proposals in order, and creates
-// clients for props in which the spawn time has passed. Executed proposals are deleted.
-//
-// See: https://github.com/cosmos/ibc/blob/main/spec/app/ics-028-cross-chain-validation/methods.md#ccv-pcf-bblock-init1
-// Spec tag:[CCV-PCF-BBLOCK-INIT.1]
+// BeginBlockInit iterates over the initialized consumers chains and creates clients for chains
+// in which the spawn time has passed
 func (k Keeper) BeginBlockInit(ctx sdk.Context) {
-	propsToExecute := k.GetConsumerAdditionPropsToExecute(ctx)
```
p-offtermatt, line 369: Just a remincer, GetConsumerAdditionPropsToExecute seems unused now and should be removed

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/interchain-security/v5/x/ccv/provider/types"
+	"time"
+)
+
+// ConsumerPhase captures the phases of a consumer chain according to `docs/docs/adrs/adr-018-permissionless-ics.md`
+type ConsumerPhase byte
+
+const (
+	Registered ConsumerPhase = iota
```
p-offtermatt, line 17: Would recommend some description in the code itself here, too, so that it shows up when you hover over the individual consts

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -13,16 +13,31 @@ import (
...
-	k.SetOptedIn(ctx, chainID, providerAddr)
+	registrationRecord, found := k.GetConsumerIdToRegistrationRecord(ctx, consumerId)
+	if !found {
+		return errorsmod.Wrapf(
+			types.ErrUnknownConsumerId,
+			"opting in to an unknown consumer chain, with id: %s", consumerId)
+	}
+
+	optedInToConsumerId, found := k.IsValidatorOptedInToChain(ctx, providerAddr, registrationRecord.ChainId)
+	if found && consumerId != optedInToConsumerId {
```
mpoke, line 34: This means that we allow validators to opt in twice on the same consumerId?

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+		}
+	}
+
+	return consumerIds
+}
+
+// IsValidatorOptedInToChain checks if the validator with `providerAddr` is opted into the chain with the specified `chainId`.
+// It returns `found == true` and the corresponding chain's `consumerId` if the validator is opted in. Otherwise, it returns an empty string
+// for `consumerId` and `found == false`.
+func (k Keeper) IsValidatorOptedInToChain(ctx sdk.Context, providerAddr types.ProviderConsAddress, chainId string) (consumerId string, found bool) {
```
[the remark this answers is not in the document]
mpoke:
```suggestion
func (k Keeper) IsValidatorOptedInToChainId(ctx sdk.Context, providerAddr types.ProviderConsAddress, chainId string) (consumerId string, found bool) {
```

x/ccv/provider/keeper/proposal.go
```diff
@@ -193,49 +144,59 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, chainID string, closeChan boo
...
+	consumerId string,
 ) (gen ccv.ConsumerGenesisState, nextValidatorsHash []byte, err error) {
-	chainID := prop.ChainId
+	initializationRecord, found := k.GetConsumerIdToInitializationRecord(ctx, consumerId)
+	if !found {
+		return gen, nil, errorsmod.Wrapf(types.ErrNoInitializationRecord,
+			"initialization record for consumer chain: %s is missing", consumerId)
+
+	}
+	updateRecord := k.GetConsumerIdToUpdateRecordOrDefault(ctx, consumerId, DefaultUpdateRecord)
```
mpoke, line 198: If there is no updateRecord, then just set TopN to default as it's the only field used here.

x/ccv/provider/keeper/proposal.go
```diff
@@ -531,77 +436,76 @@ func (k Keeper) DeletePendingConsumerRemovalProps(ctx sdk.Context, proposals ...
...
 func (k Keeper) BeginBlockCCR(ctx sdk.Context) {
-	propsToExecute := k.GetConsumerRemovalPropsToExecute(ctx)
-
-	for _, prop := range propsToExecute {
+	for _, consumerId := range k.GetLaunchedConsumersReadyToStop(ctx) {
 		// stop consumer chain in a cached context to handle errors
-		cachedCtx, writeFn, err := k.StopConsumerChainInCachedCtx(ctx, prop)
+		cachedCtx, writeFn := ctx.CacheContext()
+
+		stopTime, found := k.GetConsumerIdToStopTime(ctx, consumerId)
```
mpoke, line 449: GetLaunchedConsumersReadyToStop should already check the stopTime.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
 }
 
-func (k msgServer) ConsumerModification(goCtx context.Context, msg *types.MsgConsumerModification) (*types.MsgConsumerModificationResponse, error) {
-	if k.GetAuthority() != msg.Authority {
-		return nil, errorsmod.Wrapf(types.ErrUnauthorized, "expected %s, got %s", k.GetAuthority(), msg.Authority)
+// RegisterConsumer registers a consumer chain
+func (k msgServer) RegisterConsumer(goCtx context.Context, msg *types.MsgRegisterConsumer) (*types.MsgRegisterConsumerResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	consumerId := strconv.FormatUint(k.Keeper.FetchAndIncrementConsumerId(ctx), 10)
```
mpoke, line 300: FetchAndIncrementConsumerId should return a string (or whatever is the type of consumerId).

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
 	return &types.MsgSetConsumerCommissionRateResponse{}, nil
 }
 
-func (k msgServer) ConsumerModification(goCtx context.Context, msg *types.MsgConsumerModification) (*types.MsgConsumerModificationResponse, error) {
-	if k.GetAuthority() != msg.Authority {
-		return nil, errorsmod.Wrapf(types.ErrUnauthorized, "expected %s, got %s", k.GetAuthority(), msg.Authority)
+// RegisterConsumer registers a consumer chain
+func (k msgServer) RegisterConsumer(goCtx context.Context, msg *types.MsgRegisterConsumer) (*types.MsgRegisterConsumerResponse, error) {
```
mpoke, line 297: Do we need to set defaults for the rest of the consumer state? Or are the getter good enough to return defaults when no state was set?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	consumerId := msg.ConsumerId
+
+	phase, found := k.Keeper.GetConsumerIdToPhase(ctx, consumerId)
+	if !found || (phase != Registered && phase != Initialized) {
+		return nil, errorsmod.Wrapf(types.ErrInvalidPhase,
+			"chain with consumer id: %s has to be in its registered or initialized phase", consumerId)
 	}
 
+	ownerAddress, _ := k.Keeper.GetConsumerIdToOwnerAddress(ctx, consumerId)
+	if k.GetAuthority() == msg.Authority {
```
mpoke, line 322: This code is hard to read. I guess the assumption is that ownerAddress is `msg.Authority`, right?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+
+	k.Keeper.SetConsumerIdToRegistrationRecord(ctx, consumerId, *msg.RegistrationRecord)
+	k.Keeper.SetConsumerIdToOwnerAddress(ctx, consumerId, msg.Signer)
+	k.Keeper.SetConsumerIdToPhase(ctx, consumerId, Registered)
+
+	return &types.MsgRegisterConsumerResponse{ConsumerId: consumerId}, nil
+}
+
+// InitializeConsumer initializes a consumer chain
+func (k msgServer) InitializeConsumer(goCtx context.Context, msg *types.MsgInitializeConsumer) (*types.MsgInitializeConsumerResponse, error) {
```
mpoke, line 310: Is it possible to use this message to update only parts of the InitializationRecord? For example, send a message with all the fields in InitializationRecord unset but for the spawnTime. The result is just updating the spawnTime and leaving the other fields unchanged. 
Same for `UpdateConsumer`

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+
+// UpdateConsumer updates the record of a consumer chain
+func (k msgServer) UpdateConsumer(goCtx context.Context, msg *types.MsgUpdateConsumer) (*types.MsgUpdateConsumerResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	err := k.Keeper.HandleConsumerModificationProposal(ctx, msg)
-	if err != nil {
-		return nil, errorsmod.Wrapf(err, "failed handling ConsumerModification proposal")
+	ownerAddress, _ := k.Keeper.GetConsumerIdToOwnerAddress(ctx, msg.ConsumerId)
+
+	if k.GetAuthority() != msg.Authority && msg.Authority != ownerAddress {
```
mpoke, line 341: IIUC, this will fail if the owner is not the gov module, right?

A consumer record should be updated only if `msg.Authority == ownerAddress` (the owner signed the message). This update includes updating the owner, but only if the old owner is governance (i.e., transition from TopN to OptIn).

mpoke (commented):
Great work @insumity. See my comments below.

insumity:
> Is there a reason for the consumer id to be a string, and not an integer? Since ultimately, it's just a uint64, why isn't it used as one everywhere, but instead treated as a string in many places?
@p-offtermatt 
Simplifies things because the `chainId` that was used everywhere previously was of type `string`. Additionally, by having it as a `string` we can easily add readable info to it later on (e.g., the same way `clientId` and `chainId` are strings today).

x/ccv/provider/types/msg.go
```diff
@@ -227,100 +236,186 @@ func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
...
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgRemoveConsumer) ValidateBasic() error {
+	if err := ValidateConsumerId(msg.ConsumerId); err != nil {
+		return err
 	}
 
 	if msg.StopTime.IsZero() {
-		return errorsmod.Wrap(ErrInvalidConsumerRemovalProp, "spawn time cannot be zero")
+		return errorsmod.Wrap(ErrInvalidConsumerRemoval, "spawn time cannot be zero")
```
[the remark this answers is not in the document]
insumity: This check pre-existed Permissionless. But as you hint, there's no real benefit to it. Removed the check.
Now, when `stopTime` is 0, the chain would immediately be removed in the next block.

x/ccv/provider/keeper/keeper_test.go
```diff
@@ -251,7 +251,7 @@ func TestGetAllChannelToChains(t *testing.T) {
 	expectedGetAllOrder := []types.ChannelToChain{}
 	for i, chainID := range chainIDs {
 		channelID := fmt.Sprintf("client-%d", len(chainIDs)-i)
-		pk.SetChannelToChain(ctx, channelID, chainID)
+		pk.SetChannelToConsumerId(ctx, channelID, chainID)
```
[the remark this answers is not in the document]
insumity: Yes. Not all of them have been changed. Since you mention this test, I fixed it.

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -79,11 +82,13 @@ message MsgSubmitConsumerDoubleVoting {
   option (gogoproto.goproto_getters) = false;
 
   string submitter = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // the consumer id of the consumer chain where the double-voting took place
+  string consumer_id = 2;
```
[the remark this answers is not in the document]
insumity: Yes, I talked with the Hermes team. They estimate it's 2-3 days of work + testing.
I'll coordinate with them on this.

x/ccv/provider/types/keys.go
```diff
@@ -320,6 +337,30 @@ func getKeyPrefixes() map[string]byte {
...
+		ConsumerIdToUpdateRecordKeyName: 48,
+
+		// ConsumerIdToOwnerAddressKeyName is the key for storing the owner address for the given consumer id
+		ConsumerIdToOwnerAddressKeyName: 49,
+
+		// ConsumerIdToPhaseKeyName is the key for storing the phase of a consumer chain with the given consumer id
+		ConsumerIdToPhaseKeyName: 50,
+
+		// ConsumerIdToStopTimeKeyName is the key for storing the stop time of a consumer chain that is to be removed
+		ConsumerIdToStopTimeKeyName: 51,
```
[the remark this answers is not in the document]
insumity: I think it helps to know what the key is and what the value just by looking at the key.
For example, when we wanted to find at the keys that are associated with a `chainId`, it would have been substantially easier if keys were named `ChainIdToClientId`, etc.
In any case, I don't have strong feelings on this. We can remove `...IdTo...` and just have `ConsumerRegistrationRecord`, `ConsumerPhase`, etc.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+
+		var stopTime time.Time
+		err := stopTime.UnmarshalBinary(iterator.Value())
+		if err != nil {
+			panic(fmt.Errorf("failed to unmarshal stop stopTime: %w for consumer id: %s", err, consumerId))
+		}
+
+		// if current block time is equal to or after stop stopTime, and the chain is launched we can stop the chain
+		phase, found := k.GetConsumerIdToPhase(ctx, consumerId)
+		if !ctx.BlockTime().Before(stopTime) && (found && phase == Launched) {
```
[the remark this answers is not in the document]
insumity: You could potentially stop an initialized chain to prevent it from launching. But, not sure, I see a benefit in this: i) a consumer chain can just decide to not consume ICS security if they do not want to launch, and ii) they can leave the chain to fail to launch (e.g., if no validator opts in), iii) if they really want to prevent a launch (i.e., consumer client being created) they can set the `spawnTime` to be 100 years from now.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
insumity: > We should have an index from `chainId` to a list of `consumerIds`. Ideally, the list will contain only one element, but given the permissionless nature this will not be necessarily true.

An index from `chainId` to `consumerId`s might become extremely large (e.g., with 1M registrations for a specific chain). So I created an index from a validator `providerAddr` to its opted-in `consumerId` chains instead. Such an index can still become arbitrarily large but this implies the validator has opted-in to a multitude of chains.
If you're okay with this, feel free to resolve this conversation.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+	store := ctx.KVStore(k.storeKey)
+	bz, err := record.Marshal()
+	if err != nil {
+		panic(fmt.Errorf("failed to marshal record (%+v): %w", record, err))
+	}
+	store.Set(types.ConsumerIdToUpdateRecordKey(consumerId), bz)
+}
+
+// DeleteConsumerIdToUpdateRecord deletes the update record associated with this consumer id
+func (k Keeper) DeleteConsumerIdToUpdateRecord(ctx sdk.Context, consumerId string) {
```
p-offtermatt, line 164: Seems unused

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+		return errorsmod.Wrapf(types.ErrInvalidConsumerGenesis, "consumer genesis initial validator set is empty - no validators opted in")
+	}
+
+	// The cached context is created with a new EventManager so we merge the event
+	// into the original context
+	ctx.EventManager().EmitEvents(ctx.EventManager().Events())
+	return nil
+}
+
+func (k Keeper) UpdateConsumer(ctx sdk.Context, consumerId string) error {
```
p-offtermatt, line 317: Why is this only called in tests? I would expect this to be called during EndBlock/BeginBlock

Also, should this delete the update record?

x/ccv/provider/keeper/proposal.go
```diff
@@ -360,76 +320,21 @@ func (k Keeper) GetPendingConsumerAdditionProp(ctx sdk.Context, spawnTime time.T
...
-			// drop the proposal
-			ctx.Logger().Info("consumer client could not be created: %w", err)
-			continue
-		}
-
-		consumerGenesis, found := k.GetConsumerGenesis(cachedCtx, prop.ChainId)
-		if !found {
-			// drop the proposal
-			ctx.Logger().Info("consumer genesis could not be created")
+			ctx.Logger().Error("could not launch chain",
```
p-offtermatt, line 330: So this will retry a failing consumer chain launch in every single block?

p-offtermatt: because afaict these are not cleaned up or pruned?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -87,36 +88,19 @@ func (k msgServer) AssignConsumerKey(goCtx context.Context, msg *types.MsgAssign
...
 	}
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	err := k.Keeper.HandleConsumerRemovalProposal(ctx, msg)
-	if err != nil {
-		return nil, errorsmod.Wrapf(err, "failed handling ConsumerAddition proposal")
-	}
 
-	return &types.MsgConsumerRemovalResponse{}, nil
+	k.Keeper.SetConsumerIdToStopTime(ctx, msg.ConsumerId, msg.StopTime)
```
p-offtermatt, line 101: Does this mean I can set a stop time for a consumer chain that is not even launched yet?
Is there no way to remove the consumer chain record before it is launched?

p-offtermatt (commented):
It seems there are things that are not exported/imported in the Genesis (all the new store entries, actually?)
i.e. in this file https://github.com/cosmos/interchain-security/blob/e5e4867f0fece184b16e22416f9d4d2c7a7570f7/x/ccv/provider/keeper/genesis.go#L130

p-offtermatt:
I don't think I understand the flows in this PR very well. I am not sure if things are missing because it's just the first iteration, or because I am just missing them and they are just not where I am looking for them; I commented on the things I noticed in the first pass.
Could we do a sync session to review this? I think I need to understand the main data flows better

x/ccv/provider/keeper/partial_set_security.go
insumity: True, this was the case.
I removed the `consumerId != optedInConsumerId` so they cannot opt in to the same chain again.

x/ccv/provider/keeper/msg_server.go
insumity: Not sure I follow. What is the rest of the consumer state? You refer to the initialization and update records?
If yes, I do not see why we set them at this stage. They would be set in due time during the `MsgInitializeConsumer` and `MsgUpdateConsumer` messages.

x/ccv/provider/keeper/msg_server.go
insumity: No, not necessarily.
```
if k.GetAuthority() == msg.Authority {
    // message is executed as part of governance proposal and hence we change the owner address
    // to be the one of the module account address
    k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, k.GetAuthority())
} else if msg.Authority != ownerAddress {
    return nil, errorsmod.Wrapf(types.ErrUnauthorized, "expected owner address %s, got %s", ownerAddress, msg.Authority)
}
````
`msg.Authority` could technically be anything. If it's neither the gov address, nor the `ownerAddress` (i.e., who registered the chain), the transaction would fail. If it is the gov address, then we set the owner address to be the gov address from now on. If it's not the gov address, it **has to be** the owner of the consumer chain.

File: x/ccv/provider/keeper/msg_server.go (hunk 19 above)
Comment (insumity) [reply x/ccv/provider/keeper/msg_server.go:316]: This is possible. Seems safer if they have to set **everything** so they do not by accident pick up an old parameter that they did not want to use. But, I've no strong feelings on this. Let me know if you've a preference.

File: x/ccv/provider/keeper/msg_server.go (hunk 20 above)
Comment (insumity) [reply x/ccv/provider/keeper/msg_server.go:362]: Indeed, there was a problem in this condition. Updated it to be the same as the one we have for `InitializeConsumer`.

File: x/ccv/provider/keeper/msg_server.go [hunk 30]
```diff
@@ -309,16 +293,55 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+
+	phase, found := k.Keeper.GetConsumerIdToPhase(ctx, consumerId)
+	if !found || (phase != Registered && phase != Initialized) {
+		return nil, errorsmod.Wrapf(types.ErrInvalidPhase,
+			"chain with consumer id: %s has to be in its registered or initialized phase", consumerId)
 	}
 
+	ownerAddress, _ := k.Keeper.GetConsumerIdToOwnerAddress(ctx, consumerId)
+	if k.GetAuthority() == msg.Authority {
+		// message is executed as part of governance proposal and hence we change the owner address
```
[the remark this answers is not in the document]
insumity: This won't be the provider module authority. `GetAuthority` has the following godoc comment:
```
GetAuthority returns the x/ccv/provider module's authority.
```
Because of this, not sure if a clarification is needed here. Let me know if I misunderstood or feel free to resolve if it's okay.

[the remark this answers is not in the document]
p-offtermatt: I don't get it. You're saying `GetAuthority` won't be the provider module authority, but the godoc comment states that it is?

x/ccv/provider/keeper/proposal.go
insumity: Yes. Moved `RemoveConsumerFromToBeLaunchedConsumers` higher so that we delete consumer chains that we tried to initialize and failed as follows:
```
record, err := k.GetConsumerInitializationRecord(ctx, consumerId)
if err != nil {
        ...
	continue
}
k.RemoveConsumerFromToBeLaunchedConsumers(ctx, consumerId, record.SpawnTime)

cachedCtx, writeFn := ctx.CacheContext()
...
```

This is also not bullet proof because `GetConsumerInitializationRecord` might return an error.
Let me know if you have any concerns.

x/ccv/provider/keeper/msg_server.go
insumity: Yes, changed it to only be able to remove consumer if the chain is `Launched`.
> Is there no way to remove the consumer chain record before it is launched?
What would be the benefit of this?

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+	phase, found := k.GetConsumerIdToPhase(ctx, consumerId)
+	if !found || phase == Stopped {
+		return errorsmod.Wrapf(types.ErrInvalidPhase,
+			"cannot update stopped or not existing chain: %s", consumerId)
+	}
+
+	updateRecord, found := k.GetConsumerIdToUpdateRecord(ctx, consumerId)
```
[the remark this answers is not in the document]
insumity: Deprecated  the top N, validators power cap, validator set cap, etc. state in favor of the `ConsumerUpdateRecord`. However, I still have not deprecated the allowlist, and denylist state. It makes sense in my view to keep those stored in parallel with `ConsumerUpdateRecord` because we can easily access them as `ProviderConsAddr`esses and not merely as `string`s. Additionally, it's easier to query whether a validator is allowlisted without having to traverse through a list of allowlisted addresses every time.

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -144,9 +144,17 @@ service Query {
         option (google.api.http).get =
             "/interchain_security/ccv/provider/blocks_until_next_epoch";
   }
+
+  // QueryConsumerIdFromClientId returns the consumer id of the chain
+  // associated with the provided client id
+  rpc QueryConsumerIdFromClientId(QueryConsumerIdFromClientIdRequest)
```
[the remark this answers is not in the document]
insumity: Added the CLI support. Future queries would be added in a different PR.

insumity:
@coderabbitai resolve
@coderabbitai stop

x/ccv/provider/keeper/msg_server.go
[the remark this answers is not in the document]
insumity: Good catch. I misread this, sorry. Actually, look [here](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/keeper.go#L38):
```
// address capable of executing gov messages (gov module account)
authority string
```
Maybe I'm missing something here. I'll need to look more into this (something that would come up during E2E testing as well)>

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,440 @@
...
+type ConsumerPhase byte
+
+const (
+	Registered ConsumerPhase = iota
+	Initialized
+	Launched
+	Stopped
+)
+
+var DefaultUpdateRecord = types.ConsumerUpdateRecord{
```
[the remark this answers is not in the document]
insumity: Removed `DefaultUpdateRecord` altogether.

> Maybe it makes sense to have all these records in a different proto file than tx.proto, e.g., provider.proto and then have types/provider.go.

This is an option. Had all the records under `tx.proto` because they're part of the messages but we could have them under `provider.proto`. Let me know if you want them this way.

insumity:
@coderabbitai resolve
@coderabbitai pause

x/ccv/provider/client/cli/tx.go
```diff
@@ -209,6 +214,221 @@ Example:
...
+	cmd := &cobra.Command{
+		Use:   "register-consumer [path/to/registration-record.json]",
+		Short: "register a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Register a consumer chain to get back the assigned consumer id of this chain.
+Note that the one that signs this message is the owner of this consumer chain. The owner can be later
+changed by updating the consumer chain.
+
+Example:
+%s tx provider register-cosumer [path/to/registration-record.json] --from node0 --home ../node0 --chain-id $CID
```
bermuell, line 227:
```suggestion
%s tx provider register-consumer [path/to/registration-record.json] --from node0 --home ../node0 --chain-id $CID
```

x/ccv/provider/client/cli/tx.go
```diff
@@ -209,6 +214,221 @@ Example:
...
+func NewInitializeConsumerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "initialize-consumer [consumer-id] [path/to/initialization-record.json]",
+		Short: "initialize a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Initialize a consumer chain to get it ready to launch
+Note that only the owner of the chain can initialize it.
+
+Example:
+%s tx provider initialize-cosumer [consumer-id] [path/to/initialization-record.json] --from node0 --home ../node0 --chain-id $CID
```
bermuell, line 280:
```suggestion
%s tx provider initialize-consumer [consumer-id] [path/to/initialization-record.json] --from node0 --home ../node0 --chain-id $CID
```

x/ccv/provider/client/cli/tx.go
```diff
@@ -209,6 +214,221 @@ Example:
...
+func NewUpdateConsumerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-consumer [consumer-id] [path/to/update-record.json]",
+		Short: "update a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Update a consumer chain to change its running parameters (e.g., the allow list).
+Note that only the owner of the chain can initialize it.
+
+Example:
+%s tx provider update-cosumer [consumer-id] [path/to/update-record.json] --from node0 --home ../node0 --chain-id $CID
```
bermuell, line 334:
```suggestion
%s tx provider update-consumer [consumer-id] [path/to/update-record.json] --from node0 --home ../node0 --chain-id $CID
```

x/ccv/provider/client/cli/tx.go
```diff
@@ -209,6 +214,221 @@ Example:
...
+func NewRemoveConsumerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "remove-consumer [consumer-id] [stop-time-layout] [stop-time-value]",
+		Short: "remove a consumer chain",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Removes (and stops) a consumer chain. Note that only the owner of the chain can remove it.
+Stop time is parsed by using the layout and the value (see https://pkg.go.dev/time#Parse).
+
+Example:
+%s tx provider remove-cosumer [consumer-id] [stop-time-layout] [stop-time-value] --from node0 --home ../node0 --chain-id $CID
```
bermuell, line 388:
```suggestion
%s tx provider remove-consumer [consumer-id] [stop-time-layout] [stop-time-value] --from node0 --home ../node0 --chain-id $CID
```

testutil/keeper/unit_test_helpers.go
```diff
@@ -229,40 +230,79 @@
...
+		ChainId:     "chainID",
+	}
+}
+
+func GetTestInitializationRecord() providertypes.ConsumerInitializationRecord {
+	return providertypes.ConsumerInitializationRecord{
+		InitialHeight:                     clienttypes.NewHeight(4, 5),
+		GenesisHash:                       []byte("gen_hash"),
+		BinaryHash:                        []byte("bin_hash"),
+		SpawnTime:                         time.Now(),
```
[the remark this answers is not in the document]
mpoke: @bernd make sure this will not be a source of non-determinism for the tests.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
mpoke: @insumity The records are used for storing stuff in state. The messages are used for transactions. For simplicity, the messages contain the record as a field to avoid duplicating code. So IMO, records should go to provider.proto (which contains definition of data structures needed by the provider) and messages go in tx.proto.

x/ccv/provider/types/keys.go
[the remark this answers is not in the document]
mpoke: Not a strong preference either. If you think it helps, it's fine.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
mpoke: Does that mean that every time we need to add a new power shaping config we'll need to migrate the existing `ConsumerUpdateRecords` ?

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
mpoke: The benefit is not having extra logic for only allowing launched consumers to be stopped. Also, allowing all consumers to be stopped it's an easy way to do garbage collection.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
mpoke: That's the exact case we want to prevent -- malicious behavior that would try to scam validators into believing they are opting in on a real consumer chain.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -13,16 +13,31 @@ import (
...
+			"opting in to an unknown (or stopped) consumer chain, with id: %s", consumerId)
 	}
 
-	k.SetOptedIn(ctx, chainID, providerAddr)
+	registrationRecord, found := k.GetConsumerIdToRegistrationRecord(ctx, consumerId)
+	if !found {
+		return errorsmod.Wrapf(
+			types.ErrUnknownConsumerId,
+			"opting in to an unknown consumer chain, with id: %s", consumerId)
+	}
```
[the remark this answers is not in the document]
mpoke: `MsgRegisterConsumer` is the message that populates the `RegistrationRecord`. The fact we use `RegistrationRecord` as a field of the `MsgRegisterConsumer` is just an optimization. We could easily have ``MsgRegisterConsumer` have explicit fields. I want to decouple records that are meant for storing state and messages that are meant for users to submit input to the chain. Keeping them tightly couple will lead to weird design decisions.

x/ccv/provider/keeper/msg_server.go
mpoke: What happens if a user is sending a MsgInitializeConsumer without sending a MsgUpdateConsumer? Would the consumer start without power shaping configs? What would that even mean?

x/ccv/provider/keeper/msg_server.go
mpoke:
```golang
if k.GetAuthority() == msg.Authority {
    // message is executed as part of governance proposal and hence we change the owner address
    // to be the one of the module account address
    k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, k.GetAuthority())
```
Wouldn't this allow the Hub governance to take over any consumer chain?

x/ccv/provider/keeper/msg_server.go
mpoke: My view is that users don't need to query the chain to get the values and then to avoid race condition. If they just one to update one field, they should be able to do so without needing the know the other fields.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,836 @@
...
+		k.RemoveConsumerFromToBeLaunchedConsumers(ctx, consumerId, *previousSpawnTime)
+	}
+
+	k.AppendSpawnTimeForConsumerToBeLaunched(ctx, consumerId, spawnTime)
+}
+
+// CanLaunch checks on whether the consumer with `consumerId` has set all the initialization parameters set and hence
+// is ready to launch
+func (k Keeper) CanLaunch(ctx sdk.Context, consumerId string) bool {
+	if initializationParameters, err := k.GetConsumerInitializationParameters(ctx, consumerId); err != nil {
```
bermuell, line 738: shouldn't that be err==nil ?
Would be good to log an error in that case and for the positive case a info log would be good to see due to which 
parameter the chain is not ready to launch

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +305,126 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
 
+	if k.CanLaunch(ctx, consumerId) {
+		k.SetConsumerPhase(ctx, consumerId, Initialized)
+		k.PrepareConsumerForLaunch(ctx, consumerId, nil, *msg.InitializationParameters.SpawnTime)
+	} else {
+		k.SetConsumerPhase(ctx, consumerId, Registered)
+	}
+
+	gasAfter := ctx.GasMeter().GasConsumed()
+	ctx.GasMeter().ConsumeGas(gasAfter-initGas, "creating a chain has an additional cost during spawn time, "+
```
mpoke, line 344: Why double? This seems very random.

mpoke (commented):
Partial review. See my comments below.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +305,126 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	consumerId := msg.ConsumerId
+
+	ownerAddress, err := k.Keeper.GetConsumerOwnerAddress(ctx, consumerId)
 	if err != nil {
-		return nil, errorsmod.Wrapf(err, "failed handling ConsumerModification proposal")
+		// TODO (PERMISSIONLESS): use a better error message
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrInvalidConsumerId, "cannot retrieve owner address %s", ownerAddress)
+	}
+
+	if msg.PowerShapingParameters != nil && msg.PowerShapingParameters.IsTop_NSet && msg.PowerShapingParameters.Top_N > 0 && msg.Signer != k.GetAuthority() {
```
mpoke, line 361: This should be part of the update PowerShapingParameters logic, i.e., 
```
if msg.PowerShapingParameters != nil {
    // check if transition from Opt-In to TopN is possible 
    if msg.PowerShapingParameters.IsTop_NSet && msg.PowerShapingParameters.Top_N > 0 && msg.Signer != k.GetAuthority() {
        // error -- this transition needs to go through governance
    }

    // check if transition from TopN to Opt-In is possible
    if msg.PowerShapingParameters.IsTop_NSet && msg.PowerShapingParameters.Top_N == 0 {
        if oldPowerShapingParameters, err := k.Keeper.GetConsumerPowerShapingParameters(ctx, msg.ConsumerId); err == nil {
            if oldPowerShapingParameters.Top_N > 0 {
                if msg.NewOwnerAddress is a valid address {
                    // chain becomes an Opt In chain, hence we change the owner address
                    k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, msg.NewOwnerAddress)
                }
            }
        }
    }

    k.Keeper.SetConsumerInitializationParameters(ctx, msg.ConsumerId, *msg.InitializationParameters)
}
```

mpoke: Before update the PowerShapingParameters, you need to authenticate the message:
```
if msg.Signer == ownerAddress { 
    if ownerAddress != k.GetAuthority() {
        // the gov module shouldn't change ownership of a TopN chain to another address 
        if msg.NewOwnerAddress is a valid address { 
            k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, msg.NewOwnerAddress) 
        }
    }
} else if msg.Signer == k.GetAuthority() {
    if ownerAddress != k.GetAuthority() {
        // in what conditions can the governance take over an opt-in chain? 
    }
} else {
    // error: ErrUnauthorized (expecting either owner address or gov module address) 
}
```
The main question here is in what conditions can the governance take over an opt-in chain? If always, then we just update the consumer's ownership `k.GetAuthority()`.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,699 @@
...
+			remainingConsumerIds := len(result) + len(consumerIds) - int(limit)
+			if len(consumerIds[:len(consumerIds)-remainingConsumerIds]) == 0 {
+				return result
+			}
+			return append(result, consumerIds[:len(consumerIds)-remainingConsumerIds]...)
+		} else {
+			result = append(result, consumerIds...)
+		}
+
+		consumerIds = append(consumerIds, string(iterator.Key()[1+8:]))
```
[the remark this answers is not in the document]
mpoke: This seems like it shouldn't be here.

mpoke (commented):
Finished reviewing `UpdateConsumer`. See comments below.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,836 @@
...
+	consumerIds = append(consumerIds, consumerId)
+
+	// sort so that we avoid getting a consumer id in front of everyone else and delying the removal of a chain
+	sort.Slice(consumerIds, func(i, j int) bool {
+		return consumerIds[i] < consumerIds[j]
+	})
+
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err = enc.Encode(consumerIds)
```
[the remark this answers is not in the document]
insumity: Are you referring to the [slash acks](https://github.com/cosmos/interchain-security/blob/main/x/ccv/provider/keeper/keeper.go#L560) in the `keeper.go`? We can use proto un/marshalling there because all the `[]string`s is in the proto-generated `struct` `SlashAcks`.
I do not see a benefit in creating a proto-generated struct `ConsumerIds`. Am I missing something?

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,836 @@
+package keeper
+
+import (
+	"bytes"
+	errorsmod "cosmossdk.io/errors"
+	storetypes "cosmossdk.io/store/types"
+	"encoding/binary"
+	"encoding/gob"
```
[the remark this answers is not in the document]
insumity: See comment [below](https://github.com/cosmos/interchain-security/pull/2117#discussion_r1726880337). Since it's for the same thing, I'm resolving this comment.

x/ccv/provider/keeper/keeper.go
```diff
@@ -728,101 +742,79 @@ func (k Keeper) BondDenom(ctx sdk.Context) (string, error) {
...
-) (uint32, bool) {
-	store := ctx.KVStore(k.storeKey)
-	buf := store.Get(types.TopNKey(chainID))
-	if buf == nil {
-		return 0, false
-	}
-	return binary.BigEndian.Uint32(buf), true
+	consumerId string,
+) uint32 {
+	powerShapingParameters, _ := k.GetConsumerPowerShapingParameters(ctx, consumerId)
```
mpoke, line 770: Do not ignore errors.

x/ccv/provider/types/codec.go
```diff
@@ -24,15 +24,16 @@ func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
...
 	registry.RegisterImplementations(
 		(*sdk.Msg)(nil),
 		&MsgAssignConsumerKey{},
+		&MsgCreateConsumer{},
+		&MsgUpdateConsumer{},
+		&MsgRemoveConsumer{},
+		// keeping old messages (for now) so that existing proposals can be correctly deserialized
 		&MsgConsumerAddition{},
 		&MsgConsumerRemoval{},
+		&MsgConsumerModification{},
```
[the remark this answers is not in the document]
insumity: Removed.

x/ccv/provider/keeper/msg_server.go
insumity: Agreed that it's arbitrary. I believe the discussion was that we should have something like this. Whether it should be double, triple, etc. I do not know.
My take is that, our life would be substantially simpler if `MsgCreateConsumer` was quite expensive (~200 ATOMs) in regards to thinking about what happens when we're spammed, but this ideas was not favored. Without something like this, a malicious validator could easily prevent a chain from ever launching, etc. and I do not see an easy way around this.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
mpoke: @sainoe @MSalopek what do you guys think about this?

x/ccv/provider/keeper/msg_server.go
insumity: Discussed offline with Marius. If this ends up a problem, we'll deal with it later on.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	// initialization parameters are optional and hence could be nil
+	if msg.InitializationParameters != nil {
+		if err := k.SetConsumerInitializationParameters(ctx, consumerId, *msg.InitializationParameters); err != nil {
+			return &types.MsgCreateConsumerResponse{}, err
+		}
+	}
+
+	// power-shaping parameters are optional and hence could be null
+	if msg.PowerShapingParameters != nil {
+		if err := k.SetConsumerPowerShapingParameters(ctx, consumerId, *msg.PowerShapingParameters); err != nil {
```
mpoke, line 328: Check first that TopN is zero.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	if msg.Signer != ownerAddress {
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrUnauthorized, "expected owner address %s, got %s", ownerAddress, msg.Signer)
+	}
+
+	if msg.PowerShapingParameters != nil && msg.PowerShapingParameters.Top_N > 0 && ownerAddress != k.GetAuthority() {
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrInvalidTransformToTopN,
+			"an update to a Top N chain can only be done if chain is owner is the gov module")
+	}
+
+	k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, msg.NewOwnerAddress)
```
mpoke, line 362: What if `msg.NewOwnerAddress` is not set, i.e., is an empty field? Does the user need to always submit MsgUpdateConsumer with NewOwnerAddress set to the existing address?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
 	if err != nil {
-		return nil, errorsmod.Wrapf(err, "failed handling ConsumerModification proposal")
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrNoOwnerAddress, "cannot retrieve owner address %s", ownerAddress)
+	}
+
+	if msg.Signer != ownerAddress {
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrUnauthorized, "expected owner address %s, got %s", ownerAddress, msg.Signer)
+	}
+
+	if msg.PowerShapingParameters != nil && msg.PowerShapingParameters.Top_N > 0 && ownerAddress != k.GetAuthority() {
```
mpoke, line 357: nit: move in the `if msg.PowerShapingParameters != nil {}` block below.

x/ccv/provider/keeper/msg_server.go
mpoke: What if the current owner is the gov module and this message changes ownership to another address without changing TopN to zero?
The provider module should provide the following guarantees:
- only the existing owner can pass ownership to another account 
- a consumer with TopN > 0 MUST have the owner address set to the gov module account (this means that changing ownership from gov module entails automatically changing TopN to zero)

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	if msg.InitializationParameters != nil {
+		k.Keeper.SetConsumerInitializationParameters(ctx, msg.ConsumerId, *msg.InitializationParameters)
+	}
+
+	if msg.PowerShapingParameters != nil {
+		oldTopN := k.Keeper.GetTopN(ctx, consumerId)
+		k.Keeper.SetConsumerPowerShapingParameters(ctx, msg.ConsumerId, *msg.PowerShapingParameters)
+		k.Keeper.PopulateAllowlist(ctx, consumerId)
+		k.Keeper.PopulateDenylist(ctx, consumerId)
+		k.Keeper.PopulateMinimumPowerInTopN(ctx, consumerId, oldTopN)
```
mpoke, line 377: Handle errors.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
sainoe: My opinion is to use either `json` or `gob` formats. I have a slight preference for the former as it's widely used, including in the SDK.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+
+	if msg.PowerShapingParameters != nil && msg.PowerShapingParameters.Top_N > 0 && ownerAddress != k.GetAuthority() {
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrInvalidTransformToTopN,
+			"an update to a Top N chain can only be done if chain is owner is the gov module")
+	}
+
+	k.Keeper.SetConsumerOwnerAddress(ctx, consumerId, msg.NewOwnerAddress)
+
+	if msg.Metadata != nil {
+		k.Keeper.SetConsumerMetadata(ctx, msg.ConsumerId, *msg.Metadata)
```
mpoke, line 365: ditto: handle erros

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
-	return &types.MsgConsumerModificationResponse{}, nil
+	if msg.InitializationParameters != nil {
+		k.Keeper.SetConsumerInitializationParameters(ctx, msg.ConsumerId, *msg.InitializationParameters)
+	}
+
+	if msg.PowerShapingParameters != nil {
+		oldTopN := k.Keeper.GetTopN(ctx, consumerId)
+		k.Keeper.SetConsumerPowerShapingParameters(ctx, msg.ConsumerId, *msg.PowerShapingParameters)
+		k.Keeper.PopulateAllowlist(ctx, consumerId)
+		k.Keeper.PopulateDenylist(ctx, consumerId)
```
mpoke, line 376: Pass *msg.PowerShapingParameters.Allowlist and *msg.PowerShapingParameters.Denylist to these functions to avoid having to get the PowerShapingParameters that were just set.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,742 @@
...
+		k.RemoveConsumerFromToBeLaunchedConsumers(ctx, consumerId, previousParameters.SpawnTime)
+	}
+
+	k.AppendSpawnTimeForConsumerToBeLaunched(ctx, consumerId, spawnTime)
+}
+
+// CanLaunch checks on whether the consumer with `consumerId` has set all the initialization parameters set and hence
+// is ready to launch
+// TODO (PERMISSIONLESS): could remove, all fields should be there because we validate the initialization parameters
+func (k Keeper) CanLaunch(ctx sdk.Context, consumerId string) bool {
```
mpoke, line 730: What about spawnTime?

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+		k.Keeper.PopulateMinimumPowerInTopN(ctx, consumerId, oldTopN)
+	}
+
+	if k.CanLaunch(ctx, consumerId) {
```
mpoke, line 380: This will returned true for an already launched chain, which means the phase will switch back to initialized.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +303,96 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
-	err := k.Keeper.HandleConsumerModificationProposal(ctx, msg)
+	consumerId := msg.ConsumerId
+
+	ownerAddress, err := k.Keeper.GetConsumerOwnerAddress(ctx, consumerId)
 	if err != nil {
-		return nil, errorsmod.Wrapf(err, "failed handling ConsumerModification proposal")
+		return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrNoOwnerAddress, "cannot retrieve owner address %s", ownerAddress)
+	}
+
+	if msg.Signer != ownerAddress {
```
bermuell, line 353: This check does not work for governance driven proposals as it's always the gov authority which signs the message. 
Can we do something like this to make governance proposals work ?
```suggestion
	if msg.Signer != ownerAddress && msg.Signer != k.authority {
```

x/ccv/provider/keeper/msg_server.go
insumity: This is already checked in `MsgCreateConsumer`s `ValidateBasic`, but we can put this here as well.

x/ccv/provider/keeper/msg_server.go
insumity: Sure, we can do that, but it is not clear to me why you are suggesting this (?)

x/ccv/provider/keeper/msg_server.go
insumity: Good catch. Added a check inside `CanLaunch`.

x/ccv/provider/keeper/permissionless.go
insumity: Note that `spawnTime` is not optional anymore. In any case, added a check that `spawnTime` is in the future.

x/ccv/provider/keeper/msg_server.go
insumity: >What if msg.NewOwnerAddress is not set, i.e., is an empty field? Does the user need to always submit MsgUpdateConsumer with NewOwnerAddress set to the existing address?

A transaction with an empty `NewOwnerAddress` would fail `MsgUpdateConsumer`s `ValidateBasic`. But we can add the same empty check here.

>Does the user need to always submit MsgUpdateConsumer with NewOwnerAddress set to the existing address?

Yes.

> What if the current owner is the gov module and this message changes ownership to another address without changing TopN to zero?

Can happen. The governance can "kill" this. But, adding a check here to guarantee that it can never happen.

> only the existing owner can pass ownership to another account

This is done in:
```
if msg.Signer != ownerAddress {
    return &types.MsgUpdateConsumerResponse{}, errorsmod.Wrapf(types.ErrUnauthorized, "expected owner address %s, got %s", ownerAddress, msg.Signer)
}	
```

> a consumer with TopN > 0 MUST have the owner address set to the gov module account (this means that changing ownership from gov module entails automatically changing TopN to zero)

This was not there. Added a check (i.e., `Top_N > 0 && OwnerAddress != Gov module`) on this and fail the transaction if the check succeeds, instead of setting the `Top_N` to 0.

x/ccv/provider/keeper/permissionless.go
mpoke: Spawn time is not optional, but the entire ConsumerInitializationParameters data structure is, right?

x/ccv/provider/keeper/msg_server.go
mpoke: To have logical blocks connected. This condition is testing whether the message can update TopN, which is a power shaping param.

x/ccv/provider/keeper/msg_server.go
mpoke: >> Does the user need to always submit MsgUpdateConsumer with NewOwnerAddress set to the existing address?

> Yes.

This is not ideal.

x/ccv/provider/keeper/permissionless.go
insumity: > Spawn time is not optional, but the entire ConsumerInitializationParameters data structure is, right?

Yes. But `CanLaunch` checks `ConsumerInitializationParameters` exists, right? So, internally there should be a `spawnTime`.

x/ccv/provider/types/msg.go
```diff
@@ -227,11 +234,176 @@ func (msg MsgSubmitConsumerDoubleVoting) GetSigners() []sdk.AccAddress {
...
+// Route implements the sdk.Msg interface.
+func (msg MsgUpdateConsumer) Route() string { return RouterKey }
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgUpdateConsumer) ValidateBasic() error {
+	if err := ValidateConsumerId(msg.ConsumerId); err != nil {
+		return err
+	}
+
+	const maxNewOwnerAddress = 500
```
mpoke, line 330: This shouldn't not be validated like this. See https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/x/staking/keeper/msg_server.go#L161 for an example.

x/ccv/provider/keeper/permissionless.go
```diff
@@ -0,0 +1,746 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/interchain-security/v5/x/ccv/provider/types"
+	"strconv"
+	"time"
+)
+
+// ConsumerPhase captures the phases of a consumer chain according to `docs/docs/adrs/adr-018-permissionless-ics.md`
+type ConsumerPhase byte
+
+const (
```
sainoe, line 19: I'd use `protobuf` `enum` instead so that they're defined using `string`s.

x/ccv/provider/types/msg.go
insumity: Thanks!
Your points seems to be for validator addresses. I used:
```
		_, err := sdk.AccAddressFromBech32(msg.NewOwnerAddress)
```
for general accounts. If that's okay, feel free to resolve.

x/ccv/provider/keeper/permissionless.go
insumity: Talked offline. `CanLaunch` checks that `spawnTime` is in the future.

x/ccv/provider/keeper/msg_server.go
insumity: Fixed mentioned issues.

x/ccv/provider/keeper/permissionless.go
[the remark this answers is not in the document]
MSalopek: To remain in sync with the rest of the ecosystem please define a proto struct and use it.

Something like this should do the trick.

```proto
message ConsumerIds {
  repeated string ids;
}
```

x/ccv/provider/types/msg.go
mpoke: Then use this example from MsgSend:
 
- https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/proto/cosmos/bank/v1beta1/tx.proto#L46
- https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/x/bank/keeper/msg_server.go#L36

mpoke (approved):
Great work @insumity. I think the PR is good enough to be merged so that we can start the Gaia integration in parallel. The unresolved comments can be addressed in subsequent PRs.

PR #2122, opened 2024-08-02
test: Fix use of hardcoded cosmos... address

p-offtermatt:
## Description

Closes: #N/A

Changes handling of a hardcoded address to be done with fixed prefix

This should fix an issue that was surfaced by consumer chains that run this test, since their bech32 prefix is different

This is to address an issue that was surfaced by a consumer chain when upgrading to ICSv5.1.1

In general, if we want to offer our tests to consumer chains to test against, we cannot hardcode cosmos... addresses (at least in the integration and unit tests, where it seems this will mess things up)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved accuracy of the "IBC Transfer recipient is not the consumer rewards pool address" test case by using a dynamic address instead of a hardcoded one.
  
- **Tests**
	- Enhanced test readability and functionality through streamlined setup functions.
	- Ensured consistent passing of the test suite context in test cases for better integration.
	- Introduced a new variable for improved test accuracy in the IBC Transfer middleware scenario.

### DIFF
--- a/tests/integration/distribution.go
+++ b/tests/integration/distribution.go
@@ -138,8 +138,7 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
 	consuValsRewards := consumerValsOutstandingRewardsFunc(s.providerCtx())
 
 	// increase the block height so validators are eligible for consumer rewards (see `IsEligibleForConsumerRewards`)
-	numberOfBlocksToStartReceivingRewards :=
-		providerKeeper.GetNumberOfEpochsToStartReceivingRewards(s.providerCtx()) * providerKeeper.GetBlocksPerEpoch(s.providerCtx())
+	numberOfBlocksToStartReceivingRewards := providerKeeper.GetNumberOfEpochsToStartReceivingRewards(s.providerCtx()) * providerKeeper.GetBlocksPerEpoch(s.providerCtx())
 
 	for s.providerCtx().BlockHeight() <= numberOfBlocksToStartReceivingRewards {
 		s.providerChain.NextBlock()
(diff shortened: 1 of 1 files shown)


p-offtermatt:
Nope, doesn't look like the way to do it actually

p-offtermatt:
Reopening, because indeed it *does* seem like the way to fix the issue. The test failure with https://github.com/cosmos/interchain-security/pull/2123 confirmed that on consumer chains, in integration tests the provider chain nonetheless seems to use the consumer bech32 prefix

tests/integration/distribution.go
```diff
@@ -554,22 +553,23 @@ func (s *CCVTestSuite) TestIBCTransferMiddleware() {
 		},
 		{
 			"IBC Transfer recipient is not the consumer rewards pool address",
-			func(ctx sdk.Context, keeper *providerkeeper.Keeper, bankKeeper icstestingutils.TestBankKeeper) {
-				data.Receiver = "cosmos149lw9fktlqfed3zt8ah48r5czmsug5s7kw77u9" // random acct address
+			func(s *CCVTestSuite, ctx sdk.Context, keeper *providerkeeper.Keeper, bankKeeper icstestingutils.TestBankKeeper) {
+				data.Receiver = s.providerChain.SenderAccount.GetAddress().String() // random acct address
```
mpoke, line 557: This is not a random address. It seems that it will always be the same address. Can't we use a variable declared before `testCases` and avoid adding a new param to the setup function?

tests/integration/distribution.go
```diff
@@ -138,8 +138,7 @@ func (s *CCVTestSuite) TestRewardsDistribution() {
 	consuValsRewards := consumerValsOutstandingRewardsFunc(s.providerCtx())
 
 	// increase the block height so validators are eligible for consumer rewards (see `IsEligibleForConsumerRewards`)
-	numberOfBlocksToStartReceivingRewards :=
-		providerKeeper.GetNumberOfEpochsToStartReceivingRewards(s.providerCtx()) * providerKeeper.GetBlocksPerEpoch(s.providerCtx())
+	numberOfBlocksToStartReceivingRewards := providerKeeper.GetNumberOfEpochsToStartReceivingRewards(s.providerCtx()) * providerKeeper.GetBlocksPerEpoch(s.providerCtx())
```
mpoke, line 141: nit: there seems to be no change on this line. If so, could you please revert to the original?

tests/integration/distribution.go
p-offtermatt: The comment was misleading, I adjusted it.

I don't understand the concrete proposal. In all testcases except this one, it is important to keep the address as-is

p-offtermatt: Also adjusted the logic to not need the testsuite here.

mpoke (approved):

insumity (approved):

Issue #2108, opened 2024-07-29
EPIC: Permissionless ICS


insumity:
## Problem

This is a meta issue to track the work of [Permissinoless ICS](https://github.com/cosmos/interchain-security/blob/main/docs/docs/adrs/adr-018-permissionless-ics.md). Permissionless ICS allows _Opt In_ consumer chains to join ICS by simply issuing transactions, without having to go through a governance proposal.

## Task list

```[tasklist]
### Introduce new messages
- [x] new `MsgCreateConsumer` and `MsgUpdateConsumer` messages
```

```[tasklist]
### Re-key state
- [x] change all messages by renaming `chainId` to `consumerId` and use accordingly but make sure the error mesasges point out that `consumerId` is now needed (e.g., all previously working `MsgOptIn` messages would now fail because there's no consumer chain with a `stride-1` `consumerId`)
- [x] fix (and add new) unit/integration refactored to take into account `consumerId` instead of `chainId`
- [ ] fix (and add new) E2E tests refactored to take into account `consumerId` instead of `chainId`
```

```[tasklist]
### Introduce new state
- [x] `consumerId` counter stores somewhere that is never recycled and always increments
- [x] (`consumerId` to `chainId`) needed for misbehavior evidence (e.g., `HandleConsumerDoubleVoting`)
- [x] (`clientId` to `consumerId`) to be able to find the right consumer chain when setting up a channel, etc.
- [x] potentially additional state for the phase of a consumer chain, e.g., whether it's registered, initialized, etc. so that instead of doing `GetConsumerClientId( ...)` to check if a chain has launched, we have something akin `IsLaunched(...)`
- [x] add `owner_address` for each consumer state
```

```[tasklist]
### Deprecate proposals
- [x] deprecate legacy consumer proposals. Specifically, deprecate `ConsumerAdditionProposal` which means getting rid of `ConsumerAdditionProposal`s and using just the `consumerId`s instead that flow around (e.g., when checking which chain to launch) (see [#2130](https://github.com/cosmos/interchain-security/pull/2130))
- [x] deprecate `MsgConsumerAddition` and replace with a proposal that contains both `MsgInitializeConsumer` and `MsgUpdateConsumer`
```

```[tasklist]
### Modify relayer (optional)
- [x] contact Hermes to use the `consumerId` instead of the `chainId` for misbehaviour messages (see [hermes#4153](https://github.com/informalsystems/hermes/issues/4153))
[...]

PR #2171, opened 2024-08-26
feat!: Permissionless ICS 

mpoke:
## Description

Closes: #2108

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/provider/keeper/msg_server.go
```diff
@@ -309,16 +307,163 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	// initialization parameters are optional and hence could be nil
+	if msg.InitializationParameters != nil {
+		if err := k.Keeper.SetConsumerInitializationParameters(ctx, consumerId, *msg.InitializationParameters); err != nil {
+			return &types.MsgCreateConsumerResponse{}, errorsmod.Wrapf(types.ErrInvalidConsumerInitializationParameters,
+				"cannot set consumer initialization parameters: %s", err.Error())
+		}
+	}
+
+	// power-shaping parameters are optional and hence could be null
+	if msg.PowerShapingParameters != nil {
```
mpoke, line 334: @insumity Update allowlist and denylist inside this block. See `UpdateConsumer`, i.e., 
```golang
    k.Keeper.UpdateAllowlist(ctx, consumerId, msg.PowerShapingParameters.Allowlist)
    k.Keeper.UpdateDenylist(ctx, consumerId, msg.PowerShapingParameters.Denylist)
```
It is fine to not update the min power for TopN (see `UpdateMinimumPowerInTopN`) as the chain is still in Registered phase. The min power will be computed and stored in state on `MakeConsumerGenesis` at spawn time. Maybe add this as a comment.

x/ccv/provider/proposal_handler.go
mpoke, line 1: @bermuell This entire file needs to be removed. This will affect the e2e tests though.

x/ccv/provider/keeper/msg_server.go
```diff
@@ -293,14 +292,14 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
 		return nil, err
 	}
 
-	if err := k.HandleSetConsumerCommissionRate(ctx, msg.ChainId, types.NewProviderConsAddress(consAddr), msg.Rate); err != nil {
+	if err := k.HandleSetConsumerCommissionRate(ctx, msg.ConsumerId, types.NewProviderConsAddress(consAddr), msg.Rate); err != nil {
 		return nil, err
 	}
 
 	ctx.EventManager().EmitEvents(sdk.Events{
```
mpoke, line 299: @insumity Add more events for all new messages.

mpoke: Done in https://github.com/cosmos/interchain-security/pull/2221

insumity (approved):
As discussed with @mpoke. I did **not** review but accepting (✅) to unblock cutting. I'll review at a later stage.

MSalopek (approved):
Admin approval.
Review will follow.

proto/interchain_security/ccv/provider/v1/tx.proto
```diff
@@ -23,34 +23,45 @@ service Msg {
...
   // The validator address on the provider
   string provider_addr = 2 [ (gogoproto.moretags) = "yaml:\"address\"" ];
   // The consensus public key to use on the consumer.
   // in json string format corresponding to proto-any, ex:
   // `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"Ui5Gf1+mtWUdH8u3xlmzdKID+F3PK0sfXZ73GZ6q6is="}`
   string consumer_key = 3;
 
-  // Tx signer address
-  string signer = 4 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string submitter = 4 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
MSalopek, line 61: Revert the changes to not break compatibility. If you do this you will not be able to unmarshall historic Txs.

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+
+		// if current block time is equal to or after spawnTime, and the chain is initialized, we can launch the chain
+		consumerIds, err := k.GetConsumersToBeLaunched(ctx, spawnTime)
+		if err != nil {
+			k.Logger(ctx).Error("failed to retrieve consumers to launch",
+				"spawn time", spawnTime,
+				"error", err)
+			continue
+		}
+		if len(result)+len(consumerIds.Ids) >= int(limit) {
```
MSalopek, line 128: Simplify this. This is causing a bug at the callsite.

At the callsite, you are removing ALL items under a spawn time.

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+		initializationParameters, err := k.GetConsumerInitializationParameters(ctx, consumerId)
+		if err != nil {
+			ctx.Logger().Error("could not retrieve initialization record",
+				"consumerId", consumerId,
+				"error", err)
+			continue
+		}
+		// Remove consumer to prevent re-trying launching this chain.
+		// We would only try to re-launch this chain if a new `MsgUpdateConsumer` message is sent.
+		err = k.RemoveConsumerToBeLaunched(ctx, consumerId, initializationParameters.SpawnTime)
```
MSalopek, line 76: Here, you are removing ALL ids under a spawn time index. You should be removing a partial list.

See: [GetConsumersReadyToLaunch](https://github.com/cosmos/interchain-security/pull/2171/files/129164c31d796125fcab3d1d7ff54f941dc6b6a4#r1747167739)

x/ccv/provider/keeper/msg_server.go
```diff
@@ -293,32 +313,342 @@ func (k msgServer) SetConsumerCommissionRate(goCtx context.Context, msg *types.M
...
+	}
+	if err := k.Keeper.SetConsumerInitializationParameters(ctx, consumerId, initializationParameters); err != nil {
+		return &resp, errorsmod.Wrapf(types.ErrInvalidConsumerInitializationParameters,
+			"cannot set consumer initialization parameters: %s", err.Error())
+	}
+	if !initializationParameters.SpawnTime.IsZero() {
+		// add SpawnTime event attribute
+		eventAttributes = append(eventAttributes,
+			sdk.NewAttribute(types.AttributeConsumerSpawnTime, initializationParameters.SpawnTime.String()))
+	}
```
p-offtermatt, line 390: Feels like this should go in line 409 after the If on InitializeConsumer and use the SpawnTime from there

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+			k.Logger(ctx).Error("failed to retrieve consumers to launch",
+				"spawn time", spawnTime,
+				"error", err)
+			continue
+		}
+		if len(result)+len(consumerIds.Ids) >= int(limit) {
+			remainingConsumerIds := len(result) + len(consumerIds.Ids) - int(limit)
+			if len(consumerIds.Ids[:len(consumerIds.Ids)-remainingConsumerIds]) == 0 {
+				return result
+			}
```
p-offtermatt, line 132: This is complicated and uses so much index math, but it is just a check that len(result) == int(limit)

The math:
len(consumerIds.Ids) - remainingConsumerIds =
len(consumerIds.Ids) - len(result) - len(consumerIds.Ids) + int(limit) =
int(limit) - len(result)

the only way for consumerIds.Ids[:int(limit)-len(result)] to be empty is if int(limit) - len(result) == 0

Would suggest refactoring the surrounding if block with this in mind and to avoid the complicated index math

x/ccv/provider/types/msg.go
```diff
@@ -3,521 +3,594 @@ package types
...
-	for _, denom := range msg.DenomsToAdd {
-		if !sdk.NewCoin(denom, math.NewInt(1)).IsValid() {
-			return fmt.Errorf("invalid change reward denoms proposal: %s is not a valid denom", denom)
-		}
+	return nil
+}
+
+func ValidateByteSlice(hash []byte, maxLength int) error {
+	if len(hash) > maxLength {
+		return fmt.Errorf("hash is too long; got: %d, max: %d", len(hash), MaxHashLength)
```
p-offtermatt, line 569: Tiny nit/bug here

```suggestion
		return fmt.Errorf("hash is too long; got: %d, max: %d", len(hash), maxLength)
```

proto/interchain_security/ccv/provider/v1/provider.proto
```diff
@@ -362,3 +366,122 @@ message ConsumerRewardsAllocation {
...
 }
+
+// ConsumerMetadata contains general information about the registered chain
+message ConsumerMetadata {
+  // the name of the chain
+  string name = 1;
+  // the description of the chain
+  string description = 2;
+  // the metadata (e.g., GitHub repository URL) of the chain
+  string metadata = 3;
```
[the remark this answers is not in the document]
mpoke: We leave it as a string for now.

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+		return err
+	}
+
+	// Set minimum height for equivocation evidence from this consumer chain
+	k.SetEquivocationEvidenceMinHeight(ctx, consumerId, initializationRecord.InitialHeight.RevisionHeight)
+
+	// Consumers start out with the unbonding period from the consumer addition prop
+	consumerUnbondingPeriod := initializationRecord.UnbondingPeriod
+
+	// Create client state by getting template client from parameters and filling in zeroed fields from proposal.
```
p-offtermatt, line 189: Same, proposal seems outdated

x/ccv/provider/types/throttle.go
```diff
@@ -1,17 +0,0 @@
-package types
-
-import (
-	"time"
-)
-
-// NewGlobalSlashEntry creates a new GlobalSlashEntry.
-func NewGlobalSlashEntry(recvTime time.Time, consumerChainID string,
```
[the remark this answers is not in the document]
mpoke: No longer used after the introduction of throttling v2.

x/ccv/provider/keeper/consumer_lifecycle.go
mpoke: See https://github.com/cosmos/interchain-security/pull/2236

p-offtermatt:
Also outdated here:
https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/params.go#L14-L18

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+	if err != nil {
+		return gen, nil, fmt.Errorf("unable to create validator set from updates computed from key assignment in MakeConsumerGenesis: %s", err)
+	}
+	hash := tmtypes.NewValidatorSet(updatesAsValSet).Hash()
+
+	consumerGenesisParams := ccv.NewParams(
+		true,
+		initializationRecord.BlocksPerDistributionTransmission,
+		initializationRecord.DistributionTransmissionChannel,
+		"", // providerFeePoolAddrStr,
```
p-offtermatt, line 334: I feel like this needs a bit more of a comment here, looks a bit sketchy and like someone just forgot to uncomment this without any extra comment

x/ccv/provider/keeper/consumer_lifecycle.go
mpoke: This was fixed in https://github.com/cosmos/interchain-security/pull/2236

mpoke:
@p-offtermatt I addressed comments until here in https://github.com/cosmos/interchain-security/pull/2237

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -0,0 +1,670 @@
...
+	k.DeleteDenylist(ctx, consumerId)
+	k.DeleteAllOptedIn(ctx, consumerId)
+	k.DeleteConsumerValSet(ctx, consumerId)
+
+	k.DeleteConsumerRewardsAllocation(ctx, consumerId)
+	k.DeleteConsumerRemovalTime(ctx, consumerId)
+
+	// TODO (PERMISSIONLESS) add newly-added state to be deleted
+	// Note that we do not delete ConsumerIdToChainIdKey and ConsumerIdToPhase, as well
+	// as consumer metadata, initialization and power-shaping parameters.
```
p-offtermatt, line 517: Please add some minimal explanation of why we don't delete this here

mpoke: see https://github.com/cosmos/interchain-security/pull/2237/commits/1ab477bc7061e89f51cac19fed9d287f12c630bf

p-offtermatt:
There is a bug around the spawntime in UpdateConsumer; essentially it's possible fo a user to set a consumer chain that will attempt to launch in every block.

Assume we have a Registered consumer chain with a SpawnTime.
Now, a user sends a MsgUpdateConsumer with InitializationParameters where SpawnTime==0.
(This is not validated against in ValidateBasic https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/types/msg.go#L524)

The new InitializationParameters get set here: https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/msg_server.go#L507

However, the InitializeConsumer call here https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/msg_server.go#L567 will fail because the SpawnTime is zero, see this check https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/consumer_lifecycle.go#L54

Thus, the PrepareConsumerForLaunch call here https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/msg_server.go#L568 does not update the SpawnTime in the ConsumerToBeLaunched store.

So the InitializationParameters will say SpawnTime==0, but the store will have SpawnTime=oldSpawnTime.

During BeginBlock, when the consumers are launched, this RemoveConsumerToBeLaunched call here will thus not remove the consumer https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/consumer_lifecycle.go#L76 and error, which causes the loop to continue.

In the extreme, someone can create 200 such consumers, and those will be re-tried to be launched in every block and no other consumers will ever launch

p-offtermatt:
Another find:
The owner can update launched consumer chains; in particular the owner can arbitrarily modify the InitializationParameters of the already launched chain. afaict this doesn't have any further impact, but it means the InitializationParameters returned by QueryConsumerChain can be arbitrary, and it would look like the chain launched with different parameters than it actually launched with (i.e. someone malicious could make it look like the chain had e.g. very large ConsumerRedistributionFraction

mpoke:
> So the InitializationParameters will say SpawnTime==0, but the store will have SpawnTime=oldSpawnTime.
> 
> During BeginBlock, when the consumers are launched, this RemoveConsumerToBeLaunched call here will thus not remove the consumer
> 
> https://github.com/cosmos/interchain-security/blob/5b6252c7726809f0c56beb346de328a152f6d4fc/x/ccv/provider/keeper/consumer_lifecycle.go#L76
> 
> and error, which causes the loop to continue.
> In the extreme, someone can create 200 such consumers, and those will be re-tried to be launched in every block and no other consumers will ever launch

Nice catch @p-offtermatt. I do think that https://github.com/cosmos/interchain-security/pull/2236 somehow fixed this, but we should probably not have inconsistencies between the initialization params and the time queues used to store the chains to be launched.

PR #2177, opened 2024-08-27
feat!: migrations for permissionless

insumity:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -35,20 +35,36 @@ service Query {
         "/interchain_security/ccv/provider/consumer_chains";
   }
 
+  // [DEPRECATED]: Use `QueryConsumersThatAreAboutToStart` instead
   // QueryConsumerChainStarts queries consumer chain start proposals.
   rpc QueryConsumerChainStarts(QueryConsumerChainStartProposalsRequest)
       returns (QueryConsumerChainStartProposalsResponse) {
+    option deprecated = true;
```
insumity, line 42: I deprecate the `QueryConsumerChainStarts` and `QueryConsumerChainStops` queries.
These queries seem to be returning consumers that are about to start or stop respectively. However, I do not not really like the naming of their arguments `...Proposals...`  because with permissionless we have chains that are about to start or stop outside a proposal. Because of this, I deprecate the queries in favor of `QueryConsumersThatAreAboutToStart` and `QueryConsumersThatAreAboutToStop`.

x/ccv/provider/keeper/grpc_query_test.go
```diff
@@ -384,3 +384,7 @@ func TestQueryConsumerIdFromClientId(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, expectedConsumerId, res.ConsumerId)
 }
+
+func TestQueryProposedConsumerChainIDs(t *testing.T) {
+	// TODO (PERMISSIONLESS)
```
insumity, line 389: @mpoke This might get quite tricky because we need to mock governance, proposals, etc. I wonder if we should go back to a solution where we store the association between `consumerId` to a `proposalId` and then just implement the `QueryProposedConsumerChainIDs` by looking at that.

x/ccv/provider/migrations/v8/migrations.go
```diff
@@ -211,37 +218,86 @@ func MigrateLaunchedConsumerChains(ctx sdk.Context, store storetypes.KVStore, pk
...
 		}
+
 		validatorSetCapKey := append([]byte{LegacyValidatorSetCapKeyPrefix}, []byte(chainId)...)
 		var validatorSetCap uint32 = 0
 		buf = store.Get(validatorSetCapKey)
 		if buf != nil {
 			validatorSetCap = binary.BigEndian.Uint32(buf)
 		}
+
+		bech32PrefixConsAddr := sdk.GetConfig().GetBech32ConsensusAddrPrefix()
```
insumity, line 236: @mpoke If we are going to set the allowlists and denylists, we might as well add them in the power-shaping parameters. The way I see it, is that a launched chain might execute successfully a `ConsumerModification` and hence allowlists and denylists would be set. If this is the case we want this to show in the `powerShapingParameters` as well.

x/ccv/provider/migrations/v8/migrations.go
```diff
@@ -138,70 +140,75 @@ func MigrateLaunchedConsumerChains(ctx sdk.Context, store storetypes.KVStore, pk
...
 	for _, chainId := range chainIds {
-		// create new consumerID
+		// create new consumerId
 		consumerId := pk.FetchAndIncrementConsumerId(ctx)
 
 		// re-key store
 
-		// chainId -> clientId
-		rekeyFromChainIdToConsumerId(store, LegacyChainToClientKeyPrefix, chainId, consumerId)
+		// channelId -> chainId
```
insumity, line 148: @mpoke Verified that we re-key everything that is needed in this method.

proto/interchain_security/ccv/provider/v1/query.proto
mpoke: Why deprecate and not just remove them?

proto/interchain_security/ccv/provider/v1/query.proto
```diff
@@ -35,20 +35,36 @@ service Query {
...
     option (google.api.http).get =
         "/interchain_security/ccv/provider/consumer_chain_start_proposals";
   }
 
+  rpc QueryConsumersThatAreAboutToStart(QueryConsumersThatAreAboutToStartRequest)
```
mpoke, line 47: Rename to `ConsumersAboutToStart`. Same for `QueryConsumersThatAreAboutToStop`.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -105,39 +107,34 @@ func (k Keeper) GetConsumerChain(ctx sdk.Context, consumerId string) (types.Chai
 	}, nil
 }
 
-// TODO (PERMISSIONLESS)
 func (k Keeper) QueryConsumerChainStarts(goCtx context.Context, req *types.QueryConsumerChainStartProposalsRequest) (*types.QueryConsumerChainStartProposalsResponse, error) {
```
mpoke, line 110: Remove instead of deprecating. It will be a CLI breaking change, but we have a lot of them.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -105,39 +107,34 @@ func (k Keeper) GetConsumerChain(ctx sdk.Context, consumerId string) (types.Chai
...
+	return nil, status.Error(codes.Unimplemented, "This query is not supported anymore. Use `QueryConsumersThatAreAboutToStart` instead")
+}
+
+func (k Keeper) QueryConsumersThatAreAboutToStart(goCtx context.Context, req *types.QueryConsumersThatAreAboutToStartRequest) (*types.QueryConsumersThatAreAboutToStartResponse, error) {
```
mpoke, line 114: So a better name for this query is `QueryConsumersReadyToLaunch`, right? But the `GetInitializedConsumersReadyToLaunch` method returns consumers that are ready to launch now. This means that they will be launched in the next block. I'm not sure I understand the reason we have this query.

proto/interchain_security/ccv/provider/v1/query.proto
mpoke: Also, let's not focus now on queries that are not needed for Forge.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -105,39 +107,34 @@ func (k Keeper) GetConsumerChain(ctx sdk.Context, consumerId string) (types.Chai
...
+	consumerIds := k.GetInitializedConsumersReadyToLaunch(ctx, math.MaxUint32)
+	return &types.QueryConsumersThatAreAboutToStartResponse{ConsumerIds: consumerIds}, nil
 }
 
-// TODO (PERMISSIONLESS)
 func (k Keeper) QueryConsumerChainStops(goCtx context.Context, req *types.QueryConsumerChainStopProposalsRequest) (*types.QueryConsumerChainStopProposalsResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "This query is not supported anymore. Use `QueryConsumersThatAreAboutToStop` instead")
+}
+
+func (k Keeper) QueryConsumersThatAreAboutToStop(goCtx context.Context, req *types.QueryConsumersThatAreAboutToStopRequest) (*types.QueryConsumersThatAreAboutToStopResponse, error) {
```
mpoke, line 129: ditto: This query will return consumers that will stop in the next `BeginBlock`. Not sure I understand the use case for it.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -232,18 +229,40 @@ func (k Keeper) QueryRegisteredConsumerRewardDenoms(goCtx context.Context, req *
 	}, nil
 }
 
-// TODO (PERMISSIONLESS)
 func (k Keeper) QueryProposedConsumerChainIDs(goCtx context.Context, req *types.QueryProposedChainIDsRequest) (*types.QueryProposedChainIDsResponse, error) {
```
mpoke, line 232: What's the use case for having this query? Without a clear use case, this query should be removed. All of this can be done on the frontend -- query all the proposals, filter only the ones still in voting and contain a MsgUpdateConsumer. The MsgUpdateConsumer will have the consumerId.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -232,18 +229,40 @@ func (k Keeper) QueryRegisteredConsumerRewardDenoms(goCtx context.Context, req *
...
 	if req == nil {
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	chains := k.GetAllProposedConsumerChainIDs(ctx)
+	var proposals []govv1.Proposal
+	err := k.govKeeper.Proposals.Walk(ctx, nil, func(proposalId uint64, proposal govv1.Proposal) (stop bool, err error) {
+		proposals = append(proposals, proposal)
```
mpoke, line 241: It's safe to execute the logic within the walk without storing the proposals in a separate array.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -232,18 +229,40 @@ func (k Keeper) QueryRegisteredConsumerRewardDenoms(goCtx context.Context, req *
...
+		status := proposal.Status
+		if status != govv1.ProposalStatus_PROPOSAL_STATUS_DEPOSIT_PERIOD && status != govv1.ProposalStatus_PROPOSAL_STATUS_VOTING_PERIOD {
+			// we only care about active proposals
+			continue
+		}
+
+		messages := proposal.GetMessages()
+		for _, msg := range messages {
+			if sdkMsg, isMsgUpdateConsumer := msg.GetCachedValue().(*types.MsgUpdateConsumer); isMsgUpdateConsumer {
+				proposedChains = append(proposedChains, types.ProposedChain{ProposalID: proposal.Id, ConsumerId: sdkMsg.ConsumerId})
```
mpoke, line 259: Why is the chainId in the ProposedChain deprecated?

mpoke (commented):
Partial review on queries.

mpoke (commented):
Completed the review. See my comments below.

proto/interchain_security/ccv/provider/v1/query.proto
insumity: It won't be backward compatible if I just remove them. I don't know who might be using those queries and I wouldn't want to break their builds.

proto/interchain_security/ccv/provider/v1/query.proto
insumity: Ok, I removed the new queries.

proto/interchain_security/ccv/provider/v1/query.proto
mpoke: As long as we add clear changelogs is fine to break compatibility, especially for queries.

x/ccv/provider/keeper/grpc_query.go
insumity: See my comment [here](https://github.com/cosmos/interchain-security/pull/2177/files#r1734382529).

x/ccv/provider/keeper/grpc_query.go
insumity: Removed those queries altogether.

x/ccv/provider/keeper/grpc_query.go
insumity: Removed these queries altogether.

x/ccv/provider/keeper/grpc_query.go
insumity: Why was this query there in the first place? Do we know if someone is using it? I don't have the answers to these questions. The changes here were meant to keep the same behavior we had before. 
But if you do not see a use case, I'll go ahead and remove it.

Note that a multitude of things (e.g., [this one](https://github.com/cosmos/interchain-security/pull/2162)) could be done in the front end but nevertheless the front-end team asked us for exact queries (not this one though) to simplify their life.

x/ccv/provider/keeper/grpc_query.go
insumity: Removed this query; see comment [here](https://github.com/cosmos/interchain-security/pull/2177/files#r1734448221).

x/ccv/provider/keeper/grpc_query.go
insumity: Removed this query; see comment [here](https://github.com/cosmos/interchain-security/pull/2177/files#r1734448221).

x/ccv/provider/keeper/grpc_query_test.go
insumity: This test was removed.

x/ccv/provider/migrations/v8/migrations.go
```diff
@@ -138,70 +140,75 @@ func MigrateLaunchedConsumerChains(ctx sdk.Context, store storetypes.KVStore, pk
...
 		// chainId -> SlashAcks
 		rekeyFromChainIdToConsumerId(store, SlashAcksKeyPrefix, chainId, consumerId)
+
 		// chainId -> InitChainHeight
 		rekeyFromChainIdToConsumerId(store, InitChainHeightKeyPrefix, chainId, consumerId)
+
 		// chainId -> PendingVSCs
 		rekeyFromChainIdToConsumerId(store, PendingVSCsKeyPrefix, chainId, consumerId)
-		// chainId -> EquivocationEvidenceMinHeight
-		rekeyFromChainIdToConsumerId(store, EquivocationEvidenceMinHeightKeyPrefix, chainId, consumerId)
```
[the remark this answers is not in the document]
insumity: To respect the order of keys in `keys.go`.

x/ccv/provider/keeper/grpc_query.go
mpoke: Comment it for now with a TODO and we handle it later in a discussion with the FE team.

mpoke: Or just remove it and we add it later if we need.

x/ccv/provider/migrations/v8/migrations.go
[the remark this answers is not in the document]
mpoke: It makes it very hard to review the PR :|

x/ccv/provider/migrations/v8/migrations.go
```diff
@@ -138,70 +140,75 @@ func MigrateLaunchedConsumerChains(ctx sdk.Context, store storetypes.KVStore, pk
...
 
 		// chainId -> ConsumerAddrsToPruneV2
 		rekeyChainIdAndTsKey(store, providertypes.ConsumerAddrsToPruneV2KeyPrefix(), chainId, consumerId)
 
-		// set channelId -> consumerId mapping
-		channelId, found := pk.GetConsumerIdToChannelId(ctx, consumerId)
-		if !found {
-			return errorsmod.Wrapf(ccv.ErrInvalidConsumerState, "cannot find channel ID associated with consumer ID: %s", consumerId)
-		}
-		pk.SetChannelToConsumerId(ctx, channelId, consumerId)
```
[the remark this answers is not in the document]
insumity: I moved it up front to respect the order of the keys as seen in `keys.go`.

> Also, the old ChannelToChainId key needs to be deleted. I forgot to do that.

I do not believe so. We're overwriting what is stored under the specific `channelId` to be the `consumerId` and not the `chainId` anymore, so I do not believe we need a deletion.

x/ccv/provider/migrations/v8/migrations.go
```diff
@@ -211,37 +218,86 @@ func MigrateLaunchedConsumerChains(ctx sdk.Context, store storetypes.KVStore, pk
...
 		}
 
-		// TODO other keys
-		// - ConsumerIdToStopTimeKey
-		// - ProviderConsAddrToOptedInConsumerIdsKey
-		// - ...
+		// set phase to launched
+		pk.SetConsumerPhase(ctx, consumerId, providerkeeper.Launched)
 
+		// An already-launched chain might haven been set to be removed. After the migration, we won't be looking
```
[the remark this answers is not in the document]
insumity: Sounds good. Removed this part.

proto/interchain_security/ccv/provider/v1/query.proto
insumity: The cost of deprecating them seems almost 0. I do not see the benefit of removing them. But if you insist, I can go ahead and remove them. Let me know.

x/ccv/provider/migrations/v8/migrations.go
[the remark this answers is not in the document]
mpoke: > I do not believe so. We're overwriting what is stored under the specific channelId to be the consumerId and not the chainId anymore, so I do not believe we need a deletion.

You're right. We are just replacing the value, not the key.

x/ccv/provider/migrations/v8/migrations.go
insumity: If the allowlist is empty in a `MsgUpdateConsumer`, then the chain won't have an allowlist anymore. It won't have an allowlist in either the `allowlist` state or the state under `PowerShapingParameters.allowlist`.

> So we basically have the same info store twice?
We do. Information is stored in a slightly different ways between `allowlist` and `PowerShapingParameters.allowlist`. We could only use the latter but this would imply we alwyas have to traverse an allowlist when trying to see if a validator can validate a chain.

proto/interchain_security/ccv/provider/v1/query.proto
mpoke: As the logic changed. SDK does the same thing. It's fine to break some things to reduce code complexity.

x/ccv/provider/migrations/v8/migrations.go
mpoke: Okay, so we need to make sure the two are always synced -- the fields in the `PowerShapingParameters` and the values under the `allowlist` and `denylist` keys. Right?

insumity: Yes.

x/ccv/provider/migrations/v8/migrations.go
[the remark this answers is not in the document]
insumity: I can see at least 2 options:
1. Have them in the order they were that does not respect the `keys.go` order so you can review it and then we leave it as is?
2. Have them in the order they were that does not respect the `keys.go` order so you can review it and then reorder them based on the order of `keys.go`? I would assume that the second change would need another review, so we would end up at the same spot.

Let me know which option you prefer or if you have any other recommendation.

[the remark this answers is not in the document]
mpoke: Now is too late. I already reviewed them in the new order. Let them be as they are in this PR.

mpoke (approved):
LGTM

PR #2185, opened 2024-08-29
fix: permissionless added event + cli changes

bermuell:
## Description

Closes: #XXXX
Changes to enable/facilitate e2e tests:
- Added event on CreateConsumer to get `consumer-id` on tx
- Changed `update-consumer` and `create-consumer` cli to accept only one json file as input

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Updated command-line interface (CLI) commands for enhanced clarity, focusing on consumer identification.
	- Simplified `create-consumer` and `update-consumer` commands to require a single JSON file for input.
	- Added event emission for consumer registration to improve system notifications.

- **Improvements**
	- Enhanced validation and usability of CLI commands by consolidating input requirements.
	- Introduced new event types and attributes for better tracking of consumer interactions.

- **Bug Fixes**
	- Adjusted parameter names in commands to ensure consistency and reduce user errors.

### DIFF
--- a/x/ccv/provider/client/cli/tx.go
+++ b/x/ccv/provider/client/cli/tx.go
@@ -223,9 +223,9 @@ Note that the one that signs this message is the owner of this consumer chain. T
 changed by updating the consumer chain.
 
 Example:
-%s tx provider create-consumer [chain-id] [path/to/metadata.json] [path/to/initialization-parameters.json] [path/to/power-shaping-parameters.json] --from node0 --home ../node0 --chain-id $CID
+%s tx provider create-consumer [path/to/create_consumer.json] --from node0 --home ../node0 --chain-id $CID
 `, version.AppName)),
-		Args: cobra.ExactArgs(4),
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
@@ -240,37 +240,16 @@ Example:
 
 			signer := clientCtx.GetFromAddress().String()
 
-			chainId := args[0]
-
-			metadata := types.ConsumerMetadata{}
-			metadataJson, err := os.ReadFile(args[1])
+			consCreateJson, err := os.ReadFile(args[0])
 			if err != nil {
 				return err
 			}
-			if err = json.Unmarshal(metadataJson, &metadata); err != nil {
-				return fmt.Errorf("metadata unmarshalling failed: %w", err)
+			consCreate := types.MsgCreateConsumer{}
+			if err = json.Unmarshal(consCreateJson, &consCreate); err != nil {
+				return fmt.Errorf("consumer data unmarshalling failed: %w", err)
 			}
 
-			initializationParameters := types.ConsumerInitializationParameters{}
-			initializationParametersJson, err := os.ReadFile(args[2])
-			if err != nil {
-				return err
-			}
-			if err = json.Unmarshal(initializationParametersJson, &initializationParameters); err != nil {
-				return fmt.Errorf("initialization parameters unmarshalling failed: %w", err)
-			}
-
-			powerShapingParameters := types.PowerShapingParameters{}
-
-			powerShapingParametersJson, err := os.ReadFile(args[3])
-			if err != nil {
-				return err
-			}
-			if err = json.Unmarshal(powerShapingParametersJson, &powerShapingParameters); err != nil {
-				return fmt.Errorf("power-shaping parameters unmarshalling failed: %w", err)
-			}
-
-			msg, err := types.NewMsgCreateConsumer(signer, chainId, metadata, &initializationParameters, &powerShapingParameters)
+			msg, err := types.NewMsgCreateConsumer(signer, consCreate.ChainId, consCreate.Metadata, consCreate.InitializationParameters, consCreate.PowerShapingParameters)
 			if err != nil {
 				return err
 			}
--- a/x/ccv/provider/client/cli/tx.go
+++ b/x/ccv/provider/client/cli/tx.go
@@ -277,9 +277,9 @@ func NewUpdateConsumerCmd() *cobra.Command {
 Note that only the owner of the chain can initialize it.
 
 Example:
-%s tx provider update-consumer [consumer-id] [owner-address] [path/to/metadata.json] [path/to/initialization-parameters.json] [path/to/power-shaping-parameters.json] --from node0 --home ../node0 --chain-id $CID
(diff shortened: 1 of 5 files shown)


x/ccv/provider/client/cli/tx.go
```diff
@@ -223,9 +223,9 @@ Note that the one that signs this message is the owner of this consumer chain. T
 changed by updating the consumer chain.
 
 Example:
-%s tx provider create-consumer [chain-id] [path/to/metadata.json] [path/to/initialization-parameters.json] [path/to/power-shaping-parameters.json] --from node0 --home ../node0 --chain-id $CID
+%s tx provider create-consumer [path/to/create_consumer.json] --from node0 --home ../node0 --chain-id $CID
```
mpoke, line 226: Why do we need one json file as input?

mpoke: If you change, also change the Use description above.

x/ccv/provider/client/cli/tx.go
```diff
@@ -298,9 +277,9 @@ func NewUpdateConsumerCmd() *cobra.Command {
 Note that only the owner of the chain can initialize it.
 
 Example:
-%s tx provider update-consumer [consumer-id] [owner-address] [path/to/metadata.json] [path/to/initialization-parameters.json] [path/to/power-shaping-parameters.json] --from node0 --home ../node0 --chain-id $CID
+	%s tx provider update-consumer [path/to/consumer-update.json] --from node0 --home ../node0 --chain-id $CID
```
mpoke, line 280: ditto

x/ccv/provider/client/cli/tx.go
```diff
@@ -505,7 +469,7 @@ func NewOptOutCmd() *cobra.Command {
 
 func NewSetConsumerCommissionRateCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "set-consumer-commission-rate [consumer-chain-id] [commission-rate]",
+		Use:   "set-consumer-commission-rate [consumer-id] [commission-rate]",
```
mpoke, line 472: Below on the example the consumer-id is `consumer-1`. It needs to be a number.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -201,7 +201,7 @@ func (k Keeper) QueryValidatorProviderAddr(goCtx context.Context, req *types.Que
 	}
 	consumerAddr := types.NewConsumerConsAddress(consumerAddrTmp)
 
-	providerAddr, found := k.GetValidatorByConsumerAddr(ctx, req.ChainId, consumerAddr)
+	providerAddr, found := k.GetValidatorByConsumerAddr(ctx, req.ConsumerId, consumerAddr)
```
mpoke, line 204: Nice catch. @insumity why do we have deprecated fields in query requests. Given that both chanId and consumerId have the same type, it's easy to have bugs where we call functions with a chainId.

mpoke (approved):

x/ccv/provider/keeper/grpc_query.go
insumity: For backward compatibility. We try to prevent `chainIds` from being used by doing:
```
} else if req.ChainId != "" {
    return nil, status.Errorf(codes.InvalidArgument, "ChainId has been deprecated. Use ConsumerId instead.")
}
```
But in this case we unfortunately missed it.

x/ccv/types/events.go
```diff
@@ -15,6 +15,7 @@ const (
 	EventTypeExecuteConsumerChainSlash  = "execute_consumer_chain_slash"
 	EventTypeFeeDistribution            = "fee_distribution"
 	EventTypeConsumerSlashRequest       = "consumer_slash_request"
+	EventTypeConsumerRegistration       = "consumer_registration"
```
insumity, line 18: nit: maybe `EventTypeConsumerCreation` instead

insumity (approved):

x/ccv/provider/client/cli/tx.go
bermuell: easier from a usability perspective IMO, as you have to deal with only one file.

x/ccv/provider/keeper/grpc_query.go
mpoke: > For backward compatibility.

Can you please give more details about this. What's the use case that we want to keep compatibility with?

x/ccv/provider/client/cli/tx.go
```diff
@@ -215,17 +215,17 @@ Example:
 
 func NewCreateConsumerCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create-consumer [chain-id] [metadata] [initialization-parameters] [power-shaping-parameters]",
+		Use:   "create-consumer [consumer-parameters]",
```
mpoke, line 218: How will the user differentiate between the json they need to provide for create-consumer and the one for update-consumer?

x/ccv/provider/keeper/grpc_query.go
insumity: We could imagine that Hermes is using this query (`QueryValidatorProviderAddr`) so they use our protos to build Hermes. If we were to simply go and remove `chain_id`, we would break their builds. It seems, in those cases, the right approach is to just mark `chain_id` as deprecated. It is true that the actual query is still broken from their side but this seems less severe.
We could remove all those `chain_id`s if we were to bump the proto version to `v2` but this seems like it would be more time consuming at this stage.

mpoke: Bumping the proto version is something we should talk to @MSalopek. 

Regarding breaking the queries for Hermes for example, as you stated it will be broken anyway. All the users of these queries will need to update their code. Leaving chainId there seems even more dangerous to me as there is a chance that we forget to return an error and the query goes through and returns weird stuff. So my recommendation would be to remove deprecated fields from the queries.

x/ccv/provider/client/cli/tx.go
bermuell: right, we need to provide an example.
I can also imagine to have the common part's in a json file (e.g. metadata, init-, powershaping-parameters) and those which differ as mandatory/optional arguments (optional would be 'new_owner').
If you're ok I'll keep this change for later once we have a preference. 
In any case we'll end up with a 'one-file solution' I'd say.

x/ccv/provider/keeper/grpc_query.go
MSalopek: I would avoid `v2` and inform hermes that there will be a compatibility break and that they should update their code.

PR #2189, opened 2024-08-29
fix!: several permissionless changes

mpoke:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for retrieving consumer power shaping parameters.
	- Introduced a new method for determining if a consumer is active based on its phase.
	- Updated logic to streamline handling of validator participation in consumer chains.

- **Bug Fixes**
	- Improved robustness of tests by ensuring proper error checks during consumer operations.

- **Documentation**
	- Clarified error messages in initialization and power shaping parameters retrieval.

- **Refactor**
	- Consolidated logic for managing consumer states and validator parameters, removing redundant methods.

- **Tests**
	- Updated tests to reflect changes in consumer power shaping parameters and validator handling.


x/ccv/provider/keeper/msg_server.go
```diff
@@ -357,58 +367,59 @@ func (k msgServer) CreateConsumer(goCtx context.Context, msg *types.MsgCreateCon
...
 		}
 	}
 
 	// get the previous spawn time so that we can use it in `PrepareConsumerForLaunch`
-	var previousSpawnTime time.Time
-	previousInitializationParameters, err := k.Keeper.GetConsumerInitializationParameters(ctx, msg.ConsumerId)
-	if err == nil {
-		previousSpawnTime = previousInitializationParameters.SpawnTime
+	previousInitializationParameters, err := k.Keeper.GetConsumerInitializationParameters(ctx, consumerId)
+	if err != nil {
```
insumity, line 414: I think the first time the initialization parameters are set, this would return `err != nil` and hence we would fail although we should not.

x/ccv/provider/keeper/partial_set_security.go
```diff
@@ -372,16 +401,21 @@ func (k Keeper) FulfillsMinStake(ctx sdk.Context, consumerId string, providerAdd
 }
 
 // ComputeNextValidators computes the validators for the upcoming epoch based on the currently `bondedValidators`.
-func (k Keeper) ComputeNextValidators(ctx sdk.Context, consumerId string, bondedValidators []stakingtypes.Validator, minPowerToOptIn int64) []types.ConsensusValidator {
+func (k Keeper) ComputeNextValidators(
+	ctx sdk.Context,
+	consumerId string,
+	bondedValidators []stakingtypes.Validator,
+	powerShapingParameters types.PowerShapingParameters,
```
insumity, line 408: If I recall correctly, in some previous PR, it was mentioned that we shouldn't pass `powerShapingParameters` or `initializationParameters` if the `consumerId` is also passed as an argument because we can retrieve those from `consumerId`. I prefer it this way.
Do we have a rule on when to pass and when not?

x/ccv/provider/keeper/msg_server.go
mpoke: When creating a consumer (via CreateConsumer), all the params are initialized even if not provided. So if this get is returning error, then there is an issue with the state.

x/ccv/provider/keeper/grpc_query.go
```diff
@@ -110,7 +110,10 @@ func (k Keeper) GetConsumerChain(ctx sdk.Context, consumerId string) (types.Chai
 	}
 
 	clientID, _ := k.GetConsumerClientId(ctx, consumerId)
-	topN := k.GetTopN(ctx, consumerId)
+	powerShapingParameters, err := k.GetConsumerPowerShapingParameters(ctx, consumerId)
```
sainoe, line 113: `GetConsumerPowerShapingParameters` returns an error when params aren't set for the chain.
The error shouldn't be checked here.

x/ccv/provider/keeper/partial_set_security.go
mpoke: The rule is to try to avoid accessing the state multiple times for no reason. In this case, the powerShapingParameters are available everytime when calling ComputeNextValidators, so it makes sense to pass it. The consumerId is needed for the other methods, i.e., FilterValidators and CanValidateChain.

x/ccv/provider/keeper/grpc_query.go
mpoke: I modified the CreateConsumer method to always set power shaping params (if not provided by the user, then it will just set the defaults -- zero values).

PR #2208, opened 2024-09-03
fix!: remove `stopTime` from `MsgRemoveConsumer`

insumity:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


testutil/keeper/unit_test_helpers.go
```diff
@@ -266,15 +271,6 @@ func TestProviderStateIsCleanedAfterConsumerChainIsStopped(t *testing.T, ctx sdk
 	acks := providerKeeper.GetSlashAcks(ctx, consumerId)
 	require.Empty(t, acks)
 
-	// in case the chain was successfully stopped, it should not contain a Top N associated to it
-	ps, err := providerKeeper.GetConsumerPowerShapingParameters(ctx, consumerId)
```
insumity, line 270: We do not delete the power-shaping or initialization parameters during a chain deletion anymore.

x/ccv/provider/migrations/v5/migrations.go
```diff
@@ -10,9 +10,11 @@ import (
 // This migration only takes already registered chains into account.
 // If a chain is in voting while the upgrade happens, this is not sufficient,
 // and a migration to rewrite the proposal is needed.
+//
+// TODO (PERMISSIONLESS): this migration needs to be fix or removed
```
mpoke, line 14: This needs to be fixed as the keeper methods changed. This means that nobody can run the v5 (or v6) migration without going first through v8. Thus, these migration require having legacy keeper methods.

proto/interchain_security/ccv/provider/v1/genesis.proto
```diff
@@ -78,6 +78,8 @@ message ConsumerState {
   repeated interchain_security.ccv.v1.ValidatorSetChangePacketData
       pending_valset_changes = 6 [ (gogoproto.nullable) = false ];
   repeated string slash_downtime_ack = 7;
+  // the phase of the consumer chain
+  ConsumerPhase phase = 9;
```
mpoke, line 82: Why is this added here? Making the genesis export / import work requires more fields, so probably it makes sense to do it in its own PR.

x/ccv/provider/keeper/relay.go
```diff
@@ -209,7 +213,7 @@ func (k Keeper) QueueVSCPackets(ctx sdk.Context) error {
 		return fmt.Errorf("getting provider active validators: %w", err)
 	}
 
-	for _, consumerId := range k.GetAllRegisteredConsumerIds(ctx) {
+	for _, consumerId := range k.GetAllConsumersWithIBCClients(ctx) {
```
mpoke, line 216: Need to check if the consumer is launched and not stopped.

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -353,36 +354,58 @@ func (k Keeper) MakeConsumerGenesis(
...
+
+	// state of this chain is removed once UnbondingPeriod elapses
+	unbondingPeriod, err := k.stakingKeeper.UnbondingTime(ctx)
+	if err != nil {
+		return err
+	}
+	removalTime := ctx.BlockTime().Add(unbondingPeriod)
+
+	if err := k.SetConsumerRemovalTime(ctx, consumerId, removalTime); err != nil {
+		return errorsmod.Wrapf(types.ErrInvalidRemovalTime, "cannot set removal time: %s", err.Error())
```
mpoke, line 372: Let's use `ErrInvalid...` error message for invalid input. Here the removalTime is computed internally. 

```suggestion
		return fmt.Errorf("cannot set removal time (%s): %s", removalTime.String(), err.Error())
```

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -353,36 +354,58 @@ func (k Keeper) MakeConsumerGenesis(
...
+				"error", err.Error())
 			continue
 		}
 
-		err = k.StopConsumerChain(cachedCtx, consumerId, true)
+		// Remove consumer to prevent re-trying removing this chain.
+		err = k.RemoveConsumerToBeRemoved(ctx, consumerId, removalTime)
 		if err != nil {
-			k.Logger(ctx).Error("consumer chain could not be stopped",
+			ctx.Logger().Error("could not remove consumer from to-be-stopped queue",
```
mpoke, line 396: nit
```suggestion
			ctx.Logger().Error("could not remove consumer from to-be-removed queue",
```

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -353,36 +354,58 @@ func (k Keeper) MakeConsumerGenesis(
...
 
-		k.SetConsumerPhase(cachedCtx, consumerId, types.ConsumerPhase_CONSUMER_PHASE_STOPPED)
-
-		err = k.RemoveConsumerToBeStopped(ctx, consumerId, stopTime)
+		// delete consumer chain in a cached context to abort deletion in case of errors
+		cachedCtx, writeFn := ctx.CacheContext()
+		err = k.DeleteConsumerChain(cachedCtx, consumerId)
 		if err != nil {
-			ctx.Logger().Error("could not remove consumer from to-be-stopped queue",
+			k.Logger(ctx).Error("consumer chain could not be stopped",
```
mpoke, line 406: nit:
```suggestion
			k.Logger(ctx).Error("consumer chain could not be removed",
```

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -403,26 +426,26 @@ func (k Keeper) BeginBlockStopConsumers(ctx sdk.Context) {
...
 		}
-		if stopTime.After(ctx.BlockTime()) {
+		if removalTime.After(ctx.BlockTime()) {
 			return result
 		}
 
-		consumers, err := k.GetConsumersToBeStopped(ctx, stopTime)
+		consumers, err := k.GetConsumersToBeRemoved(ctx, removalTime)
 		if err != nil {
 			k.Logger(ctx).Error("failed to retrieve consumers to stop",
```
mpoke, line 447: nit
```suggestion
			k.Logger(ctx).Error("failed to retrieve consumers to remove",
```

x/ccv/provider/keeper/msg_server.go
```diff
@@ -469,25 +469,12 @@ func (k msgServer) RemoveConsumer(goCtx context.Context, msg *types.MsgRemoveCon
 
 	phase := k.Keeper.GetConsumerPhase(ctx, consumerId)
 	if phase != types.ConsumerPhase_CONSUMER_PHASE_LAUNCHED {
```
mpoke, line 471: Why only stop launched consumers? This means that there is no way to stop and delete a consumer that is just registered. We should be able to stop and delete consumers that are not yet launched without waiting to the unbonding period.

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -493,9 +512,16 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, consumerId string, closeChan
 	k.DeleteAllOptedIn(ctx, consumerId)
 	k.DeleteConsumerValSet(ctx, consumerId)
 
+	k.DeleteConsumerRewardsAllocation(ctx, consumerId)
+	k.DeleteConsumerOwnerAddress(ctx, consumerId)
+	k.DeleteConsumerRemovalTime(ctx, consumerId)
+
 	// TODO (PERMISSIONLESS) add newly-added state to be deleted
```
mpoke, line 519: Is this TODO still accurate?

x/ccv/provider/keeper/consumer_lifecycle.go
```diff
@@ -493,9 +512,16 @@ func (k Keeper) StopConsumerChain(ctx sdk.Context, consumerId string, closeChan
 	k.DeleteAllOptedIn(ctx, consumerId)
 	k.DeleteConsumerValSet(ctx, consumerId)
 
+	k.DeleteConsumerRewardsAllocation(ctx, consumerId)
+	k.DeleteConsumerOwnerAddress(ctx, consumerId)
```
mpoke, line 516: Probably worth leaving this in as well.

x/ccv/provider/keeper/keeper.go
```diff
@@ -211,23 +211,25 @@ func (k Keeper) DeleteConsumerIdToChannelId(ctx sdk.Context, consumerId string)
...
 	defer iterator.Close()
 
 	for ; iterator.Valid(); iterator.Next() {
 		// remove 1 byte prefix from key to retrieve consumerId
 		consumerId := string(iterator.Key()[1:])
-		consumerIds = append(consumerIds, consumerId)
+
+		// A chain might have stopped, but we might not yet have its consumer id to client id association deleted.
+		// To avoid returning stopped chains, we check the phase of the consumer chain and only return launched chains.
+		if k.GetConsumerPhase(ctx, consumerId) == types.ConsumerPhase_CONSUMER_PHASE_LAUNCHED {
```
mpoke, line 230: Remove this from here and have it in the actual logic, e.g., QueueVSCPackets. For example, in ditribution, iterating overl consumers that are stopped but not yet deleted is useful as the rewards can still be distributed.

x/ccv/provider/keeper/relay.go
```diff
@@ -30,8 +30,7 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
 			"error", err,
 		)
 		if consumerId, ok := k.GetChannelIdToConsumerId(ctx, packet.SourceChannel); ok {
-			// stop consumer chain and release unbonding
-			return k.StopConsumerChain(ctx, consumerId, false)
```
mpoke, line 34: If you try to close the channel here it will result in an error as the channel will already be closed in the IBC module. That's the reason we had the `closeChan = false` argument. What's the reason for removing it?

mpoke (changes requested):
See my comments below.

proto/interchain_security/ccv/provider/v1/genesis.proto
insumity: Was added because genesis work was failing. I agree that more things need to be added but if I didn't add this here, we would have tests failing and I wanted to this PR to be coherent (test-wise).

x/ccv/provider/keeper/msg_server.go
insumity: Discussed offline. Decided to "Ignore for now. Low priority"

My point of view:
> This adds additional complexity (check on whether we can delete everything now or later) in my view for no real benefit. Additionally, if you haven't started yet, there's probably not much to delete to start with, so no reason to start a deletion. Also, I do not think conceptually it makes sense to remove a registered chain. If you don't like it, just leave it as is and later re-register something another chain.

x/ccv/provider/keeper/consumer_lifecycle.go
insumity: Yes. Precisely deciding what we should delete would need some time and it's not that important at this point. I think we can all agree that we need to keep consumer metadata, power-shaping, and initialization parameters.

x/ccv/provider/keeper/consumer_lifecycle.go
insumity: Kept the owner address. I do not see any reason to keep the consumer rewards allocation since after the chain gets deleted we won't be able to get rewards in any case.

x/ccv/provider/keeper/keeper.go
insumity: Removed and added an additional check in `QueueVSCPackets`.

x/ccv/provider/keeper/relay.go
insumity: Discussed offline. There's no `panic` when trying to close the channel, so there's no need for `closeChan` any longer.

x/ccv/provider/keeper/consumer_lifecycle.go
mpoke: Owner address. The RewardsAllocation should be deleted.

mpoke (approved):

Issue #2198, opened 2024-09-02
e2e tests: support version specific actions for compatbility tests


bermuell:
e2e tests need to support version specific actions

Actions on how to perfrom specific test steps change over ICS versions, especially with Permissionless ICS.
E2E framework needs to be adapted to support compatibility testing when test actions changed accross versions.

Bonus:
- introduce a waitForTx() util function for e2e test which waits until a given transaction is seen
- change `SpawnTime` in all actions to more appropriate e.g. `TimeUntilSpawn` as it is used as a duration and not a specific point in time 
- ChainIDs are used to map configs to chain identifiers. With permissionless ICS, chains with same `ChainID` are supported which leads to the situation that the ChainID id will be different than the one in in the configuration related configuration. To be easier to differentiate the two ChainIDs (config, key),  the map used for the chain configs should be adapted to use a different type for the key.

PR #2223, opened 2024-09-05
test: Add e2e tests for permissionless ICS

bermuell:
## Description

Add basic e2e tests for permissionless ICS
Change e2e tests to allow chain configurations with same ChainID

Limitation:
Current hermes version does not support two chains with same 'ChainID'. Therefore the tests Initializes two
chains with same ChainID and launches only one of them.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/tests/e2e/steps_permissionless_ics.go
+++ b/tests/e2e/steps_permissionless_ics.go
@@ -9,7 +9,7 @@ import (
 
 // stepsPermissionlessICS tests
 // - starting multiple permissionless consumer chains with the same chain ID
-// - opt-in of a validator on two different chains with the same chain ID
+// - that a validator CANNOT opt-in on two different chains with the same chain ID
 // - taking ownership of a consumer chain
 func stepsPermissionlessICS() []Step {
 	s := concatSteps(
@@ -39,8 +39,6 @@ func stepsPermissionlessICS() []Step {
 			// - create the consumer chain
 			// - opt-in a validator
 			// - launch the chain
-			// - TODO (PERMISSIONLESS): 'start the chain'. Note: due to hermes we can't yet start two chains with the same ChainID as it's not yet supported
-
 			{
 				Action: CreateConsumerChainAction{
 					Chain:         ChainID("provi"),
--- a/tests/e2e/steps_permissionless_ics.go
+++ b/tests/e2e/steps_permissionless_ics.go
@@ -1,0 +1,213 @@
+package main
+
+import (
+	"time"
+
+	clienttypes "github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
+	e2e "github.com/cosmos/interchain-security/v6/tests/e2e/testlib"
+)
+
+// stepsPermissionlessICS tests
+// - starting multiple permissionless consumer chains with the same chain ID
+// - opt-in of a validator on two different chains with the same chain ID
+// - taking ownership of a consumer chain
+func stepsPermissionlessICS() []Step {
+	s := concatSteps(
+		[]Step{
+			// Start the provider chain
+			{
+				Action: StartChainAction{
+					Chain: ChainID("provi"),
+					Validators: []StartChainValidator{
+						{Id: ValidatorID("alice"), Stake: 100000000, Allocation: 10000000000},
+						{Id: ValidatorID("bob"), Stake: 200000000, Allocation: 10000000000},
+						{Id: ValidatorID("carol"), Stake: 300000000, Allocation: 10000000000},
+					},
+				},
+				State: State{
+					ChainID("provi"): ChainState{
+						ValPowers: &map[ValidatorID]uint{
+							ValidatorID("alice"): 100,
+							ValidatorID("bob"):   200,
+							ValidatorID("carol"): 300,
+						},
+					},
+				},
+			},
+
+			// Initialize a permissionless chain with ChainID `consu`
+			// - create the consumer chain
+			// - opt-in a validator
+			// - launch the chain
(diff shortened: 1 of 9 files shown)


tests/e2e/steps_permissionless_ics.go
```diff
@@ -0,0 +1,213 @@
...
+import (
+	"time"
+
+	clienttypes "github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
+	e2e "github.com/cosmos/interchain-security/v6/tests/e2e/testlib"
+)
+
+// stepsPermissionlessICS tests
+// - starting multiple permissionless consumer chains with the same chain ID
+// - opt-in of a validator on two different chains with the same chain ID
```
insumity, line 12: Mention that it tests that this "does not work" since it is not allowed. Someone reading it might think that this is possible.

tests/e2e/steps_permissionless_ics.go
```diff
@@ -0,0 +1,213 @@
...
+						},
+					},
+				},
+			},
+
+			// Initialize a permissionless chain with ChainID `consu`
+			// - create the consumer chain
+			// - opt-in a validator
+			// - launch the chain
+			// - TODO (PERMISSIONLESS): 'start the chain'. Note: due to hermes we can't yet start two chains with the same ChainID as it's not yet supported
```
insumity, line 42: Why do we need to start the second chain? I think from testing perspective, it's okay if only one is up and running (?)
Is this a general Hermes issue (cannot connect 2 chains with the same `chainId`)?

tests/e2e/steps_permissionless_ics.go
```diff
@@ -0,0 +1,213 @@
...
+							ValidatorID("bob"):   {"consu"},
+							ValidatorID("carol"): {},
+						},
+					},
+				},
+			},
+		},
+		// test chain hijacking prevention
+		[]Step{
+			// Try to change owner of chain and change deny-/allowlist
```
insumity, line 160: If my understanding is correct, the hijacking prevention is indirectly checked by checking that `bob` is not denylisted. Couldn't we test this by having an `ExpectedError` in `UpdateConsumerChainAction` similarly to `OptInAction`?

tests/e2e/steps_permissionless_ics.go
```diff
@@ -0,0 +1,213 @@
...
+					ConsumerChain: ChainID("cons2"), // test chain "cons2" is configured with ChainID "consu"
+					SpawnTime:     uint(time.Minute * 3),
+					InitialHeight: clienttypes.Height{RevisionNumber: 0, RevisionHeight: 1},
+					TopN:          0,
+				},
+				State: State{},
+			},
+			{
+				Action: OptInAction{
+					Chain:     ChainID("cons2"),
```
insumity, line 57: Having this named `ChainID` while the underlying chain is is `consu` could be hard to parse. I wonder if we could rename this to something else, and introduce `ConsumerId ConsumerID` in all the actions instead of using `ChainID`.

insumity (approved):

tests/e2e/steps_permissionless_ics.go
bermuell: yes, confirmed with hermes team that the identifier used in the hermes config is also used to double check that this matches the chain-id.
From an e2e testing perspective it's better IMO to have the 2nd chain also started to verify that there are no side effects, especially with issues reported by hermes. As this can't be tested due to hermes adoption to permissionless ICS the current setup is the best we can get at the moment.

tests/e2e/steps_permissionless_ics.go
bermuell: ConsumerID would also be misleading as it is not possible at that stage to have a match with the consumer ID we get when creating the consumer chain. But I do agree to come up with something which helps to differntiate this.
Added this to #2198

tests/e2e/steps_permissionless_ics.go
bermuell: yes, your understanding is correct. the error check would be more the whitelist approach but possible as well and might just be an additional check beside the one verifying that bob is not denylisted.

tests/e2e/steps_permissionless_ics.go
MSalopek: I'm not sure starting the second chain is relevant.

Any side effects can be confirmed in unit and integration tests.

Per IBC rules it is fine to have 2 chains with the same chain_id. Hermes may have issues connecting 2 chains with the same chain_id, but that's not relevant for ICS.

MSalopek (approved):
LGTM!

We can confirm that haing 2 chains with same chain-id does not break ICS through less costly tests. Now that we have that case added to e2e we should keep it.

EDIT:
I would add steps that transition a chain from opt-in to topN via a gov proposal.

tests/e2e/steps_start_chains.go
```diff
@@ -29,6 +32,144 @@ func stepStartProviderChain() []Step {
...
+		State: State{
+			ChainID(consumerName): ChainState{
+				ValBalances: &valBalance,
+			},
+		},
+	}
+	s = append(s, step)
+
+	// Establish IBC connection
+	steps := []Step{
```
MSalopek, line 147: I would add steps that transition a chain from opt-in to topN via a gov proposal.

bermuell: we have that tested already in stepsModifyChain() in `steps_partial_set_security()`
see https://github.com/cosmos/interchain-security/blob/50bbbfedeb82ee2f826569cf03bfb04d5e49e60b/tests/e2e/steps_partial_set_security.go#L2148

PR #2235, opened 2024-09-06
fix!: revert compat breaks; cleanup

MSalopek:
This PR reverts backward compatibility breaks introduced by https://github.com/cosmos/interchain-security/pull/2171.

Changes to `signer` were reverted because they prevent parsing of existing messages stored in previous blocks.

Unused handlers were removed because they are dead code.

Unused proposals were marked as deprecated in protos. RPC methods that are no longer available were deleted because they cannot be used anyway.


sainoe (approved):

mpoke (approved):
LGTM

PR #2265, opened 2024-09-12
test:  extend slashing window in E2E-tests

sainoe:
The slashing window param was too small and caused the e2e-test `stepsAssignConsumerKeyOnStartedChain` to fail randomly.


mpoke (approved):
Thanks @sainoe.

insumity (approved):

PR #2284, opened 2024-09-18
test: add local testnet script for permissionless ICS

sainoe:
Add a local testnet script which spins a `TopN` consumer with a few nodes.


MSalopek (approved):
Thanks!

mpoke (approved):

Issue #2397, opened 2024-11-12
[Feature]: Require one active validator to opt-in for a consumer chain to launch


mpoke:
### Summary

Currently, the requirement for a consumer chain to launch is to have at least one eligible validator opted in. This includes inactive validators. Given the relative low cost of creating an inactive validator (see https://www.mintscan.io/cosmos/validators), an attacker could launch a large number of consumer chains. Currently, ICS is a "push" protocol, i.e., the provider sends validator updates to all the launched consumer chains. Since this happens in the EndBlock of the provider module, there is a risk of DOS attacks. 

### Proposed Feature

A long-term solution is to make ICS a "pull" protocol. Basically, consumer chains will need to query the provider validator updates using interchain queries. This will take time to implement. 

In the meantime, a short-term solution is to require at least one active validator to opt-in on a consumer chain for that chain to launch. 

### Work Breakdown

```[tasklist]
#### Must have
- [ ] add extra condition to the consumer launch logic in provider module
```

mpoke:
@insumity could you please also limit the number of consumer chains a validator could opt in? It should be a provider param and the default to something like 1000.

PR #2399, opened 2024-11-13
feat!: only allow chains with at least one active validator to launch

insumity:
## Description

Closes: #2397 

**Testing**
Introduced an interchain test.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/.changelog/unreleased/features/2399-opt-in-chain-needs-active-validator-to-launch.md
+++ b/.changelog/unreleased/features/2399-opt-in-chain-needs-active-validator-to-launch.md
@@ -1,0 +1,2 @@
+- `[x/provider]` Prevent Opt-In chains from launching, unless at least one active validator has opted-in to them.
+  ([\#2101](https://github.com/cosmos/interchain-security/pull/2399))


tests/interchain/chainsuite/config.go
```diff
@@ -32,12 +32,14 @@ const (
...
 	ValidatorFunds            = 30_000_000
-	ValidatorCount            = 1
-	FullNodeCount             = 0
-	ChainSpawnWait            = 155 * time.Second
-	CosmosChainType           = "cosmos"
-	GovModuleAddress          = "cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn"
-	TestWalletsNumber         = 15 // Ensure that test accounts are used in a way that maintains the mutual independence of tests
+	// ValidatorCount is set to 2, so we have one active and one inactive (i.e., outside the active set) validator.
+	// Note that the provider has at most 1 validator (see `chain_spec_provider.go`).
+	ValidatorCount    = 2
```
mpoke, line 37: Is this done for al the tests or only for the new one introduced? We should avoid adding complexity to all tests. If you need multiple validators for a certain test, we should just add a new configuration.

mpoke (commented):
The logic LGTM. Just one comment about the tests.

tests/interchain/provider_test.go
```diff
@@ -101,6 +101,52 @@ func (s *ProviderSuite) TestProviderCreateConsumerRejection() {
 	s.Require().Error(err)
 }
 
+// TestOptInChainCanOnlyStartIfActiveValidatorOptedIn tests that only if an active validator opts in to an Opt-In chain, the chain can launch.
+// Scenario 1: Inactive validators opts in, the chain does not launch.
+// Scenario 2: Active validator opts in, the chain launches.
+func (s *ProviderSuite) TestOptInChainCanOnlyStartIfActiveValidatorOptedIn() {
+	testAcc := s.Provider.TestWallets[2].FormattedAddress()
```
stana-miric, line 108: I'll create a separate PR so that we don't have to be aware of which index is in use during the tests, and it will be handled under the hood. For now, I would use an account that isn't used in other tests, for example, account s.Provider.TestWallets[12] (just to avoid an error where an invalid account number is sent). I'll try to fix this as soon as possible since the current solution isn't the most elegant.

tests/interchain/chainsuite/config.go
stana-miric: This will affect all tests in the suite that use this specification, i.e., the entire ProviderSuite. We could create a separate suite to support running tests on a suite with multiple nodes while keeping the existing suite where the provider chain has only one node (ProviderSuite, ProviderMultiNodeSuite).
Another option is to keep it as Karolos implemented and handle the optional nodes in the suite setup, so tests that need additional nodes can simply start the extra validator(s) as the interchaintest framework supports that.
The difference is that with separate suites, we don’t need to start/stop nodes dynamically, but the provider chain will be started twice. I'm not sure what the optimal solution would be.
If you'd like, we can leave it as Karolos implemented in this PR, and I can split it or add dynamic node management in a follow-up PR. I also plan to improve how "independent" test accounts are fetched, so I can handle everything in that PR.

mpoke: Sounds good. Let's do that. 
Regarding what option to choose, go with the one easier to implement.

mpoke (approved):

stana-miric (approved):

PR #2433, opened 2024-12-12
feat!: Add tombstone field to SlashJailParameters

stana-miric:
SlashJailParameters are extended with tombstone field, which indicates whether the validator should be tombstoned when slashed. This parameter will be considered only for double sign infraction, while for downtime will not be used(validator would not be tombstoned)


mpoke (approved):
LGTM. Maybe add a warning when a user is trying to set the tombstone to true for downtime (something like it has no effect for downtime infraction).

PR #2518, opened 2025-02-18
refactor!: Upgrade ICS to IBCv10

Eric-Warehime:
## Description

Upgrade IBC dependency to v10

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/ccv/types/denom_helpers.go
```diff
@@ -0,0 +1,188 @@
...
+
+	errorsmod "cosmossdk.io/errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	cmtbytes "github.com/cometbft/cometbft/libs/bytes"
+	cmttypes "github.com/cometbft/cometbft/types"
+)
+
+// The code in this file is removed in ibc v9. It is copied from ibc v8 to here in order to support the migration to v9
```
vladjdk, line 20: We should speak to the IBC team and figure out why this was pushed downstream, since most IBC-implementing applications seam to use this

app/consumer-democracy/app.go
```diff
@@ -517,7 +500,7 @@ func New(
 	// create static IBC router, add transfer route, then set and seal it
 	ibcRouter := porttypes.NewRouter()
 	ibcRouter.AddRoute(ibctransfertypes.ModuleName, ibcmodule)
-	ibcRouter.AddRoute(consumertypes.ModuleName, consumerModule)
+	ibcRouter.AddRoute(ccvtypes.ConsumerPortID, consumerModule)
```
vladjdk, line 503: Would this not break for chains that are trying to access the original route with the "CCV" name? This looks like it effectively changes the module name to "consumer". Not sure if this has any implications for chains that use "CCV" already

app/consumer/app.go
```diff
@@ -414,7 +388,7 @@ func New(
 	// create static IBC router, add transfer route, then set and seal it
 	ibcRouter := porttypes.NewRouter()
 	ibcRouter.AddRoute(ibctransfertypes.ModuleName, ibcmodule)
-	ibcRouter.AddRoute(ibcconsumertypes.ModuleName, consumerModule)
+	ibcRouter.AddRoute(ccvtypes.ConsumerPortID, consumerModule)
```
vladjdk, line 391: Same question here as above

app/provider/app.go
```diff
@@ -535,6 +504,9 @@ func New(
```
vladjdk, line 504: Out of scope, but on new line 502, we keep using the ModuleName instead of the port id const (which doesn't exist in the provider). Curious to know why it's different here

app/sovereign/app.go
```diff
@@ -445,14 +444,12 @@ func New(
...
 		app.GetSubspace(ibctransfertypes.ModuleName),
 		app.IBCKeeper.ChannelKeeper,
 		app.IBCKeeper.ChannelKeeper,
-		app.IBCKeeper.PortKeeper,
 		app.AccountKeeper,
 		app.BankKeeper,
-		scopedTransferKeeper,
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 	)
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
```
vladjdk, line 455: Same question as above, for new line 460

vladjdk (commented):
Generally looks good! Thanks for the massive push on this. Had a few small questions regarding the new router module names and their implications for chains that use the older name (or if there are any implications at all)

app/consumer-democracy/app.go
Eric-Warehime: Yeah I have the same concern. Had a previous convo w/ @AdityaSripal on this 

[When a channel is created in ICS](https://github.com/cosmos/interchain-security/blob/994a94a5f1c29de17b36dac4dcdf1e28a8891682/x/ccv/consumer/keeper/genesis.go#L97) it uses consumer as the Port ID, but in the port router the key is ccvconsumer  which I believe is [created in each app.](https://github.com/cosmos/interchain-security/blob/994a94a5f1c29de17b36dac4dcdf1e28a8891682/app/consumer/app.go#L417)  
The problem then is that [this logic is not going to find the module](https://github.com/cosmos/ibc-go/commit/16a88b4d6fd14dd2b7e0cc9ccb0103b41b809605#diff-91fb52aa2c91092fdd10a9a17c92247d443a378b977c5f808f6f0cdafedce9acR96-R111)
And the prefix stuff added at the bottom is looking for `ccvconsumer` as a substring of `consumer` (which it's not).


However, we agreed that it should be fine to change the route registration here since it's not stored in the keeper anywhere and is just a runtime cache. This feels like a breaking change that's caused by the IBC changes and should be immediately evident when consumer chains are upgrading.

app/provider/app.go
Eric-Warehime: For provider, the port ID is `provider` and the module name is `provider`. The change in routing logic that broke consumers happened because the module name is `ccvconsumer` and the port is `consumer`. So I had to change it in consumers but not in providers.

app/consumer-democracy/app.go
vladjdk: Sweet, sounds good

vladjdk (approved):
lgtm! 

Thanks for the answers to the questions above. If routes are just checked during runtime, then nothing should break. We'll still keep an eye out during upgrades.

Great stuff!

